Issue #80, opened 2020-04-22
Swap positions of Quick Launch vs Cluster Selector on the Terminal Nodes screen


alex-kovoy:
Per Ev's request, we need to swap positions of Quick Launch vs Cluster Selector on the Terminal Nodes screen. The focus should be automatically set on the Quick Launch.

alex-kovoy:
@alexwolfe ^^^  this is a minor change

Issue #79, opened 2020-04-22
Allow quick navigation between terminal tabs 


alex-kovoy:
A user should be able to use basic hot keys (alt+1..n) to switch between open terminal tabs.

PR #81, opened 2020-04-23
Allow switching tabs with shorcut alt + numkeys

kimlisa:
resolves https://github.com/gravitational/webapps/issues/79
resolves https://github.com/gravitational/webapps/issues/80

#### Description
- Created custom hook that handles:
  - ubuntu: alt + numKeys
  - mac: ctrl + numKeys
- Unit tested custom hook
- Swapped quick launcher with cluster selector
- Autofocus on quick launcher

#### Manual Testing
- Tested in ubuntu firefox and chrome
- Tested in mac firefox, chrome, and safari

#### Screen
chrome:
Peek 2020-04-23 14-22

swapped quick launcher with cluster selector:
Screenshot from 2020-04-23 17-23-10


packages/teleport/src/lib/term/terminal.js
```diff
@@ -54,6 +55,14 @@ class TtyTerminal {
       cursorBlink: false,
     });
 
+    // TODO deprecated, use attachCustomKeyEventHandler when we upgrade xterm
+    this.term.attachCustomKeydownHandler(event => {
+      // prevents xterms default alt/cmd <1-9>
+      if ((event.altKey || event.metaKey) && DIGITS.indexOf(event.key) !== -1) {
```
alex-kovoy, line 61: It's OK to duplicate  [1..10] values here.

packages/teleport/src/console/useOnAltKeyDown.ts
```diff
@@ -0,0 +1,59 @@
...
+// DIGITS defines event.key values on keyboard from 1 - 9.
+export const DIGITS = ['1', '2', '3', '4', '5', '6', '7', '8', '9'];
+
+/**
+ * useOnAltKeyDown switches current console tab to the tab that
+ * user indicated by numbers 1-9.
+ *
+ * @param ctx data that is shared between Console related components
+ */
+const useOnAltKeyDown = (ctx: ConsoleContext) => {
```
alex-kovoy, line 29: `useOnAltKeyDown` -> `useHotKeyNavigation`

packages/teleport/src/console/useOnAltKeyDown.ts
```diff
@@ -0,0 +1,59 @@
...
+  React.useEffect(() => {
+    /**
+     * handleKeyDown listens and handles keyboard events: alt/command + <1-9>.
+     * Checks if alt or cmd was pressed with a number and if a document exists
+     * on the indicated tab, it will go to that tab.
+     */
+    const handleKeyDown = event => {
+      // windows/ubuntu uses alt key
+      // mac uses command key (meta key)
+      if (!event.altKey && !event.metaKey) return;
```
alex-kovoy, line 39:
```js
  const index =   ['1', '2', '3', '4', '5', '6', '7', '8', '9'].indexOf(event.key);
  const isHotKey = event.altKey || event.metaKey;  
  if (index !== -1 || !isHotKey)  {
    return;
  }

  event.preventDefault();

  ```

File: packages/teleport/src/console/useOnAltKeyDown.ts [hunk 4]
```diff
@@ -0,0 +1,59 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+import ConsoleContext from './consoleContext';
+
+// DIGITS defines event.key values on keyboard from 1 - 9.
+export const DIGITS = ['1', '2', '3', '4', '5', '6', '7', '8', '9'];
```
pierrebeaucamp, line 21: I'm slightly leaning towards making this a function instead:

```ts
const isNumeric(event: KeyboardEvent): bool =>
  ['1', '2', '3', '4', '5', '6', '7', '8', '9'].includes(event.key);
```

packages/teleport/src/console/useOnAltKeyDown.ts
pierrebeaucamp: I actually prefer `useOnAltKeyDown`. "HotKey" is a pretty ambiguous term introduced by Windows. "ModifierKey" would also work.

packages/teleport/src/console/useOnAltKeyDown.ts
pierrebeaucamp, line 39: I wouldn't use meta key here. "Meta+Number" switches between browser tabs, it's similar to "Ctrl+Number" on Linux. Browsers already map Mac's option key to `altKey` (see https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/altKey), so I would just use `altKey` for both systems

alex-kovoy: > "HotKey" is a pretty ambiguous term introduced by Windows. "ModifierKey" would also work.

This function is about quick navigation on some "hotKey" command. HotKey these days mean a combination of keys providing quick access to a particular function within a program. Calling this function (which handles quick navigation) as ModifiyKey or AltKey feels like a leaked technical implementation.

packages/teleport/src/lib/term/terminal.js
```diff
@@ -18,6 +18,7 @@ import 'xterm/dist/xterm.css';
 import { debounce, isInteger } from 'lodash';
 import Logger from 'shared/libs/logger';
 import { TermEventEnum } from './enums';
+import { DIGITS } from 'teleport/console/useOnAltKeyDown';
```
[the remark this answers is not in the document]
kimlisa: removed

packages/teleport/src/console/useOnAltKeyDown.ts
kimlisa: I didn't use this approach, b/c I need the index for getting the document in the docs array.

packages/teleport/src/console/useOnAltKeyDown.ts
kimlisa: revised

alex-kovoy: We can also name it as `useKeyboardNav`

packages/teleport/src/console/useOnAltKeyDown.ts
kimlisa: I figured we will add more hot key events later, so I renamed to `useHotKeyNavigation`

packages/teleport/src/console/useOnAltKeyDown.ts
kimlisa: removed meta key

packages/teleport/src/console/useOnAltKeyDown.ts
kimlisa: i used `includes` on checking if its numeric though

packages/design/src/utils/platform.ts
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * isMac determines if a user is using a mac platform.
+ */
+export const isMac = () =>
+  ['Macintosh', 'MacIntel', 'MacPPC', 'Mac68K'].includes(navigator.platform);
```
kimlisa, line 21: Got these strings from: https://github.com/xtermjs/xterm.js/blob/master/src/common/Platform.ts#L26

packages/teleport/src/lib/term/terminal.js
```diff
@@ -54,6 +55,19 @@ class TtyTerminal {
       cursorBlink: false,
     });
 
+    // TODO deprecated, use attachCustomKeyEventHandler when we upgrade xterm
+    this.term.attachCustomKeydownHandler(event => {
```
alex-kovoy, line 59: Actually, we can keep this code closer to the feature. We can move it [here](https://github.com/gravitational/webapps/blob/master/packages/teleport/src/console/components/DocumentSsh/Terminal/Terminal.tsx#L29) and subscribe to xterm events from there.

Then we can add a helper method to useKeyboardNav.ts and export it.

```js

import { getPlatform } from 'design/theme/utils';

export function getMappedAction(event) {  
  const index = ['1', '2', '3', '4', '5', '6', '7', '8', '9'].indexOf(
    event.key
  );

  const { isMac } = getPlatform();
  const isModifierKey = (isMac && event.ctrlKey) || event.altKey;
  if (!isModifierKey || index == -1) {
    return {};
  }

  return {
    tabSwitch: {
      index
    },
  };
}
```

for attachCustomKeydownHandler:

```js
const action = getMappedAction(event)
if (!action.tabSwitch){
   return false;
}
```

and for useKeyboardNav
```js
const action = getMappedAction(event)
if (!action.tabSwitch){
   return;   
}

event.preventDefault();
const doc = ctx.getDocuments()[action.tabSwitch.index + 1];
```

kimlisa: refactored

PR #91, opened 2020-05-05
Clean up active sessions list

kimlisa:
resolves https://github.com/gravitational/teleport/issues/3544

#### Description
- Removed address, type column
- Renamed `created` to `started (utc)`
- Added duration, session ID columns
- Modified Node and Users to display name and address without port
- Modified action cell to use a dotted button dropdown menu
- Updated session.story snapshot

#### Screen
Peek 2020-05-06 10-55

### DIFF
--- a/packages/teleport/src/services/ssh/ssh.ts
+++ b/packages/teleport/src/services/ssh/ssh.ts
@@ -44,7 +44,7 @@ const service = {
   fetchSession({ clusterId, sid }: FetchSessionParams) {
     return api
       .get(cfg.getTerminalSessionUrl({ sid, clusterId }))
-      .then(response => makeSession(response));
+      .then(makeSession);
   },
 
   fetchSessions(clusterId = cfg.clusterName) {
--- a/packages/teleport/src/services/ssh/ssh.ts
+++ b/packages/teleport/src/services/ssh/ssh.ts
@@ -42,20 +42,11 @@ const service = {
   },
 
   fetchSession({ clusterId, sid }: FetchSessionParams) {
-    // To retrieve a hostname, we are fetching all cluster nodes.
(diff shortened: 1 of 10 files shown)


kimlisa:
@benarent I didn't add the `options` (the actions column) because currently we only have `join` session feature implemented, so this column was left unchanged.

benarent:
@kimlisa The reason for adding the options, was to make joining not so easy. It's been a common request that the join button is too easy to access and join, then they leave and sometimes kill the other session. I think it's Ok to add only one option in the drop down. Unless @alexwolfe  has any other input.

alex-kovoy:
@benarent @alexwolfe  

"Option" button looks random because we use  image 
in all other places. So if you want to use action button with only 1 action item then lets be consistent.

benarent:
Sorry busy day, but the `...` for now and then fixing https://github.com/gravitational/teleport/issues/3671 seems like a good path forwards.

kimlisa:
@benarent I changed to the three dotted drop down menu and I also changed the title

packages/teleport/src/services/ssh/ssh.ts
```diff
@@ -42,12 +42,9 @@ const service = {
...
   fetchSession({ clusterId, sid }: FetchSessionParams) {
-    return Promise.all([
-      api.get(cfg.getTerminalSessionUrl({ sid, clusterId })),
-    ]).then(response => {
-      const [sessionJson] = response;
-      return makeSession(sessionJson);
-    });
+    return api
+      .get(cfg.getTerminalSessionUrl({ sid, clusterId }))
+      .then(response => makeSession(response));
```
alex-kovoy, line 47:
```
.then(response => makeSession(response));
```

can be replaced with .then(makeSession);

Issue #92, opened 2020-05-13
Copy/paste issue in Teleport Web Console


travelton:
### Issue:
When using the Teleport Web Console, and attempting to copy/paste a string that exceeds the length of the console window, xterm.js inserts a line break between the wrapped lines. This breaks any output that may be a URL. You need to copy and paste the output in to a text editor, remove the line breaks, then copy/paste in your browser. This issue does not occur on a native terminal. 

### Repro Steps:
```
$ echo 'http://www.google.com/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa' > test.txt

$ cat test.txt
http://www.google.com/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
aaaaaa
```
### Potential Fix:
This issue was fixed in xterm.js v2.8.0 via https://github.com/xtermjs/xterm.js/pull/693. It [appears](https://github.com/gravitational/webapps/blob/master/packages/teleport/package.json#L22) we are currently on version v2.4.0.

benarent:
It looks like X-Term has improved a lot since the 2.x times, with current being 4.x. Any reason to not just bump version?  We should at least consider upgrading to 2.9.2 for extra bug fixes.

alex-kovoy:
@benarent, never versions have some API changes which need to be addressed on our side. 
For example, newly versions have some issues when unwinding the session in the player. But I agree that it's about time to update it. If anything, we can use 2 versions: one for session player and another for active sessions.

As for this bug fix, we can do the following:

[...]

PR #95, opened 2020-06-03
Update xterm to 2.8.1

kimlisa:
fixes https://github.com/gravitational/webapps/issues/92

#### Description
Update xterm to the minimum version (`2.8.1`), that fixes the wrap problem when copying and pasting

- Did some light manual testing on the terminal for possible udpate bug: `ls`, `exit`, session replay, alt switch tabs

#### Manual Test
no-wrap-copy


travelton:
Nice!!

Issue #97, opened 2020-06-04
Do not prompt a user with a confirmation dialog on logout redirect


alex-kovoy:
We automatically redirect a user to the login screen when session expires. When it happens, the `useOnExitConfirmation` hook should not be asking a user for confirmation.

PR #100, opened 2020-06-11
Check for expired session before resources unload (redirect)

kimlisa:
fixes https://github.com/gravitational/webapps/issues/97

#### Description
- Check for expired session before prompting confirmation
- Test expired session don't prompt


kimlisa:
retest this please

PR #102, opened 2020-06-13
Tweak styling for topbar auto scrollX and text alignment

kimlisa:
#### Description
- Cluster info and support cluster info has their text aligned
- Put more spacing between number of users joined in session
- Put space between cluster and quick launch component in console
- Set Topbar to `overflowX` to scroll 
- Updated snapshots

#### Related PR
https://github.com/gravitational/webapps.e/pull/11

### DIFF
--- a/packages/teleport/src/cluster/components/Cluster.tsx
+++ b/packages/teleport/src/cluster/components/Cluster.tsx
@@ -18,7 +18,7 @@ import React from 'react';
 import styled from 'styled-components';
 import { Redirect, Switch, Route } from 'teleport/components/Router';
 import { useAttempt } from 'shared/hooks';
-import { Indicator } from 'design';
+import { Indicator, Flex } from 'design';
 import { Failed } from 'design/CardError';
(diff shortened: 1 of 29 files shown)


packages/teleport/src/cluster/components/Cluster.tsx
```diff
@@ -71,14 +71,16 @@ export default function Cluster() {
...
         <SideNav />
-        <Switch>
-          <Redirect
-            exact
-            from={cfg.routes.cluster}
-            to={cfg.routes.clusterNodes}
-          />
-          {$features}
-        </Switch>
+        <Flex width="100%" style={{ overflowX: 'auto' }}>
```
alex-kovoy, line 74: Lets remove this part as it feels unnecessary. I will take a look at the layout myself and put it back if needed.

kimlisa: I reverted the styles back to original for both dashboard and cluster layout

PR #103, opened 2020-06-15
Fix assortment of user issues

kimlisa:
fixes https://github.com/gravitational/teleport/issues/3827
fixes https://github.com/gravitational/teleport/issues/3836
fixes https://github.com/gravitational/teleport/issues/3835
fixes https://github.com/gravitational/teleport/issues/3837

#### Description
- Set sorting to ASC as default, fix Nodelist sort to be ASC on initial render
- On exit confirmation for console, `leave session` for parties > 1, else `terminate session`
- Prevent incrementing page numbers when there are no data
- Provide description for `trusted_cluster_token.created` event

#### Screenshots
Sort ASC default:
Screenshot from 2020-06-15 14-26-56

Leave (user > 1) and Terminate (user == 1):
screenshot_2
screenshot_3

Prevent incrementing page numbers when there are no data:
Screenshot from 2020-06-15 18-18-11

Provide description for trusted_cluster_token.created:
Screenshot from 2020-06-15 20-13-58

### DIFF
--- a/packages/teleport/src/components/NodeList/NodeList.tsx
+++ b/packages/teleport/src/components/NodeList/NodeList.tsx
@@ -35,7 +35,7 @@ function NodeList(props: NodeListProps) {
   const { nodes = [], onLoginMenuOpen, onLoginSelect, pageSize = 100 } = props;
   const [searchValue, setSearchValue] = React.useState('');
   const [sortDir, setSortDir] = React.useState<Record<string, string>>({
-    hostname: SortTypes.ASC,
+    hostname: SortTypes.DESC,
   });
 
   function sortAndFilter(search) {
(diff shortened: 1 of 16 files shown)


packages/teleport/src/components/NodeList/NodeList.tsx
```diff
@@ -48,7 +48,7 @@ function NodeList(props: NodeListProps) {
 
     const columnKey = Object.getOwnPropertyNames(sortDir)[0];
     const sorted = sortBy(filtered, columnKey);
-    if (sortDir[columnKey] === SortTypes.ASC) {
+    if (sortDir[columnKey] !== SortTypes.ASC) {
```
alex-kovoy, line 51: Lets just change the sorting order of the hostname column without changing the default sorting in the Table component because that would impact everything else that uses that table. Just change 

`hostname: SortTypes.ASC` to `hostname: SortTypes.DESC`

```tsx
const [sortDir, setSortDir] = React.useState<Record<string, string>>({
    hostname: SortTypes.DESC,
  });
```

PR #105, opened 2020-06-17
Replace session button with quicklauncher

kimlisa:
resolves https://github.com/gravitational/teleport/issues/3852

#### Screenshot
Screenshot from 2020-06-16 17-08-48


packages/teleport/src/cluster/components/Nodes/Nodes.tsx
```diff
@@ -16,17 +16,18 @@ limitations under the License.
...
   FeatureBox,
   FeatureHeader,
   FeatureHeaderTitle,
 } from 'teleport/components/Layout';
 import { useTeleport } from 'teleport/teleportContextProvider';
-import { useParams } from 'teleport/components/Router';
 import cfg from 'teleport/config';
 import NodeList from 'teleport/components/NodeList';
 import useClusterNodes from './useClusterNodes';
+import QuickLaunch from 'teleport/console/components/DocumentNodes/QuickLaunch';
```
kimlisa, line 29: maybe we can move the `QuickLaunch` location to `teleport/console/components`

kimlisa: relocated to `teleport/Components`

PR #106, opened 2020-06-18
Account for empty hostname and server addrs

kimlisa:
#### Description
Place checks for empty hostname/addr when a older teleport version is used (which doesn't set/retr hostname/addr)


packages/teleport/src/cluster/components/Sessions/SessionList/sessionTableCells.test..tsx
```diff
@@ -0,0 +1,87 @@
...
+  serverId: 'server-id',
+  hostname: undefined,
+  clusterId: undefined,
+  parties: undefined,
+  addr: undefined,
+};
+const data = [session];
+const tableRow = document.createElement('tr');
+
+test('empty hostname and addr check in NodeCell', () => {
```
alex-kovoy, line 38: I think these unit-tests are too much for such a basic thing. 

1. Create 1 data sample with different cases (3 is enough)
2. Render the table and create a snapshot.

PR #107, opened 2020-06-18
Replace '+' symbols with empty spaces

kimlisa:
part of https://github.com/gravitational/teleport/issues/3774

#### Description
- Use native URLSearchParams to handle escape symbols
- Add unit test

#### Screenshot
Screenshot from 2020-06-17 21-47-43

### DIFF
--- a/packages/teleport/src/services/history.js
+++ b/packages/teleport/src/services/history.js
@@ -118,15 +118,7 @@ const match = url => route => {
 export default history;
 
 export function getUrlParameter(name, path) {
-  path = path || window.location.search;
-  const query = path.substring(1);
-  const vars = query.split('&');
-  for (var i = 0; i < vars.length; i++) {
-    var pair = vars[i].split('=');
-    if (decodeURIComponent(pair[0]) == name) {
-      return decodeURIComponent(pair[1]).replace(/\+/g, ' ');
-    }
-  }
-
-  return '';
+  const params = new URLSearchParams(path || window.location.search);
+  const value = params.get(name);
+  return value || '';
 }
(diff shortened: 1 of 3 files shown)


packages/teleport/src/services/history.js
```diff
@@ -124,7 +124,7 @@ export function getUrlParameter(name, path) {
   for (var i = 0; i < vars.length; i++) {
     var pair = vars[i].split('=');
     if (decodeURIComponent(pair[0]) == name) {
-      return decodeURIComponent(pair[1]);
+      return decodeURIComponent(pair[1]).replace(/\+/g, ' ');
```
alex-kovoy, line 127: decodeURIComponent should correctly decode the text. If it has + then it should preserve it. The question is why is there "+" in the first place?

kimlisa: as disucssed in slack, replaced with URLSearchParams

packages/teleport/src/services/history.js
```diff
@@ -118,15 +118,7 @@ const match = url => route => {
...
-  const vars = query.split('&');
-  for (var i = 0; i < vars.length; i++) {
-    var pair = vars[i].split('=');
-    if (decodeURIComponent(pair[0]) == name) {
-      return decodeURIComponent(pair[1]);
-    }
-  }
-
-  return '';
+  const params = new URLSearchParams(path || window.location.search);
```
alex-kovoy, line 121: Lets remove `window.location.search` and good to go!

kimlisa: revised

PR #119, opened 2020-07-08
Update for cluster view styles

alexwolfe:
@alex-kovoy I added title tags and updated the styles for the nodes link. See image below: 
image

### DIFF
--- a/packages/teleport/src/dashboard/components/Clusters/ClusterList/ClusterList.tsx
+++ b/packages/teleport/src/dashboard/components/Clusters/ClusterList/ClusterList.tsx
@@ -173,9 +173,9 @@ export function NodeCountCell(props) {
   return (
     <Cell>
       {nodeCount && (
-        <StyledLink target="_blank" as="a" href={terminalURL}>
-          {nodeCount}
-        </StyledLink>
+        <StyledConsoleLink title="Launch New Session" target="_blank" as="a" href={terminalURL}>
+          <Cli /> {nodeCount}
+        </StyledConsoleLink>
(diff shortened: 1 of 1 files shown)


benarent:
Thanks for the quick turn around on this @alexwolfe .

My thoughts are.. 
1. How does this look in proxy mode, where we currently report 0 nodes. 
image

2. Similar note, for K8s users, but we might want to roll this into  https://docs.google.com/document/d/1cS6J2d_xBcJMWPewWPjdOZyrDHLKdqmgF1QLLo4E1YI/edit

3. What happens with very high counts, do we show exact counts or count it up / down. e.g. 8, 120, 1k, 3k, 10k nodes?

alex-kovoy:
@benarent , these are good outstanding questions which we can address in 4.4

But do we want to add these sight improvements to the node count hyperlinks in 4.3?

benarent:
I think If we do this; we should at least address the zero issue. If zero; we can just make this look more like a button ( since this design allows it) ; vs giving the illusion of zero nodes but there are many openssh nodes

alex-kovoy:
@benarent @alexwolfe 

> giving the illusion of zero nodes but there are many openssh nodes

How is it different from showing "1 node" where there are still many openssh nodes in the wild? So I guess the confusing part is with the terminal button coupled with the teleport node count.

What if we open the terminal on the cluster name click? And to manage a cluster, a user would use the options dropdown by clicking on "Manage this cluster" option?

alex-kovoy:
has been merged

PR #122, opened 2020-07-11
Add alternate way to set clusterID

kimlisa:
fixes https://github.com/gravitational/teleport/issues/4012

#### Description
Teleport node's running < v4.3 does not set clusterID for session info, which is needed to form session URL. This PR handles this incompatibility.

#### Manual Testing
Reproduced issue and tested with Proxy/Auth v4.3 with Node v4.2.11

### DIFF
--- a/packages/teleport/src/services/ssh/makeSession.ts
+++ b/packages/teleport/src/services/ssh/makeSession.ts
@@ -17,9 +17,10 @@ limitations under the License.
 import moment from 'moment';
 import { map } from 'lodash';
 import { Session, Participant } from './types';
+import cfg from 'teleport/config';
 
 export default function makeSession(json): Session {
-  const clusterId = json.cluster_name;
+  const clusterId = json.cluster_name || getClusterIDFromURL();
   const created = new Date(json.created);
   const duration = moment(new Date()).diff(created);
   const durationText = moment.duration(duration).humanize();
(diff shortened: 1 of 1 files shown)


packages/teleport/src/services/ssh/makeSession.ts
```diff
@@ -53,4 +54,13 @@ export function makeParticipant(json): Participant {
   };
 }
 
+/**
+ * getClusterIDFromURL solves empty clusterID result with Teleport
+ * Node's using versions < 4.3 (clusterId is not set for older versions).
+ */
+const getClusterIDFromURL = () => {
```
alex-kovoy, line 61: We cannot reach out to the browser URL from any place in the code and manually parse parameters. Normally you access the URL from UI layer (React components + React Router). If you look at how makeSession is invoke you will see that clusterID is coming from UI components where  it already uses the clusterID to make a call to the web API. 

That said, I now think that it's better to handle this use-case on the WebAPI layer. That would be a better place for the temporary fix as this is where missing id is coming from (this is not UI bug).

kimlisa: created PR to apply patch on the web apiserver instead: https://github.com/gravitational/teleport/pull/4027

kimlisa:
created PR to apply patch on the web apiserver instead: gravitational/teleport#4027

PR #123, opened 2020-07-14
Lisa/minor fixes

kimlisa:
fixes https://github.com/gravitational/teleport/issues/4008
fixes https://github.com/gravitational/teleport/issues/4013
fixes https://github.com/gravitational/teleport/issues/4044

#### Description
- Open new tab when clicking on active sessions icon
- Simplify QuickLaunch regex to just check for white spaces
- Remove using ServerId as fallback in place of empty hostname when displaying inform (this is now handled from webapi server)
- Use serverID's to start session instead of hostname

#### Dependent PR
https://github.com/gravitational/teleport/pull/4027

### DIFF
--- a/packages/teleport/src/components/QuickLaunch/QuickLaunch.jsx
+++ b/packages/teleport/src/components/QuickLaunch/QuickLaunch.jsx
@@ -31,11 +31,11 @@ export default function FieldInputSsh({
   function onKeyPress(e) {
     const value = e.target.value;
     if ((e.key === 'Enter' || e.type === 'click') && value) {
-      const valid = check(value);
-      setHasError(!valid);
-      if (valid) {
-        const [login, serverId] = value.split('@');
-        onPress(login, serverId);
+      const match = check(value);
+      setHasError(!match);
+      if (match) {
+        const { username, host } = match.groups;
+        onPress(username, host);
       }
     } else {
       setHasError(false);
@@ -63,10 +63,14 @@ export default function FieldInputSsh({
   );
 }
 
-const SSH_STR_REGEX = /(^(\w+-?\w+)+@(\S+)$)/;
+// SSH_STR_REGEX is a modified regex from teleport's lib/sshutils/scp/scp.go.
+// Captures two named groups: username and host.
+const SSH_STR_REGEX = /(?:(?<username>[-.\w@]+)@)(?<host>[-.\w]+)$/;
 const check = value => {
-  const match = SSH_STR_REGEX.exec(value);
-  return match !== null;
+  const hasWhiteSpace = /\s/.test(value);
+  if (!hasWhiteSpace) {
+    return SSH_STR_REGEX.exec(value);
+  }
 };
 
 const StyledInput = styled(Input)(
--- a/packages/teleport/src/components/QuickLaunch/QuickLaunch.jsx
+++ b/packages/teleport/src/components/QuickLaunch/QuickLaunch.jsx
@@ -63,14 +63,11 @@ export default function FieldInputSsh({
   );
 }
 
-// SSH_STR_REGEX is a modified regex from teleport's lib/sshutils/scp/scp.go.
-// Captures two named groups: username and host.
-const SSH_STR_REGEX = /(?:(?<username>[-.\w@]+)@)(?<host>[-.\w]+)$/;
+// Checks for spaces between chars, and
+// captures two named groups: username and host.
+const SSH_STR_REGEX = /^(?:(?<username>[^\s]+)@)(?<host>[^\s]+)$/;
 const check = value => {
-  const hasWhiteSpace = /\s/.test(value);
-  if (!hasWhiteSpace) {
-    return SSH_STR_REGEX.exec(value);
-  }
+  return SSH_STR_REGEX.exec(value.trim());
 };
 
 const StyledInput = styled(Input)(
--- a/packages/teleport/src/console/components/DocumentSsh/useSshSession.ts
+++ b/packages/teleport/src/console/components/DocumentSsh/useSshSession.ts
@@ -91,10 +91,12 @@ function handleTtyConnect(
   session: Session,
   docId: number
 ) {
-  const { hostname, login, sid, clusterId } = session;
+  const { hostname, login, sid, serverId, clusterId } = session;
   const url = cfg.getSshSessionRoute({ sid, clusterId });
   ctx.updateSshDocument(docId, {
-    title: `${login}@${hostname}`,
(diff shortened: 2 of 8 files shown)


packages/teleport/src/console/components/DocumentSsh/useSshSession.ts
```diff
@@ -91,10 +91,12 @@ function handleTtyConnect(
...
   docId: number
 ) {
-  const { hostname, login, sid, clusterId } = session;
+  const { hostname, login, sid, serverId, clusterId } = session;
   const url = cfg.getSshSessionRoute({ sid, clusterId });
   ctx.updateSshDocument(docId, {
-    title: `${login}@${hostname}`,
+    // DELETE IN 5.0.0
+    // serverId is necessary for Teleport nodes <4.3
+    title: `${login}@${hostname || serverId}`,
```
kimlisa, line 99: Same issue with missing `clusterName`, Teleport Nodes using < v4.3 has empty `hostname`.
We used `serverId` as a replacement for hostname in active sessions list, so I did the same for tabs name.
server-hostname-issue
server-hostname-issue-2

packages/teleport/src/components/QuickLaunch/QuickLaunch.jsx
```diff
@@ -63,10 +63,14 @@ export default function FieldInputSsh({
...
+// SSH_STR_REGEX is a modified regex from teleport's lib/sshutils/scp/scp.go.
+// Captures two named groups: username and host.
+const SSH_STR_REGEX = /(?:(?<username>[-.\w@]+)@)(?<host>[-.\w]+)$/;
 const check = value => {
-  const match = SSH_STR_REGEX.exec(value);
-  return match !== null;
+  const hasWhiteSpace = /\s/.test(value);
+  if (!hasWhiteSpace) {
+    return SSH_STR_REGEX.exec(value);
+  }
```
kimlisa, line 73: I couldn't keep the validation simple because mostly all special characters made the terminal look like it was hanging (but it is a `400` error), I thought it'd be simpler to prevent this error by regex: 
Screenshot from 2020-07-13 21-15-33

The regex was modeled after this: https://github.com/gravitational/teleport/blob/master/lib/sshutils/scp/scp.go#L652. @benarent mentioned this PR: https://github.com/gravitational/teleport/pull/4002, something like `root@domain@127.0.0.1` will correctly resolve to

```
username: root@domain
host: 127.0.0.1
```

Dashes and periods seems to be safe and gives correct `disconnect` errors when it can't connect:
Screenshot from 2020-07-13 22-58-52

packages/teleport/src/console/components/DocumentSsh/useSshSession.ts
```diff
@@ -91,10 +91,12 @@ function handleTtyConnect(
   session: Session,
   docId: number
 ) {
-  const { hostname, login, sid, clusterId } = session;
+  const { hostname, login, sid, serverId, clusterId } = session;
   const url = cfg.getSshSessionRoute({ sid, clusterId });
   ctx.updateSshDocument(docId, {
-    title: `${login}@${hostname}`,
+    // DELETE IN 5.0.0
```
alex-kovoy, line 97: Are we sure that this is an upgrade issue similarly to clusterId and not one of the valid use cases (when hostname is not-specified) ? If not, then lets remove this comment.

packages/teleport/src/components/QuickLaunch/QuickLaunch.jsx
alex-kovoy: > Dashes and periods seems to be safe and gives correct disconnect errors when it can't connect:

This is not up to the dashes and input field to control the correctness of the server disconnect errors. Could you check why these errors are not handled correctly on the webapi level?

alex-kovoy: > I couldn't keep the validation simple because mostly all special characters made the terminal look like it was hanging (but it is a 400 error),

Just to clarify, does it eventually show an error? (no such host, etc) ?

kimlisa: > Just to clarify, does it eventually show an error? (no such host, etc) ?

No, it just seems to send out a close event right away. I will look into the error

packages/teleport/src/console/components/DocumentSsh/useSshSession.ts
kimlisa: revised to be handled in webapi server: https://github.com/gravitational/teleport/pull/4027

packages/teleport/src/components/QuickLaunch/QuickLaunch.jsx
kimlisa: note: problem was caused by url's not matching https://github.com/gravitational/webapps/blob/master/packages/teleport/src/console/stores/storeDocs.ts#L84, fixed by encoding URL

packages/teleport/src/components/QuickLaunch/QuickLaunch.jsx
```diff
@@ -63,10 +63,14 @@ export default function FieldInputSsh({
...
 }
 
-const SSH_STR_REGEX = /(^(\w+-?\w+)+@(\S+)$)/;
+// SSH_STR_REGEX is a modified regex from teleport's lib/sshutils/scp/scp.go.
+// Captures two named groups: username and host.
+const SSH_STR_REGEX = /(?:(?<username>[-.\w@]+)@)(?<host>[-.\w]+)$/;
 const check = value => {
-  const match = SSH_STR_REGEX.exec(value);
-  return match !== null;
+  const hasWhiteSpace = /\s/.test(value);
```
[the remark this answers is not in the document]
kimlisa: removed

packages/teleport/src/components/QuickLaunch/QuickLaunch.jsx
```diff
@@ -63,10 +63,14 @@ export default function FieldInputSsh({
   );
 }
 
-const SSH_STR_REGEX = /(^(\w+-?\w+)+@(\S+)$)/;
+// SSH_STR_REGEX is a modified regex from teleport's lib/sshutils/scp/scp.go.
```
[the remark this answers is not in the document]
kimlisa: removed

packages/teleport/src/console/components/DocumentSsh/useSshSession.ts
kimlisa: reverted, handled by webapi server: gravitational/teleport#4027

PR #127, opened 2020-07-21
Work around for close events on ssh errors in Client

kimlisa:
work around for: https://github.com/gravitational/teleport/issues/4025

#### Description
Work around for now, to handle closing tabs for exec command errors.

#### Screenshot
Screenshot from 2020-07-20 23-39-38
Screenshot from 2020-07-20 23-39-47

### DIFF
--- a/packages/teleport/src/console/components/DocumentSsh/useSshSession.ts
+++ b/packages/teleport/src/console/components/DocumentSsh/useSshSession.ts
@@ -37,11 +37,20 @@ export default function useSshSession(doc: DocumentSsh) {
     function initTty(session: Session) {
       const tty = ctx.createTty(session);
 
-      // Latest stores the latest terminal payload to
-      // check for "Failed to launch" substring. This is to prevent
-      // terminal tabs from closing and hiding terminal errors from users.
-      // Currently, the backend sends a close event for exec command errors and
-      // successful exit commands.
+      /**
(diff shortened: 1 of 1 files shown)


packages/teleport/src/console/components/DocumentSsh/useSshSession.ts
```diff
@@ -37,8 +37,23 @@ export default function useSshSession(doc: DocumentSsh) {
     function initTty(session: Session) {
       const tty = ctx.createTty(session);
 
+      // Latest stores the latest terminal payload to
```
alex-kovoy, line 40:
```js
/*
Currently Teleport does not handle all errors that can occur during SSH session creation. 
It can mistakenly create a session (recording) and write SSH initialization errors 
directly into the stream as if these errors happened within an actual SSH session. It then 
proceeds with emitting the rest of audit events as if SSH session successfully started and 
ended even though it never did. 

Since we are closing the tab automatically on "end" event, there is no way a user can see 
an actual error why SSH session failed to start. 

In here we are trying to detect this scenario by looking at the last received payload before 
closing the terminal tab. If it's empty or it has special "keywords" in it then do not close the 
tab automatically on "close" event and let a user see the error.
*/
```

kimlisa: revised, and yes i rebuilt teleport from master branch to test it out

PR #139, opened 2020-08-19
Pass hostname when creating new session

kimlisa:
fixes https://github.com/gravitational/teleport/issues/4143

#### Description
When creating new session, set the hostname, so that the dummy session returned contains the hostname
- refactored how we pass around data related to c/n to a node
- removed unused api routes
- manually tested


kimlisa:
closing b/c it does not solve the root problem (fix in backend)

PR #141, opened 2020-08-20
Remove unnecessary css

kimlisa:
fixes https://github.com/gravitational/teleport/issues/4235

#### Description
- `flex: 0 0` was canceling `height: 32px`, so the height never mattered here (regardless of browser), so removed both
- `flex: 0 0` the third value is flex-basis which is set to 0. In safari this makes the ActionBar fall/collapse behind the xterm container.

I think this is what we wanted to achieve: https://css-tricks.com/almanac/properties/f/flex/#flex-none
But it isn't necessary b/c the icons sizes are fixed.

#### Testing
manually tested the icons showed in safari/firefox/chrome in mac/linux

### DIFF
--- a/packages/teleport/src/console/components/DocumentSsh/ActionBar/ActionBar.jsx
+++ b/packages/teleport/src/console/components/DocumentSsh/ActionBar/ActionBar.jsx
@@ -27,7 +27,7 @@ export default function ActionBar({
 }) {
   const isScpDisabled = isDownloadOpen || isUploadOpen || !isConnected;
   return (
-    <Flex alignItems="center">
+    <Flex flex="none" alignItems="center" height="24px">
       <ButtonIcon
         disabled={isScpDisabled}
         size={0}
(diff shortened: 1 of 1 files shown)


packages/teleport/src/console/components/DocumentSsh/ActionBar/ActionBar.jsx
```diff
@@ -27,7 +27,7 @@ export default function ActionBar({
 }) {
   const isScpDisabled = isDownloadOpen || isUploadOpen || !isConnected;
   return (
-    <Flex flex="0 0" alignItems="center" height="32px">
+    <Flex alignItems="center">
```
alex-kovoy, line 30: It intent was to have this bar always with fixed dimensions, so its height or width does not depend in rendered inner content.  This is to make sure that the size of terminal container does not change when we calculate its height to create a session. Otherwise, if action bar height changes, the scrollbars will appear in the terminal. 

In other words, the size of the action bar should be always fixed even when it's empty. Will it be still the case?

kimlisa: I see. No it doesn't. I added back the flexing to prevent growth/shrinkage and height (since the button was height 24px, I went with the same value)

PR #142, opened 2020-08-20
Prevent closing when there is an error with deleting user

kimlisa:
part of https://github.com/gravitational/teleport/issues/4240

#### Description
- [updated] Manually set attempts to prevent error about `updating unmounted component`, and to handle (prevent) closing dialog when there are errors

#### Related PR
https://github.com/gravitational/teleport.e/pull/157

#### Screenshot
Screenshot from 2020-08-20 16-49-09

### DIFF
--- a/packages/teleport/src/dashboard/People/Users/UserDelete/UserDelete.tsx
+++ b/packages/teleport/src/dashboard/People/Users/UserDelete/UserDelete.tsx
@@ -50,7 +50,7 @@ export function UserDelete({
           <Alert kind="danger" children={attempt.statusText} />
         )}
         <Text mb={4} mt={1}>
-          You are about to delete a user
+          You are about to delete user
           <Text bold as="span">
             {` ${username}`}
           </Text>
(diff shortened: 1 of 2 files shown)


packages/teleport/src/dashboard/People/Users/UserDelete/UserDelete.tsx
```diff
@@ -74,8 +74,8 @@ export function UserDelete({
 function useDialog(props: Props) {
   const { attempt, run } = useAttemptNext();
   function onDelete() {
-    return run(() => props.onDelete(props.username)).then(() =>
-      props.onClose()
+    return run(() =>
+      props.onDelete(props.username).then(() => props.onClose())
```
alex-kovoy, line 78: it will through warnings in the console about updating unmounted component. So in this case you need to use `setAttmpt({ status: 'XX'})` explicitly. See how it's done in UserAddEdit hook.

kimlisa: revised, and tested in UI

PR #144, opened 2020-08-22
Display SSO users

kimlisa:
resolves https://github.com/gravitational/cloud/issues/93#issuecomment-678394185

#### Description
- Display connector under `Identity Provider`
- Hide user action options for non `local` users
- Set/Define new field in User struct `idp`

#### Related PR
https://github.com/gravitational/teleport.e/pull/158 (merge first)

#### Test
Screenshot from 2020-08-25 16-04-13


benarent:
Let's keep Teleport Local User, but try `Type` for now. We can come back later and review, I don't want to block this PR

PR #162, opened 2020-09-28
Add Application Launcher (AAP)

### LINKED ISSUE #164: Teleport V5: Implement AAP Launch logic
Teleport V5: Implement AAP Launch logic

alex-kovoy:
This PR adds a route that creates a new session for aap application and then redirects a user to this URL.

closes #164


packages/teleport/src/AppLauncher/AppLauncher.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export function AppLauncher(props: ReturnType<typeof useAppLauncher>) {
+  if (props.status === 'failed') {
+    return <AccessDenied message={props.statusText} />;
+  }
+
+  return (
+    <Flex height="180px" justifyContent="center" alignItems="center" flex="1">
+      <Indicator />
```
kimlisa, line 34: i think we should put a message saying "redirecting"? i ran these in storybook and the blank states was not very helpful. maybe there is a polish coming up.

packages/teleport/src/Apps/AppList/AppList.tsx
```diff
@@ -52,7 +55,7 @@ function Item({ app, ...rest }) {
         tabIndex={-1}
         target="_blank"
         color="text.primary"
-        href={app.addr}
+        href={app.activationUrl}
```
kimlisa, line 58: `activationUrl` sounds kind of misleading, that this app needs to be activated/verfified. Maybe we can call it `launchUrl`?

kimlisa (approved):

PR #166, opened 2020-09-30
[teleport] Handle null value response from retrieving audit logs

kimlisa:
fixes https://github.com/gravitational/teleport/issues/4393

#### Description
- Check if response is null before running map


packages/teleport/src/cluster/components/Audit/Audit.story.test.tsx
```diff
@@ -8,3 +8,10 @@ test('overflow', async () => {
 
   expect(container.firstChild).toMatchSnapshot();
 });
+
+test('null response still renders table', async () => {
```
alex-kovoy, line 12: This is a good example when it's bad to use snapshots for testing. If you prefer to unit-test it, please unit-test the service itself as it should never return null.

kimlisa: revised

alex-kovoy (approved):

alex-kovoy:
Could you also update the 4.3 (push the master as it's currently the same) ?

And then update the webassets in `teleport master` and `teleport 4.3.x` ?

PR #171, opened 2020-10-09
Snapshot update

kimlisa:
#### Description

I forgot to update my snapshot (forgot to run yarn test) from my last commit from https://github.com/gravitational/webapps/pull/165/commits/4069fd41f2395edeaa00b11442d18d03436dc161.

Also placed a placeholder comment to check if user has access to list apps.


kimlisa:
being fixed by alexey

PR #191, opened 2020-11-12
Extra content for Application modal and new url to docs page.

benarent:
- [x] Added correct tctl command and `ttl` flag when manually creating tokens.
- [x] Added link to Application docs ( Will launch with 5.0 ) 
- [x] Changed some modal copy

### DIFF
--- a/packages/teleport/src/Apps/AddApp/Automatically/Automatically.tsx
+++ b/packages/teleport/src/Apps/AddApp/Automatically/Automatically.tsx
@@ -73,7 +73,7 @@ export default function Automatically(props: Props) {
             </Flex>
             {!cmd && (
               <Text mb="3">
-                Provide the name and URL of your application to generate our auto-installer script.
+                Teleport can automatically setup application access. Provide the name and URL of your application to generate our auto-installer script. This script will install the Teleport node service and provide secure access to your application.
               </Text>
             )}
             {attempt.status === 'failed' && (
--- a/packages/teleport/src/Apps/AddApp/Automatically/Automatically.tsx
+++ b/packages/teleport/src/Apps/AddApp/Automatically/Automatically.tsx
@@ -73,9 +73,7 @@ export default function Automatically(props: Props) {
             </Flex>
             {!cmd && (
               <Text mb="3">
-                Provide the name and URL of your application to generate a
-                script that will automatically install and configure the App
-                service on the server that can access your application.
+                Provide the name and URL of your application to generate our auto-installer script.
               </Text>
             )}
             {attempt.status === 'failed' && (
--- a/packages/teleport/src/Apps/AddApp/Automatically/Automatically.tsx
+++ b/packages/teleport/src/Apps/AddApp/Automatically/Automatically.tsx
@@ -73,7 +73,10 @@ export default function Automatically(props: Props) {
             </Flex>
             {!cmd && (
               <Text mb="3">
-                Teleport can automatically setup application access. Provide the name and URL of your application to generate our auto-installer script. This script will install the Teleport node service and provide secure access to your application.
+                Teleport can automatically setup application access. Provide
+                the name and URL of your application to generate our
+                auto-installer script. This script will install the Teleport
+                node service and provide secure access to your application.
               </Text>
             )}
             {attempt.status === 'failed' && (
--- a/packages/teleport/src/Apps/AddApp/Manually/Manually.tsx
+++ b/packages/teleport/src/Apps/AddApp/Manually/Manually.tsx
@@ -64,7 +64,7 @@ export default function Manually({ version, onClose }: Props) {
             <Text bold as="span">
               Step 4
             </Text>
-            {` - Install Teleport on target server, start Teleport with the following parameters`}
+            {` - Install Teleport node service, then start Teleport with the following parameters`}
             <TextSelectCopy
               mt="2"
               text={`teleport start --roles=app --token=<generated-app-join-token> --auth-server=${cfg.proxyCluster} `}
--- a/packages/teleport/src/Apps/AddApp/Manually/Manually.tsx
+++ b/packages/teleport/src/Apps/AddApp/Manually/Manually.tsx
@@ -57,17 +57,17 @@ export default function Manually({ version, onClose }: Props) {
             <Text bold as="span">
               Step 3
             </Text>
-            {' - Generate a join token'}
-            <TextSelectCopy mt="2" text="tctl tokens add --type=app" />
+            {' - Generate an application join token'}
+            <TextSelectCopy mt="2" text="tctl tokens add --type=app --ttl=1h" />
           </Box>
           <Box>
             <Text bold as="span">
               Step 4
             </Text>
-            {` - Install Teleport on target server, and start it with the following parameters`}
+            {` - Install Teleport on target server, start Teleport with the following parameters`}
             <TextSelectCopy
               mt="2"
-              text={`teleport start --roles=node --token=<generated-node-join-token> --auth-server=${cfg.proxyCluster} `}
+              text={`teleport start --roles=app --token=<generated-app-join-token> --auth-server=${cfg.proxyCluster} `}
             />
           </Box>
         </Box>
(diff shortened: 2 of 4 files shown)


packages/teleport/src/Apps/AddApp/Automatically/Automatically.tsx
```diff
@@ -73,9 +73,7 @@ export default function Automatically(props: Props) {
             </Flex>
             {!cmd && (
               <Text mb="3">
-                Provide the name and URL of your application to generate a
-                script that will automatically install and configure the App
-                service on the server that can access your application.
+                Provide the name and URL of your application to generate our auto-installer script.
```
alex-kovoy, line 76: @benarent dont we want to at least tell a user what that script is about to do? If anything to be installed on my machine/box I at least want to know _something_ about what is going to happen.

packages/teleport/src/Apps/AddApp/Manually/Manually.tsx
```diff
@@ -57,17 +57,17 @@ export default function Manually({ version, onClose }: Props) {
...
-            <TextSelectCopy mt="2" text="tctl tokens add --type=app" />
+            {' - Generate an application join token'}
+            <TextSelectCopy mt="2" text="tctl tokens add --type=app --ttl=1h" />
           </Box>
           <Box>
             <Text bold as="span">
               Step 4
             </Text>
-            {` - Install Teleport on target server, and start it with the following parameters`}
+            {` - Install Teleport on target server, start Teleport with the following parameters`}
```
alex-kovoy, line 67: Do we need to say anything about DNS and certificate?

packages/teleport/src/Apps/AddApp/Automatically/Automatically.tsx
alex-kovoy: @webvictim what would be your expectations from the user perspective? Will you be more comfortable knowing what this script is about to do?

packages/teleport/src/Apps/AddApp/Manually/Manually.tsx
kimlisa: Also
`Install Teleport on target server, start Teleport with the following parameters`
to
`Install Teleport on target server, then start Teleport with the following parameters`

packages/teleport/src/Nodes/AddNode/Manually/Manually.tsx
```diff
@@ -54,14 +54,14 @@ export default function Manually({ version, ...styles }: Props) {
...
-        <TextSelectCopy mt="2" text="tctl tokens add --type=node" />
+        {' - Generate a node join token'}
+        <TextSelectCopy mt="2" text="tctl tokens add --type=node --ttl=1h" />
       </Box>
       <Box>
         <Text bold as="span">
           Step 4
         </Text>
-        {` - Install Teleport on target server, and start it with the following parameters`}
+        {` - Install Teleport on target server, start Teleport with the following parameters`}
```
kimlisa, line 64: same here, `...then start Teleport`

packages/teleport/src/Apps/AddApp/Manually/Manually.tsx
benarent: > Do we need to say anything about DNS and certificate?

Cloud users won't need to make any changes to DNS since Applications will be on the subdomain of their cloud account. e.g. `internal-app.acme.teleport.sh` . Cloud will need obtain a cert for `*.acme.teleport.sh, acme.teleport.sh`

Re-naming: We've had a few versions. @klizhentas What do you think about

`Install the Teleport node service, then start Teleport with the following parameters`

packages/teleport/src/Apps/AddApp/Automatically/Automatically.tsx
```diff
@@ -73,9 +73,10 @@ export default function Automatically(props: Props) {
             </Flex>
             {!cmd && (
               <Text mb="3">
-                Provide the name and URL of your application to generate a
-                script that will automatically install and configure the App
-                service on the server that can access your application.
+                Teleport can automatically setup application access. Provide
+                the name and URL of your application to generate our
+                auto-installer script. This script will install the Teleport
+                node service and provide secure access to your application.
```
webvictim, line 79: Not sure if Application Access should be uppercased as it's the product name - it looks weird though.

```suggestion
                Teleport can automatically set up application access. Provide
                the name and URL of your application to generate our
                auto-installer script. This script will install the Teleport
                app service to provide secure access to your application.
```

packages/teleport/src/Apps/AddApp/Manually/Manually.tsx
```diff
@@ -57,17 +57,17 @@ export default function Manually({ version, onClose }: Props) {
...
-            <TextSelectCopy mt="2" text="tctl tokens add --type=app" />
+            {' - Generate an application join token'}
+            <TextSelectCopy mt="2" text="tctl tokens add --type=app --ttl=1h" />
           </Box>
           <Box>
             <Text bold as="span">
               Step 4
             </Text>
-            {` - Install Teleport on target server, and start it with the following parameters`}
+            {` - Install Teleport node service, then start Teleport with the following parameters`}
```
webvictim, line 67:
```suggestion
            {` - Install Teleport binary, then start Teleport with the following parameters`}
```

packages/teleport/src/Nodes/AddNode/Manually/Manually.tsx
```diff
@@ -54,14 +54,14 @@ export default function Manually({ version, ...styles }: Props) {
...
-        <TextSelectCopy mt="2" text="tctl tokens add --type=node" />
+        {' - Generate a node join token'}
+        <TextSelectCopy mt="2" text="tctl tokens add --type=node --ttl=1h" />
       </Box>
       <Box>
         <Text bold as="span">
           Step 4
         </Text>
-        {` - Install Teleport on target server, and start it with the following parameters`}
+        {` - Install Teleport node service, then start Teleport with the following parameters`}
```
webvictim, line 64:
```suggestion
        {` - Install Teleport binary, then start Teleport with the following parameters`}
```

packages/teleport/src/Apps/AddApp/Automatically/Automatically.tsx
webvictim: I would obviously like to know exactly what the script does, but I think this box is far too small to explain that in any detail. If somebody really wants to know exactly what the script does (and really they should) then they can just download the URL manually (without running it via `bash`) and read it to see what it does.

If we had the script's source in the OSS repo I'd consider linking to it.

packages/teleport/src/Apps/AddApp/Manually/Manually.tsx
webvictim: I think that's too many uses of "Teleport" in a row.

Also "node service" is a bit misleading when it's actually going to be running the Teleport app service; it should probably be called the `Teleport daemon`, `Teleport binary` or even `Teleport agent` which is probably the best description of what it actually does in this instance.

We have vastly overloaded the use of `service` as a noun :)

webvictim:
:man_shrugging:

packages/teleport/src/Apps/AddApp/Manually/Manually.tsx
alex-kovoy: > Cloud users won't need to make any changes to DNS since Applications will be on the subdomain of their cloud account. e.g. internal-app.acme.teleport.sh . 

@benarent there is no cloud version of teleport. We have E and OSS. So, what would be the manual steps for E users?

alex-kovoy: @benarent could we combine E and cloud instructions?

```
1. If you are running Teleport onprem, make sure that Teleport proxy has the certificate [docs] installed. You can skip this step if you are running a cloud version.
```

alex-kovoy:
we addressed above changes in https://github.com/gravitational/webapps/pull/196 
closing this one

PR #198, opened 2020-11-14
Disable AddServer and AddApp buttons on leaf clusters

alex-kovoy:
A minor change that disables create buttons on leaf clusters

### DIFF
--- a/packages/teleport/src/Apps/ButtonAdd.tsx
+++ b/packages/teleport/src/Apps/ButtonAdd.tsx
@@ -20,10 +20,12 @@ import { ButtonPrimary } from 'design';
 const docUrl = 'https://gravitational.com/teleport/docs/';
 
 export default function ButtonAdd(props: Props) {
-  if (!props.isEnterprise) {
+  const { isEnterprise, isLeafCluster, ...rest } = props;
+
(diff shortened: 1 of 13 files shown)


packages/teleport/src/Apps/ButtonAdd.tsx
```diff
@@ -36,8 +38,17 @@ export default function ButtonAdd(props: Props) {
   }
 
   if (props.canCreate) {
+    const title = isLeafCluster
+      ? 'Adding an application to the leaf cluster is not supported'
```
webvictim, line 42:
```suggestion
      ? 'Adding an application to a leaf cluster is not supported'
```

packages/teleport/src/Nodes/ButtonAdd.tsx
```diff
@@ -36,9 +38,18 @@ export default function ButtonAdd(props: Props) {
     );
   }
 
-  if (props.canCreate) {
+  if (canCreate) {
+    const title = isLeafCluster
+      ? 'Adding a server to the leaf cluster is not supported'
```
webvictim, line 43:
```suggestion
      ? 'Adding a server to a leaf cluster is not supported'
```

PR #204, opened 2020-11-25
Refactor

kimlisa:
#### Description
- Move AccessStrategy to e (seen as deleted file in this PR)
- Refactored feature flags
- Allow `Teleport` to accept a ctx


packages/teleport/src/Main/useMain.ts
```diff
@@ -16,11 +16,14 @@ limitations under the License.
...
-import useTeleport from 'teleport/useTeleport';
 import { Feature } from 'teleport/types';
+import teleportContext from 'teleport/teleportContext';
+import teleportEContext from 'e-teleport/teleportEContext';
 
-export default function useMain(features: Feature[]) {
-  const ctx = useTeleport();
+export default function useMain(
+  features: Feature[],
+  ctx: teleportContext | teleportEContext
```
alex-kovoy, line 25: `ctx: teleportContext | teleportEContext`

why? Assuming that E version would be extending the OSS it would still work

```tsx
class TeleportContext extends OSSContext {
}
```

packages/teleport/src/Main/Main.tsx
```diff
@@ -26,10 +26,12 @@ import SideNav from 'teleport/SideNav';
 import TopBar from 'teleport/TopBar';
 import getFeatures from 'teleport/features';
 import useMain, { State } from './useMain';
+import useTeleport from 'teleport/useTeleport';
 
 export default function Container() {
   const [features] = React.useState(() => getFeatures());
-  const state = useMain(features);
+  const ctx = useTeleport();
```
alex-kovoy, line 33: is it only for consistency reasons you prefer to inject the ctx this way or there is something else?

packages/teleport/src/Main/useMain.ts
```diff
@@ -16,11 +16,14 @@ limitations under the License.
 
 import { useState } from 'react';
 import useAttempt from 'shared/hooks/useAttemptNext';
-import useTeleport from 'teleport/useTeleport';
 import { Feature } from 'teleport/types';
+import teleportContext from 'teleport/teleportContext';
+import teleportEContext from 'e-teleport/teleportEContext';
```
alex-kovoy, line 21: > import teleportEContext from 'e-teleport/teleportEContext';

devs wont be able to build the OSS version. OSS should not have _any_ references to E.

packages/teleport/src/Users/useUsers.ts
```diff
@@ -78,7 +78,7 @@ export default function useUsers() {
 
   useEffect(() => {
     function fetchRoles() {
-      if (ctx.isRolesEnabled()) {
+      if (ctx.getFeatureAccess().roles) {
```
alex-kovoy, line 81: These are feature flags, not the access. Feature flags are used to disable/enable parts of functionality. It could be because of permissions or because of other reasons. For example, workflows are not available to OSS users not because of access but because of the product type.

packages/teleport/src/Main/useMain.ts
kimlisa: revised

packages/teleport/src/Main/Main.tsx
kimlisa: so i can pass in enterprise context for enterprise Main

packages/teleport/src/Main/useMain.ts
kimlisa: ah got it

packages/teleport/src/Users/useUsers.ts
kimlisa: got it, i see the difference now

packages/teleport/src/Main/Main.tsx
alex-kovoy: > so i can pass in enterprise context for enterprise Main

But all the components will continue referencing the OSS context because this is what remains in the ContextProvider. 
For E, you need to initialize TeleportContextProvider with the E version.


```tsx
export default function TeleportE({ history }) {
  const [ctx] = React.useState( () => new TeleportE() );
  return (
    <Teleport history={history} ctx={ctx}>
      <AccessStrategy>
        <LicenseEnforcer />
```

Then in every place where we have `useTeleport()` the E version will be returned. 

https://reactjs.org/docs/context.html

packages/teleport/src/Main/useMain.ts
```diff
@@ -16,11 +16,10 @@ limitations under the License.
 
 import { useState } from 'react';
 import useAttempt from 'shared/hooks/useAttemptNext';
-import useTeleport from 'teleport/useTeleport';
 import { Feature } from 'teleport/types';
+import teleportContext from 'teleport/teleportContext';
 
-export default function useMain(features: Feature[]) {
-  const ctx = useTeleport();
+export default function useMain(features: Feature[], ctx: teleportContext) {
```
alex-kovoy, line 22: you can undo `useMain` changes

alex-kovoy (approved):
with comments

packages/teleport/src/Main/useMain.ts
kimlisa: reverted

PR #210, opened 2020-12-29
Replace app name check with regex

kimlisa:
resolves https://github.com/gravitational/teleport/issues/4953

#### Description
- replace URL object app name check with regex (same regex as the backend uses):
https://github.com/kubernetes/apimachinery/blob/release-1.20/pkg/util/validation/validation.go#L219

### DIFF
--- a/packages/teleport/src/Apps/AddApp/Automatically/Automatically.tsx
+++ b/packages/teleport/src/Apps/AddApp/Automatically/Automatically.tsx
@@ -67,7 +67,7 @@ export default function Automatically(props: Props) {
                 width="320px"
                 mr="3"
                 onKeyPress={e => handleEnterPress(e, validator)}
-                onChange={e => setName(e.target.value)}
+                onChange={e => setName(e.target.value.toLowerCase())}
               />
               <FieldInput
                 rule={requiredAppUri}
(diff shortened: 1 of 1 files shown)


packages/teleport/src/Apps/AddApp/Automatically/Automatically.tsx
```diff
@@ -180,20 +189,18 @@ const requiredAppName = value => () => {
...
-    if (tmp.hostname.split('.').length > 1) {
-      throw new Error();
-    }
-  } catch {
+  const match = value.match(REGEX_DNS1035LABEL);
+  if (!match) {
     return {
       valid: false,
-      message: 'Invalid',
+      message: 'Invalid name',
```
alex-kovoy, line 203: lets change it to `Should be a valid DNS sub-domain name`

alex-kovoy: could you also ensure that the text inside app name field is also a lower case? 

I think the css rule would do.

```
text-transform: lowercase;
```

alex-kovoy (approved):
with comments

packages/teleport/src/Apps/AddApp/Automatically/Automatically.tsx
kimlisa: revised as discussed in slack

PR #212, opened 2020-12-31
Clear local storage on invite login form

kimlisa:
#### Description
Clears local storage when user goes to a invite link.

If a previous user was logged in and didn't log out, it leads to a bug where if this invited user required a waiting room and the previous user required a waiting room, it will bypass the auto request creation b/c of what existed in the local storage.


alex-kovoy (commented):
Clearing a localStorage on fetch is confusing and after reading a description it seems this is a very rare use-case to be fixed in this way.  Lets close this PR in favor of clean code.

PR #216, opened 2021-01-20
Add more fields to user context

kimlisa:
#### Description
- update user context: grouped `suggestedReviewers` and `requestableRoles` under new type `AccessCapabilities` (similar to backend)
- allow table td element to use title attribute, this was needed to allow on hover to display full text on shortened access request reasons
- replace `flex-shrink:0` with `white-space: nowrap` and update snapshots
- add global font setting for html element textarea
- add access request review event

### DIFF
--- a/packages/design/src/DataTable/Table.jsx
+++ b/packages/design/src/DataTable/Table.jsx
@@ -149,12 +149,13 @@ class Column extends React.Component {
 }
 
 const Cell = props => {
-  const { isHeader, children, align, style, className } = props;
+  const { isHeader, children, align, style, className, title } = props;
   const childProps = {
     children,
     align,
     style,
     className,
+    title,
   };
 
   if (isHeader) {
--- a/packages/teleport/src/Support/Support.story.tsx
+++ b/packages/teleport/src/Support/Support.story.tsx
@@ -18,6 +18,7 @@ import React from 'react';
 import Support from './Support';
 import * as teleport from 'teleport';
 import cfg from 'teleport/config';
+import { userContext } from 'teleport/Main/fixtures';
 
 export default {
   title: 'Teleport/Support',
@@ -26,7 +27,7 @@ export default {
 export const SupportOSS = () => {
   cfg.isEnterprise = false;
   const ctx = new teleport.Context();
-  ctx.storeUser.state = state;
+  ctx.storeUser.state = userContext;
 
   return (
     <teleport.ContextProvider ctx={ctx}>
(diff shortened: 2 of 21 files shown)


packages/teleport/src/components/Layout/Layout.jsx
```diff
@@ -40,7 +40,7 @@ FeatureHeader.defaultProps = {
  * Header Title
  */
 const FeatureHeaderTitle = styled(Text)`
-  flex-shrink: 0;
+  white-space: nowrap;
```
kimlisa, line 43: `flex-shrink: 0` prevented ellipsis from working b/c flexed containers wouldn't shrink.

I think we use flex shrink 0 here to prevent titles from wrapping, so `white-space: no wrap` achieved the same behavior.

alex-kovoy: `flex-shrink:0` is used to prevent strange squeezing behavior of this container in Safari. Try out minimizing the window height when a page has some content in it (populated table for many rows for example). Maybe we can try minHeight instead.

kimlisa: I think i'm blind, I'm not able to reproduce the behavior in safari... I tried by removing both flex-shrink and white-space. I'm thinking it got fixed elsewhere? I still need the `white-space: nowrap` to prevent title wrapping though.

packages/teleport/src/services/user/user.test.ts
```diff
@@ -67,7 +67,8 @@ test('fetch user context, null response gives proper default values', async () =
   });
 
   expect(response.accessStrategy).toStrictEqual(defaultStrategy);
-  expect(response.requestableRoles).toStrictEqual([]);
+  expect(response.accessCapabilities.requestableRoles).toStrictEqual([]);
+  expect(response.accessCapabilities.suggestedReviewers).toStrictEqual([]);
```
alex-kovoy, line 71: What about this -> `expect(response).toStringEqual(needed)` ?

packages/design/src/DataTable/Table.jsx
```diff
@@ -149,12 +149,13 @@ class Column extends React.Component {
...
 
 const Cell = props => {
-  const { isHeader, children, align, style, className } = props;
+  const { isHeader, children, align, style, className, title } = props;
   const childProps = {
     children,
     align,
     style,
     className,
+    title,
```
alex-kovoy, line 158: Where is it used? `title`

packages/teleport/src/Support/Support.story.tsx
```diff
@@ -18,6 +18,7 @@ import React from 'react';
 import Support from './Support';
 import * as teleport from 'teleport';
 import cfg from 'teleport/config';
+import { userContext } from 'teleport/Main/fixtures';
```
alex-kovoy, line 21: this dependency on the Main fixtures looks strange because Main story would reach out to its children component for the same. This cross referencing seems unnecessary. Also, what's special about userContext that it needs to be imported from the fixtures to render a basic Support screen?

packages/teleport/src/services/user/user.test.ts
kimlisa: i revised this test to test for matching response with expected object

packages/teleport/src/Support/Support.story.tsx
kimlisa: refactored Support so fixtures and context wrappers are no longer needed

packages/design/src/DataTable/Table.jsx
kimlisa: here, it's to enable on hover show full text: https://github.com/gravitational/webapps.e/pull/31/files#diff-bd72ddd2c5cda80ddbf61e0b7a50791c0f022afd31eeb78517fdbab8df343cc2R180

alex-kovoy (approved):

PR #218, opened 2021-01-27
Check for browser u2f support

kimlisa:
resolves https://github.com/gravitational/teleport/issues/5397

#### Description
For screens that used u2f (Login, Invite/Reset), on submit, checks if u2f object exists in browser window before making calls, and returns proper error message. 

From research, there is still no support for fido u2f on safari, however there is support for fido2//webauthn (as with other major browsers)

Tested u2f login on:
- mac browsers: safari, firefox, google
- linux browsers: firefox, google

#### Screenshot
safari
image

Screen Shot 2021-01-27 at 9 43 34 AM

### DIFF
--- a/packages/teleport/src/services/auth/auth.js
+++ b/packages/teleport/src/services/auth/auth.js
@@ -26,7 +26,9 @@ const auth = {
       return null;
     }
 
-    return new Error('your browser does not support FIDO U2F protocol');
+    return new Error(
+      'this browser does not support U2F required for hardware tokens, please try Chrome or Firefox instead'
+    );
   },
 
   login(userId, password, token) {
--- a/packages/teleport/src/services/auth/auth.js
+++ b/packages/teleport/src/services/auth/auth.js
@@ -26,7 +26,7 @@ const auth = {
       return null;
     }
 
-    return new Error('Your browser does not support FIDO U2F protocol.');
+    return new Error('your browser does not support FIDO U2F protocol');
   },
 
   login(userId, password, token) {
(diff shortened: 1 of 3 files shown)


packages/teleport/src/services/auth/auth.js
```diff
@@ -21,6 +21,14 @@ import cfg from 'teleport/config';
 import makePasswordToken from './makePasswordToken';
 
 const auth = {
+  u2fBrowserSupported() {
+    if (window.u2f) {
+      return null;
+    }
+
+    return new Error('Your browser does not support FIDO U2F protocol.');
```
alex-kovoy, line 29: our errors are lower case without `.`

`your browser does not support FIDO U2F protocol`

alex-kovoy (approved):
with a minor comment

packages/teleport/src/services/auth/auth.js
kimlisa: revised

klizhentas:
@kimlisa @alex-kovoy let's improve error message. We need to make it more user-friendly. Lots of users do not know what U2F protocol is and what should they do about it.

How about this:

`This browser does not support U2F required for hardware tokens. Please try Chrome or Firefox instead`

klizhentas (approved):

shaneshort:
apologies if this is the wrong place for this, but I notice this is coming up on modern (15+) versions of Safari even though it's documented that Safari now supports FIDO/U2F: https://caniuse.com/u2f / http://yubico.com/au/works-with-yubikey/catalog/safari/

Is there something that needs changing to allow this to function?

kimlisa:
hi @shaneshort, as I understood it, safari supports fido2 with webauthn (which is backwards compatible with fido u2f). We now support webauthn with our latest v8 release: https://goteleport.com/docs/ver/8.0/access-controls/guides/webauthn/

Issue #220, opened 2021-02-03
Open source RBAC components


alex-kovoy:
We need enable the following functionality for open source users:

1) Role Add|Edit
2) User Add|Edit
3) Github Connector Add|Edit

This also means that we need to move E webapi endpoints to the OSS branch as well.

I will update this ticket with more details.

alex-kovoy:
closed by https://github.com/gravitational/webapps/pull/222

PR #222, opened 2021-02-11
Open source and refactor resources

kimlisa:
part of https://github.com/gravitational/webapps/issues/220

#### Description
- Renamed AuthConnectors to GithubConnectors
- Refactored AuthConnectors, Roles, TrustedCluster:
  - converted remaining js files to ts
  - convert to the next useAttempt
- Convert resource service to a class and refactored methods

#### Screenshot
empty
image

loaded
image


alex-kovoy (commented):
@kimlisa lets not rename the feature and directory. Keep it as "Auth Connectors" similarly to E. The `E` may reuse some of the components (like dialogs) from OSS "Auth Connectors". Also do not change the Side Nav items. Keep the side nav item as it was so the E version would look as extension.

alex-kovoy (approved):
Looks good. I see that you performed some refactoring. So I assume that this has been tested and no React warnings in the dev console have been found :)

packages/teleport/src/Apps/Empty/Empty.tsx
```diff
@@ -53,7 +53,14 @@ export default function Empty(props: Props) {
...
 
   return (
-    <Card maxWidth="700px" mx="auto" py={4} as={Flex} alignItems="center">
+    <Card
+      maxWidth="700px"
+      mx="auto"
+      py={4}
+      as={Flex}
+      alignItems="center"
+      flex="0 0 auto"
```
kimlisa, line 62: `flex="0 0 auto"` fixes this squishing flex box behavior in safari

image

Issue #221, opened 2021-02-09
Fetch API params vs headers


andrejtokarcik:
https://github.com/gravitational/webapps/blob/b628daab1d49f19cc162ccfabc3232d1ff7e7978/packages/teleport/src/services/api/api.js#L82-L83

Are we sure these should be set as headers and not directly as `fetch` options? See e.g. https://javascript.info/fetch-api

PR #226, opened 2021-02-17
Refactor error handling

kimlisa:
fixes https://github.com/gravitational/webapps/issues/221
resolves https://github.com/gravitational/teleport/issues/5493 
part of https://github.com/gravitational/teleport.e/issues/208

#### Description
- Fix dev-server (copy and paste)
- Correct fetch option setting
- Refactor err handling when login fails
- Convert to ts
- Prevent sending refer field in header: also manually added rel attribute `noreferrer` to anchor tags for wider browser coverage:
  - https://caniuse.com/rel-noreferrer
  - https://caniuse.com/mdn-html_elements_meta_name_referrer

#### Dependency
https://github.com/gravitational/teleport/pull/5599

#### Screenshot
no referer header
image

same-origin setting
image

no-store cache setting (in chrome, cache control field is not there meaning no store, firefox and safari is more verbal as below)
image


packages/build/index.ejs
```diff
@@ -1,9 +1,8 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <meta charset="utf-8">
-    <meta http-equiv="X-UA-Compatible" content="IE=edge">
-    <meta name="referrer" content="origin" />
+    <meta charset="utf-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
```
alex-kovoy, line 5: You need to bring it back. This is to prevent leaking the referrer information when clicking on the URL links. https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noreferrer

This is separate from REST calls.

packages/teleport/src/config.ts
```diff
@@ -61,9 +62,9 @@ const cfg = {
     player: '/web/cluster/:clusterId/session/:sid',
     sessionAuditPlayer: '/web/cluster/:clusterId/session/:sid/player',
     sessionAuditCmds: '/web/cluster/:clusterId/session/:sid/commands',
-    error: '/web/msg/error/:type?',
     login: '/web/login',
     loginFailed: '/web/msg/error/login_failed',
+    loginFailedCallback: '/web/msg/error/login_failed_callback',
```
alex-kovoy, line 67: Instead, lets have a route

```
loginError: '/web/msg/error/login/',
loginErrorCallback: '/web/msg/error/login/call_back'
```

```tsx
 <Route path={cfg.routes.loginError} component={LoginError}  />
```

The `LoginError` will handle all it's children routes itself as well its own index route.

packages/build/index.ejs
kimlisa: revised

packages/teleport/src/config.ts
kimlisa: revised

packages/teleport/src/Teleport.tsx
```diff
@@ -44,9 +44,16 @@ const Teleport: React.FC<Props> = ({ ctx, history, children }) => {
...
-              path={cfg.routes.loginFailed}
-              component={LoginFailed}
+              path={cfg.routes.loginError}
+              render={({ location }) => (
+                <LoginFailed
+                  isCallbackErr={
+                    location.pathname === cfg.routes.loginErrorCallback
+                  }
+                />
+              )}
```
alex-kovoy, line 54:
```tsx
  return (
    <Switch>
      <Route path={cfg.routes.loginErrorCallback}>
        <LoginFailed message="unable to process callback"/>
      </Route>
      <Route component={LoginFailed} />
    </Switch>
  );
}

export function LoginFailed() {
  const message = getUrlParameter('details', window.location.search);
  return (
    <>
      <LogoHero />
      <CardFailed loginUrl={cfg.routes.login} message={message} />
    </>
  );
}

```

then

```tsx
  import LoginFailed from 'xxx';

 <Route path={cfg.routes.loginError} component={LoginFailed} />
 <Route path={cfg.routes.loginErrorLegacy} component={LoginFailed} />
 ```
 
 LoginFailed component will handle its own internal routes without exposing its logic to Teleport.tsx

Comment (kimlisa) [reply packages/teleport/src/Teleport.tsx]: revised

**Review by alex-kovoy — APPROVED** [review]

File: packages/design/src/Link/Link.jsx [hunk 4]
```diff
@@ -29,19 +29,17 @@ Link.defaultProps = {
...
   color: ${({ theme }) => theme.colors.link};
   font-weight: normal;
   background: none;
   text-decoration: underline;
   text-transform: none;
 
-  &:hover,
-  &:focus {
-    background: ${({ theme }) => theme.colors.primary.light};
-  }
```
kimlisa, line 42: The highlighting affect for hover/focus had no affect b/c the color is the same as the content background color so I removed it

packages/teleport/src/config.ts
```diff
@@ -118,11 +120,13 @@ const cfg = {
   },
 
   getAuthProviders() {
-    return cfg.auth && cfg.auth.providers ? cfg.auth.providers : [];
+    return cfg.auth && cfg.auth.providers
+      ? (cfg.auth.providers as AuthProvider[])
```
alex-kovoy, line 124: instead of casting in every place, could you do the following instead in the config.ts?

```ts
  auth: {
    localAuthEnabled: true,
    providers: [] as AuthProvider[],
    second_factor: 'off' as Auth2faType,
    authType: 'local',
  },
```

alex-kovoy (approved):
with minor comment

packages/teleport/src/config.ts
kimlisa: revised

PR #232, opened 2021-02-25
Disable use of web workers in ace editor

kimlisa:
#### Description
Fixes error messages produced when clicking on "Details" on event dialogue:
image

image

Disable use of web workers (syntax validators) b/c they are not used. Currently, we only use two modes json and yaml. json is used as read only mode, and yaml doesn't have syntax validating workers

### DIFF
--- a/packages/shared/components/TextEditor/TextEditor.jsx
+++ b/packages/shared/components/TextEditor/TextEditor.jsx
@@ -57,7 +57,7 @@ class TextEditor extends React.Component {
     session.setUndoManager(undoManager);
     session.setUseWrapMode(false);
     session.setMode(mode);
-    session.setOptions({ tabSize, useSoftTabs: true });
+    session.setOptions({ tabSize, useSoftTabs: true, useWorker: false });
     return session;
   }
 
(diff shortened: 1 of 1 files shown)


packages/shared/components/TextEditor/TextEditor.jsx
```diff
@@ -72,11 +77,15 @@ class TextEditor extends React.Component {
...
     this.setActiveSession(0);
   }
 
+  initSessionsReadOnly(data = []) {
+    this.sessions = data.map(item => this.createSessionReadOnly(item));
+    this.setActiveSession(0);
+  }
+
   componentDidMount() {
     const { data, readOnly, theme = 'ace/theme/monokai' } = this.props;
```
kimlisa, line 90: I was wondering why we send data as an array

alex-kovoy:
The browser tells exactly what's missing - CSP Header policy to execute a webworker. Please write a simple webworker and study it's registration behavior.

alex-kovoy:
@kimlisa lets just disable the workers for editor with 1 liner.

alex-kovoy (approved):

PR #235, opened 2021-02-26
Implement OAuth-style state token for AAP auth flow

andrejtokarcik:


packages/teleport/src/AppLauncher/useAppLauncher.ts
```diff
@@ -31,9 +32,15 @@ export default function useAppLauncher() {
         // make a redirect to the requested app auth endpoint
         const location = window.location;
         const port = location.port ? ':' + location.port : '';
-        window.location.replace(
-          `https://${result.fqdn}${port}/x-teleport-auth#value=${result.value}`
-        );
+        const state = getUrlParameter('state', location.search);
+        const authUrl = `https://${result.fqdn}${port}/x-teleport-auth`;
```
klizhentas, line 36: @andrejtokarcik @alex-kovoy any reason we are not using 

https://javascript.info/url#creating-a-url

klizhentas: Feel free to ignore me if I'm missing something here

andrejtokarcik: Hey @klizhentas this PR is part of a public security release I'm preparing at this very moment. I'll be able to address this only later, though I personally don't see a benefit to using `URL` in this case.

klizhentas: I'm assuming URL should be safer because it can escape the parameters that contain special characters.

alex-kovoy: That string enforces the `https://` schema which tells a browser about URL (this is to avoid `javascript://` or other stuff that might come from the server).

packages/teleport/src/AppLauncher/useAppLauncher.ts
```diff
@@ -31,9 +32,15 @@ export default function useAppLauncher() {
...
+        if (state === '') {
+          const clusterId = params.clusterId ? params.clusterId : '';
+          const publicAddr = params.publicAddr ? params.publicAddr : '';
+          window.location.replace(`${authUrl}?cluster=${clusterId}&addr=${publicAddr}`);
+        } else {
+          window.location.replace(`${authUrl}?state=${state}#value=${result.value}`);
+        }
```
alex-kovoy, line 43:
```ts
if (state === '') {
   const clusterId = params.clusterId ? params.clusterId : '';
   const publicAddr = params.publicAddr ? params.publicAddr : '';
   window.location.replace(`${authUrl}?cluster=${clusterId}&addr=${publicAddr}`);
 } else {
  window.location.replace(`${authUrl}?state=${state}#value=${result.value}`);
}
```

@kimlisa could you re-write above code with below? 

```ts
let url = ${authUrl}?state=${state}#value=${result.value}`;
if (!state){
  url = `${authUrl}?cluster=${params.clusterId}&addr=${params.publicAddr}`
}

window.location.replace(url);
```

alex-kovoy:
```js
const url = new URL(`https://{fqdn});
if (state) {
  url.searchParams.set("cluster", params.clusterId)
  url.searchParams.set("addr", params.publicAddr)
}
url.hash = `#value=${result.value}`;

window.location.replace(s)
```

andrejtokarcik:
@kimlisa I need to push this very same commit to `teleport-v6` and `master` as well. Please add any improvements you make here to those branches too.

packages/teleport/src/AppLauncher/useAppLauncher.ts
andrejtokarcik: I'd rather not pass valid sessions around unless necessary, i.e. set `#value=` only if `state` is set as it's done in my code.

PR #237, opened 2021-03-04
Handle new MFA audit events

awly:
Support for mfa.add and mfa.delete events.


kimlisa (approved):
LGTM, the addition to the snapshot looks good to me as well, can you update the snapshot that is failing by running `yarn test -u`? Its failing (or updated) b/c we added data to the fixtures which is used in storybook, and we take a snapshot of its rendered state

PR #247, opened 2021-03-18
Bring back cluster details

kimlisa:
resolves https://github.com/gravitational/teleport/issues/4734

#### Description
- brings back cluster details from before
- added last connected and status column to table

#### Dependency
https://github.com/gravitational/teleport/pull/6049

#### Screenshot
image


packages/teleport/src/services/clusters/types.ts
```diff
@@ -14,11 +14,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+type StatusEnum = 'online' | 'offline';
```
alex-kovoy, line 17: `StatusEnum -> Status`

`Enum` has a special meaning in typesscript. The fact you used this suffix to name a type is confusing. 
In Javascript we used `Enum` suffix as a convention to indicate enumerations, no need to bring this convention to TS as TS has a built-in support for it.

packages/teleport/src/Clusters/fixtures/index.ts
```diff
@@ -321,7 +321,7 @@ export const clusters = [
     clusterId: 'barrynelson110',
     nodeCount: 23,
     authVersion: '1.2.199',
-    status: 'offline',
+    status: 'offline' as any,
```
alex-kovoy, line 324: can we avoid repeating it `status: 'offline' as any` ?

Issue #250, opened 2021-03-24
[Bug] Unable to close terminal tab when there is terminal error


kimlisa:
#### Reproduce
This was how i encountered the error, sure there are other ways:

Set `require_session_mfa` on:
```
authentication:
    type: local
    second_factor: optional
    require_session_mfa: yes
```
There is currently an error with chrome on how `require session mfa` feature works in UI, and this results in timing out error. After 30 seconds, trying to exit out won't allow you to b/c numUsers is undefined: https://github.com/gravitational/webapps/blob/master/packages/teleport/src/Console/useOnExitConfirmation.ts#L73

PR #251, opened 2021-03-26
Mfa fixes

kimlisa:
fixes https://github.com/gravitational/webapps/issues/250
fixes https://github.com/gravitational/teleport/issues/6156
fixes https://github.com/gravitational/teleport/issues/6139

#### Description
- Removed special styling to mfa dropdown
- FormPassword now checks for auth2faTypes on and optional
  - Place a check for `window.u2f` being null for safari
- Prompt user to tap their u2f key when connecting to a node with require_session_mfa turned on
- useOnExitConfrimation checks if there is any participants before accessing length
- Update e-ref

#### Test
firefox worked as same as before

safari:

`require_session_mfa` turned on
image

`auth2faType` set to on or optional
image

chrome:
image

### DIFF
--- a/packages/shared/components/FormPassword/FormPassword.tsx
+++ b/packages/shared/components/FormPassword/FormPassword.tsx
@@ -15,10 +15,12 @@ limitations under the License.
 */
 
 import React from 'react';
-import { Card, ButtonPrimary } from 'design';
+import { Card, ButtonPrimary, Flex, Box } from 'design';
 import * as Alerts from 'design/Alert';
 import useAttempt from 'shared/hooks/useAttempt';
+import { Option } from 'shared/components/Select';
 import FieldInput from '../FieldInput';
+import FieldSelect from '../FieldSelect';
 import Validation, { Validator } from '../Validation';
 import {
   requiredToken,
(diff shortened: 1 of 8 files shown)


packages/shared/components/FormPassword/FormPassword.tsx
```diff
@@ -77,11 +92,21 @@ function FormPassword(props: Props) {
...
+    setMfaType(option);
+  }
+
   return (
     <Validation>
       {({ validator }) => (
         <Card as="form" bg="primary.light" width="456px" p="6">
-          <Status isU2F={u2fEnabled} attempt={attempt} />
+          <Status
+            isU2F={u2fEnabled || (mfaEnabled && mfaType.value === 'u2f')}
```
alex-kovoy, line 107: > u2fEnabled || (mfaEnabled && mfaType.value === 'u2f'

This condition gets repeated in a few places. Can we name it like 
```js
const isU2FSelected = u2fEnabled || (mfaEnabled && mfaType.value === 'u2f'` 
```

packages/teleport/src/Console/useOnExitConfirmation.ts
```diff
@@ -66,11 +66,8 @@ function useOnExitConfirmation(ctx: ConsoleContext) {
   /**
    * confirmCloseSession prompts user to confirm to close.
    */
-  function confirmCloseSession(doc) {
-    if (!doc || !doc.sid) return false;
-
-    const numUsers = ctx.storeParties.state[doc.sid];
-    if (numUsers.length > 1) {
+  function confirmCloseSession(participants: number) {
```
alex-kovoy, line 69: hm...there is no much value in this function now. It accepts an integer and checks for > 0.  You might very well delete it and move the alerting to the `verifyAndConfirm`

alex-kovoy (approved):
with minor comments

russjones (approved):
Bot.

PR #261, opened 2021-04-08
Update renew session response and renew URL

kimlisa:
part of https://github.com/gravitational/teleport/issues/4937

#### Description
- Renew session endpoint now returns a new field `sessionExpires`. This is the absolute time the web session will expire for the most recently consumed access request. This is the value used in the switchback countdown
- Refactored how bearer token is being set

~~- Added local storage reload broadcaster to be able to reload opened tabs after user assumes a role or switches back to their default roles. This will be important when user switches back from higher privilege to lower privilege especially when they are connected to a node.~~
~~- Added ability to force reload to prevent confirmation pages (terminal) from reloading tabs~~

### DIFF
--- a/packages/teleport/src/services/session/types.ts
+++ b/packages/teleport/src/services/session/types.ts
@@ -1,0 +1,4 @@
+export type RenewSessionRequest = {
+  requestId?: string;
+  switchback?: boolean;
+};
--- a/packages/teleport/src/services/session/types.ts
+++ b/packages/teleport/src/services/session/types.ts
@@ -2,3 +2,14 @@ export type RenewSessionRequest = {
   requestId?: string;
   switchback?: boolean;
 };
+
+export type BearerToken = {
+  accessToken: string;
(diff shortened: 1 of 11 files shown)


packages/teleport/src/services/session/types.ts
```diff
@@ -0,0 +1,15 @@
+export type RenewSessionRequest = {
+  requestId?: string;
+  switchback?: boolean;
```
alex-kovoy, line 3: switchback -> `onlyStaticRoles` ?

alex-kovoy (approved):

packages/teleport/src/services/session/types.ts
kimlisa: I prefer to stick with `switchback` b/c its shorter/simpler and i don't foresee `switchback` term being used for other meanings

alex-kovoy:
@kimlisa LGTM

Issue #268, opened 2021-04-20
In nodes list, allow searching for exact labels


kimlisa:
#### Problem
From Servers tab (aka Nodes), users are unable to search for exact text content of labels under `LABELS` column.
For example given a label, `arch: x86_64`, users cannot search for exact `arch: x86_64`. This is b/c the value we use to search  and value shown to the user is not the same. 

Value we use to search is shaped like [this](https://github.com/gravitational/webapps/blob/master/packages/teleport/src/services/nodes/types.ts#L17) , and we search by each field [seperately](https://github.com/gravitational/webapps/blob/master/packages/teleport/src/components/NodeList/NodeList.tsx#L112)

And for table display, we format this value to look like `value: name` as shown [here](https://github.com/gravitational/webapps/blob/master/packages/teleport/src/components/NodeList/NodeList.tsx#L166)

![image](https://user-images.githubusercontent.com/43280172/115339836-6b0bdc00-a15a-11eb-8ba3-2847e23ecb8e.png)

#### How to resolve
Format the value (like how we do [here](https://github.com/gravitational/webapps/blob/master/packages/teleport/src/components/NodeList/NodeList.tsx#L164)) where we sanitize the response value: https://github.com/gravitational/webapps/blob/master/packages/teleport/src/services/nodes/makeNode.ts#L20

We can create another field for type [Node](https://github.com/gravitational/webapps/blob/master/packages/teleport/src/services/nodes/types.ts#L22), named `tagTexts`, typed as `string[]`. We can then use the field `tagTexts` instead of `tags` (extracted from [here](https://github.com/gravitational/webapps/blob/master/packages/teleport/src/components/NodeList/NodeList.tsx#L163)) and can remove the need to separately search for each field in a tag (remove this [block](https://github.com/gravitational/webapps/blob/master/packages/teleport/src/components/NodeList/NodeList.tsx#L112)) and remove the formatting done [here](https://github.com/gravitational/webapps/blob/master/packages/teleport/src/components/NodeList/NodeList.tsx#L166)

PR #279, opened 2021-04-22
Fix not being able to filter nodes by searching exact label

rudream:
### Purpose

This PR addresses #268 

Users should be able to search through nodes by typing in the exact label/tag in the search bar. (eg. typing `arch: x86_64` in the search bar should return all nodes with `arch: x86_64` in its table column).

### Implementation

Instead of each node storing its labels as an array of objects (eg. `[{name: value}]`), it now stores them as an array of strings (eg. `["name: value"]`). This makes it possible for the labels to be directly compared against the string entered in the search bar, allowing for better and more predictable results when searching.

### Demo

demo-search

### DIFF
--- a/packages/teleport/src/Console/DocumentNodes/DocumentNodes.story.tsx
+++ b/packages/teleport/src/Console/DocumentNodes/DocumentNodes.story.tsx
@@ -18,6 +18,7 @@ import React from 'react';
 import DocumentNodes from './DocumentNodes';
 import ConsoleCtx from './../consoleContext';
 import { TestLayout } from './../Console.story';
+import {Node} from "teleport/services/nodes/types"
 
 export default {
   title: 'Teleport/Console/DocumentNodes',
@@ -90,7 +91,7 @@ const clusters = [
   },
 ];
 
-const nodes = [
+const nodes: Node[] = [
   {
     tunnel: false,
     id: '104',
--- a/packages/teleport/src/Console/DocumentNodes/DocumentNodes.story.tsx
+++ b/packages/teleport/src/Console/DocumentNodes/DocumentNodes.story.tsx
@@ -18,7 +18,7 @@ import React from 'react';
 import DocumentNodes from './DocumentNodes';
 import ConsoleCtx from './../consoleContext';
 import { TestLayout } from './../Console.story';
-import {Node} from "teleport/services/nodes/types"
+import { Node } from 'teleport/services/nodes/types';
 
 export default {
   title: 'Teleport/Console/DocumentNodes',
@@ -98,7 +98,7 @@ const nodes: Node[] = [
     clusterId: 'cluseter-1',
     hostname: 'fujedu',
     addr: '172.10.1.20:3022',
-    tagTexts: ["cluster: one", "kernel: 4.15.0-51-generic"]
+    tagTexts: ['cluster: one', 'kernel: 4.15.0-51-generic'],
   },
   {
     tunnel: false,
@@ -106,7 +106,7 @@ const nodes: Node[] = [
     clusterId: 'cluseter-1',
     hostname: 'facuzguv',
     addr: '172.10.1.42:3022',
-    tagTexts: ["cluster: one", "kernel: 4.15.0-51-generic"]
+    tagTexts: ['cluster: one', 'kernel: 4.15.0-51-generic'],
   },
   {
     tunnel: true,
@@ -114,7 +114,7 @@ const nodes: Node[] = [
     clusterId: 'cluseter-1',
     hostname: 'duzsevkig',
     addr: '172.10.1.156:3022',
-    tagTexts: ["cluster: one", "kernel: 4.15.0-51-generic"]
+    tagTexts: ['cluster: one', 'kernel: 4.15.0-51-generic'],
   },
   {
     tunnel: true,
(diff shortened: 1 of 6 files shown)


packages/teleport/src/services/nodes/makeNode.ts
```diff
@@ -18,11 +18,11 @@ import { at } from 'lodash';
 import { Node } from './types';
 
 export default function makeNode(json): Node {
-  const [id, clusterId, hostname, tags = [], addr, tunnel = false] = at(json, [
+  const [id, clusterId, hostname, tagTexts = json.tags.map(tag => `${tag.name}: ${tag.value}`), addr, tunnel = false] = at(json, [
```
kimlisa, line 21: lets refactor this function to use the most recent way of processing responses (without lodash), similar to https://github.com/gravitational/webapps.e/blob/master/teleport/src/services/workflow/makeAccessRequest.ts#L8

rudream: Done! :+1:

packages/teleport/src/services/nodes/makeNode.ts
```diff
@@ -14,18 +14,13 @@ See the License for the specific language governing permissions and
...
-    'id',
-    'siteId',
-    'hostname',
-    'tagTexts',
-    'addr',
-    'tunnel',
-  ]);
+  const {id, clusterId, hostname, addr, tunnel} = json;
+
+  const tags = json.tags || [];
```
kimlisa, line 22: we can assign default values to destructured objects, so in line 20 we can do:
```ts
const {id, clusterId, hostname, addr, tunnel, tags = []} = json;
```

packages/teleport/src/services/nodes/makeNode.ts
```diff
@@ -14,25 +14,20 @@ See the License for the specific language governing permissions and
...
-    'siteId',
-    'hostname',
-    'tags',
-    'addr',
-    'tunnel',
-  ]);
+  const {id, clusterId, hostname, addr, tunnel} = json;
+
+  const tags = json.tags || [];
+  const tagTexts = json.tagTexts || tags.map(tag => `${tag.name}: ${tag.value}`)
```
kimlisa, line 23: the json response won't contain a field named `tagTexts`, so checking if that is undefined is not needed, and since we supplied a default value for field `tags` (if `tags` came undefined), we can just do `tags.map` without worrying about undefined/null errors

rudream: Yea, I wrote this before I found out the tests don't go through `makeNode`, so I though the `tagTexts` could be passed in

packages/teleport/src/Console/DocumentNodes/DocumentNodes.story.tsx
```diff
@@ -18,6 +18,7 @@ import React from 'react';
 import DocumentNodes from './DocumentNodes';
 import ConsoleCtx from './../consoleContext';
 import { TestLayout } from './../Console.story';
+import {Node} from "teleport/services/nodes/types"
```
kimlisa, line 21: i don't think your prettier is working. i see double quotes here, but with prettify on save, should be converted to single quotes. If you followed the readme, i think its missing config for typescript files.

Could you update our [README](https://github.com/gravitational/webapps/blob/master/README.md#setup-prettier-on-vscode) with the following:

```
// Set the default
    "editor.formatOnSave": false,
    // absolute config path
    "prettier.configPath": ".prettierrc",
    // enable per-language
    "[html]": {
        "editor.formatOnSave": true,
        "editor.defaultFormatter": "esbenp.prettier-vscode"
    },
    "[javascript]": {
        "editor.formatOnSave": true,
        "editor.defaultFormatter": "esbenp.prettier-vscode"
    },
    "[javascriptreact]": {
        "editor.formatOnSave": true,
        "editor.defaultFormatter": "esbenp.prettier-vscode",
    },
    "[typescript]": {
        "editor.formatOnSave": true,
        "editor.defaultFormatter": "esbenp.prettier-vscode"
    },
    "[typescriptreact]": {
        "editor.formatOnSave": true,
        "editor.defaultFormatter": "esbenp.prettier-vscode",
    },
    "[json]": {
        "editor.formatOnSave": true,
        "editor.defaultFormatter": "vscode.json-language-features"
    },
    "[jsonc]": {
        "editor.formatOnSave": true,
        "editor.defaultFormatter": "vscode.json-language-features"
    },
    "[markdown]": {
        "editor.formatOnSave": true,
        "editor.defaultFormatter": "esbenp.prettier-vscode",
    },
    "editor.tabSize": 2,
```

klizhentas:
@rudream congrats on the first PR!  🚀

packages/teleport/src/services/nodes/makeNode.ts
```diff
@@ -14,25 +14,19 @@ See the License for the specific language governing permissions and
...
+  const {id, clusterId, hostname, addr, tunnel, tags = [] } = json;
+
+  const tagTexts = tags.map(tag => `${tag.name}: ${tag.value}`)
 
   return {
     id,
     clusterId,
     hostname,
-    tags,
+    tagTexts,
```
alex-kovoy, line 28: i would keep the `tag` name as it is shorter. 

```ts
const tags = json.tags || [];
return {
   ...json,
   tags: tags.map(tag => `${tag.name}: ${tag.value}`),
}
```

rudream: Good idea. Done! :+1:

kimlisa (approved):
LGTM, but please wait for @alex-kovoy's approval before merge

btw, i found your gif very helpful :)

alex-kovoy (approved):

PR #280, opened 2021-04-23
Create database table and Storybook

### LINKED ISSUE #269: [DB Access 1] Table
# Table
- [x] Create a `Database` type that describes the database object under `services/database/types.ts`. We normally contain types within the directory that it is being used or in the directory where it is closest to data being originated. Later the `services/database/` will be responsible for processing responses from api endpoints.
- [x] Create a fixture that contains/exports mock data
- [x] Create a table component that lists all databases and has a search box. This table should be able to search by contents in each column, and each column should be sortable.
- [x] Create a parent component that displays the title `Databases` and renders the table component (similar to below image, minus the `add server` button and SSH input box). This component should take in a prop named `data: Database[ ]`.
- [x] Create stories for the parent component for states `Loaded` (aka success) 
  - `Loaded` is what the rendered state should look like when component successfully received data.
- [x] Snapshot the rendered storybook state and double check that snapshots are what you expected. Snapshots are enough here b/c our component is simple enough that we can create stories for different rendered states. By snapshotting these rendered states, we can catch unintentional side effects from making changes to components/css styling that may affect this component. If a change we make elsewhere affects these rendered states, the snapshot test will break, and then we can analyze the changes and decide if its expected or needs to be fixed. Snapshotting saves us a lot of time on repetitive behavioral testing. We save behavioral testing for components that are more complex.

(image took from Storybook > Teleport > Nodes > Loaded)
image

### Tip
I would start by creating bare minimum components, that takes in `data` prop (a database list `Database[ ]`) and create a storybook. 

For example:
1. Create the table component `DatabaseList.tsx` that takes in a `data` prop and just renders a "hello world". `console.log` the `data` prop here to later test that this component received the expected data.
2. Create the parent component `Database.tsx` that also takes in a `data` prop and import/render the table component. 
3. Create a storybook, import the parent component, feed the `data` prop with mock data. When you open this story, open dev console to check that the table component outputted the correct data.
4. Start fleshing it out. Anytime you save, storybook will auto reload so you can see your changes as you code.

### Mock Data For Table
This is a mocked data that would be passed to the table component. Used for storybooking:
```tsx
[
 {
    name: 'aurora',
    desc: 'PostgreSQL 11.6: AWS Aurora ',
    protocol: 'postgres',
    type: 'rds',
    uri: 'postgres-aurora-instance-1.c1xpjrob56xs.us-west-1.rds.amazonaws.com:5432',
    labels: [
      { name: 'cluster', value: 'root' },
      { name: 'env', value: 'aws' },
    ],
    labelTexts: [ 'cluster: root', 'env: aws' ],
  },
  {
    name: 'postgres-gcp',
    desc: 'MPostgreSQL 9.6: Google Cloud SQL',
    protocol: 'postgres',
    type: 'gcp',
    uri: '35.193.70.43:5432 ',
    labels: [{ name: 'env', value: 'gcp' }],
    labelTexts: [ 'cluster: env', 'value: gcp' ],
  },
  {
    name: 'mysql-aurora-56',
    desc: 'MySQL 5.6: AWS Aurora Longname For SQL',
    protocol: 'mysql',
    type: 'self-hosted',
    uri: 'mysql-aurora-56.cluster-c1xpjrob56xsadckjUkjk.us-west-1.rds.amazonaws.com:3306',
    labels: [
      { name: 'cluster', value: 'root' },
      { name: 'env', value: 'aws' },
    ],
    labelTexts: [ 'cluster: root', 'env: aws' ],
  }
]
````
### Resulting table
The below is how the data should be presented in table (ignore the difference in this table's data from mock). 

```
Name            Description                      Protocol Type        URI                                                                      Labels                    
--------------- -------------------------------- -------- ----------- ------------------------------------------------------------------------ -------------------------
aurora          PostgreSQL 11.6: AWS Aurora      postgres rds         postgres-aurora-instance-1.c1xpjrob56xs.us-west-1.rds.amazonaws.com:5432 cluster: root, env: aws   
local           PostgreSQL 13.0: Local           postgres self-hosted localhost:5432                                                           cluster: root, env: local 
mysql           MySQL 8.0: Local                 mysql    self-hosted localhost:3306                                                           cluster: root, env: local 
mysql-aurora    MySQL 5.7: AWS Aurora            mysql    rds         mysql-aurora.cluster-c1xpjrob56xs.us-west-1.rds.amazonaws.com:3306       cluster: root, env: aws   
mysql-aurora-56 MySQL 5.6: AWS Aurora            mysql    rds         mysql-aurora-56.cluster-c1xpjrob56xs.us-west-1.rds.amazonaws.com:3306    cluster: root, env: aws   
postgres-gcp    PostgreSQL 9.6: Google Cloud SQL postgres gcp         35.193.70.43:5432                                                        env: gcp                  
rds             PostgreSQL 12.4: AWS RDS         postgres rds         postgres-rds.c1xpjrob56xs.us-west-1.rds.amazonaws.com:5432               cluster: root, env: aws   
redshift        Amazon Redshift                  postgres redshift    redshift-cluster-1.ccr9ykwgjl84.us-east-1.redshift.amazonaws.com:5439    cluster: root, env: aws   
```

### PR DESCRIPTION
**Pull request by rudream**
### Purpose

This PR resolves #269 

Create and render a table that displays database nodes with details about them such as name, protocol, uri, etc. Also implement sorting and searching/filtering to narrow down the databases displayed. Also add it to Storybook for interactive testing.

### Implementation

Followed the same structure as the components used to render `Nodes`. Following similar structure and patterns as other similar components creates uniformity and makes code much more maintainable, and more readable for other engineers. Also created a Storybook for the `Loaded` state as well as added a basic test using a snapshot.

#### Notes:
- Since the `LabelCell` component is going to be implemented identically in `NodeList`, `DatabaseList` and `KubeList`, I moved it to `/design/src/DataTable` so that we can simply import from there and reuse it wherever necessary.
- I removed the `labels` and `labelTexts` fields of the `Database` interface and replaced them with a `tags` field just like I did [here](https://github.com/gravitational/webapps/blob/3ebfc3886ed5d222671ae11a63192419afbf71f5/packages/teleport/src/Nodes/fixtures/index.ts#L26) for `Node`. Keeping the naming convention and functionality consistent, and to prevent re-encountering the same issue I solved in #279 or `NodeList` caused by the `[{name: value}]` structure for tags.
- I noticed that the majority of the code for the lists in our tables could be reused (eg. layout, styling, searching logic, sorting logic). This is the case for:
   - NodeList.tsx
   - DatabaseList.tsx
   - KubeList.tsx (?)
   - EventList.tsx
   - RecordList.tsx
   - SessionList.tsx

  These lists really only differ in:
  1. The data and type of data (eg. databases)
  2. The number of columns (eg. 4 columns)
  3. The types of cells (eg. AddressCell, TextCell, LabelCell)
  
  This makes me think it could be worth considering creating a more generic `List` component that accepts the aforementioned 
  differences as props to render the correct list accordingly, this way we won't have to rewrite the same code in so many places. 
  Please let me know your opinion :slightly_smiling_face:  

### Demo

database-table-demo

### CODE REVIEW

File: packages/design/src/DataTable/Table.jsx [hunk 1]
```diff
@@ -170,6 +171,18 @@ const TextCell = props => {
   return <Cell {...rest}>{data[rowIndex][columnKey]}</Cell>;
 };
 
+const LabelCell = props => {
```
alex-kovoy, line 174: This component requires that an object must have a `tags` field which does not look to be generic enough to be part of the Table. Instead you could declare a function like below:

```ts
export function renderLabelCell(labels: []string){
 const $labels = tags.map(label => (
    <Label mb="1" mr="1" key={label} kind="secondary">
      {`${label}`}
    </Label>
  ));

  return <Cell>{$labels}</Cell>;  
}
```

And then use it like so in various places. 

```tsx
function LabelCell(props){
 const { rowIndex, data } = props;
 const { tags } = data[rowIndex];
 return renderLabelCell(tags);
}
```

Also, reasonable code duplication is OK :)

rudream: Good suggestion, done :+1:

packages/teleport/src/Databases/Databases.story.tsx
```diff
@@ -0,0 +1,37 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import React from 'react';
+import Databases from './Databases';
+import { databases } from './fixtures';
+
+export function Loaded() {
+  return render();
```
kimlisa, line 22: This file contains only one story, so we don't need to use a helper.

```
export const Loaded = () => <Databases databases={databases} rest of props.../>
```

You are probably thinking of next issue, but @alex-kovoy recently mentioned a shorter/simpler way of rendering same story with different props, so keep in mind (for the next issue) that this will be the new way we will define stories: https://storybook.js.org/docs/react/writing-stories/args

more about bind: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind

packages/teleport/src/Databases/Databases.tsx
```diff
@@ -0,0 +1,53 @@
...
+import {
+  FeatureBox,
+  FeatureHeader,
+  FeatureHeaderTitle,
+} from 'teleport/components/Layout';
+import { Flex } from 'design';
+import InputSearch from 'teleport/components/InputSearch';
+import DatabaseList from 'teleport/components/DatabaseList';
+
+export default function Databases(props) {
```
kimlisa, line 27: props needs to be typed here

packages/teleport/src/Databases/Databases.tsx
```diff
@@ -0,0 +1,53 @@
...
+        flex="0 0 auto"
+        justifyContent="space-between"
+      >
+        <InputSearch
+          mr="3"
+          onChange={e => {
+            setSearchValue(e);
+          }}
+        />
+      </Flex>
```
kimlisa, line 49: Lets put this inside the list component b/c the search box is for the table (i looked up NodesList and realized that the search box wasn't a part of the component, it looks like it was extracted for flexibility of placement b/c in `DocumentNodes` we have two input boxes to show above the table: search and quick launcher). 

Lookup `UserList.tsx` or `RequestList.tsx` and you'll see how it is done there.

packages/teleport/src/Databases/fixtures/index.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export const databases: Database[] = [
+  {
+    name: 'aurora',
+    desc: 'PostgreSQL 11.6: AWS Aurora ',
+    protocol: 'postgres',
+    type: 'rds',
+    uri:
+      'postgres-aurora-instance-1.c1xpjrob56xs.us-west-1.rds.amazonaws.com:5432',
+    tags: ['cluster: root', 'env: aws'],
```
kimlisa, line 27: actually, could we keep this as `labels`? I was mulling it over and not sure how `tags` originated since label is more accurate

packages/teleport/src/components/DatabaseList/DatabaseList.tsx
```diff
@@ -0,0 +1,156 @@
...
+  TextCell,
+  SortTypes,
+} from 'design/DataTable';
+import Table from 'design/DataTable/Paged';
+import isMatch from 'design/utils/match';
+import { sortBy } from 'lodash';
+import { Database } from 'teleport/services/database';
+
+function DatabaseList(props: Props) {
+  const { databases = [], searchValue = '', pageSize = 100 } = props;
```
kimlisa, line 33: hm, 100 seems a little too big a number before paging, i think we use 20 everywhere else

packages/teleport/src/components/DatabaseList/DatabaseList.tsx
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  function onSortChange(columnKey: string, sortDir: string) {
+    setSortDir({ [columnKey]: sortDir });
+  }
+
+  const data = sortAndFilter(searchValue);
+
+  return (
+    <div>
```
kimlisa, line 63: the `StyledTable` doesn't need to be wrapped with anything b/c it has no siblings. (i know in NodesList it does, not sure why, could be left over code that we forgot to clean up after making changes).

packages/teleport/src/components/DatabaseList/DatabaseList.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const { databases = [], searchValue = '', pageSize = 100 } = props;
+
+  const [sortDir, setSortDir] = React.useState<Record<string, string>>({
+    name: SortTypes.DESC,
+  });
+
+  function sortAndFilter(search) {
+    const filtered = databases.filter(obj =>
+      isMatch(obj, search, {
+        searchableProps: ['name', 'desc', 'protocol', 'type', 'uri', 'tags'],
```
kimlisa, line 42: we dont use a sort header cell for `tags` so it shouldn't be included in this list (i don't think it really makes sense to anyways and we don't in other tables that uses labels)

kimlisa (changes requested):

packages/teleport/src/Databases/fixtures/index.ts
kimlisa: lets get @alex-kovoy approval before you start renaming.. only to rename it back ;)

rudream: I agree, the [teleport docs refer to them as labels ](https://goteleport.com/docs/admin-guide/#labeling-nodes-and-applications) so it's more accurate. In #268 `tags` was being used so I just applied the same. I will also change the `NodeList` logic to use `labels` instead of tags so things are more consistent :slightly_smiling_face:

packages/teleport/src/components/DatabaseList/DatabaseList.tsx
rudream: You're right but it'll need a wrapper now that it has `<SearchInput />` as a sibling

packages/teleport/src/components/DatabaseList/DatabaseList.tsx
rudream: It needs to be in the list since it's a searchable prop (users can search by tags in the search box), the `searchAndFilterCb()` needs it

kimlisa: oh you are right, my bad :)

packages/teleport/src/Databases/fixtures/index.ts
rudream: Sounds good, I just reverted that commit

kimlisa (approved):
LGTM, but please wait for @alex-kovoy approval.

You might already know, but you don't have to wait for this PR to be approved to work on the next issue. Let me know if you need help with that.

alex-kovoy (approved):

PR #284, opened 2021-04-26
Use dedicated API for app FQDN resolving

andrejtokarcik:
Fixes https://github.com/gravitational/teleport/issues/6551, based on https://github.com/gravitational/teleport/pull/6449.

I tried to incorporate the recommendations from #235, feel free to adapt further according to your coding conventions.

### DIFF
--- a/packages/teleport/src/AppLauncher/useAppLauncher.ts
+++ b/packages/teleport/src/AppLauncher/useAppLauncher.ts
@@ -25,49 +25,49 @@ export default function useAppLauncher() {
   const params = useParams<UrlLauncherParams>();
   const { attempt, setAttempt } = useAttempt('processing');
 
-  const location = window.location;
-  const port = location.port ? ':' + location.port : '';
-  const state = getUrlParameter('state', location.search);
-
   React.useEffect(() => {
-    if (!state) {
-      service
-        .getAppFqdn(params)
-        .then(result => {
-          const url = new URL(`https://${result.fqdn}${port}/x-teleport-auth`);
-          if (params.clusterId) {
-            url.searchParams.set("cluster", params.clusterId);
-          }
-          if (params.publicAddr) {
-            url.searchParams.set("addr", params.publicAddr);
-          }
-          location.replace(url.toString());
-        })
-        .catch((err: Error) => {
-          setAttempt({
-            status: 'failed',
-            statusText: err.message,
-          });
+    resolveRedirectUrl(params)
+      .then(url => {
+        window.location.replace(url);
+      })
+      .catch((err: Error) => {
+        setAttempt({
+          status: 'failed',
+          statusText: err.message,
         });
-    } else {
-      service
-        .createAppSession(params)
-        .then(result => {
-          const url = new URL(`https://${result.fqdn}${port}/x-teleport-auth`);
-          url.searchParams.set("state", state);
-          url.hash = `#value=${result.value}`;
-          location.replace(url.toString());
-        })
-        .catch((err: Error) => {
-          setAttempt({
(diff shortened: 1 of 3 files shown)


packages/teleport/src/services/apps/apps.ts
```diff
@@ -39,11 +39,19 @@ const service = {
...
+      .then(json => ({
+        fqdn: json.fqdn as string,
+      }));
+  },
 };
 
 export default service;
 
-type CreateAppSessionParams = {
+type AppParams = {
```
kimlisa, line 54: Since we exported the `UrlAppParams` defined in our `config.ts`, we can import it here and use that instead and avoid dup type

kimlisa (commented):
@andrejtokarcik thank you for remembering about the recommendations and applying it here! :)

packages/teleport/src/config.ts
```diff
@@ -76,6 +76,7 @@ const cfg = {
 
   api: {
     aapSession: '/v1/webapi/sessions/app',
+    aapFqdnPath: '/v1/webapi/apps/:fqdn/:clusterId?/:publicAddr?',
```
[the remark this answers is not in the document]
alex-kovoy: I agree with @kimlisa, lets keep it consistent.

packages/teleport/src/AppLauncher/useAppLauncher.ts
```diff
@@ -25,29 +25,46 @@ export default function useAppLauncher() {
...
-          window.location.replace(`${authUrl}?cluster=${clusterId}&addr=${publicAddr}`);
-        } else {
-          window.location.replace(`${authUrl}?state=${state}#value=${result.value}`);
-        }
-      })
-      .catch((err: Error) => {
-        setAttempt({
-          status: 'failed',
-          statusText: err.message,
+    if (!state) {
```
alex-kovoy, line 33: @andrejtokarcik  Could you update this function with below code and add some comments `// comment here` ? 

```ts
export default function useAppLauncher() {
  const params = useParams<UrlLauncherParams>();
  const { attempt, setAttempt } = useAttempt('processing');

  React.useEffect(() => {
    resolveRedirectUrl(params)
      .then(url => {
        window.location.replace(url);
      })
      .catch((err: Error) => {
        setAttempt({
          status: 'failed',
          statusText: err.message,
        });
      });
  }, []);

  return {
    ...attempt,
  };
}

function resolveRedirectUrl(params: UrlLauncherParams) {
  const state = getUrlParameter('state', location.search);

  // comment here
  if (!state) {
    return service.getAppFqdn(params).then(result => {
      const url = new URL(`https://${result.fqdn}${port}/x-teleport-auth`);
      if (params.clusterId) {
        url.searchParams.set('cluster', params.clusterId);
      }
      if (params.publicAddr) {
      }

      return url.toString();
    });
  }

  // comment here
  return service.createAppSession(params).then(result => {
    const url = new URL(`https://${result.fqdn}${port}/x-teleport-auth`);
    url.searchParams.set('state', state);
    url.hash = `#value=${result.value}`;
    return url.toString();
  });
}
```

packages/teleport/src/config.ts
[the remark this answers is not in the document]
andrejtokarcik: Did a general `aap => app` rename, hope that's alright!

packages/teleport/src/AppLauncher/useAppLauncher.ts
andrejtokarcik: Done.

alex-kovoy (approved):

Issue #271, opened 2021-04-21
[DB Access 3] Define route, endpoint, and service


kimlisa:
# Define route, api endpoint, and service

In `config.ts`:
- [x] Define a database `route` and `api endpoint` and its getters. Since database viewing is cluster specific, remember to pass in the `clusterId` to this route and endpoint.

`services/database/` is a directory that contains all logic for retrieving/processing/typing response values from api:
- [x] Create a database class that will define a fetch method to get a list of all database servers for the specified clusterId.
- [x] The response value from the fetch method (or any api responses really) will need to be processed. The convention is to create a separate file named `make.ts`, and it will contain methods named `makeXXX` that will be responsible for processing the responses we get from api. There are variety of reasons why we want to process the response values:
   1. Arrays may come as `undefined` and needs to be replaced with an empty array to avoid accessing undefined props errors. 
   1. Setting default values to objects if undefined
   1. Sometimes json response contains field names containing underscore ie. `expires_in`. JS conventions dictates we use camelCase, so we can rename this field to `expiresIn: json.expires_in`
   1. Sometimes with the response given, we want to add an extra field. For example, if the response value contained a field `date: Date`, and we wanted a formatted string version of this date, we can add a field `dateText: stringValue` (while keeping the original data). These text values often helps with table searching.
   1. Sometimes we want to reformat the response value
   1. All of the above helps create consistent responses sent to our components
- [x] Instantiate this class in our `teleportContext.tsx` file, so that children components can have access to it.
- [x] Test the class you made, to ensure the response received from api, is processed and formatted as you expected.

### Table Cell formatting
- [x] Format the database columns `protocol` and `type` cells to the following:

#### Protocol Formatting
```
postgres -> PostgreSQL
mysql -> MySQL
```

#### Type Formatting
```
self-hosted -> Self-hosted
rds -> RDS
gcp -> GCP
redshift -> Redshift
```

### Mocked response
This is what response data would look like from the api
```tsx
[
 {
    name: 'aurora',
    desc: 'PostgreSQL 11.6: AWS Aurora ',
    protocol: 'postgres',
    type: 'rds',
    uri: 'postgres-aurora-instance-1.c1xpjrob56xs.us-west-1.rds.amazonaws.com:5432',
    labels: [
      { name: 'cluster', value: 'root' },
      { name: 'env', value: 'aws' },
    ],
  },
]
````

This is what data should look like after being processed:
```tsx
[
 {
    name: 'aurora',
[...]

PR #287, opened 2021-04-26
Add database table attempt states (Success, Loading, Failed)

### LINKED ISSUE #270: [DB Access 2] Table's Attempt States
# Table's Attempt States
Currently our table assumes only a `loaded` (success) state and does not have states for loading or errors. 

When we are retrieving data from the api, the response time could potentially be a few seconds or more depending on the internet or size of data. To indicate to user we are waiting for a response, we want to show a `loading` state of the component similar to below:

(image took from Storybook > Teleport > Nodes > Loading)
image

And when we get an error response, we want to show the user that there was an error, and display the error message, similar to below.

(image took from Storybook > Teleport > Nodes > Failed)
image

We have a [useAttemptNext](url) hook that handles the response's three states: `success`, `processing`, and `failed`. 

For this PR, lets not use the `useAttempt` hook, but provide it as a prop named `attempt: Attempt` to `Database/Database.tsx` component, so that we can manipulate the attempts states to display them on storybook. (We will be using the hook later when we create our custom hook).

- [x] Add prop `attempt: Attempt` ([Attempt type](https://github.com/gravitational/webapps/blob/master/packages/shared/hooks/useAttemptNext.ts#L54))
- [x] Create stories to showcase `Loading` state and `Error` state
    - `Loading` is what the rendered state should look like when components are still waiting to receive data
    - `Failed` is what the rendered state should look like when we received an error
- [x] Create snapshot of Error state (skip `Loading`)

rudream:
### Purpose

This PR resolves #270 

Add attempt states for the Database table. If the state is `Success`, display the list of databases. If the state is `Loading`, display  an animated loading indicator. If the state is `Failed` display an alert box with the error message. Add snapshot for `Failed` state.

### Implementation

Use prop `attempt` to contain information about the attempt (eg. status, statusText). When rendering the `Databases`, render the appropriate UI based on the `attempt.state`. Create a snapshot test for `Failed` state.

#### Loaded State

successState

#### Loading State

loadingState

#### Failed State

failedState

### DIFF
--- a/packages/teleport/src/Databases/Databases.story.tsx
+++ b/packages/teleport/src/Databases/Databases.story.tsx
@@ -18,8 +18,28 @@ import React from 'react';
 import Databases from './Databases';
 import { databases } from './fixtures';
 
-export const Loaded = () => <Databases databases={databases} />;
+type PropTypes = Parameters<typeof Databases>[0];
+
+export const Loaded = () => render({ status: 'success' });
+
+export const Loading = () => render({ status: 'processing' });
+
+export const Failed = () =>
+  render({ status: 'failed', statusText: 'Server Error' });
 
 export default {
   title: 'Teleport/Databases',
 };
+
+function render(attemptOptions: Partial<PropTypes['attempt']>) {
+  const props = {
+    databases,
+    attempt: {
+      status: '' as any,
+      statusText: '',
+      ...attemptOptions,
+    },
+  };
+
+  return <Databases {...props} />;
+}
--- a/packages/teleport/src/Databases/Databases.story.tsx
+++ b/packages/teleport/src/Databases/Databases.story.tsx
@@ -18,28 +18,21 @@ import React from 'react';
 import Databases from './Databases';
 import { databases } from './fixtures';
 
-type PropTypes = Parameters<typeof Databases>[0];
+export const Loaded = () => (
+  <Databases databases={databases} attempt={{ status: 'success' }} />
+);
 
-export const Loaded = () => render({ status: 'success' });
+export const Loading = () => (
+  <Databases databases={databases} attempt={{ status: 'processing' }} />
+);
 
-export const Loading = () => render({ status: 'processing' });
-
-export const Failed = () =>
-  render({ status: 'failed', statusText: 'Server Error' });
+export const Failed = () => (
+  <Databases
+    databases={databases}
+    attempt={{ status: 'failed', statusText: 'Server Error' }}
+  />
+);
 
 export default {
   title: 'Teleport/Databases',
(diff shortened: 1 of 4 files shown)


packages/teleport/src/Databases/Databases.story.tsx
```diff
@@ -18,8 +18,28 @@ import React from 'react';
...
+    databases,
+    attempt: {
+      status: '' as any,
+      statusText: '',
+      ...attemptOptions,
+    },
+  };
+
+  return <Databases {...props} />;
+}
```
kimlisa, line 45: Because our component is not very complicated and has only two props, we can just pass in props manually:

```
export const Failed = () => <Databases databases={databases} attempt={{ status: 'failed', statusText: 'Server Error' }}/>

...etc
```

We could've also simplified our helper like this (this is in fact how we normally do it):

```
export const Failed = () => <Databases {...props} attempt={{ status: 'failed', statusText: 'Server Error' }}/>

const props = {
  attempt: { 
   status: 'success' as any,
   statusText: ''
 },
 databases,
}
```
(still too much for our simple component though). I know the example in Nodes do it the way you've done it, but sometimes codes can be simplified.

rudream: Done :+1:

packages/teleport/src/Databases/Databases.tsx
```diff
@@ -15,27 +15,37 @@ limitations under the License.
...
       <FeatureHeader alignItems="center" justifyContent="space-between">
         <FeatureHeaderTitle>Databases</FeatureHeaderTitle>
       </FeatureHeader>
-      <DatabaseList databases={databases} />
+      {attempt.status === 'processing' && (
+        <Box textAlign="center" m={10}>
+          <Indicator />
+        </Box>
+      )}
+      {attempt.status === 'failed' && <Danger>{attempt.statusText} </Danger>}
```
kimlisa, line 42:
```suggestion
      {attempt.status === 'failed' && <Danger>{attempt.statusText}</Danger>}
```

kimlisa (approved):

r0mant:
@kimlisa @rudream Table is looking great! Could I make a small suggestion to format protocol and type with proper capitalization please? I.e.

```
postgres -> PostgreSQL
mysql -> MySQL
self-hosted -> Self-hosted
rds -> RDS
gcp -> GCP
redshift -> Redshift
```

Thanks!

kimlisa:
@r0mant included with issue 3: https://github.com/gravitational/webapps/issues/271 

also @rudream please make sure to delete your branches after merge

PR #288, opened 2021-04-27
Create database route, endpoint and service

### LINKED ISSUE #271: [DB Access 3] Define route, endpoint, and service

### LINKED ISSUE #283: [Kb Access 3] Define route, endpoint, and service
# Define route, api endpoint, and service

In `config.ts`:
- [x] Define a kubernetes `route` and `api endpoint` and its getters. Since kubernetes viewing is cluster specific, remember to pass in the `clusterId` to this route and endpoint.

`services/kubernetes/` is a directory that contains all logic for retrieving/processing/typing response values from api:
- [x] Create a kubernetes class that will define a fetch method to get a list of all kube servers for the specified clusterId.
- [x] The response value from the fetch method (or any api responses really) will need to be processed. The convention is to create a separate file named `make.ts`, and it will contain methods named `makeXXX` that will be responsible for processing the responses we get from api. There are variety of reasons why we want to process the response values:
   1. Arrays may come as `undefined` and needs to be replaced with an empty array to avoid accessing undefined props errors. 
   1. Setting default values to objects if undefined
   1. Sometimes json response contains field names containing underscore ie. `expires_in`. JS conventions dictates we use camelCase, so we can rename this field to `expiresIn: json.expires_in`
   1. Sometimes with the response given, we want to add an extra field. For example, if the response value contained a field `date: Date`, and we wanted a formatted string version of this date, we can add a field `dateText: stringValue` (while keeping the original data). These text values often helps with table searching.
   1. All of the above helps create consistent responses sent to our components
- [x] Instantiate this class in our `teleportContext.tsx` file, so that children components can have access to it.
- [x] Test the class you made, to ensure the response received from api, is processed and formatted as you expected.

### Mocked response
This is what response data would look like from the api
```tsx
[
  {
    name: 'tele.logicoma.dev-prod',
    labels: [
      { name: 'kernal', value: '4.15.0-51-generic' },
      { name: 'env', value: 'prod' },
    ],
  },
]
````

This is what data should look like after being processed:
```tsx
[
  {
    name: 'tele.logicoma.dev-prod',
    labels: [ 'kernal: 4.15.0-51-generic', 'env: prod' ],
  },
]
````

### Note
We won't be using this service until the next issue.

### PR DESCRIPTION
**Pull request by rudream**
### Purpose

This PR resolves #271  

Create a route and endpoint to for fetching databases. Process the response so that all fields are properly set up and formatted for front-end displaying and searching. 

### Implementation

Define an api endpoint in `config.ts` to be used by our database service for fetching databases for a given `:clusterId`. Create `makeDatabase.ts` to process the response data such that:
- An undefined `labels` field defaults to an empty array `[]`
- Labels formatted as `[{name: value}]` are turned to `["name: value"]`
- Type and protocol fields are optimally formatted (eg. 'postgres' -> 'PostgreSQL')

Call this `makeDatabase()` to process the response when fetching databases in the database service.

The route and route getter I defined in `config.ts` will be used in the next issue when registering the feature.

### Demo

If `makeDatabase()` is given the following data:

```tsx
{
    name: 'aurora',
    desc: 'PostgreSQL 11.6: AWS Aurora ',
    protocol: 'postgres',
    type: 'rds',
    uri: 'postgres-aurora-instance-1.c1xpjrob56xs.us-west-1.rds.amazonaws.com:5432',
    labels: [
      { name: 'cluster', value: 'root' },
      { name: 'env', value: 'aws' },
    ],
}
````

it will return:

```tsx
{
    name: 'aurora',
    desc: 'PostgreSQL 11.6: AWS Aurora ',
    protocol: 'PostgreSQL',
    type: 'RDS',
    uri: 'postgres-aurora-instance-1.c1xpjrob56xs.us-west-1.rds.amazonaws.com:5432',
    tags: ['cluster: root',  'env: aws']
}
````

formatteddatabases

### CODE REVIEW

File: packages/teleport/src/services/database/databases.ts [hunk 1]
```diff
@@ -0,0 +1,30 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { map } from 'lodash';
+import api from 'teleport/services/api';
+import cfg from 'teleport/config';
+import makeDatabase from './makeDatabase';
+
+const service = {
```
kimlisa, line 22: hmm... is this a class?

packages/teleport/src/services/database/makeDatabase.ts
```diff
@@ -0,0 +1,47 @@
...
+    case 'self-hosted':
+      return 'Self-hosted';
+    case 'rds':
+      return 'RDS';
+    case 'gcp':
+      return 'GCP';
+    case 'redshift':
+      return 'Redshift';
+  }
+};
```
kimlisa, line 47: lets separate these to `formatProtocol` and `formatType`, if none match, return the value as is

alex-kovoy:
image


Lets combine these 2 fields into 1 and call it `TYPE` and make sure that search works when a user types `rds/p` for example.

```
RDS/PosgreSQL
GCP/PosgreSQL
MySQL
```

rudream:
### Update

Upon @alex-kovoy 's suggestion, the type and protocol fields are now combined. 

If `makeDatabase()` is given:

```tsx
{
    name: 'aurora',
    desc: 'PostgreSQL 11.6: AWS Aurora ',
    protocol: 'postgres',
    type: 'rds',
    uri: 'postgres-aurora-instance-1.c1xpjrob56xs.us-west-1.rds.amazonaws.com:5432',
    labels: [
      { name: 'cluster', value: 'root' },
      { name: 'env', value: 'aws' },
    ],
}
````

it will return:

```tsx
{
    name: 'aurora',
    desc: 'PostgreSQL 11.6: AWS Aurora ',
    type: 'RDS/PostgreSQL',
    uri: 'postgres-aurora-instance-1.c1xpjrob56xs.us-west-1.rds.amazonaws.com:5432',
    tags: ['cluster: root',  'env: aws']
}
````

### Screenshot of new table

updateddatabases

**Comment by r0mant** [comment]
@rudream @kimlisa I think combining type/protocol is a good idea. A couple of suggestions:

* Let's remove the slash? So it would say `Self-hosted MySQL`, `RDS PostgreSQL`, etc. which reads a bit better.
* Can we please also rename `GCP` to `Cloud SQL`? It would be more accurate. Ideally I need to rename it on the backend as well but I'm stuck with it for now. So for the frontend just translate "gcp" to "Cloud SQL" is possible.

**Comment by r0mant** [comment]
Also, didn't we agree not to include the database endpoint (URI) in the table? I don't think it has much value for the user - e.g. `tsh db ls` doesn't show it by default.

**Comment by kimlisa** [comment]
@r0mant the connect button is in a separate issue: https://github.com/gravitational/webapps/issues/277 (add database button is also a separate issue) 

and @rudream along with other changes, lets remove the URI column, i must've forgotten about removing it, sorry about that :)

**Comment by rudream** [comment]
### Update

After implementing @r0mant 's suggestions:

The `DatabaseList` no longer displays the URI column. 

If `makeDatabase()` is given:

```tsx
{
    name: 'aurora',
    desc: 'PostgreSQL 11.6: AWS Aurora ',
    protocol: 'postgres',
    type: 'rds',
    uri: 'postgres-aurora-instance-1.c1xpjrob56xs.us-west-1.rds.amazonaws.com:5432',
    labels: [
      { name: 'cluster', value: 'root' },
      { name: 'env', value: 'aws' },
    ],
}
````

it will return:

```tsx
{
    name: 'aurora',
    desc: 'PostgreSQL 11.6: AWS Aurora ',
    type: 'RDS PostgreSQL',
    uri: 'postgres-aurora-instance-1.c1xpjrob56xs.us-west-1.rds.amazonaws.com:5432',
    tags: ['cluster: root',  'env: aws']
}
````

### Screenshot of new table

updatedupdatedtable

File: packages/teleport/src/services/database/types.ts [hunk 3]
```diff
@@ -17,7 +17,6 @@ limitations under the License.
 export interface Database {
   name: string;
   desc: string;
-  protocol: string;
   type: string;
   uri: string;
```
kimlisa, line 21: don't need the uri field anymore

rudream: Thought we might need it later for connecting to the DB

rudream:
### Update

After removing the `uri` field as suggested by @kimlisa 

If `makeDatabase()` is given:

```tsx
{
    name: 'aurora',
    desc: 'PostgreSQL 11.6: AWS Aurora ',
    protocol: 'postgres',
    type: 'rds',
    uri: 'postgres-aurora-instance-1.c1xpjrob56xs.us-west-1.rds.amazonaws.com:5432',
    labels: [
      { name: 'cluster', value: 'root' },
      { name: 'env', value: 'aws' },
    ],
}
````

it will return:

```tsx
{
    name: 'aurora',
    desc: 'PostgreSQL 11.6: AWS Aurora ',
    type: 'RDS PostgreSQL',
    tags: ['cluster: root',  'env: aws']
}
````

File: packages/teleport/src/services/database/types.ts (hunk 3 above)
Comment (kimlisa) [reply packages/teleport/src/services/database/types.ts]: We can add it back in the next issue if need

File: packages/teleport/src/services/database/databases.test.ts [hunk 4]
```diff
@@ -0,0 +1,51 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import DatabaseService from './databases';
+import api from 'teleport/services/api';
+
+test('correct formatting of database fetch response', async () => {
```
kimlisa, line 20: i'd check all possible combo of protocol type and instead of matching the entire object, i'd just check the `type` field

packages/teleport/src/services/database/databases.test.ts
```diff
@@ -33,6 +33,31 @@ test('correct formatting of database fetch response', async () => {
...
+    ${'rds'}             | ${'mysql'}               | ${'RDS MySQL'}
+    ${'gcp'}             | ${'mysql'}               | ${'Cloud SQL MySQL'}
+    ${'redshift'}        | ${'mysql'}               | ${'Redshift MySQL'}
+    ${'self-hosted'}     | ${'postgres'}            | ${'Self-hosted PostgreSQL'}
+    ${'rds'}             | ${'postgres'}            | ${'RDS PostgreSQL'}
+    ${'gcp'}             | ${'postgres'}            | ${'Cloud SQL PostgreSQL'}
+    ${'redshift'}        | ${'postgres'}            | ${'Redshift PostgreSQL'}
+    ${'some other type'} | ${'some other protocol'} | ${'some other type some other protocol'}
+  `(
+    'should combine type and protocol correctly',
```
kimlisa, line 49: if i remember correctly, when we run this test, the output will be:
```
correct formatting of all type...
   should combine
   should combine
   ....
```

There is no differentiating messages for each test, so when it fails, i won't know what test fails. global search for `test.each` and you can see how to access make a more unique message per test

packages/teleport/src/services/database/databases.test.ts
```diff
@@ -33,6 +33,31 @@ test('correct formatting of database fetch response', async () => {
...
+    async ({ type, protocol, combined }) => {
+      jest.spyOn(api, 'get').mockResolvedValue({ items: [{ type, protocol }] });
```
kimlisa, line 51: also, the api response won't be wrapped in `items`. I know some responses do, some don't (the ones i implemented recently), i double checked with alexey and it isn't necessary

packages/teleport/src/services/database/databases.test.ts
rudream: It does show which test fails, for example:

image

I can still add the unique message per test for convenience though

packages/teleport/src/services/database/databases.test.ts
```diff
@@ -58,19 +58,17 @@ describe('correct formatting of all type and protocol combos', () => {
...
-      uri:
-        'postgres-aurora-instance-1.c1xpjrob56xs.us-west-1.rds.amazonaws.com:5432',
-      labels: [
-        { name: 'cluster', value: 'root' },
-        { name: 'env', value: 'aws' },
-      ],
-    },
-  ],
-};
+const mockResponse = [
```
kimlisa, line 61: Also add a test, if response comes back undefined. Arrays in particular may come as undefined.

kimlisa: If you want to know why: https://github.com/golang/go/issues/37711

kimlisa (approved):
one small nit: rename `services/database/` to `services/databases/` and you're good to go!

PR #289, opened 2021-04-27
Define route, endpoint and service for kubernetes

### LINKED ISSUE #283: [Kb Access 3] Define route, endpoint, and service

amha19:
**Description**

- Defines a kubernetes route and api endpoint that takes in a cluster id.
- It defines getter method for the route and the api.
- Creates a fetch method and fetch response processing method.

Given this:
```
[
  {
    name: 'tele.logicoma.dev-prod',
    tags: [
      { name: 'kernal', value: '4.15.0-51-generic' },
      { name: 'env', value: 'prod' },
    ],
  },
]
```

Will output this:

```
[
  {
    name: 'tele.logicoma.dev-prod',
    tags: [ 'kernal: 4.15.0-51-generic', 'env: prod' ],
  },
]
```

And it adds this service to the teleport context.

This PR closes #283


packages/teleport/src/services/kube/kubernetes.ts
```diff
@@ -0,0 +1,30 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { map } from 'lodash';
+import api from 'teleport/services/api';
+import cfg from 'teleport/config';
+import makeKubernetes from './makeKubernetes';
+
+const service = {
```
kimlisa, line 22: hmm... is this a class?

amha19: No, it is not :disappointed:. I was following Nodes structure, I forgot it was supposed to be a class.

packages/teleport/src/services/kube/index.ts
```diff
@@ -14,4 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+import KubernetesService from './kube';
```
kimlisa, line 17: Lets rename to `KubeService`

packages/teleport/src/config.ts
```diff
@@ -282,6 +288,10 @@ const cfg = {
     return generatePath(cfg.api.rolesPath, { name });
   },
 
+  getClusterKubernetesUrl(clusterId: string) {
```
kimlisa, line 291:
```suggestion
  getKubernetesUrl(clusterId: string) {
```

packages/teleport/src/services/kube/makeKube.ts
```diff
@@ -0,0 +1,26 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { Kube } from './types';
+
+export default function makeKube(json): Kube {
+  const { name, tags = [] } = json;
```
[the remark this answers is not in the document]
amha19: I've now changed the coming json field from `tags` to `labels`.

packages/teleport/src/services/kube/kube.test.ts
```diff
@@ -0,0 +1,42 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import KubeService from './kube';
+import api from 'teleport/services/api';
+
```
kimlisa, line 19: Also add a test, if response comes back undefined. Arrays in particular may come as undefined.

kimlisa: if you want to know why: https://github.com/golang/go/issues/37711

packages/teleport/src/services/kube/makeKube.ts
```diff
@@ -17,7 +17,8 @@ limitations under the License.
 import { Kube } from './types';
 
 export default function makeKube(json): Kube {
-  const { name, labels = [] } = json;
+  const { name } = json;
+  const labels = json.labels ? json.labels : [];
```
kimlisa, line 21:
```suggestion
  const labels = json.labels || [];
```

simplified. as stated earlier, `undefined` and `null` are falsey values, so if false, empty array will be assigned

packages/teleport/src/services/kube/kube.test.ts
```diff
@@ -17,18 +17,43 @@ limitations under the License.
...
-  const kubeService = new KubeService();
-  const response = await kubeService.fetchKubernetes('clusterId');
-
-  expect(response).toEqual([
-    {
-      name: 'tele.logicoma.dev-prod',
-      tags: ['kernal: 4.15.0-51-generic', 'env: prod'],
-    },
-  ]);
+describe('correct processed fetch response', () => {
```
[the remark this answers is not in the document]
amha19: Good to know. Thanks!

kimlisa (approved):
LGTM

PR #292, opened 2021-04-29
Create useDatabases hook, container, Register feature and update Main.story

### LINKED ISSUE #272: [DB Access 4] Define custom hook, register feature, update Main.story
# Define custom hook, register feature, update Main.story
You'll notice throughout our components that we use this pattern where we encapsulate logic required to retrieve the data from the api and retrieve the data that is required from our `TeleportContext`. We encapsulate like this to make storybooking and testing easier, it removes dependencies like mocking endpoints and contexts when we render the main component, and allows us to manipulate data easier (by providing our own values). Any data that we need to manipulate for storybooking and testing can be placed into this custom hook to be manually given to our main component.

- [x] create a custom hook that on mount, sets `data`, uses `useAttemptNext` hook, and export fields that the main component may need
- [x] In `Database.tsx` create a component named `Container`. This will be the default exported Container that runs the custom hook and pass the data given back from hook as props to main `Database` component.
- [x] Register database feature to our dashboard sidenav (as shown in image)
- [x] Use database Icon for the sidenav. All our icons are displayed in Storybook > Design > Icon > List of Icons, and the name of the icon is the name of the icon component
- [x] Test your custom hook integrating with our main component in Main.story.tsx. `Main.story` is like running webapps locally, but in storybook. In this story, you should see your database feature on the sidenav, be able to click on it, and it should render the database screen and be able to interact with it.
- [x] Any OS feature, is also a Enterprise feature. We need to also include this feature in Enterprise. This will require you to make a PR against `webapps.e` as well, just for this small code bit. (This also requires you to update `Main.story.ts` for enterprise version)
- [x] Update `webapps.e` reference since we made changes to `webapps.e`

image

rudream:
### Purpose

This PR resolves #272  

Create a custom `useDatabases` hook. Create `Container` component. Register databases as a feature and add it to the side nav. Update Main.story to include it and to use the hook. Updated the enterprise submodule as well.

### Implementation

Created a `useDatabases` hook that uses `useAttemptNext` to set the `attempt` state based on the success of the database fetching, in order to display an error or a loading icon. Registered databases as a feature by creating a `FeatureDatabases` class in `features.ts`, which also adds it to the side nav. Also added it to `Main.story`. Updated the `feature.ts` and `Main.story` of `webapps.e` and made a PR to that repo which you can find [here](https://github.com/gravitational/webapps.e/pull/47). The e-ref will be updated later when merging.

Also added support for the edge case where the `labels` field is `null` because of a quirk with Go that causes responses to contain `null` for arrays instead of `undefined`. 

### Demo

image

### DIFF
--- a/packages/teleport/src/Databases/Databases.tsx
+++ b/packages/teleport/src/Databases/Databases.tsx
@@ -34,7 +34,7 @@ export default function Container() {
   return <Databases {...data} />;
 }
 
-export function Databases(props: Props) {
+export function Databases(props: ReturnType<typeof useDatabases>) {
   const { databases, attempt } = props;
 
   return (
@@ -52,8 +52,3 @@ export function Databases(props: Props) {
     </FeatureBox>
   );
 }
-
-type Props = {
-  databases: Database[];
-  attempt: Attempt;
-};
--- a/packages/teleport/src/Databases/Databases.tsx
+++ b/packages/teleport/src/Databases/Databases.tsx
@@ -17,15 +17,13 @@ limitations under the License.
 import React from 'react';
 import { Indicator, Box } from 'design';
 import { Danger } from 'design/Alert';
+import useTeleport from 'teleport/useTeleport';
 import {
   FeatureBox,
   FeatureHeader,
   FeatureHeaderTitle,
 } from 'teleport/components/Layout';
-import useTeleport from 'teleport/useTeleport';
-import { Attempt } from 'shared/hooks/useAttemptNext';
 import DatabaseList from './DatabaseList';
-import { Database } from 'teleport/services/databases';
 import useDatabases from './useDatabases';
 
 export default function Container() {
--- a/packages/teleport/src/Databases/Databases.tsx
+++ b/packages/teleport/src/Databases/Databases.tsx
@@ -24,7 +24,7 @@ import {
   FeatureHeaderTitle,
 } from 'teleport/components/Layout';
 import DatabaseList from './DatabaseList';
-import useDatabases from './useDatabases';
+import useDatabases, { Props } from './useDatabases';
 
 export default function Container() {
   const ctx = useTeleport();
@@ -32,7 +32,7 @@ export default function Container() {
   return <Databases {...data} />;
 }
 
-export function Databases(props: ReturnType<typeof useDatabases>) {
+export function Databases(props: Props) {
   const { databases, attempt } = props;
 
   return (
--- a/packages/teleport/src/Databases/Databases.tsx
+++ b/packages/teleport/src/Databases/Databases.tsx
@@ -22,13 +22,15 @@ import {
   FeatureHeader,
   FeatureHeaderTitle,
 } from 'teleport/components/Layout';
+import useTeleport from 'teleport/useTeleport';
+import { Attempt } from 'shared/hooks/useAttemptNext';
 import DatabaseList from './DatabaseList';
 import { Database } from 'teleport/services/databases';
-import { Attempt } from 'shared/hooks/useAttemptNext';
 import useDatabases from './useDatabases';
 
 export default function Container() {
-  const data = useDatabases();
+  const ctx = useTeleport();
+  const data = useDatabases(ctx);
   return <Databases {...data} />;
 }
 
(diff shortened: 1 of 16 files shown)


packages/teleport/src/services/databases/databases.test.ts
```diff
@@ -67,6 +67,24 @@ describe('correct formatting of all type and protocol combos', () => {
...
+  jest.spyOn(api, 'get').mockResolvedValue([{ labels: null }]);
+
+  const database = new DatabaseService();
+  const response = await database.fetchDatabases('im-a-cluster');
+
+  expect(response[0].tags).toEqual([]);
+});
+
+test('undefined labels field in database fetch response', async () => {
+  jest.spyOn(api, 'get').mockResolvedValue([{ labels: undefined }]);
```
kimlisa, line 80: we don't need to test for undefined from our api responses. `null` is an assignable value that says "no value" vs `undefined` that says value hasn't been assigned yet: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined#description

kimlisa:
```
Also added support for the edge case where the labels field is null 
because of a bug with Go that causes responses to contain null for arrays instead of undefined.
```

Not a bug, that's just how [go json marshaller](https://golang.org/pkg/encoding/json/) works:
```
Array and slice values encode as JSON arrays, except that []byte 
encodes as a base64-encoded string, and a nil slice encodes as the null JSON value.
```

packages/teleport/src/services/databases/databases.test.ts
kimlisa: also: https://stackoverflow.com/a/14946821

packages/teleport/src/Databases/Databases.tsx
```diff
@@ -22,13 +22,15 @@ import {
   FeatureHeader,
   FeatureHeaderTitle,
 } from 'teleport/components/Layout';
+import useTeleport from 'teleport/useTeleport';
```
kimlisa, line 25: check the order of imports

packages/teleport/src/Databases/Databases.tsx
```diff
@@ -22,13 +22,15 @@ import {
...
+import { Attempt } from 'shared/hooks/useAttemptNext';
```
kimlisa, line 26: since we have a custom hook, that exports fields needed by our main component, we can use the return type of our custom hook as our prop now

packages/teleport/src/features.ts
```diff
@@ -376,10 +377,37 @@ export class FeatureTrust {
...
+  }
+
+  route = {
+    title: 'Databases',
+    path: cfg.routes.databases,
+    exact: true,
+    component: Databases,
+  };
+
+  register(ctx: Ctx) {
```
kimlisa, line 392: db feature is similar to apps where it can be enabled/disabled depending on the users RBAC rules. Their role can say `disable database feature`, then we shouldn't allow user to access this feature on UI. Examine how Apps is being registered

packages/teleport/src/services/user/makeAcl.ts
```diff
@@ -30,6 +30,7 @@ export default function makeAcl(json): Acl {
   const tokens = json.tokens || defaultAccess;
   const accessRequests = json.accessRequests || defaultAccess;
   const billing = json.billing || defaultAccess;
+  const databases = json.databases || defaultAccess;
```
kimlisa, line 33: lets rename to `dbServers`

packages/teleport/src/services/user/makeAcl.ts
```diff
@@ -30,6 +30,7 @@ export default function makeAcl(json): Acl {
   const tokens = json.tokens || defaultAccess;
   const accessRequests = json.accessRequests || defaultAccess;
   const billing = json.billing || defaultAccess;
+  const dbServers = json.dbServers || defaultAccess;
```
kimlisa, line 33: we need to test this. we always try to test our api responses, so if we make a change to an existing `make*`, there is usually a test for it, if not, we make one

kimlisa (approved):

PR #300, opened 2021-05-05
Add connect to database button and popup dialogue

### LINKED ISSUE #277: [DB Access 6] Add Connect Button and Popup Dialogue
### Description
For the database access table, render a button for each row that says `connect`, that when clicked upon, a dialogue shows up with preopulated fields.

For example, given this row, and our logged in user is `kimlisa` and my current cluster is `asteroid-sun.teleport.sh`:
```
Name            Description                      Protocol Type        URI                                                                      Labels                  
--------------- -------------------------------- -------- ----------- ------------------------------------------------------------------------ ---------------------- ------- 
aurora          PostgreSQL 11.6: AWS Aurora      postgres rds         postgres-aurora-instance-1.c1xpjrob56xs.us-west-1.rds.amazonaws.com:5432 cluster=root,env=aws    <btn>  
```
Clicking on button for this row will render a dialogue:
image

### CLI template
Step 1 is same as found in `Apps` and `Nodes` component.

For step 2 CLI template: `tsh db login <db name>`

For step 3 CLI template, depending on `protocol`:

psql (the quotes are required): `psql "service=<clustername>-<db name> user=[user] dbname=[dbname]"`
mysql: `mysql --defaults-group-suffix=_<clustername>-<db name> --user=[user] --database=[database]`

NOTE: The `angle brackets` are values we need to populate for user, the `square bracket` are values that user will need to fill out for themselves (so leave it as is).

### Link for the GUI
https://goteleport.com/docs/database-access/guides/gui-clients/

rudream:
### Purpose

This PR resolves #277  

Create a `connect` button on each row of the database list, when clicked, a popup dialogue is opened with instructions on how to connect to that database.

### Implementation

- Created a `ConnectDatabase` dialog component with similar structure as the `AddDatabase` component. 
- This dialog renders CLI instructions for the user that generates commands based on the selected database's info (eg. protocol, cluster name, etc.)
- Added tests and stories for the component

### Demo

connectdatabase


kimlisa:
@rudream could you rebase this after you merge the other PR (or rebase it from your other branch) b/c this PR contains those commits and its hard for me to filter

rudream:
@kimlisa done rebasing!

packages/teleport/src/Databases/AddDatabase/AddDatabase.tsx
```diff
@@ -30,7 +30,7 @@ export default function AddDatabase({ user, version, onClose }: Props) {
       onClose={onClose}
       open={true}
     >
-      <DialogTitle mb={4} mr="auto">
+      <DialogTitle mb="4" mr="auto">
```
kimlisa, line 33: i think curly brackets more accurate b/c its a number, it refers to the indexes of this array: https://github.com/gravitational/webapps/blob/master/packages/design/src/theme/theme.js#L22, the values are in pixels, and keep in mind, anytime for spacings we want to use these values (multiples of 8) for consistency

packages/teleport/src/Databases/ConnectDatabase/ConnectDatabase.story.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import React from 'react';
+import ConnectDatabase from './ConnectDatabase';
```
kimlisa, line 18: Starting to think we should rename these to `ConnectDialogue` and `AddDialogue`, bc the root folder is `Databases` and it makes it clearer what the component is, what do you think?

packages/teleport/src/Databases/ConnectDatabase/ConnectInstructions/ConnectInstructions.tsx
```diff
@@ -0,0 +1,94 @@
...
+  clusterId,
+  onClose,
+  dbInfo,
+}) {
+  const { name, protocol } = dbInfo;
+  const { hostname, port } = window.document.location;
+  const host = `${hostname}:${port || '443'}`;
+
+  return (
+    <>
```
kimlisa, line 35: I don't see a reason to have this in a separate file. I was going to say something about `Manually.tsx` but i decided to let that one as is, but normally i avoid implementing to accommodate future addons like `Automatically`, b/c it may or may not happen.

packages/teleport/src/Databases/ConnectDatabase/ConnectInstructions/ConnectInstructions.tsx
```diff
@@ -0,0 +1,94 @@
...
+        </Box>
+      </DialogContent>
+      <DialogFooter>
+        <ButtonSecondary onClick={onClose}>Close</ButtonSecondary>
+      </DialogFooter>
+    </>
+  );
+}
+
+export function generateDbConnectCmd(
```
kimlisa, line 84: I'd avoid exporting functions just to test it and its always usually better to do behavioral testing with components

packages/teleport/src/Databases/DatabaseList/DatabaseList.tsx
```diff
@@ -124,6 +139,24 @@ function LabelCell(props) {
   return renderLabelCell(tags);
 }
 
+function ConnectDBButton(props) {
```
kimlisa, line 142: I'd just named this `ConnectButton`, its defined in `DatabaseList.tsx`, we can infer it means to connect to a database

packages/teleport/src/Databases/DatabaseList/DatabaseList.tsx
```diff
@@ -145,6 +178,13 @@ function searchAndFilterCb(
 type Props = {
   databases: Database[];
   pageSize?: number;
+  user: string;
+  clusterId: string;
+};
+
+export type DbConnectInfo = {
+  name: string;
+  protocol: 'mysql' | 'postgres';
```
kimlisa, line 187: we have a type for this

packages/teleport/src/Databases/ConnectDatabase/ConnectDatabase.story.tsx
rudream: I agree, `ConnectDatabase` `AddDatabase` might not be descriptive enough

packages/teleport/src/Databases/ConnectDatabase/ConnectInstructions/ConnectInstructions.tsx
rudream: There's no interactivity in the connect database dialog, so there isn't any behaviour I could mimick. The only way to behaviourally test this would be to render the database list and then have the test click on the `connect` button and then check the rendered dialog

packages/teleport/src/Databases/ConnectDatabase/ConnectDatabase.story.tsx
rudream: I think `ConnectDialog` is more consistent with the way we spell `Dialog` in other places in the code than `Dialogue`, eg. `Dialog`, `DialogTitle`, `DialogContent`, `dialogCss`

packages/teleport/src/Databases/DatabaseList/DatabaseList.tsx
```diff
@@ -113,7 +116,19 @@ function DatabaseList(props: Props) {
...
+        <Column
+          header={<Cell />}
+          cell={<ConnectButton setDbConnectInfo={setDbConnectInfo} />}
+        />
       </StyledTable>
+      {dbConnectInfo && (
+        <ConnectDialog
+          user={user}
+          clusterId={clusterId}
+          dbConnectInfo={dbConnectInfo}
```
kimlisa, line 128: I'd just pass it as individual props to avoid exporting and importing the dbConnectInfo type. I think the `ConnectDialog` importing the type is kind of weird b/c ConnectDialog should be the one dictating the type.

packages/teleport/src/Databases/Databases.tsx
```diff
@@ -65,9 +66,11 @@ export function Databases(props: State) {
         </Box>
       )}
       {attempt.status === 'failed' && <Danger>{attempt.statusText}</Danger>}
-      {attempt.status === 'success' && <DatabaseList databases={databases} />}
+      {attempt.status === 'success' && (
+        <DatabaseList databases={databases} user={user} clusterId={clusterId} />
+      )}
       {isAddDatabaseVisible && (
-        <AddDatabase user={user} version={version} onClose={hideAddDatabase} />
+        <AddDialog user={user} version={version} onClose={hideAddDatabase} />
```
kimlisa, line 73: I'd also rename the flags `isAdd...`

packages/teleport/src/Databases/DatabaseList/DatabaseList.tsx
```diff
@@ -29,16 +29,22 @@ import {
...
     name: SortTypes.DESC,
   });
 
-  const [searchValue, setSearchValue] = React.useState<string>('');
+  const [searchValue, setSearchValue] = useState<string>('');
+
+  const [dbConnectInfo, setDbConnectInfo] = useState<{
+    name: string;
+    protocol: DbProtocol;
+  }>(undefined);
```
kimlisa, line 47: i'd use `null` when we are intentionally wanting to assign something as no value

kimlisa (commented):
almost there, also i'd remove the `packages/webapps.e` from this commit

kimlisa (approved):

PR #304, opened 2021-05-11
Create Kubernetes access screen

amha19:
This PR simply squashes all the commits made regarding the Kubernetes access screen into one commit with a summarized body.

- Creates Kubernetes access table

- Defines route, endpoint and service for kubernetes

- Defines kube custom hook, register feature, update Main.story

- Creates buttons and popup dialogue

PR dealing with registering Kubernetes feature to enterprise version can be found [here](https://github.com/gravitational/webapps.e/pull/53)


kimlisa (commented):
language fix

packages/teleport/src/Kubes/ConnectDialog/ConnectDialog.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  let text: string;
+
+  if (authType === 'local') {
+    text = `tsh login --proxy=${host} --auth=${authType} --user=${username}`;
+  }
+
+  if (authType === 'sso') {
+    text = `tsh login --proxy=${host}`;
+  }
```
kimlisa, line 41: the two text's share one common base: `tsh login --proxy=${host}`, i'd extend this base if the `authType` is local

kimlisa: i'd either add a snapshot test or a render and look for the specific output test

packages/teleport/src/Kubes/ConnectDialog/ConnectDialog.story.test.tsx
```diff
@@ -0,0 +1,24 @@
+/*
```
kimlisa, line 1: i would also add a test for the other case: either a storybook snapshot test or a render and look for the specific output test

packages/teleport/src/Kubes/ConnectDialog/ConnectDialog.tsx
amha19: I've changed it to a single string literal.

packages/teleport/src/Kubes/ConnectDialog/ConnectDialog.story.test.tsx
amha19: Added a story for sso case and applied a snapshot test.

packages/teleport/src/Kubes/ConnectDialog/ConnectDialog.tsx
```diff
@@ -56,26 +46,33 @@ function ConnectDialog(props: Props) {
...
           </Text>
           {' - Login to Teleport'}
-          <TextSelectCopy mt="2" text={text} />
+          <TextSelectCopy
+            mt="2"
+            text={`tsh login --proxy=${host} ${
+              authType === 'local'
+                ? `--auth=${authType} --user=${username}`
+                : ''
+            }`}
```
kimlisa, line 55: this is hard to read, i would pull this out into a variable

packages/teleport/src/Kubes/KubeList/KubeList.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+  function onSortChange(columnKey: string, sortDir: string) {
+    setSortDir({ [columnKey]: sortDir });
+  }
+
+  const data = sortAndFilter(searchValue);
+
+  return (
+    <>
+      <Flex flex="0 0 auto" mb={4}>
```
alex-kovoy, line 69: @amha19 
The search box does not really belong to this list, it needs to be moved 1 level up. One use case is that you can switch between multiple clusters and preserve the search value. With this, it will be reset each time a user changes a cluster.

It also a bit inconsistent with Nodes/Events/Etc where the search box does not disappear when content is being fetched.

packages/teleport/src/Kubes/Kubes.tsx
```diff
@@ -0,0 +1,67 @@
...
+import useTeleport from 'teleport/useTeleport';
+import useKubes, { State } from './useKubes';
+
+export default function Container() {
+  const ctx = useTeleport();
+  const state = useKubes(ctx);
+  return <Kubes {...state} />;
+}
+
+const docUrl = 'https://goteleport.com/docs/kubernetes-access';
```
alex-kovoy, line 35: string/numeric constants, defined on the global scope, need to be capitalized  `const DOC_URL = `

alex-kovoy (commented):
with a minor note (move the search box 1 level up)

packages/teleport/src/Kubes/KubeList/KubeList.tsx
amha19: Initially, it was in the parent component, but later I moved it to the list for simplicity. 
now I've moved it back and updated the search value state.

alex-kovoy (approved):

PR #305, opened 2021-05-12
Backports

kimlisa:


alex-kovoy (commented):
Are we missing 
https://github.com/gravitational/webapps/blob/master/packages/teleport/src/Roles/templates/role.yaml
?

alex-kovoy (approved):

PR #317, opened 2021-05-18
Refactor default dropdown css

### LINKED ISSUE #265: Update default CSS for the dropdown component
#### Used in several places
billing cycle:
image

cluster list:
image

cluster list console:
image

what a default dropdown looks like (you can see size disrepency compared to the cluster list dropdown, look at storybook for more)
image

For each of these we duplicate almost the same CSS. We need to update the default dropdown css instead so there is fewer overrides when extending from default.  The new default item size needs to be the same as in the cluster dropdown.

The `default` css for dropdowns is the one defined in: https://github.com/gravitational/webapps/blob/master/packages/shared/components/Select/Select.tsx. We want this default css to be the same as the css for our cluster dropdown in terms of sizing, spacing, fonts, hover color (we are keeping the white bg though). Add pointer cursor on hovers too.

Billing cycle dropdown should be converted to the default (white background) as well. The cluster list will remain the same look.

amha19:
**Description**

This PR refactors the default select component styling.
Creates a new styling component to override the default select component for darker themes. 

The purpose is to reduce code duplication, have less code overriding the default styling and consistent styling across components that use the default drop-down.

This PR resolves #265 

### Screenshot
image


packages/shared/components/Select/Select.story.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+const StyledSelect = styled(DarkStyledSelect)(
+  ({ theme }) => `
+  .react-select__control,
+  .react-select__control--is-focused {
+    min-height: 40px;
+  }
+  .react-select__input {
+    color: ${theme.colors.text.primary}
+  }
```
kimlisa, line 86: this should be a dark default, otherwise every time we use `DarkStyledSelect`, we have to repeat this style

packages/shared/components/Select/Select.story.tsx
```diff
@@ -0,0 +1,88 @@
+/*
```
kimlisa, line 1: since we have this file to showcase light vs dark, lets create a normal searchable selector too (not multis)

kimlisa:
Couple of things:

- we wanted to retain the coloring of what it was before. See the current and before image for the cluster selector. We should just make it `theme.colors.text.primary`. 
- The placeholder text is still grey and I didn't see this addressed in our dark styled component, i think we agreed on `theme.colors.text.secondary`

Current:
image

This is before:
image

packages/shared/components/Select/Select.story.tsx
```diff
@@ -0,0 +1,108 @@
...
+            options={options}
+            placeholder="Click to select a role"
+          />
+        </StyledSelect>
+      </Box>
+    </Flex>
+  );
+}
+
+const StyledSelect = styled(DarkStyledSelect)`
```
kimlisa, line 103: I'd put a note here saying why we increased the height

packages/shared/components/Select/Select.story.tsx
```diff
@@ -0,0 +1,108 @@
...
+limitations under the License.
+*/
+
+import React from 'react';
+import styled from 'styled-components';
+import { Flex, Box, LabelInput } from 'design';
+import Select, { DarkStyledSelect } from '../Select';
+
+export default {
+  title: 'Shared/Select',
```
kimlisa, line 23: Since its just one file for now, rename to `Shared`, and rename `SelectTypes` as `Selects` so there is no nesting

packages/teleport/src/Console/DocumentNodes/ClusterSelector/ClusterSelector.tsx
```diff
@@ -104,111 +103,23 @@ function filterOptions(value = '', options: Option[] = []) {
...
-  }
-
-  .react-select__indicator-separator {
-    display: none;
-  }
-
-  .react-select__value-container{
-    height: 30px;
-    padding: 0 8px;
+  .react-select__single-value {
```
kimlisa, line 112: lets add back the `color: white`, i did a side by side test, and the white looks better on the very dark bg

packages/shared/components/Select/DarkStyledSelect.tsx
```diff
@@ -0,0 +1,121 @@
...
+    &:hover {
+      border-color: rgba(255, 255, 255, 0.24);
+      background-color: ${theme.colors.primary.lighter};
+      .react-select__dropdown-indicator {
+        color: #666;
+      }
+    }
+  }
+
+  .react-select__control--is-focused {
```
kimlisa, line 64:
```suggestion
  .react-select__control--menu-is-open {
```
I think this is better, it only highlights when its opened, and removes focus afterwards

kimlisa (commented):
could we also update this issue with a image of the selects

packages/shared/components/Select/DarkStyledSelect.tsx
amha19: this will do the trick

kimlisa (approved):
please wait for @alex-kovoy approval

kimlisa (changes requested):
I don't think we handled disabled state. I would put another side by side of disabled selects. Example below is a disabled state but doesn't appear disabled.

image

amha19:
Disabled state for default and dark style selects. 

disabled-default-dark

kimlisa (approved):

alex-kovoy (approved):

PR #323, opened 2021-05-19
Lisa/tweaks

kimlisa:
### Description 
- add `kubernetes` and `databases` to our cluster list action menu
- move error banners below titles

### Screenshot
image


kimlisa:
@alex-kovoy I don't know if it was intentional to have the cluster lists action menu to have different names then our side nav i.e `Nodes` vs `Servers` , `Apps` vs `Applications`, `Audit Logs` vs `Audit Events`, so I left it as is

alex-kovoy:
@kimlisa i agree, the names should be the same

alex-kovoy (approved):
lets fix the names and merge (better way would be to read them from the side nav store but for now it's OK)

PR #331, opened 2021-05-28
Refactor services get link return value

amha19:
This PR contains minor refactoring of `getLink` function return value in services.
The reason for the change is to export the defined const and use it for filtering purposes in webapps.e.

### DIFF
--- a/packages/teleport/src/services/links.ts
+++ b/packages/teleport/src/services/links.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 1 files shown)


packages/teleport/src/services/links.ts
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+export const DOMAIN_NAME = 'https://get.gravitational.com/';
```
kimlisa, line 17: maybe `DOWNLOAD_DOMAIN_NAME`?

alex-kovoy (approved):

PR #333, opened 2021-06-05
Empty States

rudream:
## Purpose

Add empty states for enterprise users that haven't yet added a server/application/k8s cluster/database, and provide them with a quickstart guide video to learn how to add their first one.

## Implementation

Added an `<Empty />` component to `teleport/components` that accepts a `type` prop to determine what type of access the empty state is for and what content to render. eg. `type="databases"` will render the empty state for database access. The info for all types is stored in `Empty.tsx` itself, inside the `emptyStateInfo` object.

Also wrote some tests for to test different scenarios based on `isEnterprise`, `canCreate`, `isLeafCluster` and make sure the correct message is displayed.

I also noticed other some small things and fixed them:
- Fixed `InputSearch` being inside the list, which would cause it to re-render unnecessarily whenever the table re-renders.
- Cleaned up stories and made them consistent and readable. This means `Apps` stories don't use context and hook anymore so I added a test for it.

## Notes

- The quickstart video opens in a new tab, Alexey told me to hold off on adding a popup with the embedded video for the time being.
- The description texts for Kubernetes and Databases are temporary, @klizhentas will add the correct ones
- I noticed some small issues so it's not identical to the mockups, (for example, 'ADD AN APPLICATION' being inconsistent with  'ADD SERVER'), so I made some changes with @alexwolfe's approval. 

### Servers
Screenshot from 2021-06-04 22-52-17

### Applications
Screenshot from 2021-06-04 22-52-11

### Kubernetes
Screenshot from 2021-06-04 22-52-02

### Databases
Screenshot from 2021-06-04 22-51-49


alex-kovoy (commented):
I just noticed that somebody change the `pageSize` from 100 to 20 on the node list by making people to click on "next" when browsing 1k servers. Lets be nice to the users and revert that change because dealing with > 1k table that displays 20 record is a crime.

packages/teleport/src/components/Empty/Empty.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+  expect(await findByText(/There are no servers for the/i)).toBeVisible();
+});
+
+const props: Props = {
+  type: 'nodes',
+  clusterId: 'im-a-cluster',
+  isEnterprise: true,
+  canCreate: true,
+  isLeafCluster: false,
+  onButtonClick: () => null,
```
alex-kovoy, line 35: what happens when `button` becomes a link?  "onClick" would be more preferable in this case as it does not leak implementation detail

packages/teleport/src/Kubes/Kubes.story.test.tsx
```diff
@@ -27,3 +27,15 @@ test('failed', () => {
   const { container } = render(<Failed />);
   expect(container.firstChild).toMatchSnapshot();
 });
+
+test('empty state', () => {
+  const { container } = render(<Empty />);
+
```
alex-kovoy, line 33: i would remove this extra line between a `const` and a `expect`

packages/teleport/src/Kubes/Kubes.tsx
```diff
@@ -42,10 +43,17 @@ export function Kubes(props: State) {
...
     showButton,
-    searchValue,
-    setSearchValue,
+    isEnterprise,
+    isLeafCluster,
+    clusterId,
+    canCreate,
   } = props;
 
+  const [searchValue, setSearchValue] = useState('');
```
alex-kovoy, line 52: I would move it back to the hook, no reason to break up a state this way because of 1 value (the next question would be why not move down the rest of the state fields?).

packages/teleport/src/components/Empty/Empty.tsx
```diff
@@ -0,0 +1,171 @@
...
+  as: Flex,
+  mx: '16px',
+  bg: '#01172c',
+  width: '296px',
+  height: '200px',
+  borderRadius: 8,
+  alignItems: 'center',
+  justifyContent: 'center',
+};
+
```
alex-kovoy, line 163: Box -> Container -> Flex -> defaultProps is not needed. It took me some time to understand the intent and read through the layout properties. Instead do:

```tsx
 <Flex 
   mx= "16px"
   bg= "#01172c"
   width= "296px"
   height= "200px"
   alignItems= "center"
   justifyContent= "center"
   onClick={() => window.open(videoLink)}
>
  <Image width="265px" src={graphic} />
  <Box style={{ position: 'absolute' }} mt={4}>
    <Image m="auto" mb={1} width="90px" src={playIcon} />
    <Text fontWeight={700}>WATCH THE QUICKSTART</Text>
  </Box>
</Flex>
```

 This way you do not need to jump around the file to get an idea what the layout parameters are of the internal component (not exported).

packages/teleport/src/components/Empty/Empty.tsx
```diff
@@ -0,0 +1,171 @@
...
+      'Teleport Application Access provides secure access to internal applications without the need for a VPN but with the audibility and control of Teleport.',
+    videoLink: 'https://www.youtube.com/watch?v=PCYyTecSzCY',
+    buttonText: 'ADD DATABASE',
+    graphic: emptyPng,
+    readOnly: {
+      title: 'No Databases Found',
+      message: 'There are no databases for the "',
+    },
+  },
+};
```
alex-kovoy, line 147: Instead of having 1 place that describes these parameters for all possible scenarios, let each use-case to explicitly set and define own params. This not only makes it easier to find/locate a necessary change but also let deleting the code easier without any leftovers. It becomes even more useful when code that uses it resides in different repos (oss/e)

packages/teleport/src/components/Empty/Empty.tsx
rudream: Yeah, I thought with all the those props (which are mainly just styling) it might make the component harder to read, but it looks like splitting it up with default props actually made it even harder

alex-kovoy (approved):
Looks good. Lets clean-up the images
1) Use iconmoon icon for play
2) Use 1 empty.png for empty on both screns.

and then merge.

packages/teleport/src/Apps/Apps.tsx
```diff
@@ -73,10 +75,23 @@ export function Apps(props: State) {
...
       )}
       {isAddAppVisible && <AddApp onClose={hideAddApp} />}
     </FeatureBox>
   );
 }
+
+const emptyStateInfo: EmptyStateInfo = {
+  title: 'ADD YOUR FIRST APPLICATION',
+  description:
+    'Teleport Application Access provides secure access to internal applications without the need for a VPN but with the audibility and control of Teleport.',
```
klizhentas, line 90: Quick access to web applications running behind NAT and firewalls with security and compliance. Follow [the documentation](https://goteleport.com/docs/application-access/getting-started/) to get started.

packages/teleport/src/Databases/Databases.tsx
```diff
@@ -90,3 +120,15 @@ export function Databases(props: State) {
     </FeatureBox>
   );
 }
+
+const emptyStateInfo: EmptyStateInfo = {
+  title: 'ADD YOUR FIRST DATABASE',
+  description:
```
klizhentas, line 126: Consolidate access to databases running behind NAT, prevent data exfiltration, meet compliance requirements, and have complete visibility into access and behavior. Follow [the documentation](https://goteleport.com/docs/database-access/guides/) to get started.

packages/teleport/src/Kubes/Kubes.tsx
```diff
@@ -62,23 +70,48 @@ export function Kubes(props: State) {
...
         />
       )}
     </FeatureBox>
   );
 }
+
+const emptyStateInfo: EmptyStateInfo = {
+  title: 'ADD YOUR FIRST KUBERNETES CLUSTER',
+  description:
+    'Teleport Application Access provides secure access to internal applications without the need for a VPN but with the audibility and control of Teleport.',
```
klizhentas, line 110: Fast, secure access to Kubernetes clusters. Follow [the documentation](https://goteleport.com/docs/kubernetes-access/guides/multiple-clusters/) to connect your first cluster.

packages/teleport/src/Nodes/Nodes.tsx
```diff
@@ -74,30 +79,54 @@ export function Nodes(props: State) {
...
       )}
       {isAddNodeVisible && <AddNode onClose={hideAddNode} />}
     </FeatureBox>
   );
 }
+
+const emptyStateInfo: EmptyStateInfo = {
+  title: 'ADD YOUR FIRST SERVER',
+  description:
+    'Centralized control and access to Servers with Teleport Server Access. Add labels to nodes and get full visibility into access and behavior.',
```
klizhentas, line 125: SSH access for engineers with native security and compliance.

packages/teleport/src/components/Empty/Empty.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+const props: Props = {
+  clusterId: 'im-a-cluster',
+  isEnterprise: true,
+  canCreate: true,
+  isLeafCluster: false,
+  onClick: () => null,
+  emptyStateInfo: {
+    title: 'ADD YOUR FIRST SERVER',
+    description:
+      'Centralized control and access to Servers with Teleport Server Access. Add labels to nodes and get full visibility into access and behavior.',
```
klizhentas, line 38: Instant SSH access for engineers with native security and compliance.

klizhentas (approved):
See comments for updated text and link to the docs

alex-kovoy (approved):

PR #338, opened 2021-06-11
Recovery Tokens Screen

rudream:
## Purpose

Add a screen that shows the user their recovery tokens after creating an account or resetting their password. Users can easily copy the tokens to their clipboard, or print them out.

## Implementation

The recovery tokens feature as a whole is still a WIP, so currently in order for the recovery tokens screen to show, you will need to run the Teleport backend with @kimlisa's [branch `lisa/recovery-token-gen`](https://github.com/gravitational/teleport/tree/lisa/recovery-token-gen) (enterprise). In `useInvite` I temporarily added a `!res.length` condition, this is for backwards compatibility so that the app acts the same as before if you use the old Teleport backend. Nonetheless I will hold off on merging until Lisa's backend is ready.

I refactored the `Invite` component to TypeScript and to our updated way of writing components (with hooks and containers instead of with `withState`, and with `useAttemptNext`)

The icons for copying and printing are temporary, our icomoon subscription expired so I just tried to work with what we already had, I will add correct icons once we renew our subscription.

I wrapped the view in a `PrintWrapper` component which has a `@media print` query so that when the user wants to print the tokens, we can generate a clean printout without useless items like buttons and icons. Anything with `className="no-print"` will be excluded from the print.

The UI is quite responsive, and it will work with any number of tokens (in case we decide to give users more/less than 3 recovery tokens)

## Notes

- It might be a good idea to have a confirm dialog when the user clicks the continue button, since the user will only ever see these tokens once, it shouldn't be easy to accidentally proceed before saving the tokens. Thoughts?
- It might also be beneficial to have some type of bold warning that alerts the user to the importance of the tokens and how this is the only time they will get to see them.

### Account Creation Flow

recoverycodesdemo4

### Account Reset Flow

resetflow

### Recovery Tokens Screen for New Accounts (eg. from invite)
 
image

### Recovery Tokens Screen for Existing Accounts (eg. after resetting password)

image


alex-kovoy:
@rudream we need to remove these strange button-icons (copy&print) and replace them with our standard buttons as nowhere else we use button like these. This is because we do not want to give a user wrong pointers because we are never going to use these buttons after this screen.

packages/teleport/src/Invite/RecoveryTokens/RecoveryTokens.tsx
```diff
@@ -0,0 +1,214 @@
...
+          }}
+          href="https://goteleport.com/privacy/"
+          target="_blank"
+          className="no-print"
+        >
+          Privacy Policy
+        </Link>
+      </Flex>
+    </PrintWrapper>
+  );
```
alex-kovoy, line 199: Lets remove these (some random stuff)

rudream: revised

rudream:
Replaced icons with the same buttons we use for `TextSelectCopy`. Updated demo screenshots.

rudream:
Closing because I am moving this to [webapps.e](https://github.com/gravitational/webapps.e)

PR #339, opened 2021-06-15
Use filter `session.end` to retrieve events for session recording screen

kimlisa:
In response to: https://github.com/gravitational/teleport/pull/7231

#### Description
- Use filter `session.end` to retrieve events for session recording screen
- Add pagination to session recordings list

### DIFF
--- a/packages/teleport/src/services/audit/audit.ts
+++ b/packages/teleport/src/services/audit/audit.ts
@@ -29,8 +29,9 @@ class AuditService {
     const url = cfg.getClusterEventsUrl(clusterId, {
       start,
       end,
-      limit: params.limit ? params.limit : this.maxFetchLimit,
-      startKey: params.startKey ? params.startKey : undefined,
+      limit: this.maxFetchLimit,
+      include: params.filterBy || undefined,
+      startKey: params.startKey || undefined,
     });
 
     return api.get(url).then(json => {
--- a/packages/teleport/src/useAuditEvents.ts
+++ b/packages/teleport/src/useAuditEvents.ts
@@ -35,6 +35,8 @@ export default function useAuditEvents(
     fetchStatus: '',
   });
 
+  const filterBy = eventCode ? formatters[eventCode].type : '';
+
   useEffect(() => {
     fetch();
   }, [clusterId, range]);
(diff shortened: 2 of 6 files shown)


packages/teleport/src/services/audit/audit.ts
```diff
@@ -29,8 +29,9 @@ class AuditService {
     const url = cfg.getClusterEventsUrl(clusterId, {
       start,
       end,
-      limit: params.limit ? params.limit : this.maxFetchLimit,
-      startKey: params.startKey ? params.startKey : undefined,
+      limit: this.maxFetchLimit,
+      include: params.filterBy || undefined,
```
alex-kovoy, line 33: why `indefined` , we should not be using it as a value in the code

packages/teleport/src/useAuditEvents.ts
```diff
@@ -23,7 +23,7 @@ import { Event } from 'teleport/services/audit';
 export default function useAuditEvents(
   ctx: Ctx,
   clusterId: string,
-  limit?: number
+  eventType?: string
```
alex-kovoy, line 26:
```ts
import { EventCode } from 'teleport/services/audit';

export default function useAuditEvents(
  ctx: Ctx,
  clusterId: string,
  eventCode?: EventCode
)
```

packages/teleport/src/useAuditEvents.ts
```diff
@@ -48,7 +48,11 @@ export default function useAuditEvents(
       fetchStatus: 'loading',
     });
     ctx.auditService
-      .fetchEvents(clusterId, { ...range, startKey: results.fetchStartKey })
+      .fetchEvents(clusterId, {
+        ...range,
+        startKey: results.fetchStartKey,
+        filterBy: eventType,
```
alex-kovoy, line 54: formatters[eventCode].type

alex-kovoy (approved):
with comments

packages/teleport/src/useAuditEvents.ts
kimlisa: revised

packages/teleport/src/useAuditEvents.ts
kimlisa: revised

packages/teleport/src/services/audit/audit.ts
kimlisa: it was because of this https://github.com/gravitational/webapps/blob/master/packages/teleport/src/config.ts#L122, these values can be an empty string, and passing in a empty string to `generatePath` will cause an error so it had to be undefined.

alex-kovoy (approved):
LGTM.

Lets also update the teleport master and 6.2.x branches with this change.

PR #340, opened 2021-06-18
Convert applications tiles view to table view

rudream:
## Purpose

Convert the applications UI view to table view so that it is more scalable and a user can have many apps and see them easily.

This is just the first phase of a bigger ticket to revamp our Applications UI.

This is related to https://github.com/gravitational/teleport/issues/5448

## Implementation

Used our table components to display the app list in a table format like we do for nodes/db/k8s. 

Converted labels eg, `[{name:"cluster", value: "one"}]` to tags eg. `["cluster: one"]` so that they are searchable.

For the time being, since we haven't yet added the ability to add custom icons, a basic icon is generated for each app. It just has the first letter of the app name, and one of 10 colors picked based on its name through a basic hashing function. 

## Demo

image

### DIFF
--- a/packages/teleport/src/Apps/AppList/AppList.tsx
+++ b/packages/teleport/src/Apps/AppList/AppList.tsx
@@ -19,6 +19,18 @@ import styled from 'styled-components';
 import { sortBy } from 'lodash';
 import { ButtonBorder, Flex, Text } from 'design';
 import {
+  pink,
+  teal,
+  cyan,
+  blue,
+  green,
+  orange,
+  brown,
+  red,
+  deepOrange,
+  blueGrey,
+} from 'design/theme/palette';
+import {
   Column,
   SortHeaderCell,
   Cell,
@@ -168,16 +180,16 @@ function getIconColor(appName: string) {
   }
 
   const colors = [
-    '#d81b60',
-    '#00897b',
-    '#00acc1',
-    '#1e88e5',
-    '#43a047',
-    '#ffb300',
-    '#6d4c41',
-    '#e53935',
-    '#f4511e',
-    '#546e7a',
+    pink[700],
+    teal[700],
+    cyan[700],
+    blue[700],
+    green[700],
+    orange[700],
+    brown[700],
+    red[700],
+    deepOrange[700],
+    blueGrey[700],
   ];
 
   return colors[stringValue % 10];
--- a/packages/teleport/src/Apps/AppList/AppList.tsx
+++ b/packages/teleport/src/Apps/AppList/AppList.tsx
@@ -70,8 +70,6 @@ function AppList(props: Props) {
     setSortDir({ [columnKey]: sortDir });
   }
 
-  const data = sortAndFilter(searchValue);
-
   function onAppClick(e: React.MouseEvent) {
     if (e.ctrlKey || e.shiftKey || e.altKey) {
       return;
@@ -91,6 +89,8 @@ function AppList(props: Props) {
       }
     });
   }
+
+  const data = sortAndFilter(searchValue);
 
   return (
     <StyledTable pageSize={pageSize} data={data} onClick={onAppClick}>
--- a/packages/teleport/src/Apps/AppList/AppList.tsx
+++ b/packages/teleport/src/Apps/AppList/AppList.tsx
@@ -61,35 +61,33 @@ function AppList(props: Props) {
   const data = sortAndFilter(searchValue);
 
   return (
-    <>
-      <StyledTable pageSize={pageSize} data={data}>
-        <Column header={<Cell />} cell={<AppIconCell />} />
-        <Column
-          columnKey="name"
-          header={
-            <SortHeaderCell
-              sortDir={sortDir.name}
-              onSortChange={onSortChange}
-              title="Name"
-            />
-          }
-          cell={<TextCell />}
-        />
-        <Column
-          columnKey="publicAddr"
-          header={
-            <SortHeaderCell
-              sortDir={sortDir.publicAddr}
-              onSortChange={onSortChange}
-              title="Address"
-            />
-          }
-          cell={<TextCell />}
-        />
-        <Column header={<Cell>Labels</Cell>} cell={<LabelCell />} />
-        <Column header={<Cell />} cell={<OpenButton />} />
-      </StyledTable>
-    </>
+    <StyledTable pageSize={pageSize} data={data}>
+      <Column header={<Cell />} cell={<AppIconCell />} />
+      <Column
+        columnKey="name"
+        header={
+          <SortHeaderCell
+            sortDir={sortDir.name}
+            onSortChange={onSortChange}
+            title="Name"
+          />
+        }
+        cell={<TextCell />}
+      />
+      <Column
+        columnKey="publicAddr"
+        header={
+          <SortHeaderCell
+            sortDir={sortDir.publicAddr}
+            onSortChange={onSortChange}
+            title="Address"
+          />
+        }
+        cell={<AddressCell />}
+      />
+      <Column header={<Cell>Labels</Cell>} cell={<LabelCell />} />
+      <Column header={<Cell />} cell={<OpenButton />} />
+    </StyledTable>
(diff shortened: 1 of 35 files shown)


packages/teleport/src/Apps/AppList/AppList.tsx
```diff
@@ -1,106 +1,188 @@
...
+  }
+
+  function onSortChange(columnKey: string, sortDir: string) {
+    setSortDir({ [columnKey]: sortDir });
+  }
+
+  const data = sortAndFilter(searchValue);
+
+  return (
+    <>
```
alex-kovoy, line 64: no need for `<>` here

packages/teleport/src/Databases/Databases.story.tsx
```diff
@@ -23,7 +23,16 @@ export default {
   title: 'Teleport/Databases',
 };
 
-export const Loaded = () => <Databases {...props} />;
+export const Loaded = () => {
+  const [searchValue, setSearchValue] = useState('');
```
alex-kovoy, line 27: Lets remove. I understand that you added it so it's easier to test this _one_ part of the component state _manually_ in the storybook. If so, add a unit-test instead and keep the stories primary for rendering testing.

Please do not use snapshot testing for adding these tests.

packages/teleport/src/Apps/Apps.tsx
```diff
@@ -66,7 +69,25 @@ export function Apps(props: State) {
...
+            mb={4}
+            alignItems="center"
+            flex="0 0 auto"
+            justifyContent="space-between"
+          >
+            <InputSearch
+              mr="3"
+              value={searchValue}
+              onChange={e => {
+                setSearchValue(e);
```
alex-kovoy, line 84:
```
onChange={setSearchValue}
```

packages/teleport/src/Apps/Apps.tsx
```diff
@@ -66,7 +69,25 @@ export function Apps(props: State) {
...
+      {hasApps && (
+        <>
+          <Flex
+            mb={4}
+            alignItems="center"
+            flex="0 0 auto"
+            justifyContent="space-between"
+          >
+            <InputSearch
+              mr="3"
```
alex-kovoy, line 81: `mr="3"` => `mr={3}`

packages/teleport/src/Databases/Databases.story.test.tsx
```diff
@@ -15,14 +15,22 @@ limitations under the License.
...
 test('open source loaded', () => {
   const { container } = render(<Loaded />);
   expect(container.firstChild).toMatchSnapshot();
 });
 
+test('search filter works', () => {
+  render(<Loaded searchValue="postgres-gcp" />);
+
+  expect(screen.queryByText(/value: gcp/i)).toBeInTheDocument();
+
```
alex-kovoy, line 30: it's better to group tests together so it's easer to read

```
const x;
const b;

doX
doB
renderZ

expectZ2
expectZ2
```

packages/teleport/src/Nodes/Nodes.story.test.tsx
```diff
@@ -1,12 +1,20 @@
...
   expect(container.firstChild).toMatchSnapshot();
 });
 
+test('search filter works', () => {
+  render(<Loaded searchValue="fujedu" />);
+
+  expect(screen.queryByText(/172.10.1.20:3022/i)).toBeInTheDocument();
+
+  expect(screen.queryByText(/172.10.1.1:3022/i)).toBeNull();
+});
```
alex-kovoy, line 16: Move it to `Nodes.test.tsx`

packages/teleport/src/services/apps/makeApps.ts
```diff
@@ -36,7 +36,7 @@ export default function makeApp(json): App {
     name,
     uri,
     publicAddr,
-    labels,
+    tags: labels.map(label => `${label.name}: ${label.value}`),
```
alex-kovoy, line 39: no need to you lodash here:

```tsx
export default function makeApp(json: any): App {
  json = json || {};
  const { name, uri, publicAddr, labels, clusterId, fqdn } = json;
  const id = `${clusterId}-${name}-${publicAddr}`;
  const launchUrl = cfg.getAppLauncherRoute({ fqdn, clusterId, publicAddr });
  const tags = !Array.isArray(labels)
    ? []
    : labels.map(label => `${label.name}: ${label.value}`);

  return {
    id,
    name,
    uri,
    publicAddr,
    tags,
    clusterId,
    fqdn,
    launchUrl,
  };
}
```

Notice an extra check for nullable labels, this is because our backend does not guarantee that there always be an empty array.

alex-kovoy: @rudream could you also fix floating "Play" icon?

image

alex-kovoy: + Lets also make the row clickable as this is the primary action for applications. Reaching out to the "Open" button on wide screens is painful. As for "copy&past" we can detect if mouse button is pressed to ignore the click.

alex-kovoy: @rudream lets finish with this PR by addressing the rest of the comments (like using the colors from palette)

packages/teleport/src/Users/UserReset/UserReset.story.test.tsx
```diff
@@ -29,6 +29,10 @@ test('failed state', () => {
 });
 
 test('success state', () => {
+  jest
+    .spyOn(Date, 'now')
+    .mockImplementation(() => Date.parse('2021-04-08T07:00:00Z'));
```
alex-kovoy, line 34: why is it needed to render a success card?

packages/teleport/src/Users/UserTokenLink/UserTokenLink.story.test.tsx
```diff
@@ -2,6 +2,10 @@ import React from 'react';
 import * as story from './UserTokenLink.story';
 import { render } from 'design/utils/testing';
 
+jest
+  .spyOn(Date, 'now')
+  .mockImplementation(() => Date.parse('2021-04-08T07:00:00Z'));
```
alex-kovoy, line 7: why this global mocks? (yes, jest clears up the mocks after each test by default but still there must be a good reason to do mock it this way)

alex-kovoy (commented):
@rudream I tired it out locally and it looks great. Lets make the rows clickable (so it triggers the "open" action) :-) Yes, I know that we went back&fo with this one but at the end, it's better to make it clickable (and we will figure out what to do with AWS console apps later).

packages/teleport/src/Apps/Apps.tsx
```diff
@@ -66,7 +69,19 @@ export function Apps(props: State) {
...
+      {hasApps && (
+        <>
+          <Flex
+            mb={4}
+            alignItems="center"
+            flex="0 0 auto"
+            justifyContent="space-between"
+          >
+            <InputSearch mr={3} value={searchValue} onChange={setSearchValue} />
+          </Flex>
```
alex-kovoy, line 81: why this? I guess some copy&pasta :) 


Try this instead:

```ts
{hasApps && (
  <Box>
      <InputSearch mb={4} value={searchValue} onChange={setSearchValue} />
      <AppList searchValue={searchValue} apps={apps} />
   </Box>
)}
```

packages/teleport/src/Apps/AppList/AppList.tsx
```diff
@@ -1,106 +1,207 @@
...
+    }
+
+    return sorted;
+  }
+
+  function onSortChange(columnKey: string, sortDir: string) {
+    setSortDir({ [columnKey]: sortDir });
+  }
+
+  const data = sortAndFilter(searchValue);
```
alex-kovoy, line 73: I would move this line `const data = sortAndFilter(searchValue);` below the func declarations (so functions are visually grouped)

alex-kovoy (approved):
Good to merge (with 1 comment to address)

PR #345, opened 2021-06-28
Recovery Flows

rudream:
### Purpose

Add a feature that allows cloud users to reset their accounts by using recovery codes and email. 

These are the `webapps` changes related to recovery flow feature. Please refer to [the pull request in `webapps.e` ](https://github.com/gravitational/webapps.e/pull/65) for the main changes.


packages/teleport/src/Teleport.tsx
```diff
@@ -109,4 +120,6 @@ export default hot(Teleport);
 export type Props = {
   ctx: TeleportContext;
   history: History;
+  renderPublicRoutes?(children?: any): any;
+  renderPrivateRoutes?(CustomMain?: any): any;
```
kimlisa, line 124: I think these type should be more specific, ie `renderPublicRoutes` probably can't accept like say an array of strings

kimlisa (commented):
i'd remove the e-ref commit for `webapps.e`, as it will be a different commit hash when you merge it

packages/shared/components/FormLogin/FormLogin.tsx
```diff
@@ -145,28 +150,31 @@ export default function LoginForm(props: Props) {
...
-                <ForgotLink
-                  style={{
-                    textAlign: 'right',
-                  }}
-                  to={getRecoveryStartRoute('password')}
-                >
-                  Forgot Password?
-                </ForgotLink>
-              )}
+              <Flex flexDirection="column" alignItems="flex-end" mb={3}>
```
kimlisa, line 153: Unless we need to flex items, opt for `Box`

packages/shared/components/FormLogin/FormLogin.tsx
```diff
@@ -62,7 +62,7 @@ export default function LoginForm(props: Props) {
       mfaOptions = [{ value: 'u2f', label: 'U2F' }, ...mfaOptions];
     }
 
-    if (auth2faType === 'optional') {
+    if (auth2faType === 'optional' || auth2faType === 'off') {
```
kimlisa, line 65: If `auth2faType` is off, shouldn't we be not showing the drop down?

packages/shared/components/FormLogin/FormLogin.tsx
rudream: I am using flex to have the "Forgot Password" link align to the right side

alex-kovoy (approved):
assuming that Lisa's comments will be addressed

packages/shared/components/FormLogin/FormLogin.tsx
```diff
@@ -139,81 +149,98 @@ export default function LoginForm(props: Props) {
...
-                    <FieldSelect
-                      label="Second factor"
-                      value={mfaType}
-                      options={mfaOptions}
-                      onChange={opt => onSetMfaOption(opt as Option, validator)}
-                      mr={3}
-                      mb={0}
-                      isDisabled={isProcessing}
-                    />
+              <Flex flexDirection="column" mb={isRecoveryEnabled ? 2 : 4}>
```
kimlisa, line 152:
```suggestion
              <Box mb={isRecoveryEnabled ? 2 : 4}>
```

packages/teleport/src/Teleport.tsx
```diff
@@ -109,4 +120,6 @@ export default hot(Teleport);
 export type Props = {
   ctx: TeleportContext;
   history: History;
+  renderPublicRoutes?(children?: JSX.Element[]): any;
+  renderPrivateRoutes?(CustomMain?: JSX.Element): JSX.Element;
```
kimlisa, line 124: wouldn't this typing be same as `renderPublicRoutes`?

kimlisa (approved):

packages/teleport/src/Invite/useInvite.ts
```diff
@@ -0,0 +1,73 @@
...
+*/
+
+import { useState, useEffect } from 'react';
+import useAttempt from 'shared/hooks/useAttemptNext';
+import cfg from 'teleport/config';
+import history from 'teleport/services/history';
+import auth from 'teleport/services/auth';
+
+export default function useInvite(tokenId: string) {
+  const [passwordToken, setPswToken] = useState<ResetToken>(undefined);
```
kimlisa, line 24: one more thing 
```suggestion
  const [passwordToken, setPswToken] = useState<ResetToken>();
```
should be equivalent

packages/teleport/src/Teleport.tsx
rudream: No because it's just one parent element `<WaitingRoom>...</WaitingRoom>`

rudream: Don't think so because `CustomMain` is a single parent component. Though the return type for `renderPublicRoutes` should be `JSX.Element[]`.

kimlisa (approved):

PR #348, opened 2021-07-02
Add AWS App Access Dialog

rudream:
## Purpose

Add ability for selecting an IAM role when launching AWS console apps.

## Implementation

When launching an app (by clicking on it), if the app has any `awsRoles`, it will open up a dialog where the user can select an IAM role and launch the app with it. It will add an `arn` param alongside the already existing `cluster` and `addr`.

Since the `arn`'s have special characters like `/`, it will be encoded (eg. `...::role%2FEC2ReadOnly`) and will have to be decoded on the backend.

We still haven't figured out app icons, so for the time being, there is just a simple 'AWS' circle icon rather than the logo. 

### Questions

- Is there a scenario where an AWS console app has no `awsRoles`? How should it be launched then?
- If the app only has one `awsRole`, should it just be launched with that role as the default `arn` param (skipping the dialog step)

### Dialog

image

### Opening the dialog

awsdialogdemo


rudream:
Removed dialog and added small popover menu instead

### Updated Demos:

awsappdemo

image

packages/teleport/src/Apps/AppList/AwsLaunchButton/AwsLaunchButton.tsx
```diff
@@ -0,0 +1,132 @@
+import React from 'react';
+import styled from 'styled-components';
+import { ButtonBorder, Text } from 'design';
+import { Cell } from 'design/DataTable';
+import Menu, { MenuItem } from 'design/Menu';
+import { CarrotDown } from 'design/Icon';
+import cfg from 'teleport/config';
+import { AwsRole } from 'teleport/services/apps';
+
+class AwsLaunchButton extends React.Component<AwsLaunchButtonProps> {
```
alex-kovoy, line 10: `export default class`

packages/teleport/src/Apps/AppList/AwsLaunchButton/AwsLaunchButton.tsx
```diff
@@ -0,0 +1,132 @@
...
+type AwsLaunchButtonProps = {
+  awsRoles: AwsRole[];
+  fqdn: string;
+  clusterId: string;
+  publicAddr: string;
+};
+
+export default AwsLaunchButton;
+
+const StyledMenuItem = styled(MenuItem)(
```
alex-kovoy, line 117: When looking at the recording, i can tell that the margins are off on hover effect. Lets make them  equal size.

image


Also, lets have a max-width (40) on the item text and have the text truncated with ellipsis. Make sure that tootltip comes up on hover (`title` attribute)

Change the `IAM Roles:` to:  `Select AIM role`

rudream: I'm guessing you mean `40ch`, though this seems a bit long to me. How about a max of `25ch`, what do you think?

### max-width: 40ch
image

### max-width: 25ch
image

r0mant:
@rudream @alex-kovoy I tried this out today and was able to get it working - nice job!

There was only one piece missing - passing selected role ARN as a part of the "create app session" request. I poked around for a bit and pushed the fix to this branch - https://github.com/gravitational/webapps/pull/348/commits/7eb4dc47e3290e48a5ab5b3c5a57f80d1e739c3e. PTAL and feel free to refactor as needed - I'm not sure I implemented it the way you guys would.

Regarding open questions from the ticket description.

> Is there a scenario where an AWS console app has no awsRoles? How should it be launched then?

Theoretically there could be, e.g. in case of misconfiguration. I think that Launch button should be disabled in such a case, however the trick is frontend currently determines whether this is an AWS console by the presence of awsRoles so if they're absent it would just consider it a regular app. Maybe we can change the way frontend determines this is an AWS console? On the backend I'm just basically checking whether app's URI has `https://console.aws.amazon.com` prefix - could do the same here? Or backend can return some kind of indicator.

> If the app only has one awsRole, should it just be launched with that role as the default arn param (skipping the dialog step)

I think we should keep the selector dialog with a single option in this case - to be explicit. IIRC servers login selector has the same behavior.

BTW, looking at the Servers login selector dialog - it also has an edit box for users to enter the login themselves. Can we do the same for AWS roles? This would allow people to enter ARNs which aren't explicitly configured in their roles.

Screen Shot 2021-07-12 at 12 19 23 PM

r0mant (commented):
Just a couple of tweaks:

* Can we display app description in the table as well? Same as we do for databases. I've added `description` json field to the "get apps" response items in my branch.
* How hard is it to display AWS icon for console apps? E.g. just https://a0.awsstatic.com/libra-css/images/logos/aws_logo_smile_1200x630.png in a circle instead of letters? That would make them more visually distinctive, right now they still blend in a little IMO.

packages/teleport/src/AppLauncher/useAppLauncher.ts
```diff
@@ -58,12 +59,18 @@ function resolveRedirectUrl(params: UrlLauncherParams) {
       if (params.publicAddr) {
         url.searchParams.set('addr', params.publicAddr);
       }
+      if (params.arn) {
+        url.searchParams.set('arn', params.arn);
```
alex-kovoy, line 63: can we rename this query parameter to `awsrole` ?

alex-kovoy:
@rudream for the icon, we can use one from iconmoon (which has AWS bran icon). For that we need to update our iconmoon library 
https://gravitational.slab.com/posts/update-icomoon-library-srv7qufz

rudream:
- Added description column
- Added AWS Icon, and made icons slightly bigger
- Renamed query param to `awsrole`

### Updated Demo
image

packages/teleport/src/Apps/AppList/AppList.tsx
```diff
@@ -70,55 +72,49 @@ function AppList(props: Props) {
...
-            sortDir={sortDir.publicAddr}
-            onSortChange={onSortChange}
-            title="Address"
-          />
-        }
-        cell={<AddressCell />}
-      />
-      <Column header={<Cell>Labels</Cell>} cell={<LabelCell />} />
-    </StyledTable>
+    <>
```
alex-kovoy, line 78: You do not need this extra <>

packages/teleport/src/services/apps/makeApps.ts
```diff
@@ -19,20 +19,32 @@ import cfg from 'teleport/config';
...
     name,
+    description,
     uri,
     publicAddr,
     tags: labels.map(label => `${label.name}: ${label.value}`),
     clusterId,
     fqdn,
     launchUrl,
+    awsRoles,
+    isAWSConsole,
```
alex-kovoy, line 48: lets change `isAWSConsole` -> `awsConsole` so the aws related props are scoped with this prefix.

alex-kovoy (approved):
Some minor comments to address and good to go (please rename isAWSconsole to awsConsole)

PR #349, opened 2021-07-12
Add support for optional/on auth types on Invite and Reset screens

rudream:
## Purpose

Add UI support for `second_factor: on` and `second_factor: optional` on the Invite and Reset screens. Previously, it would force the user to use `TOTP`.

## Implementation

Made the two factor selection consistent with our updated Login Form, where we always have a "Second Factor" dropdown, and its items depend on what `second_factor` setting is used in the config. 

### Invite Screen with `second_factor: optional`

authtypeoptional

### Reset Screen with `second_factor: optional`

changepasswordoptional


alex-kovoy:
@rudream is it possible to have "none" as an option?

rudream:
> @rudream is it possible to have "none" as an option?

@alex-kovoy Yes, this is what the invite looks like with `second_factor: off` in the config

image

and this is what it looks like selecting the "none" option with `second_factor: optional` in the config

noneoption

alex-kovoy (approved):

packages/shared/components/FormInvite/FormInvite.tsx
```diff
@@ -64,13 +86,19 @@ export default function FormInvite(props: Props) {
...
     } else {
       onSubmit(password, token);
     }
   }
 
+  function onSetMfaOption(option: MFAOption, validator: Validator) {
+    setToken('');
+    validator.reset();
+    setMfaType(option);
+  }
```
kimlisa, line 100: i would also clear errors, lets say I was to set up a `u2f` key and it wasn't supported and errors out (safari) or any kind of error really, and I switch to `totp`, the u2f error dialog will still be there. I think any error should disappear on select new option

packages/teleport/src/Invite/useInvite.ts
```diff
@@ -58,6 +58,7 @@ export default function useInvite(tokenId: string) {
     auth2faType,
     fetchAttempt: fetchAttempt.attempt,
     submitAttempt: submitAttempt.attempt,
+    setSubmitAttempt: submitAttempt.setAttempt,
```
kimlisa, line 61: instead of this, lets define a `clearSubmitAttempt`, that `setAttempt({status: '', statusText: ''})

kimlisa (approved):

PR #350, opened 2021-07-13
Tweaks for MongoDB support

r0mant:
A few tweaks for MongoDB and other 7.0 things for database access:

* Add support for "query failed" audit event.
* Add MongoDB and GCP MySQL to "add database" dialog.
* Update connect command to `tsh db connect` for all protocols.


packages/teleport/src/Databases/ConnectDialog/ConnectDialog.tsx
```diff
@@ -68,7 +68,7 @@ export default function ConnectDialog({
             Step 2
           </Text>
           {' - Retrieve credentials for the database'}
-          <TextSelectCopy mt="2" text={`tsh db login ${dbName}`} />
+          <TextSelectCopy mt="2" text={`tsh db login [--db-user=<user>] [--db-name=<name>] ${dbName}`} />
```
kimlisa, line 71: `[--db-user=<user>]`, does the square bracket mean optional?

kimlisa (commented):
- for the new audit event, can we add an example json to the fixtures file, which will show up in the `Audit.story`: https://github.com/gravitational/webapps/blob/master/packages/teleport/src/Audit/fixtures/index.ts
- we should place a `maxMenuHeight={220}` to make the dropdown height smaller [here](https://github.com/gravitational/webapps/blob/master/packages/teleport/src/Databases/AddDatabase/AddDatabase.tsx#L113), i missed the last option b/c i didn't realize the menu was longer than my current view in dialog
- we have a test file to test the output of the new options added: https://github.com/gravitational/webapps/blob/master/packages/teleport/src/Databases/AddDatabase/AddDatabase.test.tsx#L21

packages/teleport/src/Databases/ConnectDialog/ConnectDialog.tsx
r0mant: Yeah

kimlisa: shoot, i feel like we have to go back and change all square brackets to angle brackets to keep the meaning consistent, b/c we currently use square brackets to mean `user need to fill this out`, maybe as a refactor for later? @alex-kovoy

r0mant:
@kimlisa Addressed your feedback in https://github.com/gravitational/webapps/pull/350/commits/97cf057d0b82902862b1d3cc8a9c605f5bd1374e, PTAL.

alex-kovoy (approved):

kimlisa (approved):

PR #368, opened 2021-07-27
Temporary work around that displays all types of sessions

kimlisa:
temp for issue https://github.com/gravitational/teleport/issues/7724

#### Description
- Added a column that describes a session type: `interactive, `noninteractive, `recording disabled`
- Render play button only for interactive sessions

#### Screenshot
image


alex-kovoy (approved):
Lets make sure that these field is searchable too

rudream:
I think it would look better if instead of the `Type` column, we had a disabled "Play" button there with the `title` set to something like "Recording is disabled for this session" or "This recording is non-interactive", what do you think?

alex-kovoy:
Lisa will run it by Sasha and let us know :)

alex-kovoy:
the type is misleading because there is no type as "recording disabled" so merging this table with action makes sense. Although without the disabled button as it will increase it in side.

kimlisa:
ran it through sasha, and we are going with @alex-kovoy suggestion:
image

PR #382, opened 2021-08-17
MFA-related Language Changes & Move MFA logic to OSS 

rudream:
## Purpose

Changed the way we call MFA devices after discussing with @klizhentas and @kimlisa .

Moved MFA logic to OSS for the upcoming MFA device management feature and also to reduce the scope of the PR in `webapps.e`.

## Implementation

The back-end MFA-related endpoints have been moved to the OSS cfg, and I created a new `MfaService` for MFA-related logic, this service will also be used by the account dashboard device management.

We previously had inconsistent and sometimes confusing naming when it came to MFA devices. We now call them "Two-Factor Devices".

- "TOTP" is now "Authenticator App".
- "U2F" is now "Hardware Key".

The flows for resetting `totp` and `u2f` have been combined into one, there is no longer a "Lost U2F Key?"/"Lost Two-Factor Token?", but a single "Lost Two-Factor Device?".

### Login

login

#### U2F

image

#### TOTP

image

### DIFF
--- a/packages/shared/components/FormInvite/TwoFaInfo.tsx
+++ b/packages/shared/components/FormInvite/TwoFaInfo.tsx
@@ -15,10 +15,8 @@ limitations under the License.
 */
 
 import React from 'react';
-import { Box, Text, ButtonLink, Link } from 'design';
+import { Box, Text, Link } from 'design';
 import { Auth2faType } from 'shared/services';
-
-const U2F_HELP_URL = 'https://support.google.com/accounts/answer/6103523?hl=en';
 
 export default function TwoFAData({ auth2faType, qr, submitBtnText }: Props) {
   const imgSrc = `data:image/png;base64,${qr}`;
@@ -27,20 +25,17 @@ export default function TwoFAData({ auth2faType, qr, submitBtnText }: Props) {
     return (
       <Box width="168px">
         <Text typography="paragraph2" mb={3}>
-          Scan the bar code with Google Authenticator to generate a two-factor
-          token.
+          Scan the QR Code with any authenticator app and enter the generated
+          code.
         </Text>
         <img width="152px" src={imgSrc} style={{ border: '8px solid' }} />
-        <ButtonLink
-          width="100%"
-          kind="secondary"
-          target="_blank"
-          size="small"
-          href="https://support.google.com/accounts/answer/1066447?co=GENIE.Platform%3DiOS&hl=en&oco=0"
-          rel="noreferrer"
-        >
-          Download Google Authenticator
-        </ButtonLink>
+        <Text typography="paragraph2" color="text.secondary">
+          We recommend{' '}
+          <Link href="https://authy.com/download/" target="_blank">
+            Authy
+          </Link>
+          .
+        </Text>
       </Box>
     );
   }
@@ -49,23 +44,15 @@ export default function TwoFAData({ auth2faType, qr, submitBtnText }: Props) {
     return (
       <Box width="168px">
         <Text typography="h5" mb="2">
-          Insert your U2F key
+          Insert your hardware key
         </Text>
-        <Box color="text.primary">
-          <Text typography="paragraph2" mb={3}>
-            Press the button on the U2F key after you press the{' '}
-            <Text as="span" style={{ textTransform: 'uppercase' }} bold>
-              {submitBtnText}
-            </Text>{' '}
-            button.
-          </Text>
-          <Text typography="paragraph2" mb={3}>
-            <Link color="light" target="_blank" href={U2F_HELP_URL}>
-              Learn more
-            </Link>{' '}
-            about U2F 2-Step Verification.
-          </Text>
-        </Box>
+        <Text typography="paragraph2" mb={3}>
+          Press the button on the hardware key after you press the{' '}
+          <Text as="span" style={{ textTransform: 'uppercase' }} bold>
+            {submitBtnText}
+          </Text>{' '}
+          button.
+        </Text>
       </Box>
     );
   }
--- a/packages/shared/components/FormInvite/TwoFaInfo.tsx
+++ b/packages/shared/components/FormInvite/TwoFaInfo.tsx
@@ -48,7 +48,7 @@ export default function TwoFAData({ auth2faType, qr, submitBtnText }: Props) {
         </Text>
         <Text typography="paragraph2" mb={3}>
           Press the button on the hardware key after you press the{' '}
-          <Text as="span" style={{ textTransform: 'uppercase' }} bold>
+          <Text as="span" caps bold>
             {submitBtnText}
           </Text>{' '}
           button.
--- a/packages/shared/components/FormLogin/FormLogin.tsx
+++ b/packages/shared/components/FormLogin/FormLogin.tsx
@@ -14,13 +14,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import React, { useState } from 'react';
+import React, { useState, useMemo } from 'react';
 import styled from 'styled-components';
 import { Card, Text, Flex, ButtonLink, ButtonPrimary, Box } from 'design';
 import * as Alerts from 'design/Alert';
 import { AuthProvider, Auth2faType } from 'shared/services';
-import { Option } from 'shared/components/Select';
 import { useAttempt } from 'shared/hooks';
+import { getMfaOptions, MfaOption } from 'teleport/services/mfa/utils';
 import Validation, { Validator } from '../Validation';
 import FieldInput from '../FieldInput';
 import FieldSelect from '../FieldSelect';
@@ -42,33 +42,15 @@ export default function LoginForm(props: Props) {
     clearAttempt,
   } = props;
 
-  const mfaEnabled = auth2faType === 'on' || auth2faType === 'optional';
-  const u2fEnabled = auth2faType === 'u2f';
-  const otpEnabled = auth2faType === 'otp';
   const ssoEnabled = authProviders && authProviders.length > 0;
 
   const [pass, setPass] = useState('');
   const [user, setUser] = useState('');
   const [token, setToken] = useState('');
-  const [mfaOptions] = useState<MFAOption[]>(() => {
-    const mfaOptions: MFAOption[] = [];
 
-    if (auth2faType === 'optional') {
-      mfaOptions.push({ value: 'optional', label: 'None' });
-    }
+  const mfaOptions = useMemo<MfaOption[]>(() => getMfaOptions(auth2faType), []);
 
-    if (mfaEnabled || u2fEnabled) {
-      mfaOptions.push({ value: 'u2f', label: 'Hardware Key' });
-    }
-
-    if (mfaEnabled || otpEnabled) {
-      mfaOptions.push({ value: 'otp', label: 'Authenticator App' });
-    }
-
-    return mfaOptions;
-  });
-
-  const [mfaType, setMfaType] = useState<MFAOption>(mfaOptions[0]);
+  const [mfaType, setMfaType] = useState<MfaOption>(mfaOptions[0]);
   const [isExpanded, toggleExpander] = useState(
     !(isLocalAuthEnabled && ssoEnabled)
   );
@@ -90,7 +72,7 @@ export default function LoginForm(props: Props) {
     }
   }
 
-  function onSetMfaOption(option: MFAOption, validator: Validator) {
+  function onSetMfaOption(option: MfaOption, validator: Validator) {
     setToken('');
     clearAttempt();
     validator.reset();
@@ -180,7 +162,7 @@ export default function LoginForm(props: Props) {
                         value={mfaType}
                         options={mfaOptions}
                         onChange={opt =>
-                          onSetMfaOption(opt as MFAOption, validator)
+                          onSetMfaOption(opt as MfaOption, validator)
                         }
                         mr={3}
                         mb={0}
@@ -294,5 +276,3 @@ type Props = {
   onLoginWithU2f(username: string, password: string): void;
   onLogin(username: string, password: string, token: string): void;
 };
-
-type MFAOption = Option<Auth2faType>;
--- a/packages/shared/components/FormLogin/FormLogin.tsx
+++ b/packages/shared/components/FormLogin/FormLogin.tsx
@@ -164,7 +164,7 @@ export default function LoginForm(props: Props) {
                   <Box textAlign="right">
                     <ButtonLink
                       style={{ padding: '0px', minHeight: 0 }}
-                      onClick={() => onRecover(true)}
+                      onClick={() => onRecover('password')}
                     >
                       Forgot Password?
                     </ButtonLink>
@@ -211,7 +211,7 @@ export default function LoginForm(props: Props) {
                     <Box>
                       <ButtonLink
                         style={{ padding: '0px', minHeight: 0 }}
-                        onClick={() => onRecover(false)}
+                        onClick={() => onRecover('2fa')}
                       >
                         Lost Two-Factor Device?
                       </ButtonLink>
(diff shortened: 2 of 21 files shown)


packages/teleport/src/services/mfa/makeMfaDevice.ts
```diff
@@ -0,0 +1,25 @@
...
+  if (json.type === 'U2F') {
+    type = 'Hardware Key';
+  }
+
+  return {
+    id,
+    name,
+    type,
+    registeredDate: moment(addedAt).unix(),
+    lastUsedDate: moment(lastUsed).unix(),
```
kimlisa, line 21: hm, is this conversion required?

packages/teleport/src/services/mfa/types.ts
```diff
@@ -0,0 +1,15 @@
+export type MfaDevice = {
+  id: string;
+  type: 'Hardware Key' | 'Authenticator App';
```
kimlisa, line 3: i'm not really seeing the benefit of typing this, do we use this else where other than in `makeMfaDevice`? If not, i think it's better to just contain it inside the makeXXX?

packages/teleport/src/config.ts
```diff
@@ -114,6 +114,10 @@ const cfg = {
     nodeTokenPath: '/v1/enterprise/nodes/token',
     nodeScriptPath: '/scripts/:token/install-node.sh',
     appNodeScriptPath: '/scripts/:token/install-app.sh?name=:name&uri=:uri',
+
+    mfaSignRequestWithTokenPath: '/v1/webapi/u2f/signrequest/token',
+    mfaFetchDevicesPath: '/v1/webapi/mfadevice/:token',
+    mfaDeleteDevicePath: '/v1/webapi/mfadevice',
```
kimlisa, line 120:
```suggestion
    mfaDevicePath: '/v1/webapi/mfadevice',
```

if it was something like `/v1/webapi/mfadevice/delete` then `mfaDeleteDevicePath` would be better

packages/teleport/src/config.ts
```diff
@@ -114,6 +114,10 @@ const cfg = {
     nodeTokenPath: '/v1/enterprise/nodes/token',
     nodeScriptPath: '/scripts/:token/install-node.sh',
     appNodeScriptPath: '/scripts/:token/install-app.sh?name=:name&uri=:uri',
+
+    mfaSignRequestWithTokenPath: '/v1/webapi/u2f/signrequest/token',
+    mfaFetchDevicesPath: '/v1/webapi/mfadevice/:token',
```
kimlisa, line 119:
```suggestion
    mfaDeviceListPath: '/v1/webapi/mfadevice/:tokenId',
```
to keep it consistent elsewhere, i know there is like a few that deviates, but I go with majority

packages/teleport/src/components/MfaDeviceList/MfaDeviceList.tsx
```diff
@@ -0,0 +1,136 @@
...
+        Remove
+      </ButtonBorder>
+    </Cell>
+  );
+};
+
+type Props = {
+  devices: MfaDevice[];
+  onDelete({ id, name }: { id: string; name: string }): void;
+  isRecoveryFlow: boolean; // Whether this list is being shown in the recovery flow as opposed to in the account dashboard
```
kimlisa, line 127: i'd remove this until we need it, also long comments should go on top of the field, side comments are usually very short

packages/teleport/src/components/MfaDeviceList/MfaDeviceList.tsx
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  const mostRecentDevice = sort(devices, { registeredDate: SortTypes.ASC })[0];
+
+  const data = sort(devices);
+
+  return (
+    <StyledTable
+      data={data}
+      style={{ overflow: 'hidden', borderRadius: '8px' }}
```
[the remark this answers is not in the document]
rudream: It definitely looks better with the radius, what I will do is only have the radius if `isRecoveryFlow`.

packages/teleport/src/services/mfa/makeMfaDevice.ts
rudream: Yes. The `addedAt` and `lastUsed` returned from the backend are formatted strings eg. `"2021-08-18T18:39:23.522007Z"`. Converting them to unix timestamps makes sorting possible.

packages/teleport/src/services/mfa/types.ts
rudream: We do use it in a couple places in `webapps.e`

packages/shared/components/FormInvite/TwoFaInfo.tsx
```diff
@@ -49,23 +44,15 @@ export default function TwoFAData({ auth2faType, qr, submitBtnText }: Props) {
     return (
       <Box width="168px">
         <Text typography="h5" mb="2">
-          Insert your U2F key
+          Insert your hardware key
+        </Text>
+        <Text typography="paragraph2" mb={3}>
+          Press the button on the hardware key after you press the{' '}
+          <Text as="span" style={{ textTransform: 'uppercase' }} bold>
```
alex-kovoy, line 51: there is `caps` property that can be used instead of textTransform

packages/shared/components/FormLogin/FormLogin.tsx
```diff
@@ -300,7 +288,7 @@ type Props = {
   auth2faType?: Auth2faType;
   attempt: ReturnType<typeof useAttempt>[0];
   isRecoveryEnabled?: boolean;
-  onRecover?: (type: 'totp' | 'u2f' | 'password') => void;
+  onRecover?: (isRecoverPassword: boolean) => void;
```
alex-kovoy, line 291: keep the strings as it's way more descriptive than a boolean flag. -> `password | 2fa`

packages/teleport/src/components/MfaDeviceList/MfaDeviceList.tsx
```diff
@@ -0,0 +1,140 @@
...
+  );
+}
+
+const NameCell = props => {
+  const { data, rowIndex } = props;
+  const { name } = data[rowIndex];
+
+  return (
+    <Cell>
+      <Text style={{ maxWidth: '12ch' }} title={name}>
```
alex-kovoy, line 95: Please use `px` units as this is what we use everywhere

packages/teleport/src/components/MfaDeviceList/MfaDeviceList.tsx
```diff
@@ -0,0 +1,140 @@
...
+        {name}
+      </Text>
+    </Cell>
+  );
+};
+
+const DateCell = props => {
+  const { data, rowIndex, columnKey } = props;
+  const dateText = data[rowIndex][`${columnKey}Text`];
```
alex-kovoy, line 104: It's better to avoid any dynamically resolved property name (like this `Text` suffix). It's hard to read, it's hard to `type` (Typescript), and easy to break.

packages/teleport/src/components/MfaDeviceList/MfaDeviceList.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  return <Cell>{dateText}</Cell>;
+};
+
+const DeleteDeviceBtn = props => {
```
alex-kovoy, line 109: `DeleteDeviceBtn` is not a button but a `Cell` -> `DeleteCell`

packages/teleport/src/services/mfa/makeMfaDevice.ts
alex-kovoy: @rudream no need to use `unix` nor `moment` here, just create `Date` objects as follows
 
```
  registeredDate: new Date(addedAt),
  lastUsedDate: new Date(lastUsed),
```

packages/teleport/src/services/mfa/mfa.ts
```diff
@@ -0,0 +1,19 @@
...
+import { DeleteMfaDeviceRequest } from './types';
+
+class MfaService {
+  fetchMfaDevices(tokenId: string) {
+    return api
+      .get(cfg.getMfaDeviceListUrl(tokenId))
+      .then(devices => devices.map(makeMfaDevice));
+  }
+
+  deleteMfaDevice(data: DeleteMfaDeviceRequest) {
```
alex-kovoy, line 14: deleteMfdDevice -> `removeDevice` (the service name already has `Mfa` prefix)
fetchMfaDevices -> `fetchDevices`

packages/teleport/src/services/mfa/types.ts
alex-kovoy: this is a strange name for a type, please remove any spaces

packages/teleport/src/services/mfa/types.ts
```diff
@@ -0,0 +1,15 @@
...
+  name: string;
+  registeredDate: number;
+  lastUsedDate: number;
+  registeredDateText: string;
+  lastUsedDateText: string;
+};
+
+export type DeleteMfaDeviceRequest = {
```
alex-kovoy, line 11: DeleteMfaDeviceRequest -> `RemoveDeviceRequest`

packages/teleport/src/services/mfa/types.ts
rudream: @alex-kovoy this is just for the text on the `type` column in the table, I will rename `type` to `typeLabel` to make it more clear

packages/teleport/src/services/mfa/types.ts
```diff
@@ -0,0 +1,13 @@
+export type MfaDevice = {
+  id: string;
+  name: string;
+  typeText: 'Hardware Key' | 'Authenticator App';
```
alex-kovoy, line 4: I am a bit confused with this `typeText`. Why does it need to be `typed` ? If this is a description then `description: string` should be enough. 

```
  id: string;
  name: string;
  type:  'TOTP' | 'U2F' ...,
  description: string;
```

packages/teleport/src/services/mfa/mfa.ts
```diff
@@ -0,0 +1,18 @@
...
+import { RemoveDeviceRequest } from './types';
+
+class MfaService {
+  fetchDevices(tokenId: string) {
+    return api
+      .get(cfg.getMfaDeviceListUrl(tokenId))
+      .then(devices => devices.map(makeMfaDevice));
+  }
+
+  removeDevice(data: RemoveDeviceRequest) {
```
alex-kovoy, line 13: No need for `{ ...data} `

```ts
removeDevice(req: RemoveDeviceRequest) {
    return api.delete(cfg.api.mfaDevicePath, req);
  }
```

packages/teleport/src/services/mfa/types.ts
rudream: @alex-kovoy In the table we display either "Hardware Key" or "Authenticator App"

image

The alternative is using `type: 'TOTP' | 'U2f'` and then creating something like a `DeviceTypeCell` that checks the `type` and then for example renders `"Hardware Key"` if `type === 'U2F`, but we won't use the strings `'TOTP'` or `'U2F'` anywhere else, so I am just doing that check on the service layer and returning the device object with the formatted text already there.

packages/teleport/src/components/MfaDeviceList/MfaDeviceList.tsx
```diff
@@ -0,0 +1,150 @@
...
+export default function MfaDeviceList({
+  devices = [],
+  onDelete,
+  isRecoveryFlow,
+}: Props) {
+  const [sortDir, setSortDir] = useState<Record<string, string>>({
+    registeredDate: SortTypes.ASC,
+  });
+
+  const mostRecentDevice = useMemo(() => sort(devices)[0], []);
```
alex-kovoy, line 25: since it uses `useMemo` what happens to it when a new devices added?

packages/teleport/src/components/MfaDeviceList/MfaDeviceList.tsx
```diff
@@ -0,0 +1,150 @@
...
+      </ButtonBorder>
+    </Cell>
+  );
+};
+
+type Props = {
+  devices: MfaDevice[];
+  onDelete({ id, name }: { id: string; name: string }): void;
+  // Whether this list is being shown in the recovery flow as opposed to in the account dashboard
+  isRecoveryFlow: boolean;
```
alex-kovoy, line 140: is it really needed for this shared component to be aware of recovery flow and account dashboard?

packages/teleport/src/components/MfaDeviceList/MfaDeviceList.tsx
```diff
@@ -0,0 +1,150 @@
...
+    setSortDir({ [columnKey]: sortDir });
+  }
+
+  const data = sort(devices);
+
+  return (
+    <StyledTable
+      data={data}
+      style={isRecoveryFlow ? { overflow: 'hidden', borderRadius: '8px' } : {}}
+    >
```
alex-kovoy, line 47: instead of `isRecoveryFlow`, you can propagate the props to the table

```tsx
const {
  devices = [],
  onDelete,
 ...styles
}: Props)

return (
<StyledTable data={data} ...styles>
```    

this way, a consumer of this component can apply custom styles

```tsx
<MfaDeviceList style={...} />
```

packages/teleport/src/components/MfaDeviceList/MfaDeviceList.tsx
```diff
@@ -0,0 +1,150 @@
...
+};
+
+const DeleteCell = props => {
+  const { data, rowIndex, onDelete, mostRecentDevice, isRecoveryFlow } = props;
+  const { id, name } = data[rowIndex];
+
+  if (id === mostRecentDevice.id && isRecoveryFlow) {
+    return null;
+  }
+
```
alex-kovoy, line 126: lets show the delete button by default

rudream: @alex-kovoy This was a UX suggestion from Sasha, he wants us to hide the remove button for the most recent device

packages/teleport/src/components/MfaDeviceList/MfaDeviceList.tsx
rudream: revised

packages/teleport/src/components/MfaDeviceList/MfaDeviceList.tsx
rudream: I have changed the implementation, `mostRecentDevice` is now going to be provided by the parent `Devices` component in Recovery Flow. 

It is not possible for new devices to be added on this step since this `mostRecentDevice` is only used when the device list is in the recovery flow (not the dashboard), it is set when the list of devices is retrieved from the backend.

packages/teleport/src/services/mfa/utils.ts
```diff
@@ -0,0 +1,23 @@
+import { Auth2faType } from 'shared/services';
+import { Option } from 'shared/components/Select';
+
+export type MfaOption = Option<Auth2faType>;
+
+export function getMfaOptions(auth2faType: Auth2faType) {
+  const mfaOptions: MfaOption[] = [];
+  const mfaEnabled = auth2faType === 'on' || auth2faType === 'optional';
+
+  if (auth2faType === 'optional' || auth2faType === 'off') {
```
kimlisa, line 10: shouldn't `auth2faType === 'off'` return an empty array?

kimlisa (approved):
lgtm, please wait for @alex-kovoy approval

packages/teleport/src/services/mfa/utils.ts
rudream: When `auth2faType === 'off'` there won't be a dropdown shown, but behind the scenes this allows us to use `if(mfaOption.value === 'optional)'` both when the 'None' option is selected and when `auth2faType === 'off'` rather than needing an additional check for the latter

alex-kovoy (approved):

PR #384, opened 2021-08-26
Simplify "Add Database" Dialog

atburke:
Resolves gravitational/teleport#7537.
Backend is implemented in gravitational/teleport#7978.

# Details

Screenshot 2021-08-26 at 15-53-23 Databases

The token is generated when the dialog opens, so if a user closes the dialog and opens it again, they'll get a new token.

TODO: Finish this description!


atburke:
Screenshot 2021-08-30 at 13-24-15 Databases
Here's the updated dialog, using the dark-themed select. The Add Database button's behavior has also been reverted, so the dialog won't be available unless the user has permission to create tokens.

packages/teleport/src/Databases/AddDatabase/AddDatabase.tsx
```diff
@@ -59,6 +90,10 @@ export default function AddDatabase({
       ? `tsh login --proxy=${host}`
       : `tsh login --proxy=${host} --auth=local --user=${username}`;
 
+  const { id: token } = joinToken;
+  const useGeneratedToken = canCreate && token !== '';
```
alex-kovoy, line 94: in React community, `use` is reserved prefix used to name hooks. Also, why cannot `canCreate` be used without token check?

packages/teleport/src/Databases/AddDatabase/useAddDatabase.ts
```diff
@@ -0,0 +1,42 @@
...
+  const canCreate = ctx.storeUser.getTokenAccess().create;
+
+  useEffect(() => {
+    if (canCreate) {
+      run(() =>
+        ctx.joinTokenService.generateJoinToken(['Db']).then(setJoinToken)
+      );
+    } else {
+      setAttempt({ status: 'failed' });
+    }
```
alex-kovoy, line 35: no need to check for `canCreate` condition. It's controlled by the parent.  If it happens that user somehow opens it, there will be a server side error with access denied displayed on the screen.

```tsx
useEffect(() => {
    run(() =>
      ctx.joinTokenService.generateJoinToken(['Db']).then(setJoinToken)
    );
  }, []);
  ```

File: packages/teleport/src/Databases/AddDatabase/AddDatabase.tsx [hunk 3]
```diff
@@ -22,21 +22,52 @@ import Dialog, {
...
+export default function Container() {
+  const ctx = useTeleport();
+  const {
+    username,
+    version,
+    authType,
+    canCreate,
+    hideAddDialog,
+  } = useDatabases(ctx);
+  const { joinToken, attempt } = useAddDatabase(ctx);
```
alex-kovoy, line 50: this `useDatabases(ctx)` will kick in a new initialization flow for the `Database` feature, instead pass the upper level function via props.

I am not sure if it even works (it wont close the dialog). Have you tried running it against real Teleport instance and not the Storybook?

alex-kovoy: I did a round of cleanups and created a branch, here is my commit with cleanups https://github.com/gravitational/webapps/commit/270f9295af880e0f150a48165459ddd72a6c8969 .

alex-kovoy: Updated the link https://github.com/gravitational/webapps/commit/a39d798191d0fd1b6dc53d265b07fcb67497fcd6

PR #386, opened 2021-09-03
performance test setup

ibeckermayer:
Adds the necessary framework for loading `Websocket.onmessage` calls from json and running them in a storybook.

One issue I haven't solved is that the "recording" I'm loading from json is fit to a particular (large) screen size. We will eventually need to solve this problem for playback, but that's a relatively low priority issue. 

Regardless, we can use this to test the time from when an `onmessage` is received to actually rendering one the screen. That said, I don't see why we can't do the same when talking over a real websocket. Open for discussion.

### DIFF
--- a/packages/teleport/src/DesktopSession/DesktopSession.story.tsx
+++ b/packages/teleport/src/DesktopSession/DesktopSession.story.tsx
@@ -1,0 +1,61 @@
+/*
+Copyright 2021 Gravitational, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


alex-kovoy (approved):

ibeckermayer:
On further thought, this design is useful for controlling the throughput of the [simulated] websocket onmessage calls, so we can compare the throughput (as in, say, canvas draw calls per second) on the frontend of different web-worker architectures / canvas architectures.

I'm going to hold off on merging for now until I re-sync with Andrew about which tests we want to run.

Issue #388, opened 2021-09-04
Broken Table Layout on Session screens


alex-kovoy:
![image](https://user-images.githubusercontent.com/7816406/132078592-9e2519a5-9d94-4126-a75b-409f0919af1b.png)

PR #390, opened 2021-09-09
Fix broken paged table layout

rudream:
## Purpose

This PR resolves https://github.com/gravitational/webapps/issues/388

In some cases the table could exceed the width of its parent and cause it to be misaligned with the pager section above it. this caused it to look clipped and broken.

## Implementation

The issue was that the pager section was respecting the parent div's width, while the table section was overflowing. Adding a `min-width: min-content` to the parent div ensures that it will always be at least as wide as its widest child, and the pager section's width can match and be aligned with the table regardless of how wide it is.

Also fixed the rounded corners so that the table doesn't have rounded corners if the pager section is above it.

### Before

broken

### After

fixed


rudream:
The Drone check should stop failing once https://github.com/gravitational/webapps.e/pull/71 is merged and e-ref is updated

alex-kovoy (approved):

PR #385, opened 2021-08-30
launch desktop window

ibeckermayer:
This PR gets the remote desktop system working up to the point of being able to connect to and view a remote desktop (but you can't click or type or do anything yet). The purpose of this PR is to get a review on the general architecture and approach up to this point.

I imitated the Console's design in taking the websocket connection parameters and username from the URL. This way users can navigate directly to a remote desktop url.

#### Still missing:

- Loading behavior
- Resize behavior
- Information and control TopBar (see [Avocode](https://app.avocode.com/revisions/82719596/?design=55636388))
- Tests

#### Test the system manually by

1. [Setting up and running a Windows Server vm in VirtualBox](https://gravitational.slab.com/posts/desktop-access-srk4fa31#getting-started-with-virtual-box-and-windows)
2. Building teleport with the `desktop_access_beta` tag, i.e. `GOOS=darwin GOARCH=amd64 CGO_ENABLED=1 go build -v -tags "desktop_access_beta" -o build/teleport  -ldflags '-w -s ' ./tool/teleport`
3. Creating a `teleport.yaml` like the example below with the network address for your running Windows Server vm
4. Running teleport with the `TELEPORT_DEV_RDP_PASSWORD` set to your Windows Server vm's password, i.e. `sudo TELEPORT_DEV_RDP_PASSWORD=your_windows_vm_pwd DEBUG=1 ./build/teleport start -c path/to/teleport-desktop-beta.yaml -d --insecure`
5. Connecting to the machine through the UI by navigating to the Desktops page and entering your username in the connect dropdown and hitting Enter. This should pop open a new window that should display the login and then desktop of the remote windows desktop like in the screenshot below.

`teleport-desktop-beta.yaml`
```yaml
teleport:
  nodename: Isaiahs-MacBook-Pro.local
  log:
    output: stderr
    severity: INFO
    format:
      output: text
  ca_pin: []
  diag_addr: ""
  data_dir: /Users/ibeckermayer/teleport-config/desktop-access/data
auth_service:
  enabled: "yes"
  listen_addr: 0.0.0.0:3025
  license_file: /Users/ibeckermayer/teleport/e/fixtures/license-enterprise.pem
  authentication:
    second_factor: off
ssh_service:
  enabled: "yes"
  labels:
    env: example
  commands:
    - name: hostname
      command: [hostname]
      period: 1m0s
proxy_service:
  enabled: "yes"
  listen_addr: 0.0.0.0:3023
  https_keypairs: []
  acme: {}
[...]

PR #391, opened 2021-09-13
Desktop Access Session with Toolbar

ibeckermayer:
This PR adds the core structure for the Desktop Access window with a toolbar. 

The toolbar itself has state based on the session settings (currently these settings `clipboard` and `recording` are hardcoded, but in the future will come from the backend).

Screen Shot 2021-09-13 at 10 25 24 AM
Screen Shot 2021-09-13 at 10 25 15 AM

The `...` menu also has a `Disconnect` option for the user to explicitly Disconnect the session

Screen Shot 2021-09-13 at 10 27 23 AM

The requirement of adding this `Disconnected` state made the overall state of this component more complex than what our typical `useAttempt` hook offers. To handle this, I added
```typescript
export type DesktopSessionAttempt = {
  status: Attempt['status'] | 'disconnected';
  statusText?: string;
};
```
and a "meta attempt" state
```typescript
const [attempt, setAttempt] = useState<DesktopSessionAttempt>({
    status: 'processing',
  });
```

This meta state tracks the logical combination of the websocket/tdpclient connection, the initial api call to fetch the Desktop (which will eventually evolve to or add another call to fetch a session with session settings like `clipboard` and `recording`), and whether or not the user has purposefully disconnected the session. See the storyboard for how those states look.

See the [**Test the system manually by**](https://github.com/gravitational/webapps/pull/385#issue-722861571) heading here for how to build/use Teleport with the Desktop Access beta feature.


packages/teleport/src/DesktopSession/TopBar.tsx
```diff
@@ -0,0 +1,167 @@
...
+        </Text>
+      </>
+
+      <>
+        {/* Center element. Only one of these is shown at a time depending on attempt.status */}
+        <Text style={clipboardTextStyle}>
+          <Clipboard style={clipboardStyle} pr={2} />
+          Clipboard Sharing {clipboard ? 'Enabled' : 'Disabled'}
+        </Text>
+        <Text style={connectingStyle}>Connecting...</Text>
```
kimlisa, line 148: hm, wondering why we couldn't just inline the styles? 
```
{ attempt.status === 'processing' && <Text style={...}> }
```

or we could also pass the attempt down to a styled comp and do our checking for states there?
```
const SomeName = styled.div`
    color: ${({ attempt }) => attempt.state === 'proccessing' ? ....};
`;
```

packages/teleport/src/DesktopSession/ActionMenu.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const theme = useTheme();
+
+  return (
+    <Flex alignItems="center">
+      <MenuIcon
+        buttonIconProps={{
+          ml: 4,
+          size: 0,
+          style: { fontSize: '20px', color: theme.colors.text.secondary },
+        }}
```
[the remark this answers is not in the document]
ibeckermayer: Confused by a few things here:

- Why can I pass `color` as an object field, but need to pass `fontSize` in the `style` object?
- How/where does the string `'text.secondary'` get interpreted to mean `props.theme.colors.text.secondary`? Or is that a misunderstanding of what the `StyledThemeProvider` is passing down to child objects?

packages/teleport/src/DesktopSession/ActionMenu.tsx
```diff
@@ -0,0 +1,63 @@
...
+const menuProps = {
+  menuListCss,
+  anchorOrigin: {
+    vertical: 'center',
+    horizontal: 'center',
+  },
+  transformOrigin: {
+    vertical: 'top',
+    horizontal: 'center',
+  },
```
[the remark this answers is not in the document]
ibeckermayer: Seems not. I also removed from `packages/teleport/src/Console/ActionBar/ActionBar.tsx` and `packages/teleport/src/Player/ActionBar/ActionBar.tsx`.

packages/teleport/src/DesktopSession/DesktopSession.tsx
```diff
@@ -17,42 +17,59 @@ limitations under the License.
...
+  } = props;
   const canvasRef = React.useRef<HTMLCanvasElement>(null);
 
   // Waits for the state hook to initialize the TdpClient.
-  // Once the client is initialized, calls for the client to connect to the
-  // server and passes in the canvas for the client to render to.
+  // Once the client is initialized, sets wsAttempt to 'success'.
   React.useEffect(() => {
-    setAttempt({ status: 'processing' });
+    setWsAttempt({ status: 'processing' });
```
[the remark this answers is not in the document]
ibeckermayer: The reason its out here is that on a render event, we need to draw to the canvas, and the canvasRef necessarily lives here. My initial thought was that keeping all of the tdpClient event listeners in one place was the best design. 
```typescript
tdpClient.on('render', ({ bitmap, left, top }) => {
      const ctx = canvasRef.current.getContext('2d');
      ctx.drawImage(bitmap, left, top);
    });
```

But now that I'm thinking about it again, it would be preferable to keep all the `setXXX` state in the state hook and only house the `render` event here.

packages/teleport/src/DesktopSession/DesktopSession.tsx
```diff
@@ -17,42 +17,59 @@ limitations under the License.
...
+    // otherwise assume a server error.
+    tdpClient.on('close', (message: { userDisconnected: boolean }) => {
+      if (message.userDisconnected) {
+        setAttempt({ status: 'disconnected' });
+      } else {
+        setWsAttempt({
+          status: 'failed',
+          statusText: 'server error',
+        });
+      }
```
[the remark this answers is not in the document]
ibeckermayer: As this is written it's not only associated with the initialization attempt -- `close` can be caused by the server severing the connection at any time.

Let me think about your proposition above (https://github.com/gravitational/webapps/pull/391#discussion_r708007444). Going with that may solve the issue.

packages/teleport/src/DesktopSession/TopBar.tsx
```diff
@@ -0,0 +1,167 @@
...
+  onDisconnect: VoidFunction;
+};
+
+const RecordingIndicator = styled.div``;
+
+export default function TopBar(props: TopBarProps) {
+  const { userHost, clipboard, recording, attempt, onDisconnect } = props;
+  const theme = useTheme();
+
+  const noneUntilSuccess = () => {
```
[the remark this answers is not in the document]
ibeckermayer: I'm confused -- why does it need to be defined outside of this comp? It relies on `props.attempt`

packages/teleport/src/DesktopSession/TopBar.tsx
ibeckermayer: > hm, wondering why we couldn't just inline the styles?

We could. Initially I had it that way, but there was enough repetition that I elected for this approach. I was also finding it difficult to edit/read the JSX with so many ternary operator statements in there.

> or we could also pass the attempt down to a styled comp and do our checking for states there?

I had something like that initially but I prefer the `useTheme` hook. The styled comp has so much implicit "magic" in comparison.

packages/teleport/src/DesktopSession/useDesktopSession.tsx
```diff
@@ -36,10 +71,42 @@ export default function useDesktopSession() {
...
+    ) {
+      setAttempt({ status: 'processing' });
+    } else if (
+      wsAttempt.status === 'success' &&
+      fetchDesktopAttempt.status === 'success'
+    ) {
+      setAttempt({ status: 'success' });
+    } else {
+      setAttempt({ status: 'failed', statusText: 'unknown error' });
+    }
```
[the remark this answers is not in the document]
ibeckermayer: Great idea, that worked.

packages/teleport/src/DesktopSession/useDesktopSession.tsx
```diff
@@ -59,44 +59,67 @@ export default function useDesktopSession(ctx: Ctx) {
...
+
+        // Triggered in the event of a websocket protocol error.
+        tdpClient.on('error', () => {
+          setAttempt({
+            status: 'failed',
+            statusText: 'websocket protocol error',
+          });
+
+          tdpClient.removeAllListeners();
+        });
```
alex-kovoy, line 99: looks complicated, I wonder why did not you like the previous version where tdpClient was close to `canvas` element?

packages/teleport/src/DesktopSession/ActionMenu.tsx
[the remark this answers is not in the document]
kimlisa: > Why can I pass color as an object field, but need to pass fontSize in the style object?

Because of styled systems, if you dig into `MenuIcon` you'll see this: https://github.com/gravitational/webapps/blob/master/packages/design/src/ButtonIcon/ButtonIcon.jsx#L93, I think in order for `fontSize` to be recognized as a prop, we need to also include `{typography}`: https://styled-system.com/api/#typography

> How/where does the string 'text.secondary' get interpreted to mean props.theme.colors.text.secondary? Or is that a misunderstanding of what the StyledThemeProvider is passing down to child objects?

From my understanding, the styled-components `ThemeProvider` allows styled components to access the theme object: https://styled-components.com/docs/advanced#theming. If you dig into `MenuIcon`, you'll see we pass these styling props to a styled component: https://github.com/gravitational/webapps/blob/master/packages/design/src/ButtonIcon/ButtonIcon.jsx#L61

packages/teleport/src/DesktopSession/TopBar.tsx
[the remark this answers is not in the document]
kimlisa: ignore, it was tied to https://github.com/gravitational/webapps/pull/391#discussion_r708421885

packages/teleport/src/DesktopSession/DesktopSession.tsx
```diff
@@ -17,58 +17,45 @@ limitations under the License.
...
 
 export function DesktopSession(props: State) {
-  const { setAttempt, tdpClient } = props;
+  const { userHost, tdpClient, attempt, clipboard, recording } = props;
   const canvasRef = React.useRef<HTMLCanvasElement>(null);
 
   // Waits for the state hook to initialize the TdpClient.
-  // Once the client is initialized, calls for the client to connect to the
-  // server and passes in the canvas for the client to render to.
+  // Once the client is initialized, sets wsAttempt to 'success'.
```
kimlisa, line 35: don't think this comment is accurate anymore

packages/teleport/src/DesktopSession/useDesktopSession.tsx
```diff
@@ -14,16 +14,42 @@ See the License for the specific language governing permissions and
...
+import { useMemo, useEffect, useState } from 'react';
 import { getAccessToken, getHostName } from 'teleport/services/api';
 import { useParams } from 'react-router';
-import useAttempt from 'shared/hooks/useAttemptNext';
+import { Attempt } from 'shared/hooks/useAttemptNext';
 import cfg, { UrlDesktopParams } from 'teleport/config';
 import TdpClient from 'teleport/lib/tdp/client';
+import Ctx from 'teleport/teleportContext';
+import { Desktop } from 'teleport/services/desktops';
+import { stripRdpPort } from '../Desktops/DesktopList';
```
kimlisa, line 25: feels a bit weird to import something from `DesktopList`, maybe have `Desktops/index.ts` re-export it?

packages/teleport/src/DesktopSession/TopBar.tsx
kimlisa: i think using styled components is the better approach but we can also get a second opinion from @alex-kovoy

packages/teleport/src/DesktopSession/useDesktopSession.tsx
ibeckermayer: I think you're referring to [this version here](https://github.com/gravitational/webapps/blob/222768dd9bb95843c1950032965948aa79a45c22/packages/teleport/src/DesktopSession/DesktopSession.tsx#L44-L88). The original impetus for moving away from that design was Lisa's comment [here](https://github.com/gravitational/webapps/pull/391#discussion_r707970871), where she suggested keeping state management encapsulated in the `useDesktopSession` hook (my sense is this is standard practice in the codebase with infrequent exceptions). 

This implementation does that, while at the same time implementing Lisa's other suggestion [here](https://github.com/gravitational/webapps/pull/391#discussion_r708007444). With this, we only need to track a single `DesktopSessionAttempt` for the entire connection state, rather than tracking both `wsAttempt` and `fetchDesktopAttempt`, and use [this gross and complex logic to synchronize them](https://github.com/gravitational/webapps/blob/222768dd9bb95843c1950032965948aa79a45c22/packages/teleport/src/DesktopSession/useDesktopSession.tsx#L81-L99).

I agree though, that this is complicated. Another approach lies with making `tdpClient.connect()` [return a `Promise`](https://github.com/gravitational/webapps/blob/2790402525dcfb43a73eb7789004ae151e752c70/packages/teleport/src/lib/tdp/client.ts#L35-L69). Then, this big disgusting thing can just be
```typescript
  useEffect(() => {
    Promise.all([
      ctx.desktopService.fetchDesktops(clusterId),
      tdpClient.connect(),
    ])
      .then(vals => {
        makeUserHost(vals[0]);
        setAttempt({ status: 'success' });
      })
      .catch(err => {
        setAttempt({ status: 'failed', statusText: err.message });
      });
  }, [clusterId, username, desktopId]);
```

(I had this in a previous commit, and then created this mess based on what I now see as a mistaken premise that it was important to keep the `tdpClient` an `EventEmitter` exclusively, and not confuse a user by giving them the option to deal with a `Promise` _or_ an `'open'` Event. That doesn't hold up to scrutiny though, since nobody is confused...)

In that case, we have the option of either adding these event listeners here or in `DesktopSession` next to the `canvas`. My argument for keeping them in here is that by doing so, `tdpClient` events that create side effects by changing state are encapsulated in the hook, alongside any api calls (currently `ctx.desktopService.fetchDesktops(clusterId)`) and the initial websocket connection which do the same. Only the `'render`' event listener need be in `DesktopSession`, distinguished by the fact that it doesn't change the component's state and requires the `canvas` to draw to.

alex-kovoy: I think we are dealing with 2 states: one is the session and another one is the canvas. Sometimes it's OK to lift a state up and sometimes it does not. In this case I would keep it closer to the `canvas` as it makes it easier to reason about (because tdpClient _updates_ canvas, canvas is the state and tdpClient is an API).

russjones:
@ibeckermayer Once you have something kind of working, can you do a quick demo for @klizhentas?

packages/teleport/src/DesktopSession/useDesktopSession.tsx
```diff
@@ -14,32 +14,43 @@ See the License for the specific language governing permissions and
...
 
   return {
-    tdpClient,
+    hostname,
     attempt,
-    setAttempt,
+    // clipboard and recording settings will eventuall come from backend, hardcoded for now
+    clipboard: false,
+    recording: false,
+    ...useTdpClientCanvas(),
```
alex-kovoy, line 53: lets move it to the top

packages/teleport/src/DesktopSession/DesktopSession.tsx
```diff
@@ -15,87 +15,112 @@ limitations under the License.
...
+      connection.status === 'connected'
+    ) {
+      return null;
+    } else {
+      <Alert
+        style={{
+          alignSelf: 'center',
+        }}
+        width={'450px'}
+        my={2}
```
alex-kovoy, line 90: consider simplifying it a bit if you have time

alex-kovoy (approved):

PR #394, opened 2021-09-19
Share link for current time in player

KonradKuznicki:
Adds button on the progress bar to share player at current time. 
Implements https://github.com/gravitational/teleport/issues/996.

Peek 2021-09-19 22-34

### Behavior
When ProgressBarTty is used with linkConstructor,
ProgressBar will display additional button,
which on click shows popup with link to session player at current time.
While popup is open, playback is paused if it was playing.

#### Link construction
link is constructed based on  window.location + from=<time> this way it's independent from cfg.routes and can be reused
 
#### Rewind
it is implemented inside SshPlayer so it is not generic and PlayerNext needs it's own implementation
when SshPlayer is constructed 
and  tty connects 
tty moves to position from url and stops 

### PlayerNext
PlayerNext can be used both with tty as an argument or create tty on it's own from given url
and I couldn't find any usage of PlayerNext in the code base.
Therefore I'm not sure where to implement this "rewind" capability so "link" button is just disabled


alex-kovoy:
@KonradKuznicki thank you for submission, we will discuss this feature internally with the product and UX teams and once it's ready we will use your PR for those changes.

PR #395, opened 2021-09-20
Clear localStorage on Login

rudream:
## Purpose

The browser `localStorage` gets cleared when the user clicks the “Sign Out” button, however, it is possible to navigate to the `/web/login` page directly and log into a new user without clicking “Sign Out". When logging into a new user in this manner, the `localStorage` can still contain remnants of the previous user and cause unintended UI behavior, for example, the "role assumed" banner from the previous user can persist. This PR clears the remnants of the previous user on login.


rudream:
Closing as this functionality was added as part of this PR: https://github.com/gravitational/webapps/pull/1275

PR #409, opened 2021-10-06
`requestAnimationFrame` with buffer and performance test

ibeckermayer:
- Uses the current best iteration performance-wise for rendering the desktop.
- Adds a `saveMessages` flag to `client.ts` so that additional performance tests can be easily generated again in the future.
- Adds a `Performance` storybook story.

### DIFF
--- a/packages/teleport/src/lib/tdp/client.ts
+++ b/packages/teleport/src/lib/tdp/client.ts
@@ -42,8 +42,8 @@ export default class Client extends EventEmitter {
       this.emit('init');
     };
 
-    this.socket.onmessage = (ev: MessageEvent<ArrayBuffer>) => {
-      this.processMessage(ev.data);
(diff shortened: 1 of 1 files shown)


packages/teleport/src/DesktopSession/DesktopSession.story.tsx
```diff
@@ -135,3 +140,45 @@ export const ConnectionError = () => (
...
+  const client = fakeClient();
+  var startTime,
+    endTime,
+    i = 0;
+
+  return (
+    <DesktopSession
+      {...props}
+      fetchAttempt={{ status: 'success' }}
+      connection={{ status: 'success' }}
```
rudream, line 153: I think we should rename this `connection` prop to `connectionAttempt` to be more consistent and avoid confusion

rudream (approved):
see my comment

PR #413, opened 2021-10-08
Add database created/updated/deleted events

r0mant:
Screen Shot 2021-10-08 at 10 03 47 AM


packages/teleport/src/services/audit/audit.test.ts
```diff
@@ -75,6 +75,75 @@ test('fetch events', async () => {
         user: 'root',
       },
     },
+    {
+      codeDesc: 'Database Created',
```
kimlisa, line 79: sorry i overlooked this, can we revert this file? this is just to test correct formatting of events when fetching with just select random events

r0mant: @kimlisa Isn't this a unit test for events formatting? I can revert but I thought it's useful to make sure your event message is formatted correctly for example. There's also a storybook test which I added them to as you asked, but that one is very messy with all the html.

r0mant: 🤷‍♂️ I think it's much more convenient to test your event message formatting with this unit test than updating the storybook but it's your land :) I'll remove.

PR #414, opened 2021-10-11
userACL

ibeckermayer:
- Resolves the ACL so that the Desktops tab is not displayed if list access is not granted.
- Displays the logins allowed by the user's role's `windows_desktop_logins` field (a carbon copy of how we do the same for `logins` for ssh nodes in `NodeList.tsx`.
- Converts text that says "Desktop" into "Desktop (preview)".
- Makes the top bar shorter.
- Changes the "Domain" column to "Hostname" in the Dekstops table to avoid confusion with "LDAP domain" (same term we use for ssh nodes).

Relies on running teleport with the changes from [this branch](https://github.com/gravitational/teleport/tree/isaiah/windows-rbac/user-acl) ([PR here](https://github.com/gravitational/teleport/pull/8560))

### DIFF
--- a/packages/teleport/src/Desktops/DesktopList/DesktopList.tsx
+++ b/packages/teleport/src/Desktops/DesktopList/DesktopList.tsx
@@ -124,7 +124,7 @@ const DesktopDomainCell = props => {
   return <Cell {...rest}>{addr}</Cell>;
 };
 
-// TODO: may be able to be abstracted out from here/NodeList.tsx
+// TODO(isaiah): may be able to be abstracted out from here/NodeList.tsx
 const LoginCell: React.FC<Required<{
   onSelect?: (
     e: React.SyntheticEvent,
@@ -207,8 +207,8 @@ type Props = {
   username: string;
   clusterId: string;
   searchValue: string;
-  onLoginMenuOpen: (desktopId: string) => { login: string; url: string }[];
-  openRemoteDesktopWindow: (username: string, desktopId: string) => void;
+  onLoginMenuOpen(desktopId: string): { login: string; url: string }[];
+  openRemoteDesktopWindow(username: string, desktopId: string): void;
 };
 
 export default DesktopList;
--- a/packages/teleport/src/Desktops/useDesktops.ts
+++ b/packages/teleport/src/Desktops/useDesktops.ts
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import { useState, useEffect, useCallback } from 'react';
+import { useState, useEffect } from 'react';
 import useAttempt from 'shared/hooks/useAttemptNext';
 import Ctx from 'teleport/teleportContext';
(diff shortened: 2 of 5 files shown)


packages/teleport/src/Desktops/useDesktops.ts
```diff
@@ -26,11 +26,17 @@ export default function useDesktops(ctx: Ctx) {
   const { attempt, run } = useAttempt('processing');
   const { clusterId } = useStickyClusterId();
   const username = ctx.storeUser.state.username;
+  const windowsLogins = ctx.storeUser.getWindowsLogins();
 
   const [searchValue, setSearchValue] = useState<string>('');
 
   const [desktops, setDesktops] = useState<Desktop[]>([]);
 
+  const getWindowsLoginOptions = useCallback(
```
zmb3, line 35: Are you certain that `useCallback` is necessary here?

packages/teleport/src/Desktops/DesktopList/DesktopList.tsx
```diff
@@ -206,6 +207,7 @@ type Props = {
   username: string;
   clusterId: string;
   searchValue: string;
+  onLoginMenuOpen: (desktopId: string) => { login: string; url: string }[];
```
kimlisa, line 210: i think we prefer this syntax (i also think its easier to read)
```suggestion
  onLoginMenuOpen(desktopId: string): { login: string; url: string }[];
```

kimlisa (approved):

packages/teleport/src/Desktops/useDesktops.ts
kimlisa: not needed per alexey (could we also remove it from `useNodes`)

packages/teleport/src/services/user/makeAcl.ts
```diff
@@ -32,9 +33,11 @@ export default function makeAcl(json): Acl {
   const accessRequests = json.accessRequests || defaultAccess;
   const billing = json.billing || defaultAccess;
   const dbServers = json.dbServers || defaultAccess;
+  const desktopServers = json.desktopServers || defaultAccess;
 
   return {
     logins,
+    windowsLogins,
```
alex-kovoy, line 40: windows or desktop logins?

zmb3: In the RFD Andrew was pretty intentional about dedicated "Windows Logins" instead of using the existing field we have for *nix logins, so let's keep them separate.

packages/teleport/src/DesktopSession/TopBar.tsx
```diff
@@ -89,7 +89,7 @@ export default function TopBar(props: Props) {
   );
 }
 
-export const TopBarHeight = 56;
+export const TopBarHeight = 40;
```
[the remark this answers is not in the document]
alex-kovoy: I think you can calculate it without any hardcoded values.

PR #415, opened 2021-10-12
Add converter and new config field

kimlisa:
#### Description
None of these changes will affect anything in master.

- When receiving webauthn related challenges from server, base64URL strings need to be converted to ArrayBuffer (expected type for navigator.credentials api)
- When sending back navigator.credentials api response back to server, we need to convert ArrayBuffer back to base64URL string (expected type for server)
- Added `preferredLocalMfa` to our config file, added `webauthn` as an auth type

Note the converter functions were copied from https://github.com/github/webauthn-json/blob/main/src/base64url.ts, I copied and pasted their [license file content](https://github.com/github/webauthn-json/blob/main/LICENSE.md) to the top of file


packages/shared/utils/base64url.ts
```diff
@@ -0,0 +1,61 @@
...
+export function bufferToBase64url(buffer: ArrayBuffer): Base64urlString {
+  // Buffer to binary string
+  const byteView = new Uint8Array(buffer);
+  let str = '';
+  for (const charCode of byteView) {
+    str += String.fromCharCode(charCode);
+  }
+
+  // Binary string to base64
+  const base64String = btoa(str);
```
alex-kovoy, line 52: Will this work instead? If so, you can refactor this code and remove the license header

```ts
 const binary = String.fromCharCode.apply(null, new Uint8Array(buffer));
 const base64String = btoa(binary);
```

PR #418, opened 2021-10-14
Add webauthn methods to auth service

kimlisa:
#### Description
- Define webauth methods for changing password, login, invite/reset
- Define navigator credetial get/create response makers

#### Enterprise changes for recovery service
https://github.com/gravitational/webapps.e/pull/77

### DIFF
--- a/packages/teleport/src/services/auth/auth.ts
+++ b/packages/teleport/src/services/auth/auth.ts
@@ -22,6 +22,8 @@ import makePasswordToken from './makePasswordToken';
 import {
   makeMfaAuthenticateChallenge,
   makeMfaRegistrationChallenge,
+  makeNavCredentialsGetResponse,
+  makeNavCredentialsCreateResponse,
 } from './makeMfa';
 import { DeviceType } from './types';
 
@@ -33,6 +35,16 @@ const auth = {
 
     return new Error(
       'this browser does not support U2F required for hardware tokens, please try Chrome or Firefox instead'
+    );
+  },
+
+  webauthnBrowserSupported() {
+    if ('credentials' in navigator) {
+      return null;
+    }
+
+    return new Error(
+      'this browser does not support Webauthn required for hardware tokens, please try Chrome, Firefox or Safari with the latest update'
     );
   },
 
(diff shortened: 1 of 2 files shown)


packages/teleport/src/services/auth/makeMfa.ts
```diff
@@ -72,3 +72,61 @@ export function makeMfaAuthenticateChallenge(json): MfaAuthenticateChallenge {
...
 }
+
+// makeNavCredentialsCreateResponse takes response from navigator.credentials.create
+// and creates a credential object expected by the server with ArrayBuffer
+// fields converted to Base64URL:
+// - rawId
+// - response.attestationObject
+// - response.clientDataJSON
+export function makeNavCredentialsCreateResponse(res) {
+  // Response can be null if no Credential object can be created.
```
kimlisa, line 83: it's unclear how these null responses happen, but looking through these comments https://github.com/w3c/webauthn/issues/876, it's rare to happen. Yassine was able to trigger it for `navigator.credentials.get` which was resolved by trying again, but we don't remember _how_ it happened and couldn't reproduce it.

packages/teleport/src/services/auth/auth.ts
```diff
@@ -36,6 +38,16 @@ const auth = {
     );
   },
 
+  webauthnBrowserSupported() {
+    if ('credentials' in navigator) {
+      return null;
+    }
+
+    return new Error(
```
alex-kovoy, line 46:
```ts

checkWebauthn() {
   if ('credentials' in navigator) {
      return Promise.resolve();
    }

    return Promise.reject( new Error('this browser does not support Webauthn required for hardware tokens, please try Chrome, Firefox or Safari with the latest update')
},

 loginWithWebauthn(user: string, pass: string) {
    return this.checkWebAuthn()
       .then(() => auth.mfaLoginBegin(user, pass))
       .then(...)
}

```

packages/teleport/src/services/auth/auth.ts
```diff
@@ -36,6 +38,16 @@ const auth = {
...
+      'this browser does not support Webauthn required for hardware tokens, please try Chrome, Firefox or Safari with the latest update'
```
ibeckermayer, line 47:
```suggestion
      'this browser does not support Webauthn required for hardware tokens, please try the latest version of Chrome, Firefox or Safari'
```

packages/teleport/src/services/auth/makeMfa.ts
ibeckermayer: If I'm understanding correctly, if we add logic for a single retry on null response [here](https://github.com/gravitational/webapps/blob/lisa/define-webauthn-methods/packages/teleport/src/services/auth/auth.ts#L169), that should solve all but the rarest cases until the underlying webauthn implementation is fixed.

packages/teleport/src/services/auth/makeMfa.ts
```diff
@@ -72,3 +72,61 @@ export function makeMfaAuthenticateChallenge(json): MfaAuthenticateChallenge {
...
+// and creates a credential object expected by the server with ArrayBuffer
+// fields converted to Base64URL:
+// - rawId
+// - response.authenticatorData
+// - response.clientDataJSON
+// - response.signature
+// - response.userHandle
+export function makeNavCredentialsGetResponse(res) {
+  // Response can be null if Credential cannot be unambiguously obtained.
+  if (!res) {
```
ibeckermayer, line 113: Same as `makeNavCredentialsCreateResponse`

alex-kovoy (approved):

PR #422, opened 2021-10-19
Desktop Access Audits

ibeckermayer:
Configures the audit logs to display desktop access event

Screen Shot 2021-10-18 at 10 52 12 PM
s


packages/teleport/src/services/audit/makeEvent.ts
```diff
@@ -451,6 +448,24 @@ export const formatters: Formatters = {
...
+    type: 'windows.desktop.session.start',
+    desc: 'Windows Desktop Session Denied',
+    format: ({ user, windows_domain, desktop_addr, windows_user }) =>
+      `User [${user}] was denied access to Windows desktop [${windows_user}@${desktop_addr}] on [${windows_domain}]`,
+  },
+  [eventCodes.DESKTOP_SESSION_ENDED]: {
+    type: 'windows.desktop.session.end',
+    desc: 'Windows Desktop Session Ended',
+    format: ({ user, windows_domain, desktop_addr, windows_user }) =>
+      `User [${user}] has disconnected from Windows desktop [${windows_user}@${desktop_addr}] on [${windows_domain}]`,
```
zmb3, line 467: You could also see this event for reasons other than the user disconnecting (connection idle timeout, user locked, etc)

In these cases, you'll get additional audit events for those scenarios, but it might be helpful to reword the text here so it doesn't always look like a user initiated disconnect. 

Maybe something like "Windows desktop session X for user Joe at Administrator@desktop has ended."

zmb3 (commented):
Looks good. General comment on wording of the disconnect message.

packages/teleport/src/services/audit/makeEvent.ts
zmb3: No. For example, you could have an idle timeout of 30 minutes configured. The session could start successfully and the user could work on the desktop for 2-3 hours and then step away. After 30 minutes of inactivity, we'll close the session due to inactivity.

We wouldn't emit a `DESKTOP_SESSION_STARTED_FAILED`, because the session didn't fail to start. It started just fine.

We would emit a `DESKTOP_SESSION_ENDED` when the session is no longer active, but the text should be something more generic and not make it look like the user terminated it.

kimlisa (approved):
you might want to run `yarn test` and update snapshot

packages/teleport/src/services/audit/makeEvent.ts
ibeckermayer: `Session for Windows desktop [${windows_user}@${desktop_addr}] on [${windows_domain}] has ended for user [${user}]`

alex-kovoy (approved):

PR #440, opened 2021-11-01
Add audit log support for privilege token event

rudream:
## Purpose

Add audit log support for the `privilege_token.create` event

## Implementation

Added support for the new privilege token creation event.

Due to the fact that we've now exceeded 50 events (which was the `pageSize` we were using), new events added would push other events onto the second page and they would disappear from the snapshot (which was only of the first page). To resolve this, the `EventsList` now accepts a `pageSize` prop so we can snapshot the full list of all events. The `AllEvents` story is the list of all the events we support, while the `LoadedSample` story only has a few events so that we can snapshot the full screen including search bar etc.

## Demo

image


packages/teleport/src/Audit/EventList/EventList.tsx
```diff
@@ -27,7 +27,14 @@ import EventDialog from '../EventDialog';
...
 
 export default function EventList(props: Props) {
-  const { clusterId, events = [], search = '', fetchMore, fetchStatus } = props;
+  const {
+    clusterId,
+    events = [],
+    search = '',
+    fetchMore,
+    fetchStatus,
+    pageSize = 100,
```
kimlisa, line 36: lets default to 50, like we had before
```suggestion
    pageSize = 50,
```

packages/teleport/src/services/audit/types.ts
```diff
@@ -75,6 +75,7 @@ export const eventCodes = {
   OIDC_CONNECTOR_DELETED: 'T8101I',
   PORTFORWARD_FAILURE: 'T3003E',
   PORTFORWARD: 'T3003I',
+  PRIVILEGE_TOKEN_CREATED: 'T6002I',
```
kimlisa, line 78: nit: can we have this below line 79

packages/teleport/src/Audit/Audit.story.tsx
```diff
@@ -28,11 +29,22 @@ export default {
 export const Loaded = () => {
   const ctx = new Context();
   ctx.auditService.fetchEvents = () =>
-    Promise.resolve({ events, startKey: '' });
+    Promise.resolve({ events: events.slice(0, 20), startKey: '' });
```
kimlisa, line 32: this can break the snapshot b/c it relies on the events order, so lets make a `eventsSample` in our fixtures, where we make like 4 events, and use that here. Lets also use this sample for the `LoadedFetchMore`

To make it easier on folks who add events for us, we can place a comment for `export const events = [`, that states that this list contains all known events

packages/teleport/src/Audit/Audit.story.tsx
```diff
@@ -28,11 +29,22 @@ export default {
...
 
   return render(ctx);
 };
 
+export const AllEventsList = () => (
```
kimlisa, line 37:
```suggestion
export const AllEvents = () => (
```

and lets place this at the bottom of file, so its last in storybook list

packages/teleport/src/Audit/Audit.story.tsx
```diff
@@ -28,11 +29,22 @@ export default {
 export const Loaded = () => {
```
kimlisa, line 29:
```suggestion
export const LoadedSample = () => {
```
?

packages/teleport/src/Audit/fixtures/index.ts
```diff
@@ -16,6 +16,7 @@ limitations under the License.
 
 import { makeEvent } from 'teleport/services/audit';
 
+// 'events' contains sample JSON's of all supported event types
```
kimlisa, line 19:
```suggestion
// events contains sample JSON's of all supported event types
```

kimlisa (approved):

alex-kovoy (approved):

PR #442, opened 2021-11-02
Add created date to recovery codes respones

kimlisa:
part of https://github.com/gravitational/teleport.e/issues/323

#### Description
- Moved recovery code typing and make response to open source
- Fixes a regression bug I introduced where firefox still requires global setting of appId and challenges whereas chrome looks for it among the registered keys

#### Enterprise changes
https://github.com/gravitational/webapps.e/pull/84

### DIFF
--- a/packages/teleport/src/services/auth/makeRecoveryCodes.ts
+++ b/packages/teleport/src/services/auth/makeRecoveryCodes.ts
@@ -16,6 +16,9 @@
 
 import { RecoveryCodes } from './types';
 
+// makeRecoveryCodes makes the response from a successful user reset or invite.
+// Only teleport cloud and users with valid emails as username will receive
+// recovery codes.
 export function makeRecoveryCodes(json): RecoveryCodes {
   json = json || {};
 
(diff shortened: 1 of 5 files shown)


packages/teleport/src/services/auth/makeRecoveryCodes.ts
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { RecoveryCodes } from './types';
+
+export function makeRecoveryCodes(json): RecoveryCodes {
```
rudream, line 19: maybe add a comment here that explains that this only expects a response when it's with a cloud user? might be confusing to see this in OSS

rudream (approved):

alex-kovoy (approved):

PR #444, opened 2021-11-05
Use default terminal shell and add context menu

gzdunek:


packages/teleterm/src/services/pty/ptyService.ts
```diff
@@ -66,3 +67,19 @@ function buildOptions(settings: RuntimeSettings, cmd: PtyCommand): PtyOptions {
 
   throw Error(`Unknown pty command type: ${cmd.kind}`);
 }
+
+function getDefaultShell(): string {
```
alex-kovoy, line 71: lets move this method to the [runtimeSettings.ts](https://github.com/gravitational/webapps/blob/gzdunek/use-default-terminal-shell-and-add-context-menu/packages/teleterm/src/mainProcess/runtimeSettings.ts#L10)

packages/teleterm/src/services/pty/ptyService.ts
```diff
@@ -46,7 +47,7 @@ function buildOptions(settings: RuntimeSettings, cmd: PtyCommand): PtyOptions {
   switch (cmd.kind) {
     case 'new-shell':
       return {
-        path: 'bash',
+        path: getDefaultShell(),
```
alex-kovoy, line 50: Lets move `getDefaultShell` to the [runtimeSettings.ts](https://github.com/gravitational/webapps/blob/gzdunek/use-default-terminal-shell-and-add-context-menu/packages/teleterm/src/mainProcess/runtimeSettings.ts#L10) so then it can be used as 

```ts
switch (cmd.kind) {
    case 'new-shell':
      return {
        path: settings.defaultShell,
```

packages/teleterm/src/boot.tsx
```diff
@@ -17,6 +17,9 @@ appContext.serviceDocs = new ServiceDocs();
 appContext.serviceTerminals = new ServiceTerminals(
   electronGlobals.ptyServiceClient
 );
+appContext.ui = {
+  openContextMenu: electronGlobals.mainProcessClient.openContextMenu,
```
alex-kovoy, line 21:
```ts
appContext.mainProcessClient = electronGlobals.mainProcessClient
```

alex-kovoy (approved):
Left some comments, good to go after they are addressed

packages/teleterm/src/services/pty/ptyService.ts
gzdunek: Done

packages/teleterm/src/services/pty/ptyService.ts
gzdunek: Done

packages/teleterm/src/boot.tsx
gzdunek: Done

PR #448, opened 2021-11-09
"clusterId • username@hostname" document title

ibeckermayer:
fixes https://github.com/gravitational/teleport/issues/8854

### DIFF
--- a/packages/teleport/src/DesktopSession/useDesktopSession.tsx
+++ b/packages/teleport/src/DesktopSession/useDesktopSession.tsx
@@ -19,8 +19,6 @@ import { useParams } from 'react-router';
 import useAttempt from 'shared/hooks/useAttemptNext';
 import { UrlDesktopParams } from 'teleport/config';
 import Ctx from 'teleport/teleportContext';
-import { Desktop } from 'teleport/services/desktops';
-import { stripRdpPort } from '../Desktops/DesktopList';
 import { useTdpClientCanvas } from './TdpClientCanvas';
 
 export default function useDesktopSession(ctx: Ctx) {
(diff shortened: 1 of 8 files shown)


packages/teleport/src/DesktopSession/useDesktopSession.tsx
```diff
@@ -31,6 +31,8 @@ export default function useDesktopSession(ctx: Ctx) {
   const [hostname, setHostname] = useState<string>('');
   const clientCanvasProps = useTdpClientCanvas();
 
+  document.title = `${clusterId} • ${clientCanvasProps.username}@${hostname}`;
+
```
alex-kovoy, line 35: Instead of `  const { clusterId, desktopId } = useParams<UrlDesktopParams>();`, could you return this 2 values from `useTdpClientCanvas` as it has the same line? Or do this instead?

```ts

const { clusterId, username, desktopId } = useParams<UrlDesktopParams>();
const tdpClientCanvas = useTdpClientCanvas({ clusterId, desktopId, username})

useEffect(() => {
  document.title = `${clusterId} • ${username}@${hostname}`;
}, [hostname])

useEffect(() => {    
  run(() => 
    // TODO: add a method to the backend that returns a desktop by ID, meanwhile move that code to the desktopService.getDesktop method. Also, there is no need to carry on `stripRdpPort(desktop.addr)` all the way from backend to the client side. This is bad API design. 
    ctx.desktopService
      .getDesktop({ clusterId, desktopId)
      .then(desktop => {    
        setHostname(desktop.hostname);
      })    
  ));
}, [clusterId, desktopId]);

```

alex-kovoy (approved):

ibeckermayer:
Updated to use new backend api, which is updated with the changes suggested by @alex-kovoy in https://github.com/gravitational/webapps/pull/448#discussion_r748001617 (see https://github.com/gravitational/teleport/pull/9041)

alex-kovoy (approved):

packages/teleport/src/DesktopSession/useDesktopSession.tsx
```diff
@@ -25,11 +25,11 @@ export default function useDesktopSession(ctx: Ctx) {
   // Tracks combination of tdpclient/websocket and api call state,
   // as well as whether the tdp client for this session was intentionally disconnected.
   const { attempt: fetchAttempt, run } = useAttempt('processing');
-  const { username, desktopId, clusterId } = useParams<UrlDesktopParams>();
+  const { username, desktopName, clusterId } = useParams<UrlDesktopParams>();
```
alex-kovoy, line 28: @ibeckermayer  lets keep `desktopId` on the client side, no need to match the backend handler internal implementation. It will be aligned with other ids: `clusterId`, `serverId`. Also keep in mind that this is not actually a `name` but more of `id`.

alex-kovoy: @ibeckermayer after looking at the code it seems that we started using `name` for dbs and kubes. If so, you can keep this change then.

alex-kovoy: although using `name` is really confusing as most of our names are UIDs.

PR #454, opened 2021-11-13
Create a general select filter component

kimlisa:
part of https://github.com/gravitational/teleport/issues/5021

#### Screenshots

opened

image

with labels

image

PR #456, opened 2021-11-15
Create filter by label list component

kimlisa:

#### Description
- Create a component that allows you to filter by labels with any List component (given that data list contains a `tag` field)
- Grabs labels from URL query params in the form of `labels=name: value, name2: value2...`

#### Dependent PR
relies on changes made on https://github.com/gravitational/webapps/pull/454

### DIFF
--- a/packages/teleport/src/components/FilterByLabelList/FilterByLabelList.tsx
+++ b/packages/teleport/src/components/FilterByLabelList/FilterByLabelList.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import React, { useState, useMemo } from 'react';
+import React, { useMemo } from 'react';
 import { useLocation } from 'react-router';
 import { Option } from 'shared/components/Select';
 import SelectFilter from 'teleport/components/SelectFilter';
@@ -27,7 +27,8 @@ export default function FilterableList({
 }: Props) {
   const { search, pathname } = useLocation();
   const labels = useMemo<Option[]>(() => makeLabelOptions(data), [data]);
-  const [selectedLabels, setSelectedLabels] = useState<Option[]>(() => {
+
+  const selectedLabels = useMemo<Option[]>(() => {
     const searchParams = new URLSearchParams(search);
     const query = searchParams.get('labels');
     if (!query || query === 'null') {
@@ -37,16 +38,14 @@ export default function FilterableList({
     return decodeURIComponent(query)
       .split(',')
       .map(label => ({ value: label, label }));
-  });
-  const [filteredData, setFilteredData] = useState(() =>
-    filterData(data, selectedLabels)
+  }, [search]);
+
+  const filteredData = useMemo<Option[]>(
+    () => filterData(data, selectedLabels),
+    [data, selectedLabels]
   );
 
   function onFilterApply(labels: Option[]) {
-    const filtered = filterData(data, labels);
-
-    setFilteredData(filtered);
-    setSelectedLabels(labels);
     updateUrlQuery(labels, pathname);
   }
 
--- a/packages/teleport/src/components/FilterByLabelList/FilterByLabelList.tsx
+++ b/packages/teleport/src/components/FilterByLabelList/FilterByLabelList.tsx
@@ -26,7 +26,8 @@ export default function FilterableList({
   ...tableProps
 }: Props) {
   const { search, pathname } = useLocation();
-  const labelsFromQuery = useMemo<Option[]>(() => {
+  const labels = useMemo<Option[]>(() => makeLabelOptions(data), [data]);
+  const [selectedLabels, setSelectedLabels] = useState<Option[]>(() => {
     const searchParams = new URLSearchParams(search);
     const query = searchParams.get('labels');
     if (!query || query === 'null') {
@@ -36,14 +37,9 @@ export default function FilterableList({
     return decodeURIComponent(query)
       .split(',')
       .map(label => ({ value: label, label }));
-  }, []);
-
-  const labels = useMemo<Option[]>(() => makeLabelOptions(data), []);
-  const [selectedLabels, setSelectedLabels] = useState<Option[]>(
-    labelsFromQuery
-  );
+  });
   const [filteredData, setFilteredData] = useState(() =>
-    filterData(data, labelsFromQuery)
+    filterData(data, selectedLabels)
   );
 
   function onFilterApply(labels: Option[]) {
@@ -55,7 +51,7 @@ export default function FilterableList({
   }
 
   function onLabelClick(label: string) {
-    let copy = selectedLabels;
+    let copy = [...selectedLabels];
     const index = selectedLabels.findIndex(tag => tag.value === label);
 
     if (index > -1) {
@@ -85,7 +81,7 @@ export default function FilterableList({
 }
 
 function filterData(data = [], labels: Option[] = []) {
-  if (labels.length === 0) {
+  if (!labels.length) {
     return data;
   }
 
@@ -95,7 +91,7 @@ function filterData(data = [], labels: Option[] = []) {
 }
 
 function updateUrlQuery(filters: Option[], pathname = '') {
-  if (filters.length === 0) {
+  if (!filters.length) {
     history.replace(pathname);
   }
 
@@ -105,7 +101,7 @@ function updateUrlQuery(filters: Option[], pathname = '') {
 
 function makeLabelOptions(data = []): Option[] {
   // Test a tags field exist.
-  if (data.length === 0 || !data[0].tags) {
+  if (!data.length || !data[0].tags) {
     return [];
   }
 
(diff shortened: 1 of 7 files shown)


packages/teleport/src/components/FilterByLabelList/FilterByLabelList.tsx
```diff
@@ -0,0 +1,132 @@
...
+    const searchParams = new URLSearchParams(search);
+    const query = searchParams.get('labels');
+    if (!query || query === 'null') {
+      return [];
+    }
+
+    return decodeURIComponent(query)
+      .split(',')
+      .map(label => ({ value: label, label }));
+  }, []);
```
gzdunek, line 39: Shouldn't it depend on `search`? Currently, the UI won't react on manual URL change.

packages/teleport/src/components/FilterByLabelList/FilterByLabelList.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  const labels = useMemo<Option[]>(() => makeLabelOptions(data), []);
```
gzdunek, line 41: Similarly here, `eslint` will warn about missing deps

packages/teleport/src/components/FilterByLabelList/FilterByLabelList.tsx
```diff
@@ -0,0 +1,132 @@
...
+    updateUrlQuery(labels, pathname);
+  }
+
+  function onLabelClick(label: string) {
+    let copy = selectedLabels;
+    const index = selectedLabels.findIndex(tag => tag.value === label);
+
+    if (index > -1) {
+      // remove the label
+      copy.splice(index, 1);
```
gzdunek, line 63: It also mutates original `selectedLabels` here, I would choose to filter out that index

packages/teleport/src/components/FilterByLabelList/FilterByLabelList.tsx
```diff
@@ -0,0 +1,132 @@
...
+    return data;
+  }
+
+  return data.filter(obj =>
+    labels.every(label => obj.tags.toString().includes(label.value))
+  );
+}
+
+function updateUrlQuery(filters: Option[], pathname = '') {
+  if (filters.length === 0) {
```
gzdunek, line 98: Nit: shorter version: `!filters.length`

packages/teleport/src/components/FilterByLabelList/FilterByLabelList.tsx
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  const tagDict = {};
+  data.forEach(({ tags }) => {
+    tags.forEach(tag => {
+      if (!tagDict[tag]) {
+        tagDict[tag] = tag;
+      }
+    });
+  });
```
gzdunek, line 119: Maybe shorter version using Lodash?
```
const uniqeTags = uniq(data.flatMap(({tags}) => tags))
```

packages/teleport/src/components/FilterByLabelList/FilterByLabelList.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  return Object.keys(tagDict)
+    .sort((a, b) => a.localeCompare(b, undefined, { numeric: true }))
+    .map(t => ({ value: t, label: t }));
+}
+
+export type Props = {
+  data: any[];
```
gzdunek, line 127: > component that allows you to filter by labels with any List component (given that data list contains a tag field)

so I'd reflect this in the type:
`data: { tags: string[] }[]`

packages/teleport/src/components/FilterByLabelList/FilterByLabelList.tsx
```diff
@@ -0,0 +1,132 @@
...
+    filterData(data, labelsFromQuery)
+  );
+
+  function onFilterApply(labels: Option[]) {
+    const filtered = filterData(data, labels);
+
+    setFilteredData(filtered);
+    setSelectedLabels(labels);
+    updateUrlQuery(labels, pathname);
+  }
```
gzdunek, line 55: I'm thinking about this and we can simplify things a bit:
- `selectedLabels` are duplicating the state from `labelsFromQuery` so it can be removed 
- `filteredData` can be `useMemo` calculating new result whenever `data` or `labelsFromQueryChange`
- `onFilterApply` can be much simpler and only call `updateUrlQuery()`
Thanks to these changes URL will be our single source of truth and it will be always in sync with the UI.

Let me know what do you think about this :)

packages/teleport/src/components/FilterByLabelList/FilterByLabelList.tsx
kimlisa: no, it wasn't intended to change per change in URL, this should be combined with `selectedFilter`

packages/teleport/src/components/FilterByLabelList/FilterByLabelList.tsx
kimlisa: revised

packages/teleport/src/components/FilterByLabelList/FilterByLabelList.tsx
kimlisa: i needed a new reference to let react know something changed

packages/teleport/src/components/FilterByLabelList/FilterByLabelList.tsx
kimlisa: i prefer not to use lodash if i don't have to

packages/teleport/src/components/FilterByLabelList/FilterByLabelList.tsx
kimlisa: i think it's a great idea :) i thought about this too, but i hesitated to change b/c we are basically reformatting labels to Option type, when it was originally a Option type (it's also harder to play around with it in storybook). Though it's probably a pre-optimization thinking and doesn't matter too much

packages/teleport/src/components/FilterByLabelList/FilterByLabelList.tsx
alex-kovoy: `lodash` and `momentjs` are two libraries that we need to get rid of eventually

packages/teleport/src/components/FilterByLabelList/FilterByLabelList.tsx
```diff
@@ -0,0 +1,127 @@
...
+      <TableComponent
+        {...tableProps}
+        data={filteredData}
+        onLabelClick={onLabelClick}
+      />
+    </>
+  );
+}
+
+function filterData(data = [], labels: Option[] = []) {
```
gzdunek, line 82: We can add a type to data (similarly for `makeLabelOptions`), I'd add sth like `type FilterableListData = { tags: string[] }` and use it in `Props`)

packages/teleport/src/components/FilterByLabelList/FilterByLabelList.tsx
gzdunek: I'm not sure if I understand, you mean changing `tags` from list of objects (`{ name: 'env', value: 'prod' }`) to list of strings (`env: prod`) in storybook? But both now and before we had list of strings in `FilterByLabelList`, right?

gzdunek (approved):

packages/teleport/src/components/FilterByLabelList/FilterByLabelList.tsx
gzdunek: I'm not sure if I understand, you mean changing tags from list of objects ({ name: 'env', value: 'prod' }) to list of strings (env: prod) in storybook? But both now and before we had list of strings in FilterByLabelList, right?

kimlisa: Prior, I was manually updating labels: `setSelectedLabels(newLabels)`. Now we rely on the url change, but I am not sure how to configure storybook so that this url change can be picked up. 

instead of (what we currently have):
```ts
const selectedLabels = useMemo<Option[]>(() => {
    const searchParams = new URLSearchParams(search);
    const query = searchParams.get('labels');
    if (!query || query === 'null') {
      return [];
    }

    return decodeURIComponent(query)
      .split(',')
      .map(label => ({ value: label, label }));
}, [search]);

function onFilterApply(labels: Option[]) {
    updateUrlQuery(labels, pathname);
}
```

equivalent, but also works well with storybook:
```ts
const [selectedLabels, setSelectedLabels] = useState<Option[]>(() => {
    const searchParams = new URLSearchParams(search);
    const query = searchParams.get('labels');
    if (!query || query === 'null') {
      return [];
    }

    return decodeURIComponent(query)
      .split(',')
      .map(label => ({ value: label, label }));
});

function onFilterApply(labels: Option[]) {
    setSelectedLabels(labels)
    updateUrlQuery(labels, pathname);
}
```

gzdunek: Yeah, it doesn't work in storybook because `teleport history service` doesn't operate on history of `MemoryRouter`. Maybe we can use `const history = useHistory()`? It picks up the current history object.

kimlisa:
closing in favor of https://github.com/gravitational/webapps/pull/465

PR #458, opened 2021-11-16
Add source maps generation

gzdunek:
Enable source maps to improve debugging experience


alex-kovoy:
we deliberately disabled source maps as it makes debugging easier because debugging over source-maps is quite limited. If you like it enabled, try do so with ENV variable and keep it `off` by default.

PR #461, opened 2021-11-18
Add filter capability to list with labels

kimlisa:

#### Description
It looks like a lot of changes, but most commit contains similar changes, each commit should be self contained (except for like tweaks), so might be easier to review by commits.

- Adds filter buttons to lists with labels, and move [searchValue, setSearchValue] pair closer to table:
  - Apps
  - Databases
  - Desktops
  - Kubes
  - Nodes
  - DocumentNodes**

- To make room for the filter button, we had to move the search bar into the top part of the table, which required to change all tables with search bars

#### DocumentNode console screenshot (had to move the quick launcher upwards)
image

image

#### Everywhere else screenshot
image

image

#### Dependent PR
relies on changes on https://github.com/gravitational/webapps/pull/456


kimlisa:
closing, b/c refactoring `filterable list` to a hook, and will cause massive change

PR #463, opened 2021-11-18
Improve error message when TOTP token is not valid

mcbattirola:
Closes gravitational/cloud#243

As mentioned in the issue, the error message when the TOTP token isn't valid is not actionable.
I've fixed it in the front end, should we change it in [Teleport's api](https://github.com/gravitational/teleport/blob/master/lib/auth/password.go#L301) instead?

### DIFF
--- a/packages/teleport/src/Invite/useInvite.ts
+++ b/packages/teleport/src/Invite/useInvite.ts
@@ -39,7 +39,17 @@ export default function useInvite(tokenId: string) {
     auth
       .resetPassword(tokenId, password, otpToken)
       .then(redirect)
-      .catch(submitAttempt.handleError);
+      .catch(err => {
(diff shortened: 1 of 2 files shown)


packages/teleport/src/Invite/useInvite.ts
```diff
@@ -39,7 +39,17 @@ export default function useInvite(tokenId: string) {
...
-      .catch(submitAttempt.handleError);
+      .catch(err => {
+        if (err.response?.status === 403) {
+          submitAttempt.handleError(
+            new Error(
+              'Invalid one time token. Please check if the token has expired and try again.'
+            )
+          );
+          return;
+        }
```
alex-kovoy, line 50: can we return this error from the server side? I assume it will also fix the same issue for the tsh as well.

PR #464, opened 2021-11-18
Fix Safari Favicon & Update Docker Node Version

### LINKED ISSUE #443: Teleport favicon doesn't show up on Safari
image

It looks like it is being added to the DOM, but I guess Safari doesn't like `rel="shortcut icon"`.

rudream:
## Purpose

This PR resolves #443 

Fixes the favicon not showing up on Safari

## Implementation

After experimenting with different `rel`’s to no avail I came across [this WebKit issue](https://bugs.webkit.org/show_bug.cgi?id=75877) where this WebKit developer said: "we quite purposefully don’t allow dynamic changes/animations to the favicon"

This means that Safari doesn’t allow us to set the favicon with JavaScript so our current approach using `useFavicon()` to append the favicon to `<head>` with JavaScript wasn’t going to work, even though it might show up in the DOM inspector. 

We use `HtmlWebpackPlugin` with webpack for generating our `index.html`, and luckily it can be configured with a `favicon:` field that we can use to set the favicon.

These changes will have to be made to `webapps.e` as well.

Updated the the docker image to `node:16.3-slim` because the old version (`v12`) doesn't support optional chaining and was breaking build.

### Chrome
chrome demo

### Firefox
firefox demo

### Safari
safari demo

## Notes

- If you are using dark mode, Safari detects that the contrast is too low (since the favicon is quite dark in colour) and adds a white background instead of the transparent background. It doesn't look like there's any way to prevent this.

safari demo darkmode


packages/teleport/package.json
```diff
@@ -4,8 +4,8 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "start": "g-start",
-    "build": "g-build --progress --bail",
+    "start": "g-start --config $(pwd)/webpack.dev.config.js",
```
zmb3, line 7: The `$(pwd)` shouldn't be necessary here.

zmb3 (approved):

alex-kovoy (approved):

packages/teleport/package.json
rudream: It's required, it's unable to find it otherwise and throws an error

alex-kovoy (approved):

alex-kovoy:
@rudream can we merge this one?

PR #465, opened 2021-11-19
Create url filter hook

kimlisa:

#### Description
Refactoring of PR https://github.com/gravitational/webapps/pull/456 where instead of creating a general component, create hooks:

`useUrlFiltering`:
- extract filters from data (unique labels)
- filters data looking for matching filter

`useUrlQueryParams`:
- extract filters from url
- toggle filter (for when user clicks on labels from table)

#### Url format example:
```
https://localhost:8080/web/cluster/clusterName/nodes?q=l=country:South%20Korea+l=os:windows

<path>?q=l=<encodedName>:<encodedValue>+l=<encodedName2>:<encodedValue2>
```
The values between the colon sign are the Label {name: string, value: string} pair.

#### Hook use example:
```ts

import SelectFilter from 'teleport/components/SelectFilter';
import useUrlFiltering from 'teleport/useUrlFiltering';

export function Nodes({nodes}: State) {
  const filter = useUrlFiltering(nodes);

  return (
        <>
          <SelectFilters
            filters={filter.filters}
            appliedFilters={filter.appliedFilters}
            applyFilters={filter.applyFilters}
          />
          <NodeList
            nodes={filter.result}
            onLabelClick={filter.toggleFilter}
          />
        </>
```


packages/teleport/src/services/history/history.ts
```diff
@@ -114,16 +112,37 @@ const history = {
...
-  const params = new URLSearchParams(path);
-  const value = params.get(name);
-  return value || '';
+// regexFilterQueryParam is a regex to split a path into two groups.
+// First group captures anything before the filter param as the path we will search
+// for the target query param. The second group captures the first filter param and
+// anything else following it we assume is part of the filter.
+const regexFilterQueryParam = /(?<pathToSearch>.+?)(?<pathWithFilter>\?filter=.*)/;
+
+export function getUrlParameter(name = '', path = '', preserveFilter = false) {
```
gzdunek, line 126: IMO we don't need to call this function without params so I'd remove default values for `name` and `path`

packages/teleport/src/useUrlFiltering.tsx
```diff
@@ -0,0 +1,176 @@
...
+  const selectedLabels = useMemo<Option[]>(() => getLabelsFromUrl(search), [
+    search,
+  ]);
+
+  const filteredData = useMemo(() => filterData(data, selectedLabels), [
+    data,
+    selectedLabels,
+  ]);
+
+  function apply(labels: Option[]) {
```
gzdunek, line 44: Maybe `applyLabels`?

packages/teleport/src/useUrlFiltering.tsx
```diff
@@ -0,0 +1,176 @@
...
+    return [];
+  }
+
+  return options;
+}
+
+// filterData returns new list of data that contains the selected labels.
+// Return type is relaxed here so it can be passed down to components with
+// stricter types.
+function filterData(data: Filterable[] = [], labels: Option[] = []): any[] {
```
gzdunek, line 134: `useUrlFiltering` has a generic type so we can use it here:
`filterData<T extends Filterable>(data: T[] = [], labels: Option[] = []): T[]`

packages/teleport/src/useUrlFiltering.tsx
kimlisa: i renamed `useUrlFiltering` to `useUrlLabelFiltering` so the hook name should imply its for labels

packages/teleport/src/useUrlFiltering.test.tsx
```diff
@@ -0,0 +1,263 @@
...
+  // Test new label is added.
+  expect(hook.labels).toEqual(expect.arrayContaining([label1, label2]));
+  expect(hook.labels).toHaveLength(2);
+
+  // Test url has been updated correctly.
+  expect(history.replace).toHaveBeenCalledWith(baseUrl);
+});
+
+describe('test decoding of urls', () => {
+  const enc = s => encodeURIComponent(s);
```
gzdunek, line 136: `const enc = encodeUriComponent`?

gzdunek (approved):

PR #473, opened 2021-11-26
Update build depedencies

gzdunek:
### Done:
- updated build dependecies
- migrated to `Webpack@5`
- switched to `React Refresh` from `React Hot Reload`
### Issues:
- rebuilds of `Storybook` are a bit slower (1,5-2x) after migration to `Webpack@5` - didn't find a solution for it yet
- a need to press Ctrl-C twice to stop the dev server (first press should stop it gracefully - but it doesn't seem to work for us) https://github.com/webpack/webpack-dev-server/pull/3880/files#diff-b706bbadb3d2315d99678f05e1f4bfd78a4bb4a7a6c65f3e81f1386fee3c883a

### Others:
cleanup of Webpack configs will be done in a separate PR


alex-kovoy:
> rebuilds of Storybook are a bit slower (1,5-2x) after migration to Webpack@5 - didn't find a solution for it yet

@gzdunek it seems that Storybook now comes with docGenPlugin enabled by default, so we can disable it in `storybook/main.js`

```js
module.exports = {
  typescript: { reactDocgen: false },
```

Another optimization is to disable `ES2015/6` babel transpiring for the storybook by removing `@babel/preset-env` present (it uses the default one from the packages/build/.babelrc.js)

This should bring dev UX back to normal.

alex-kovoy (approved):

alex-kovoy:
LGTM!

PR #476, opened 2021-11-29
webpack configs cleanup

gzdunek:
- created separated fields for each config part in `webpack.base.js` to indicate that this file doesn't contain a valid Webpack config
- added TS annotations to enable code completion

### DIFF
--- a/packages/build/webpack/webpack.base.js
+++ b/packages/build/webpack/webpack.base.js
@@ -23,7 +23,81 @@ const ReactRefreshPlugin = require('@pmmmwh/react-refresh-webpack-plugin');
 const appDirectory = fs.realpathSync(process.cwd());
 const resolveApp = relativePath => path.resolve(appDirectory, relativePath);
 
-module.exports = function createConfig() {
+const configFactory = {
+  createDefaultConfig,
+  plugins: {
+    reactRefresh(options) {
+      return new ReactRefreshPlugin(options);
+    },
+    eslint(options) {
+      return new ESLintPlugin({
+        context: '../',
+        extensions: ['ts', 'tsx', 'js', 'jsx'],
+        failOnError: false,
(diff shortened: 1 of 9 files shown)


packages/build/webpack/webpack.base.js
```diff
@@ -86,83 +96,80 @@ module.exports = function createConfig() {
...
-        options: {
-          limit: 102400, // 100kb
-          name: '/assets/fonts/[name].[ext]',
-        },
-      },
+function getNoParse() {
+  return function (content) {
+    return /xterm.js$/.test(content);
+  };
+}
```
alex-kovoy, line 106: this `getNoParse` no longer required (some legacy stuff) so it can be removed

packages/build/webpack/webpack.base.js
```diff
@@ -23,7 +23,17 @@ const ReactRefreshPlugin = require('@pmmmwh/react-refresh-webpack-plugin');
 const appDirectory = fs.realpathSync(process.cwd());
 const resolveApp = relativePath => path.resolve(appDirectory, relativePath);
 
-module.exports = function createConfig() {
+module.exports = function getWebpackConfigFactory() {
+  return {
```
alex-kovoy, line 27: hm...if we decided to call it factory, let have it just provide factory methods for the rest? Here is an example https://gist.github.com/alex-kovoy/9a09f2f4bed18101d4bae061894aaf2c  Notice how json property became a function.

```ts
rules: [
  configFactory.rules.fonts(),
  configFactory.rules.svg(),
  configFactory.rules.images(),
  configFactory.rules.jsx(),
  configFactory.rules.css()]
```

gzdunek: Done

alex-kovoy (approved):

PR #475, opened 2021-11-29
Add a new "accept" endpoint that directs to signup

michaelmcallister:
This change introduces a new endpoint (/web/accept/:tokenId) that simply
prompts the user to create an account by clicking a button, when they do
so they will be redirected to /web/invite/:tokenId (which is what is
used currently)

This change is aimed at preventing web crawlers/proxies and malware link
scanners in email clients that are hitting the /invite/ endpoint
directly shortly after the user clicks the link, which invalidates the
URL.

Please see https://github.com/gravitational/cloud/issues/1079 for
further investigation and justification.

It is proposed that this link will be used for any outbound email
communication, or other situations where it's plausible that clients may
interfere with the invite link. It will not replace the invite link
entirely.

PR #479, opened 2021-11-29
Add prompt prior to form for Invite and Reset

rudream:
## Purpose

Alternate implementation of https://github.com/gravitational/webapps/pull/475

Adds a new `/continue` route for the invite/reset form that the user gets to by first clicking the button on the initial prompt. This is to prevent crawlers (eg. in gmail) from triggering the OTP QR code generation before the user.

### Invite
image

### Reset
image


michaelmcallister:
LGTM

michaelmcallister (approved):

PR #480, opened 2021-11-30
simplify connection establishment

ibeckermayer:
Updates to be compatible with https://github.com/gravitational/teleport/pull/9098


packages/teleport/src/DesktopSession/TdpClientCanvas/useTdpClientCanvas.tsx
```diff
@@ -141,6 +141,16 @@ export default function useTdpClientCanvas(props: Props) {
   };
 }
 
+// Calculates the size (in pixels) of the display.
+// Since we want to maximize the display size for the user, this is simply
+// the full width of the screen and the full height sans top bar.
+function getDisplaySize(): { width: number; height: number } {
```
alex-kovoy, line 147: lets rely on type inference here and remove this return type definition (this is an internal method so it's OK)

packages/teleport/src/config.ts
```diff
@@ -102,7 +102,7 @@ const cfg = {
     desktopsPath: `/v1/webapi/sites/:clusterId/desktops`,
     desktopPath: `/v1/webapi/sites/:clusterId/desktops/:desktopName`,
     desktopWsAddr:
-      'wss://:fqdm/v1/webapi/sites/:clusterId/desktops/:desktopName/connect?access_token=:token',
+      'wss://:fqdm/v1/webapi/sites/:clusterId/desktops/:desktopName/connect?access_token=:token&&username=:username&width=:width&height=:height',
```
alex-kovoy, line 105: double `&&`?

packages/teleport/src/DesktopSession/TdpClientCanvas/useTdpClientCanvas.tsx
```diff
@@ -32,21 +32,22 @@ export default function useTdpClientCanvas(props: Props) {
...
-    // Calculate the size of the canvas to be displayed.
-    // Setting flex to "1" ensures the canvas will fill out the area available to it,
-    // which we calculate based on the window dimensions and TopBarHeight below.
-    const width = window.innerWidth;
-    const height = window.innerHeight - TopBarHeight;
+  const syncCanvasSizeToDisplaySize = (canvas: HTMLCanvasElement) => {
+    const { width, height } = getDisplaySize();
 
     // If it's resolution does not match change it
     if (canvas.width !== width || canvas.height !== height) {
```
alex-kovoy, line 53: Is it needed `if (canvas.width !== width || canvas.height !== height) {` ? Why not always assigning these to canvas?

packages/teleport/src/DesktopSession/TdpClientCanvas/useTdpClientCanvas.tsx
```diff
@@ -141,6 +141,16 @@ export default function useTdpClientCanvas(props: Props) {
   };
 }
 
+// Calculates the size (in pixels) of the display.
+// Since we want to maximize the display size for the user, this is simply
+// the full width of the screen and the full height sans top bar.
+function getDisplaySize(): { width: number; height: number } {
+  return {
+    width: window.innerWidth,
+    height: window.innerHeight - TopBarHeight,
```
alex-kovoy, line 150: It should be possible to get these dimensions without relying either on window or TopBarHeight. Canvas could be placed inside a div element where its dimensions will define the canvas size (you can use div.clientWidth and div.clientHeight). If you have time for this challenge, would be great to accept the challenge.

packages/teleport/src/DesktopSession/TdpClientCanvas/useTdpClientCanvas.tsx
ibeckermayer: I can't recall, I think it was copied from a tutorial many moons ago. Doesn't seem necessary.

packages/teleport/src/DesktopSession/TdpClientCanvas/useTdpClientCanvas.tsx
ibeckermayer: I agree, but we are behind on session recording and so I'm going to focus on that for now.

Created an issue: https://github.com/gravitational/webapps/issues/485.

alex-kovoy (approved):

PR #495, opened 2021-12-07
Add tabs shortcuts & open new tab basing on active tab

gzdunek:


packages/teleterm/src/services/pty/ptyProcess.ts
```diff
@@ -66,6 +68,22 @@ class PtyProcess extends EventEmitter {
     this._process.kill();
   }
 
+  async getWorkingDirectory(): Promise<string> {
+    const asyncExec = promisify(childProcess.exec);
+    // works only for unix
+    const { stdout, stderr } = await asyncExec(
+      `lsof -p ${this._process.pid} | awk '$4=="cwd" {print $9}'`
```
gzdunek, line 75: I didn't find a possibility to read `cwd` directly from `xterm`

packages/teleterm/src/ui/services/docs/docs.ts
```diff
@@ -136,6 +135,13 @@ export default class DocumentService extends Store<State> {
     });
   }
 
+  addAndOpen(doc: Document) {
```
gzdunek, line 138: I added a new method because invoking `add()` and `open()` separately caused an issue in async handler (`fit()` in xterm didn't take up all the container space). Invoking these methods works correctly in non-async flows because React batches state updates and new document is added and shown at the same time.

gzdunek:
Improved CWD reading

packages/teleterm/src/ui/TabHost/useDocumentCreator.ts
```diff
@@ -0,0 +1,59 @@
...
+    if (options.activeDocumentRef.current.type === 'document_terminal') {
+      try {
+        return await options.activeDocumentRef.current.getWorkingDirectory();
+      } catch (err) {
+        return; // default cwd will be used
+      }
+    }
+  }
+
+  async function getNewDocumentBasingOnActive(): Promise<Document> {
```
alex-kovoy, line 29: Lets do this. Instead of passing around the document to grab an instance of pty to execute a method that returns a directory, instead we can do this (sample below).

```tsx
export default function createPtyService(runtimeSettings: RuntimeSettings) {
  
  const service = {

    getProcessDirectory(pid){
      return PtyProcess.getDirectory(pid)
    },

    // the rest
```

```tsx
export default function useDocumentTerminal(doc: Props['doc']) {
  const ctx = useAppContext();

  const ptyProcess = React.useMemo(() => {
    const activeDoc = ctx.serviceDocs.getActive()        
    if (activeDoc.kind === 'terminal_tsh_session' || 'terminal_shell'){
      cwd = ctx.serviceTerminals.getDirectory(activeDoc.pid);
    }

    if (doc.kind === 'terminal_tsh_session') {        
      return ctx.serviceTerminals.createPtyProcess({
        ...doc,
        cwd,
        kind: 'tsh-login',
      });
    }

    if (doc.kind === 'terminal_shell') {      
      return ctx.serviceTerminals.createPtyProcess({ kind: 'new-shell', cwd, });
    }
  }, []);

  React.useEffect(() => {
    ctx.serviceDocs.update(doc.uri, { pid:  ptyProcess.getPid() })

    const cleanup = () => {
      ptyProcess.dispose();
    };

    return cleanup;
  }, [ptyProcess]);

  return {
    ptyProcess,
  };
}
```

alex-kovoy: Btw, we need to think how we can do the same for `tsh` session (getting to know the platform host is not trivial)

packages/teleterm/src/ui/services/docs/docs.ts
gzdunek: Removed

packages/teleterm/src/ui/TabHost/TabHost.tsx
```diff
@@ -51,14 +62,18 @@ export default function TabHost(props: Props) {
...
+    const isActiveDoc = doc === docActive;
 
-  const $docs = documents.map(doc => (
-    <MemoizedDocument doc={doc} visible={doc === docActive} key={doc.uri} />
-  ));
+    return (
+      <MemoizedDocument
+        doc={doc}
+        visible={isActiveDoc}
+        activeDocumentRef={isActiveDoc ? activeDocumentRef : null}
```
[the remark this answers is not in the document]
gzdunek: Removed and switched to the solution from the comment below

packages/teleterm/src/ui/TabHost/TabHost.tsx
```diff
@@ -33,8 +33,12 @@ export default function TabHost(props: Props) {
   const documents = serviceDocs.getDocuments();
   const docActive = serviceDocs.getActive();
 
+  const handleTabNew = useCallback(() => {
+    serviceDocs.openNewDocument();
+  }, [serviceDocs]);
+
   // enable keyboard shortcuts
-  useTabShortcuts(ctx);
+  useTabShortcuts(ctx, { openNewTab: handleTabNew });
```
alex-kovoy, line 41: do we need to pass `handleNewTab` method? The `serviceDocs.openNewDocument()` could be called from within the `useTabShortcuts` as it has reference to it (via `ctx`)

packages/teleterm/src/ui/TabHost/TabHost.tsx
```diff
@@ -33,8 +33,12 @@ export default function TabHost(props: Props) {
   const documents = serviceDocs.getDocuments();
   const docActive = serviceDocs.getActive();
 
+  const handleTabNew = useCallback(() => {
```
alex-kovoy, line 36: lets remove `useCallback` :)

packages/teleterm/src/ui/services/docs/docs.ts
```diff
@@ -184,23 +189,41 @@ export default class DocumentService extends Store<State> {
...
-
   changeIndex(oldIndex: number, newIndex: number) {
     const doc = this.state.docs[oldIndex];
     const newDocs = [...this.state.docs];
     newDocs.splice(oldIndex, 1);
     newDocs.splice(newIndex, 0, doc);
     this.setState({ ...this.state, docs: newDocs });
   }
+
+  openNewDocument() {
```
alex-kovoy, line 200: `openNewDocument` -> `openNewTerminal` (as this method name is less generic, dedicated specifically for terminals)

packages/teleterm/src/ui/services/docs/docs.ts
```diff
@@ -184,23 +189,41 @@ export default class DocumentService extends Store<State> {
...
+    this.previouslyActiveLocation = this.getActive().uri;
+    const newDocument = this.getNewDocumentBasingOnActive();
+    this.add(newDocument);
+    this.setLocation(newDocument.uri);
+  }
+
+  private getNewDocumentBasingOnActive(): Document {
```
alex-kovoy, line 207: I would combine this method with `openNewTerminal`

alex-kovoy (approved):
with minor comments

packages/teleterm/src/ui/TabHost/TabHost.tsx
gzdunek: We don't need after the changes, thanks for catching this!

PR #500, opened 2021-12-08
Add context menu to clusters

### LINKED ISSUE #471: [teleterm] Add a context menu to the cluster nav items
We need to add a context menu for "online" and "offline" clusters.

For `online` clusters

```pro
Refresh      (enabled)
---------------------------
Login        (disabled)
Logout      (enabled)
---------------------------
Remove       (enabled)
```

For `offline` clusters

```pro
Refresh      (disabled)
---------------------------
Login        (enabled)
Logout     (disabled)
---------------------------
Remove       (enabled)
```

Clicking on `Remove` should display a confirmation dialog asking: 

titile: Remove Cluster `clusterName` 
body: Are you sure you want to remove cluster?

Clicking on `Refresh` should trigger cluster refresh method.

gzdunek:
Closes #471

* Added context menu for clusters 
* Added new dialog to remove cluster
Screenshot 2021-12-08 at 16 46 35

### DIFF
--- a/packages/teleterm/src/mainProcess/clusterContextMenu.ts
+++ b/packages/teleterm/src/mainProcess/clusterContextMenu.ts
@@ -11,78 +11,62 @@ type MainClusterContextMenuOptions = Pick<
 >;
 
 export function subscribeToClusterContextMenuEvent(): void {
-  ipcMain.on(
-    ClusterContextMenuEventChannel.ToMain,
+  ipcMain.handle(
+    ClusterContextMenuEventChannel,
     (event, options: MainClusterContextMenuOptions) => {
-      Menu.buildFromTemplate([
-        {
-          label: 'Refresh',
-          enabled: options.isClusterConnected,
-          click: () =>
-            event.reply(
-              ClusterContextMenuEventChannel.ToRenderer,
-              ClusterContextMenuEventType.Refresh
-            ),
-        },
-        {
-          type: 'separator',
-        },
-        {
-          label: 'Login',
-          enabled: !options.isClusterConnected,
-          click: () =>
-            event.reply(
-              ClusterContextMenuEventChannel.ToRenderer,
-              ClusterContextMenuEventType.Login
-            ),
-        },
-        {
-          label: 'Logout',
-          enabled: options.isClusterConnected,
-          click: () =>
-            event.reply(
-              ClusterContextMenuEventChannel.ToRenderer,
-              ClusterContextMenuEventType.Logout
-            ),
-        },
-        {
-          type: 'separator',
-        },
-        {
-          label: 'Remove',
-          click: () =>
-            event.reply(
-              ClusterContextMenuEventChannel.ToRenderer,
-              ClusterContextMenuEventType.Remove
-            ),
-        },
-      ]).popup();
+      return new Promise(resolve => {
+        Menu.buildFromTemplate([
+          {
+            label: 'Refresh',
+            enabled: options.isClusterConnected,
+            click: () => resolve(ClusterContextMenuEventType.Refresh),
+          },
+          {
+            type: 'separator',
+          },
+          {
+            label: 'Login',
+            enabled: !options.isClusterConnected,
+            click: () => resolve(ClusterContextMenuEventType.Login),
+          },
+          {
+            label: 'Logout',
+            enabled: options.isClusterConnected,
+            click: () => resolve(ClusterContextMenuEventType.Logout),
+          },
+          {
+            type: 'separator',
+          },
+          {
+            label: 'Remove',
+            click: () => resolve(ClusterContextMenuEventType.Remove),
+          },
+        ]).popup({
+          callback: () => resolve(undefined),
+        });
+      });
     }
   );
 }
 
-export function openClusterContextMenu(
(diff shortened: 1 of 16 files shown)


packages/teleterm/src/mainProcess/clusterContextMenu.ts
```diff
@@ -0,0 +1,107 @@
+import { ipcMain, ipcRenderer, Menu } from 'electron';
+
+enum ClusterContextMenuEventChannel {
```
gzdunek, line 3: Are we ok with enums?

packages/teleterm/src/ui/ClusterRemove/ClusterRemove.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { useAppContext } from 'teleterm/ui/appContextProvider';
+import useAsync from 'teleterm/ui/useAsync';
+
+export default function ClusterRemove({
+  onClose,
+  clusterTitle,
+  clusterUri,
+}: Props) {
+  const ctx = useAppContext();
+  const [{ status, statusText }, run] = useAsync(() => {
```
alex-kovoy, line 34: Lets rename `run` -> `removeCluster`

packages/teleterm/src/ui/ClusterRemove/ClusterRemove.tsx
```diff
@@ -0,0 +1,93 @@
...
+    return ctx.serviceClusters.removeCluster(clusterUri);
+  });
+
+  useEffect(() => {
```
alex-kovoy, line 38: it's better to have a separate hook called `useClusterRemove` that takes care of the state which we can inject via 

```tsx
import useClusterRemove, { State } from './useClusterRemove';

  const ctx = useAppContext();
  const state = useClusterRemove(ctx);
  return <ClusterRemove {...state} />;
}

export function ClusterRemove(props: <State>) {
}
```

This will make it easier to add to the storybook and add snapshot tests in the future.

packages/teleterm/src/ui/ModalsHost/ModalsHost.tsx
```diff
@@ -38,6 +39,16 @@ export default function ModalsHost() {
     );
   }
 
+  if (dialog.kind === 'cluster-remove') {
+    return (
+      <ClusterRemove
+        clusterUri={dialog.clusterUri}
+        clusterTitle={dialog.clusterTitle}
```
alex-kovoy, line 46: seems that `serviceModals.closeDialog()` is repeated for every dialog, so lets add 
const handleClose = () => serviceModals.closeDialog() ?

packages/teleterm/src/mainProcess/clusterContextMenu.ts
alex-kovoy: I would put all constants used for channel messages into the `/src/mainProcess/types.ts`. Using enums are OK.

packages/teleterm/src/mainProcess/clusterContextMenu.ts
```diff
@@ -0,0 +1,107 @@
...
+}
+
+type MainClusterContextMenuOptions = Pick<
+  ClusterContextMenuOptions,
+  'isClusterConnected'
+>;
+
+export function subscribeToClusterContextMenuEvent(): void {
+  ipcMain.on(
+    ClusterContextMenuEventChannel.ToMain,
```
alex-kovoy, line 34: Do we need separate message types per Main vs Render?

alex-kovoy (approved):

packages/teleterm/src/mainProcess/clusterContextMenu.ts
gzdunek: Done

packages/teleterm/src/ui/ClusterRemove/ClusterRemove.tsx
gzdunek: Done

packages/teleterm/src/ui/ClusterRemove/ClusterRemove.tsx
gzdunek: Done

packages/teleterm/src/ui/ModalsHost/ModalsHost.tsx
gzdunek: Done

packages/teleterm/src/mainProcess/clusterContextMenu.ts
gzdunek: Yes, I was basing on the official documentation, where they use separate channels for bidirectional communication:  https://www.electronjs.org/docs/latest/api/ipc-main 
```typescript
ipcMain.on('asynchronous-message', (event, arg) => {
  console.log(arg) // prints "ping"
  event.reply('asynchronous-reply', 'pong')
})
```

But another idea came to my mind - communication with promises & [ipc.handle](https://www.electronjs.org/docs/latest/api/ipc-main#ipcmainhandlechannel-listener)
I think I like it, it's a bit shorter and simpler
https://gist.github.com/gzdunek/fe3333102278e0738355afd7df389d98

PR #502, opened 2021-12-09
Move search bar into Table

kimlisa:

#### Description
Moves search bar into the Table panel, to make room for the filter button.

#### Related PR
https://github.com/gravitational/webapps.e/pull/99

#### Notable changes
moved quick launcher next to add button

image

moved quick launcher next to cluster selector

image

example with fetch more

image

### DIFF
--- a/packages/teleport/src/Users/UserList/UserList.tsx
+++ b/packages/teleport/src/Users/UserList/UserList.tsx
@@ -17,7 +17,7 @@
 import React, { useState } from 'react';
 import styled from 'styled-components';
 import { sortBy } from 'lodash';
-import { Flex, Label } from 'design';
+import { Label } from 'design';
 import {
   Cell,
   Column,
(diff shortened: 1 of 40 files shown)


packages/teleport/src/Users/UserList/UserList.tsx
```diff
@@ -68,59 +67,54 @@ export default function UserList({
   }
 
   const data = sortAndFilter(searchValue);
-  const tableProps = { pageSize, data };
+  const tableProps = { pageSize, data, search: searchValue, onSearchChange };
```
gzdunek, line 70: `searchValue` can be renamed to `search`

gzdunek (approved):

alex-kovoy (approved):

packages/teleport/src/Users/UserList/UserList.tsx
kimlisa: leaving as is, there is going to be table update soon

PR #507, opened 2021-12-14
Make xterm understand resize sequence

gzdunek:
Handles `\033[8;30;80t` sequence.
Imported xterm styling from the `teleport` because it has `overflow: auto` set and I wanted to avoid copy&paste.

### DIFF
--- a/packages/teleterm/src/ui/DocumentTerminal/Terminal/Terminal.tsx
+++ b/packages/teleterm/src/ui/DocumentTerminal/Terminal/Terminal.tsx
@@ -64,22 +64,21 @@ export default class Terminal extends React.Component<{
 
 const StyledXterm = styled(Box)(
   props => `
-  height: 100%;
-  width: 100%;
-  font-size: 14px;
-  line-height: normal;
-  overflow: hidden;
-  background-color: ${colors.bgTerminal};
-  .terminal {
-    font-family: ${props.theme.fonts.mono};
-    border: none;
-    font-size: inherit;
-    line-height: normal;
-    position: relative;
-  }
-  .terminal .xterm-viewport {
-    background-color: ${colors.bgTerminal};
-  }
-
-}`
+   height: 100%;
+   width: 100%;
+   font-size: 14px;
+   line-height: normal;
+   overflow: auto;
+   background-color: ${colors.bgTerminal};
+   .terminal {
+     font-family: ${props.theme.fonts.mono};
+     border: none;
+     font-size: inherit;
+     line-height: normal;
+     position: relative;
+   }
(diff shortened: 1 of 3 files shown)


packages/teleterm/src/ui/DocumentTerminal/Terminal/xterm.ts
```diff
@@ -110,10 +132,8 @@ class TtyTerminal {
     this.term.write(displayText);
   }
 
-  _requestResize() {
-    this._fitAddon.fit();
-    this._ptyProcess.resize(this.term.cols, this.term.rows);
+  private requestResize() {
```
alex-kovoy, line 135: Thank for making them `private`. Btw, do you think it makes sense to keep the prefixes? Seems more readable with them.

alex-kovoy (commented):
Right now when `teleterm` window resizes, inactive tabs send `resize` events as well. The width/height for those tabs are miscalculated (0:0) due to invisible container. Can we do the same what `web` does where inactive tab sends the resize when it becomes visible?

packages/teleterm/src/ui/DocumentTerminal/Terminal/Terminal.tsx
```diff
@@ -15,11 +15,10 @@ limitations under the License.
 */
 
 import React from 'react';
-import { Box, Flex } from 'design';
-import styled from 'styled-components';
+import { Flex } from 'design';
 import XTermCtrl from './xterm';
-import { colors } from 'teleterm/ui/ThemeProvider/colors';
 import { PtyProcess } from 'teleterm/services/pty/types';
+import StyledXterm from 'teleport/Console/StyledXterm';
```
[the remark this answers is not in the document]
gzdunek: Done.
I added checking if terminal is visible before resizing, I also had to add triggering an event when tab becomes active. From what I can see in the web version, the browser automatically sends this event when the tab is changed.

packages/teleterm/src/ui/DocumentTerminal/Terminal/xterm.ts
gzdunek: Personally, I'm not a fan of these prefixes :) I would only use TS modifiers, these style guidelines also do not recommend underlines:
[TypeScript contributing guidelines](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines)
[AirBnB JS style guide](https://github.com/airbnb/javascript#naming--leading-underscore)
[Google TypeScript Style Guide](https://google.github.io/styleguide/tsguide.html#naming-style)

alex-kovoy (approved):

PR #508, opened 2021-12-14
Use select filter in tables that uses labels

kimlisa:

#### Description
- Created a `LabelFilterCell` component that takes in a `onLabelClick` func that sends back a filter type on click
- When making server response for (apps, nodes, etc), instead of `tags`, renamed to `labels` and instead of formatting, it returns as is and is formatted whenever it is needed with `makeLabelTag`


kimlisa:
closing b/c we need to paginate from server

PR #513, opened 2021-12-16
Remove old `PlayerNext`

ibeckermayer:
`PlayerNext` was no longer being used, so I:
- Deleted `PlayerNext`.
- Deleted two paths from the config that it was using (`sessionAuditPlayer` and `sessionAuditCmds`).
- Changed the `player` path to something more clear.
- Instructed the `RecordingList` to use the `player` path rather than deprecated `sessionAuditPlayer`.


alex-kovoy:
@ibeckermayer  `PlayerNext` is the next version of `ssh` player which got postponed (after priority changed). The work on it will be resumed once we get extra resources and the old one will be replaced. But it's OK to remove it from the repo for now (to keep it clean). Alternately, we can keep it and remove only the references to it from the config and main app.

alex-kovoy (approved):

ibeckermayer:
@alex-kovoy Ah ok, I thought it was an old one since it was jsx (rather than tsx). Anyways, I'll merge this and keep a note of it so we can get back to it easily.

PR #518, opened 2021-12-20
Select filter paginate

kimlisa:
#### Description
- Allows pagination for select filter.
  - Default is to list 100 filter per page.
  - Solves UI freezing from react select trying to list every filter
  - searching searches through the entire list of filters (not just what is currently listed)
- Copy and pasted `usePages` and `Pager` from table component

#### Screenshot

Pages

image
image
image

No pages

image

### DIFF
--- a/packages/shared/components/Select/types.ts
+++ b/packages/shared/components/Select/types.ts
@@ -39,7 +39,7 @@ export type Props = {
   menuPosition?: 'fixed' | 'absolute';
   inputValue?: string;
   filterOption?(): null | boolean;
-  onInputChange?(value: string, actionMeta: { action: InputAction }): void;
+  onInputChange?(value: string, actionMeta: ActionMeta): void;
 };
 
 export type AsyncProps = Omit<Props, 'options'> & {
@@ -58,4 +58,6 @@ export type Option<T = string> = {
   label: string;
 };
 
-type InputAction = 'set-value' | 'input-change' | 'input-blur' | 'menu-close';
+export type ActionMeta = {
+  action: 'set-value' | 'input-change' | 'input-blur' | 'menu-close';
+};
--- a/packages/shared/components/Select/types.ts
+++ b/packages/shared/components/Select/types.ts
@@ -37,6 +37,9 @@ export type Props = {
   components?: any;
   customProps?: Record<string, any>;
   menuPosition?: 'fixed' | 'absolute';
+  inputValue?: string;
+  filterOption?(): null | boolean;
+  onInputChange?(value: string, actionMeta: { action: InputAction }): void;
 };
 
 export type AsyncProps = Omit<Props, 'options'> & {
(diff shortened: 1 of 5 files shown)


packages/shared/components/Select/types.ts
```diff
@@ -54,3 +57,5 @@ export type Option<T = string> = {
   // label is the value user sees in the select options dropdown.
   label: string;
 };
+
+type InputAction = 'set-value' | 'input-change' | 'input-blur' | 'menu-close';
```
alex-kovoy, line 61: This is clearer:

```ts
type ActionMeta = {
 action:  'set-value' | 'input-change' | 'input-blur' | 'menu-close'
}
```

packages/teleport/src/components/SelectFilters/usePages.ts
```diff
@@ -0,0 +1,67 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+
+export default function usePages({ pageSize = 100, data = [] }) {
```
alex-kovoy, line 19: if this is indeed `1 to 1` with existing one, perhaps it's OK to reference it directly.

alex-kovoy (approved):
with comments

packages/teleport/src/components/SelectFilters/Pager.tsx
```diff
@@ -0,0 +1,97 @@
...
+  onNext,
+}: Props) {
+  const isPrevDisabled = totalRows === 0 || startFrom === 0;
+  const isNextDisabled = totalRows === 0 || endAt === totalRows;
+  const initialStartFrom = totalRows > 0 ? startFrom + 1 : 0;
+
+  return (
+    <Flex m={2} justifyContent="flex-end">
+      <Flex alignItems="center" ml={2}>
+        <Text typography="body2" style={{ color: '#4b4b4b' }}>
```
gzdunek, line 36: I found this color in a few places, maybe it's worth to extract it to theme?

packages/shared/components/Select/types.ts
```diff
@@ -37,6 +37,9 @@ export type Props = {
   components?: any;
   customProps?: Record<string, any>;
   menuPosition?: 'fixed' | 'absolute';
+  inputValue?: string;
+  filterOption?(): null | boolean;
+  onInputChange?(value: string, actionMeta: { action: InputAction }): void;
```
gzdunek, line 42: Maybe extract type of this function to the new type (so we will be able to use it in handler in `SelectFilters.tsx` line 84)?

gzdunek (approved):

packages/teleport/src/components/SelectFilters/Pager.tsx
kimlisa: it's kind of specific color to the SelectFilter component, so what I did was remove it from here, and set its parent container color, so at least its contained in one file

PR #519, opened 2021-12-21
Teleterm bugfixes

gzdunek:


packages/teleterm/src/ui/Tabs/TabItem.tsx
```diff
@@ -43,7 +43,9 @@ export default function TabItem(props: Props) {
       style={{ ...style, opacity }}
     >
       <StyledTabButton>
-        <Text mx="auto">{name}</Text>
+        <Text css={{ direction: 'rtl' }} mx="auto">
```
gzdunek, line 46: Experimental idea to show `...files/holidays` instead of `/Users/me/fil...` when there is no space, but it will work this way for all kinds of tabs, so I'm not sure if we want it

alex-kovoy:
@gzdunek i've pushed my recent changes (sorry for that), could you resolve the conflicts and update the branch?

gzdunek:
Conflicts resolved :)

gzdunek:
I will remove `getTitle()` and revert `title`

packages/teleterm/src/ui/TabHost/TabHost.tsx
```diff
@@ -55,17 +55,25 @@ export default function TabHost(props: Props) {
...
+
   const $docs = documents.map(doc => {
     const isActiveDoc = doc === docActive;
-    return <MemoizedDocument doc={doc} visible={isActiveDoc} key={doc.uri} />;
+    return (
+      <MemoizedDocument
+        doc={doc}
+        visible={isActiveDoc}
+        key={doc.uri}
+        onTerminalContextMenu={openTerminalContextMenu}
```
alex-kovoy, line 69: Do we need pass the `onTerminalContextMenu` from here down to Terminal? Can we reference the mainProcessCelint from the  `useDocumentTerminal` instead and clean up TabHost?

alex-kovoy: No need to pass down the handleTabClose as the documents have access to `serviceDocs`

alex-kovoy (commented):
lets try to clean it up a bit

packages/teleterm/src/services/pty/ptyProcess.ts
```diff
@@ -20,99 +20,98 @@ import { PtyOptions } from './types';
...
+      this.attachedBuffer = data;
+      setTimeout(this.flushBuffer.bind(this), 10);
     }
   }
 
-  _onExit(e: { exitCode: number; signal?: number }) {
+  private onExit(e: { exitCode: number; signal?: number }) {
     this.emit(TermEventEnum.EXIT, e);
-    this._logger.info('pty has been terminated');
+    this.process = undefined;
```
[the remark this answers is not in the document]
gzdunek: It was "just-in-case" to prevent `resize` from working on the terminated process, but closing the tab on process exit should be sufficient and resolve it

packages/teleterm/src/ui/DocumentTerminal/useDocTerminal.ts
```diff
@@ -47,24 +48,40 @@ export default function useDocumentTerminal(doc: Props['doc']) {
...
 
   function updateDocumentPid(pty: PtyProcess): void {
-    ctx.serviceDocs.update(doc.uri, { pid: pty.getPid() });
+    const pid = pty.getPid();
+    if (pid) {
+      ctx.serviceDocs.update(doc.uri, { pid });
+    }
   }
 
+  const refreshDocumentCwd = useMemo(() => {
```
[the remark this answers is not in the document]
gzdunek: It would be better, but `useMemo` doesn't have the cleanup function :( 
We can move the whole `useEffect` from `DocumentTerminal` to `useDocTerminal` but I'm not sure if `useDocTerminal` should be aware of the terminal ref?

packages/teleterm/src/ui/TabHost/TabHost.tsx
gzdunek: Correct, removed handlers from `TabHost`

packages/teleterm/src/ui/DocumentTerminal/useDocTerminal.ts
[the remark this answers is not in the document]
alex-kovoy: I was referring to just this part where canceling the debouncing can be part of this hook too. 
```ts
useEffect(() => {
    return () => refreshDocumentCwd.cancel();
  }, []);
```

packages/teleterm/src/ui/DocumentTerminal/useDocTerminal.ts
```diff
@@ -47,24 +48,50 @@ export default function useDocumentTerminal(doc: Props['doc']) {
...
-        setPtyProcess(pty);
-        updateDocumentPid(pty);
+        if (!disposed) {
+          setPtyProcess(pty);
+          updateDocumentPid(pty);
+        }
       });
     }
 
+    ptyProcess?.onExit(() => {
```
alex-kovoy, line 80: lets check for the exit code `0` before closing the tab. 
if a process exits with an error code. it's better to keep the tab open to notify a user about it.

packages/teleterm/src/ui/DocumentTerminal/useDocTerminal.ts
```diff
@@ -47,24 +48,50 @@ export default function useDocumentTerminal(doc: Props['doc']) {
   }
 
   function updateDocumentPid(pty: PtyProcess): void {
-    ctx.serviceDocs.update(doc.uri, { pid: pty.getPid() });
+    const pid = pty.getPid();
+    if (pid) {
```
alex-kovoy, line 52: can `pid` not have a value?

packages/teleterm/src/ui/DocumentTerminal/useDocTerminal.ts
```diff
@@ -47,24 +48,50 @@ export default function useDocumentTerminal(doc: Props['doc']) {
...
     return () => {
+      disposed = true;
       ptyProcess?.dispose();
     };
   }, [ptyProcess]);
 
   return {
     ptyProcess,
+    refreshDocumentCwd,
+    openTerminalContextMenu,
```
alex-kovoy, line 94: I wonder if we can make `ptyProcess` _not null_ during initial render? This is because this `null` introduces  multiple edge cases to handle. If I recall, it started from `getPreviousDocumentCwdIfPossible` so  maybe we can revisit this part and simplify?

For example, can a `doc` have `cwd` populated prior `useDocumentTerminal(doc)` is called? If we look at `openNewTerminal` it seems that it clones the `doc` object which might be what we need.

alex-kovoy (approved):
With minor comments to address

packages/teleterm/src/ui/DocumentTerminal/useDocTerminal.ts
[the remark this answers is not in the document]
gzdunek: I added `useDocumentCwdUpdater` hook to have all things related to this in a single place

packages/teleterm/src/ui/DocumentTerminal/useDocTerminal.ts
gzdunek: Done

packages/teleterm/src/ui/DocumentTerminal/useDocTerminal.ts
gzdunek: Removed this part completely

packages/teleterm/src/ui/DocumentTerminal/useDocTerminal.ts
gzdunek: Done as discussed

packages/teleterm/src/services/pty/ptyProcess.ts
```diff
@@ -20,99 +20,104 @@ import { PtyOptions } from './types';
...
+        ...this.options.env,
       },
     });
 
-    this._process.onData(data => this._onData(data));
-    this._process.onExit(ev => this._onExit(ev));
+    this.onStart();
+    this.process.onData(data => this.onData(data));
+    this.process.onExit(ev => this.onExit(ev));
+    this.isReady = true;
```
alex-kovoy, line 50: `isReady` -> `disposed`

alex-kovoy (approved):

PR #522, opened 2021-12-22
Address eslint warnings

### LINKED ISSUE #497: Address eslint warnings in unit-tests and add a lint step into our build make target
We do not lint our unit-tests which we need to fix by adding an extra lint step into our build (similarly how we do type checking `tsc`).  Adding this extra step will result in lots of warnings/errors so it's OK to start addressing it by creating smaller PRs. We can enable the lint step after it's all addressed. 

image

gzdunek:
Closes #497 
- Addressed ESLint warnings
- Enabled skipped tests
- Removed `eslint-webpack-plugin` so linting won't be performed on Webpack rebuilds. This plugin operated only files which came into a bundle, so test files weren't checked. Lining will be done as a step on CI (during `yarn build`), locally please use [ESLint plugin ](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) & `yarn lint`

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,8 @@
     "build-teleport-e": "yarn workspace @gravitational/teleport.e build --output-path=../../../dist/e/teleport/app",
     "build-oss": "yarn build-teleport-oss",
     "build-e": "yarn build-teleport-e",
-    "build": "yarn type-check && yarn build-oss && yarn build-e",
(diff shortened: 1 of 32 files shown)


alex-kovoy (approved):
Looks good! Lets just add updated `e` ref to this branch as well.

Issue #523, opened 2021-12-23
Implement Table v2


rudream:
Master ticket for implementing new Table

Todo:

- [x] Add TableV2 component https://github.com/gravitational/webapps/pull/524
- [x] Use TableV2 for NodeList https://github.com/gravitational/webapps/pull/525
- [x] Use TableV2 for AppList https://github.com/gravitational/webapps/pull/535
- [x] Use TableV2 for DatabaseList https://github.com/gravitational/webapps/pull/534
- [x] Use TableV2 for KubeList https://github.com/gravitational/webapps/pull/536
- [x] Use TableV2 for DesktopList https://github.com/gravitational/webapps/pull/537
- [x] Use TableV2 for SessionList https://github.com/gravitational/webapps/pull/545
- [x] Use TableV2 for RecordingsList https://github.com/gravitational/webapps/pull/546
- [x] Use TableV2 for EventList https://github.com/gravitational/webapps/pull/541
- [x] Use TableV2 for RoleList https://github.com/gravitational/webapps/pull/542
- [x] Use TableV2 for UserList https://github.com/gravitational/webapps/pull/543
- [x] Use TableV2 for RequestList https://github.com/gravitational/webapps.e/pull/114
- [x] Use TableV2 for ClusterList https://github.com/gravitational/webapps/pull/548
- [x] Use TableV2 for MfaDeviceList https://github.com/gravitational/webapps/pull/549
- [x] Use TableV2 for UsageSummary https://github.com/gravitational/webapps.e/pull/115
- [x] Delete TableV1 Component https://github.com/gravitational/webapps/pull/573

PR #525, opened 2021-12-23
Update NodeList to use Table V2

rudream:
## Purpose

Part of #523 

Switches NodeList to using Table V2

## Demo

image

### DIFF
--- a/packages/teleport/src/components/NodeList/NodeList.tsx
+++ b/packages/teleport/src/components/NodeList/NodeList.tsx
@@ -24,7 +24,7 @@ function NodeList(props: Props) {
   const { nodes = [], onLoginMenuOpen, onLoginSelect, pageSize = 100 } = props;
 
   return (
-    <StyledTable<Node>
+    <StyledTable
       columns={[
         {
           key: 'hostname',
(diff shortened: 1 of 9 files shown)


packages/teleport/src/components/NodeList/NodeList.tsx
```diff
@@ -16,126 +16,64 @@ limitations under the License.
...
-          cell={<TextCell />}
-        />
-        <Column
-          columnKey="addr"
-          header={
-            <SortHeaderCell
-              sortDir={sortDir.addr}
-              onSortChange={onSortChange}
-              title="Address"
+    <StyledTable<Node>
```
alex-kovoy, line 27: Can we try to infer this type from the `data` prop so we do not need to explicitly specify it?

packages/teleport/src/components/NodeList/NodeList.tsx
```diff
@@ -16,126 +16,64 @@ limitations under the License.
...
+      columns={[
+        {
+          key: 'hostname',
+          headerText: 'Hostname',
+          isSortable: true,
+        },
+        {
+          key: 'addr',
```
alex-kovoy, line 35: wonder if `key` is typed as well? will it display an error if missing key is specified?

rudream: `key` is typed and has autocomplete based on the different properties of the data type (in this case `Node`). It won't return an error if a missing key is specified so you can add your own key. This is so we can still have unique keys even for columns that aren't necessarily a property in the data object, eg. `connect-btn` in this case.

```
key: keyof T & string;
```

alex-kovoy: The `key` needs to be typed so TS compiler can catch errors otherwise there is little value in it.

kimlisa (commented):
this PR should be based on your other table component branch

alex-kovoy (approved):

alex-kovoy:
@rudream rebase it off master and good to go

PR #526, opened 2021-12-28
Add basic config service

gzdunek:


gzdunek:
Removed subscribing/unsubscribing from updates

packages/teleterm/src/services/config/providers/keyboardShortcutsConfigProvider.ts
```diff
@@ -0,0 +1,71 @@
...
+  | 'tab-new'
+  | 'tab-previous'
+  | 'tab-next'
+  | 'focus-global-search';
+
+export type KeyboardShortcutsConfig = Record<KeyboardShortcutType, string>;
+
+export const keyboardShortcutsConfigProvider: ConfigServiceProvider<KeyboardShortcutsConfig> =
+  {
+    provideDefaultsForPlatform(platform) {
```
alex-kovoy, line 29: `getPlatformDefaults()` or event better `getDefaults(platform?: string)`

packages/teleterm/src/services/config/types.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export interface Config {
+  keyboardShortcuts: KeyboardShortcutsConfig;
+}
+
+export interface ConfigServiceProvider<
+  T extends Record<string, number | string | boolean>
+  > {
+  provideDefaults?(): Partial<T>;
+  provideDefaultsForPlatform?(platform: Platform): Partial<T>;
```
alex-kovoy, line 12: combine them in 1 method?

alex-kovoy (approved):
with minor comments to consider

packages/teleterm/src/services/config/providers/keyboardShortcutsConfigProvider.ts
gzdunek: Changed to `getDefaults(platform: string)` - we don't have to use `platform` parameter in the implementation, but it is always available.

packages/teleterm/src/services/config/types.ts
gzdunek: I did it this way because at the same time we can specify some of the defaults for all platforms and some of them only for a single platform. But I can agree, they can be combined into a 1 method, it will be more type-safe as we get rid of `Partials`

PR #527, opened 2021-12-29
Adds a TDP Error message

ibeckermayer:
Adds handling for the new TDP Error message. When a TDP error message is received, the websocket is closed and the error message is displayed.

Corresponds to https://github.com/gravitational/teleport/pull/9586


packages/teleport/src/lib/tdp/client.ts
```diff
@@ -30,6 +32,7 @@ export default class Client extends EventEmitter {
   logger = Logger.create('TDPClient');
   private connected = false;
   private disconnected = false;
+  private errored = false;
```
alex-kovoy, line 35: Lets do another round by revisiting this flags as their usage seems unnecessary complicated? For example, do we need this `errored` flag at all? Can we remove `connected` flag?  

```ts
private connected = false;
private disconnected = false;
private errored = false
```  

You may also consider adding a new message type to EventEmiiter 'renderError` - this can help you to distinguish protocol errors vs external errors.

Yeah, lets try to remove all above flags to see if we can do without them.

alex-kovoy (changes requested):

packages/teleport/src/lib/tdp/client.ts
ibeckermayer: The problem that `disconnected` and `errored` are trying to solve is that when the websocket it closed, the `onclose` handler needs some way to know whether or not to emit the generic `'Connection closed due to an unknown error'` error event. This state must be tracked somehow in order for this object to work how we need it to. 

It would be possible to get rid of the `errored` flag by tracking it with the `CloseEvent.code`, however this would just push the responsibility for setting this aspect of this `Client`'s state to the backend, where we would need to set the websocket's close code depending on whether or not we sent a TDP error. This is a worse design, because now we've spread this event emitter's state to the server, plus it's practically infeasible since we would need to completely refactor the proxy from blindly proxying messages to the websocket to actually decoding TDP and checking for TDP errors, and deciding a websocket code based on them.

Managing this state with websocket close codes raises a similar problem for `disconnect`, though in its case we would need to add a "Disconnect" tdp message, which the proxy would have to be refactored to read and then pass a particular close code over the websocket. 

I agree it would generally be preferable that we could easily just pass these states into `onclose`, however because of the nature of how that listener get's called its just impractical and would lead to a lot of additional complexity in the protocol and on the backend for very little gain. Of course let me know if there's some approach that I'm missing.

It may be possible to get rid of the "connect" event entirely, I'll take a closer look.

packages/teleport/src/lib/tdp/codec.test.ts
```diff
@@ -157,8 +137,29 @@ test('decodes regions', () => {
...
-  expect(region.left).toEqual(0);
-  expect(region.bottom).toEqual(64);
-  expect(region.right).toEqual(64);
+  expect(region.top).toBe(0);
+  expect(region.left).toBe(0);
+  expect(region.bottom).toBe(64);
+  expect(region.right).toBe(64);
+});
+
+// This test is skipped because codec.decodeError uses TextDecoder which jest doesn't currently support.
```
alex-kovoy, line 146: Lets fix the comment too.

packages/teleport/src/lib/tdp/client.ts
alex-kovoy: > the onclose handler needs some way to know whether or not to emit the generic 'Connection closed due to an unknown error' error event. This state must be tracked somehow in order for this object to work how we need it to.

These are 2 types of errors and could be handled separately. 
1) Show TDP error messages when they arrive
2) Show "connection has been terminated" error.

No need to combine them.

Also, lets rename `'Connection closed due to an unknown error'` -> `Connection has been terminated by the server`  as there is no way for us to report an actual error via websocket close event.

You should be able to determine when connection has been closed by the client vs the server.

packages/teleport/src/DesktopSession/useDesktopSession.tsx
```diff
@@ -19,23 +19,41 @@ import { useParams } from 'react-router';
...
-  // Tracks combination of tdpclient/websocket and api call state,
-  // as well as whether the tdp client for this session was intentionally disconnected.
   const { attempt: fetchAttempt, run } = useAttempt('processing');
+
+  // tdpConnection tracks the state of the tdpClient's TDP connection
+  // tdpConnection.status ===
+  // - 'processing' at first
+  // - 'success' once the first TdpClientEvent.IMAGE_FRAGMENT is seen
+  // - 'failed' if a TdpClientEvent.TDP_ERROR is encountered
+  const { attempt: tdpConnection, setAttempt: setTdpConnection } =
```
alex-kovoy, line 32: Do we really need this abstraction `tdpConnection` in addition to `wsConnection` ? Why the first `IMAGE_FRAGMENT`  should have any meaning in the context of connection state handling?

ibeckermayer: I think so. The `wsConnection` is instantaneous (barring an incorrect websocket address), but after the websocket connection is made, there is a potentially substantial period of time where the proxy is sending the username and screen size to the windows_service, and then the windows_service is attempting to establish an RDP connection to the windows server, during which we want to display a spinner rather than an empty canvas.

alex-kovoy: > but after the websocket connection is made, there is a potentially substantial period of time where the proxy is sending the username and screen size to the windows_service

I thought that you have fixed it where tdp handshake is done [at websocket connection creation time](https://github.com/gravitational/teleport/blob/master/lib/web/desktop.go#L135). This is why i am asking about `tdpConnection`.

ibeckermayer: The initial handshake is now done in the backend (as you linked to), but it still takes some amount of time for those to be sent from the proxy to the windows_service, and for the windows_service to connect to the windows machine via RDP and receive the first frame.

The proxy is [sending](https://github.com/gravitational/teleport/blob/master/lib/web/desktop.go#L125-L133) those initial TDP messages over the tls connection to the windows_service, and then [blindly proxying](https://github.com/gravitational/teleport/blob/master/lib/web/desktop.go#L135-L168) whatever TDP messages the windows_service sends back. 

The only way we know on the frontend that the RDP connection from the windows_service to the windows host has succeeded is that we get the first `IMAGE_FRAGMENT` back. The websocket connection itself is wrapping TDP, but a successful websocket might yet get back a `TDP_ERROR`.

alex-kovoy (approved):
We discussed it on phone, good to go now.

packages/teleport/src/lib/tdp/codec.ts
```diff
@@ -305,16 +306,28 @@ export default class Codec {
...
+      return messageType;
     } else {
       // We don't expect to need to decode any other value on the client side
       throw new Error(`invalid message type: ${messageType}`);
     }
   }
 
+  // decodeError decodes a raw tdp ERROR message and returns it as a javascript Error object
+  // | message type (9) | message_length uint32 | message []byte
+  decodeErrorMessage(buffer: ArrayBuffer): Error {
```
alex-kovoy, line 323: Better to return a string, not an actual JS Error object. `decodeErrorMessage(buffer: ArrayBuffer): string`

PR #533, opened 2022-01-07
Add context menu to tab items

gzdunek:


packages/teleterm/src/ui/TabHost/TabHost.tsx
```diff
@@ -55,6 +55,30 @@ export default function TabHost(props: Props) {
...
+        const documentIndex = documents.indexOf(document);
+        const toClose = documents.filter((_, index) => index > documentIndex);
+        serviceDocs.closeMultiple(toClose);
+      },
+      onDuplicatePty: () => {
+        const documentIndex = documents.indexOf(document);
+        const newDocument = serviceDocs.createPtyDocumentCopy(document);
+        serviceDocs.add(newDocument, documentIndex + 1);
+        serviceDocs.setLocation(newDocument.uri);
+      },
```
alex-kovoy, line 78: I think it would be better if we move these methods to the doc service (closeToTheRight, closeOthers, etc). Btw, could you also add some unit-tests to the doc service as well? (for these methods)

gzdunek: I was wondering if they are not too `tabs` specific to be placed in the doc service, but I'm fine with moving them

alex-kovoy (approved):

PR #536, opened 2022-01-10
Update KubeList to use TableV2

rudream:
## Purpose

Part of #523 

Updates KubeList to use Table V2

## Demo

image


Crypto231 (commented):
Good

PR #537, opened 2022-01-10
Update DesktopList to use TableV2

rudream:
## Purpose

Part of #523 

Updates DesktopList to use Table V2

Adds a unit test for the table.

## Demo

image


alex-kovoy (approved):

ibeckermayer (approved):
lgtm

PR #539, opened 2022-01-13
Improve the PNG rendering process

zmb3:
Prior to this, for each PNG (there are lots!) we would:

- create a new `` element in the document
- convert the PNG data from raw bytes to a base64-encoded string,
  which allocates a second (even larger) copy of the data we
  already have
- tell the `` element to render the PNG data and then draw the
  `` element on the canvas

Instead of wasting cycles converting the data to a string just so
that the browser can decode it a moment later, we generate a bitmap
with a raw bytes and draw it directly to the canvas, also bypassing
the need for an `` element for each PNG.

### DIFF
--- a/packages/teleport/src/lib/tdp/client.ts
+++ b/packages/teleport/src/lib/tdp/client.ts
@@ -97,10 +97,9 @@ export default class Client extends EventEmitter {
   // bounds and png bitmap and emit a render event.
   processFrame(buffer: ArrayBuffer) {
     const { left, top } = this.codec.decodeRegion(buffer);
-    const image = new Image();
-    image.onload = () =>
-      this.emit(TdpClientEvent.IMAGE_FRAGMENT, { image, left, top });
-    image.src = this.codec.decodePng(buffer);
+    createImageBitmap(new Blob([buffer.slice(17)], { type: 'image/png' })).then(image => {
+      this.emit(TdpClientEvent.IMAGE_FRAGMENT, { image, left, top })
+    })
   }
 
   sendUsername(username: string) {
@@ -147,7 +146,7 @@ export default class Client extends EventEmitter {
 }
 
 export type ImageFragment = {
-  image: HTMLImageElement;
+  image: ImageBitmap;
   left: number;
   top: number;
 };
--- a/packages/teleport/src/components/TdpClientCanvas/TdpClientCanvas.tsx
+++ b/packages/teleport/src/components/TdpClientCanvas/TdpClientCanvas.tsx
@@ -55,6 +55,7 @@ export default function TdpClientCanvas(props: Props) {
         if (buffer.length) {
           for (let i = 0; i < buffer.length; i++) {
             tdpCliOnImageFragment(ctx, buffer[i]);
+            buffer[i].image.close()
           }
           buffer = [];
         }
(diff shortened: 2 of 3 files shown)


zmb3:
I suspect this should be more efficient, but would love some pointers or assistance in using the dev tools to profile and verify that it makes a difference.

packages/teleport/src/lib/tdp/client.ts
```diff
@@ -97,10 +97,9 @@ export default class Client extends EventEmitter {
   // bounds and png bitmap and emit a render event.
   processFrame(buffer: ArrayBuffer) {
     const { left, top } = this.codec.decodeRegion(buffer);
-    const image = new Image();
-    image.onload = () =>
-      this.emit(TdpClientEvent.IMAGE_FRAGMENT, { image, left, top });
-    image.src = this.codec.decodePng(buffer);
+    createImageBitmap(new Blob([buffer.slice(17)], { type: 'image/png' })).then(image => {
```
zmb3, line 100: This is kind breaks the codec abstraction. I should move the 17 into codec.ts.

ibeckermayer:
I profiled this branch using the `Performance` storybook (`yarn storybook` and search for "Performance", once the desktop 
is drawn to canvas open the console and you will see a log with the time the render took) and compared it to master, and it was [slower by an order of magnitude](https://docs.google.com/spreadsheets/d/1vnNiyVi7P8ht8vxMz7hFZW3RjE1AYxOqc7xysWF23QU/edit?usp=sharing). Additionally, if I run this and connect to a desktop, the rendering is visibly choppier and slower.

IIRC I tested something like this previously; it's an interesting question as to why it's so much slower. I think it's related to the fact that `createImageBitmap` is asynchronous and [how asynchronous calls are handled by the JS event loop](https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/).

I spent a lot of time tinkering with this a few months ago, some of which is documented in Slack starting around [here](https://gravitational.slack.com/archives/C02DQ1C2BMW/p1632359774030300). It is an issue that's worth re-examining. For example, looking into Myrtille, I see that there's [a `BINARY` mode](https://github.com/cedrozor/myrtille/blob/b99671589ad0c18f7ccb0878d08e05bcb9cb15c6/Myrtille.Web/js/network.js#L117-L141) that's preferred to the `BASE64` mode. However when they actually render in `BINARY` mode, there's still an option that [has them converting to base64](https://github.com/cedrozor/myrtille/blob/b99671589ad0c18f7ccb0878d08e05bcb9cb15c6/Myrtille.Web/js/display/canvas.js#L145-L167).

I'm not sure if I ever chased this down totally before, and tried the `url = URL.createObjectURL(new Blob([data], { type: 'image/' + format }));`. The comment says that it may be slower for an idiosyncratic reason, but its also a 4 year old comment. I'll give it a go in another branch.

zmb3:
womp womp 😢

This is why we measure though!

ibeckermayer:
`URL.createObjectURL(new Blob([data], { type: 'image/' + format }))` is also OOM worse 🤷. Probably due to [this](https://github.com/cedrozor/myrtille/blob/b99671589ad0c18f7ccb0878d08e05bcb9cb15c6/Myrtille.Web/js/display/canvas.js#L147-L153).

PR #541, opened 2022-01-14
Update EventList to TableV2

rudream:
## Purpose

Part of #523 

Update EventList to Table V2

### DIFF
--- a/packages/design/src/DataTableNext/Pager/usePager.ts
+++ b/packages/design/src/DataTableNext/Pager/usePager.ts
@@ -1,4 +1,4 @@
-import { FetchingConfig } from '../types';
+import { FetchStatus } from '../types';
 
 export default function usePager({
   nextPage,
@@ -38,7 +38,7 @@ export type Props = {
   currentPage: number;
   pageSize: number;
   onFetchMore?: () => void;
-  fetchStatus?: FetchingConfig['fetchStatus'];
+  fetchStatus?: FetchStatus;
 };
 
 export type State = ReturnType<typeof usePager>;
--- a/packages/design/src/DataTableNext/Pager/usePager.ts
+++ b/packages/design/src/DataTableNext/Pager/usePager.ts
@@ -1,4 +1,4 @@
-import { PaginationConfig } from '../types';
+import { FetchingConfig } from '../types';
 
 export default function usePager({
   nextPage,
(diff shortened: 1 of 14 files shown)


packages/teleport/src/Audit/EventList/EventList.tsx
```diff
@@ -14,147 +14,101 @@ See the License for the specific language governing permissions and
...
-  searchableProps: string[];
-  colSortDirs: Record<string, string>;
-  detailsToShow?: Event;
-};
+export const ActionCell = ({
+  event,
+  onShowDetails,
+}: {
+  event: Event;
+  onShowDetails: React.Dispatch<React.SetStateAction<Event>>;
```
alex-kovoy, line 89: Why typing with dispatcher when we can do this?

`React.Dispatch<React.SetStateAction<Event>>` -> `onShowDetails: (e: Event) => void` ?

packages/design/src/DataTableNext/types.ts
```diff
@@ -17,6 +18,9 @@ type TableColumnBase<T> = {
 export type PaginationConfig = {
   pageSize?: number;
   pagerPosition?: 'top' | 'bottom';
+};
+
+export type FetchingConfig = {
```
alex-kovoy, line 23: I think it's better if `onFetchMore` and `fetchStatus` are both part of TableProps without a dedicated (which also optional) `FetchingConfig` type. But I have no strong preferences.

packages/design/src/DataTableNext/Pager/usePager.ts
```diff
@@ -37,8 +37,8 @@ export type Props = {
   paginatedData: Array<Array<any>>;
   currentPage: number;
   pageSize: number;
-  onFetchMore: () => void;
-  fetchStatus: PaginationConfig['fetchStatus'];
+  onFetchMore?: () => void;
+  fetchStatus?: FetchingConfig['fetchStatus'];
```
alex-kovoy, line 41: Lets export a proper type:
```ts
export type FetchStatus =  'loading' | 'disabled' | '';
```

alex-kovoy (approved):
with minor comments

packages/teleport/src/Audit/EventList/EventList.tsx
rudream: The type with dispatcher is technically the more accurate type, but it doesn't make much difference

packages/design/src/DataTableNext/types.ts
rudream: I think grouping them up makes it more clear that they come together. Similar to how we do for pagination. Also, this way we can make the fetching config optional, but require both fields in case the user does want to add it.

PR #543, opened 2022-01-14
Update UserList to TableV2

rudream:
## Purpose

Part of #523

Updates UserList to use TableV2


alex-kovoy (approved):
This is not related but could you make sure that we close the AWS-login pop-up on Apps screen after it's been clicked? :) thanks.

PR #547, opened 2022-01-14
Add recently opened documents to workspace

gzdunek:


packages/teleterm/src/ui/DocumentTerminal/DocumentTerminal.tsx
```diff
@@ -14,12 +14,50 @@ See the License for the specific language governing permissions and
...
 
-export default function DocumentTerminal(props: Props & { visible: boolean }) {
+export default function DocumentTerminalContainer(props: Props) {
+  const ctx = useAppContext();
+  const { doc } = props;
+
+  useEffect(() => {
+    if (
+      doc.kind === 'doc.terminal_tsh_node' &&
+      ctx.serviceClusters.findCluster(`/clusters/${doc.rootClusterId}`)
```
gzdunek, line 31: I'm not sure about it, is there any better way to find cluster by id?

packages/teleterm/src/ui/Navigator/ExpanderConnections/useExpanderConnections.ts
```diff
@@ -18,81 +18,108 @@ import * as Icons from 'design/Icon';
...
-  return [...tshItems, ...gwItems, ...demoHistoryItems];
+    document,
+  };
 }
 
 export interface ConnectionItem extends types.NavItem {
-  status: 'connected' | 'disconnected';
-  kind: 'nav.connection-tsh' | 'nav.connection-gw';
+  readonly status: 'connected' | 'disconnected';
+  document: Document;
```
gzdunek, line 122: IMO we don't need this additional kind (as we have a document type)

packages/teleterm/src/ui/ClusterLogin/useClusterLogin.ts
```diff
@@ -85,7 +85,10 @@ export default function useClusterLogin(props: Props) {
 
     if (loginAttempt.status === 'success') {
       onClose();
-      serviceDocs.open(clusterUri);
+      if (!customOnSuccess) {
+        serviceDocs.open(clusterUri);
+      }
+      customOnSuccess?.();
```
alex-kovoy, line 91: Passing this callback through many components/actions seem too fragile. Lets try to add a 

```ts
    postLoginActions: new Map<{clusterUri, cb: ()=> void > ()
```

to the `clustersService` and allow to register a single post-action callback per cluster. The `commandLauncher` would be the place to initialize these post login actions, something like below:

```ts
  'cluster-login': {
    displayName: '',
    description: '',
    run(ctx: IAppContext, args: { clusterUri?: string, done }) {
      if(done){
        ctx.clustersService.setPostLoginAction(clusterUri, done)
      }else{
        // default action
        ctx.clustersService.setPostLoginAction(clusterUri, () => {
          serviceDocs.open(clusterUri);
        })
      }
      
      if (!args.clusterUri) {
        ctx.modalsService.openAddClusterDialog();
      } else {
        ctx.modalsService.openLoginDialog(args.clusterUri);
      }
    },
  },
 ```

File: packages/teleterm/src/services/workspace/workspaceService.ts [hunk 4]
```diff
@@ -38,6 +39,24 @@ export function createWorkspaceService(options: {
       });
   }
 
+  function addToRecentDocuments(document: Document): void {
+    const maxRecentDocumentsCount = 5;
```
alex-kovoy, line 43: no need for this limit, allow as many as it gets.

packages/teleterm/src/ui/DocumentTerminal/DocumentTerminal.tsx
alex-kovoy: I would try to replace  `login`, `rootClusterId`, and `leafClusterId` with `serverUri` (or just add serverUri to existing props) then you can do `ctx.clustersService.findClusterByResource(serverUri)` to find the cluster.

gzdunek: Good idea, thanks :) 
I won't be removing these properties for now, they seem useful.

packages/teleterm/src/ui/ClusterLogin/useClusterLogin.ts
gzdunek: Hmm it won't work for `openAddClusterDialog()` case - after adding a cluster it opens login dialog, but as we didn't  have a `clusterUri` we couldn't set an action for it (event default).

I can agree that passing down props has downsides, but I'm afraid that proposed solution can increase complexity (some of the props are passed directly to the modal, some properties are stored globally; where should we add new properties?). I'd like to avoid a global state and try to keep the modal as _clean_ as possible.

What would you say for renaming `customOnSuccess` to `onSuccess` and moving `serviceDocs.open(clusterUri);` from `useClusterLogin` to `commandLauncher`? By doing this we will remove that high-level logic from lower-level components.

packages/teleterm/src/ui/ResourceReconnect/ResourceReconnect.tsx
```diff
@@ -0,0 +1,41 @@
...
+export function ResourceReconnect(props: ReconnectProps) {
+  const ctx = useAppContext();
+  const { visible, connected, children, serverUri } = props;
+
+  function reconnect() {
+    if (serverUri) {
+      const cluster = ctx.clustersService.findClusterByResource(serverUri);
+      ctx.modalsService.openLoginDialog(cluster.uri, () => {
+        props.afterReconnect();
+      });
```
alex-kovoy-test, line 23: we should use the `commandLauncher` command instead.

packages/teleterm/src/ui/Navigator/ExpanderConnections/useExpanderConnections.ts
```diff
@@ -18,81 +18,108 @@ import * as Icons from 'design/Icon';
...
   ctx.docsService.useState();
 
+
+  //TODO: think how to use useStore for services from context bridge
+  useEffect(() => {
+    function onChange() {
+      rerender({});
+    }
+
+    ctx.workspaceService.subscribe(onChange);
```
alex-kovoy-test, line 40: curious if it needs to subscribe to the workspaceService? Would it be enough to use the service only for initial state?

packages/teleterm/src/ui/Navigator/ExpanderConnections/useExpanderConnections.ts
```diff
@@ -18,81 +18,108 @@ import * as Icons from 'design/Icon';
...
-      kind: 'nav.connection-tsh',
-      uri: '/nothing/402',
-      Icon: Icons.Keypair,
-      items: [],
-      group: false,
-      status: 'disconnected',
+function getWorkspaceItems(ctx: AppContext): ConnectionItem[] {
+  return [
+    ...ctx.workspaceService
+      .get()
```
alex-kovoy-test, line 54: `ctx.workspaceService.get()` -> `ctx.workspaceService.getRecentDocuments()`

packages/teleterm/src/ui/Navigator/ExpanderConnections/useExpanderConnections.ts
alex-kovoy-test: This would introduce a dependency on the `Document` on unnecessary levels. I would take what i need from the document to initiate the items and then let go of the document after that (if possible). This way rendering of an icon for the item will not require any Document references.

packages/teleterm/src/ui/Navigator/ExpanderConnections/useExpanderConnections.ts
```diff
@@ -18,81 +18,108 @@ import * as Icons from 'design/Icon';
...
-    {
-      title: 'pearl@node1',
-      kind: 'nav.connection-tsh',
-      uri: '/nothing/2cf',
-      Icon: Icons.Keypair,
-      items: [],
-      group: false,
-      status: 'disconnected',
+    get uri() {
+      return document.uri;
```
alex-kovoy-test, line 108: Lets keep simple objects without property accessors as I am not sure how it would behave with spread operators and such.

packages/teleterm/src/ui/ClusterLogin/useClusterLogin.ts
alex-kovoy: > Hmm it won't work for openAddClusterDialog() case - after adding a cluster it opens login dialog

Oh yeah, that was for illustration purposes to give an idea, `openAddClusterDialog()` needs to be removed from that command (or be a separate command instead). 

> I can agree that passing down props has downsides, but I'm afraid that proposed solution can increase complexity (some of the props are passed directly to the modal, some properties are stored globally; where should we add new properties?).

My thinking was to decouple `dialogs` from execution of post login actions by moving `postLoginActions` to the `clusterService`. This way it is possible to subscribe to these events without dependency on the dialog itself (via cb). The logout actions would work similarly.

> (some of the props are passed directly to the modal, some properties are stored globally; where should we add new properties?)

`dialog` has own properties, `clusterServices`, and `commandLauncher` commands theirs. `Command` encapsulates the high level relationships between these. To open a login dialog, one should execute a command with arguments. The command then opens a dialog and interacts with other services to carry out its execution.  

> What would you say for renaming customOnSuccess to onSuccess and moving serviceDocs.open(clusterUri); from useClusterLogin to commandLauncher

Yeah, that was the main idea - to decouple dialog from `docsService` or any "on-login" event handlers. If `postLoginActions` idea seems raw at the moment, lets postpone the discussion and proceed with `onSuccess` callback per your suggestion.

packages/teleterm/src/ui/Navigator/ExpanderConnections/useExpanderConnections.ts
alex-kovoy: This would introduce a dependency on the `Document` on unnecessary levels. I would take what i need from the document to initiate the items and then let go of the document after that (if possible). This way rendering of an icon for the item will not require any Document references.

packages/teleterm/src/ui/Navigator/ExpanderConnections/useExpanderConnections.ts
alex-kovoy: curious if it needs to subscribe to the workspaceService? Would it be enough to use the service only for initial state?

packages/teleterm/src/ui/Navigator/ExpanderConnections/useExpanderConnections.ts
gzdunek: Agree, no need to pass the `Document` down

packages/teleterm/src/ui/ClusterLogin/useClusterLogin.ts
gzdunek: I understand your points, some questions:
1. How would we handle `openAddClusterDialog()` to preserve current behavior (opening add dialog -> opening login dialog -> opening document)? It's used in `ExpanderCluster` for clicking an add icon. We can't set a callback per `clusterUri` as we don't know it yet.
2. Should an action be cleared after running it? For example, if someone needs to just run `clusterService.login()` directly, with no side effects, should he run `ctx.clustersService.setPostLoginAction(clusterUri, null)` first?

packages/teleterm/src/services/workspace/workspaceService.ts
gzdunek: Done

packages/teleterm/src/ui/Navigator/ExpanderConnections/useExpanderConnections.ts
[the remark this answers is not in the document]
gzdunek: Done

packages/teleterm/src/ui/DocumentTerminal/DocumentTerminal.tsx
```diff
@@ -14,12 +14,50 @@ See the License for the specific language governing permissions and
...
+    if (
+      doc.kind === 'doc.terminal_tsh_node' &&
+      ctx.clustersService.findClusterByResource(doc.serverUri)
+        ?.connected
+    ) {
+      ctx.docsService.update(doc.uri, { status: 'connected' });
+    }
+  }, []);
+
+  const documentTerminal = (
```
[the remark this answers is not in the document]
gzdunek: Done

packages/teleterm/src/ui/ResourceReconnect/ResourceReconnect.tsx
```diff
@@ -0,0 +1,41 @@
...
+  afterReconnect(): void;
+};
+
+export function ResourceReconnect(props: ReconnectProps) {
+  const ctx = useAppContext();
+  const { visible, connected, children, serverUri } = props;
+
+  function reconnect() {
+    if (serverUri) {
+      const cluster = ctx.clustersService.findClusterByResource(serverUri);
```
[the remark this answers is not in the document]
gzdunek: I will update it after deciding whether to use callback or `postLoginActions`

PR #549, opened 2022-01-14
Update MfaDeviceList to use TableV2

rudream:
## Purpose

Part of #523 

Updates MfaDeviceList to use TableV2

## Demo

image

### DIFF
--- a/packages/teleport/src/components/MfaDeviceList/MfaDeviceList.tsx
+++ b/packages/teleport/src/components/MfaDeviceList/MfaDeviceList.tsx
@@ -41,7 +41,7 @@ export default function MfaDeviceList({
         {
           key: 'name',
           headerText: 'Device Name',
-          render: ({ name }) => <NameCell name={name} />,
+          render: renderNameCell,
         },
         {
           key: 'registeredDate',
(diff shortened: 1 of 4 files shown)


packages/teleport/src/components/MfaDeviceList/MfaDeviceList.tsx
```diff
@@ -34,80 +26,56 @@ export default function MfaDeviceList({
...
-        cell={
-          <RemoveCell
-            remove={remove}
-            mostRecentDevice={mostRecentDevice}
-            mfaDisabled={mfaDisabled}
-          />
-        }
-      />
-    </StyledTable>
+    <StyledTable
```
alex-kovoy, line 33: Does it still need this `<StyledTable>` ?

alex-kovoy (approved):
with comment

packages/teleport/src/components/MfaDeviceList/MfaDeviceList.tsx
rudream: Yes because here we add some styles that aren't default `vertical-align: middle` and `height: 32px`

PR #550, opened 2022-01-14
Clean up custom cells in Tables

rudream:
## Purpose

Cleans up custom cells and their types and makes them easier to read.

### DIFF
--- a/packages/teleport/src/components/NodeList/NodeList.tsx
+++ b/packages/teleport/src/components/NodeList/NodeList.tsx
@@ -43,7 +43,8 @@ function NodeList(props: Props) {
         },
         {
           altKey: 'connect-btn',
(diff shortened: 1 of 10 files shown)


packages/teleport/src/components/NodeList/NodeList.tsx
```diff
@@ -64,25 +56,21 @@ function NodeList(props: Props) {
...
-  onSelect,
-  onOpen,
-  serverId,
-}: {
-  onSelect?: (e: React.SyntheticEvent, login: string, serverId: string) => void;
-  onOpen: (serverId: string) => LoginItem[];
-  serverId: string;
-}) => {
+const renderLoginCell = (
+  { id }: Node,
```
alex-kovoy, line 60: if `id` only needed then `renderLoginCell(id: string, ...)`

alex-kovoy (approved):
with comments

PR #558, opened 2022-01-19
List recordings

ibeckermayer:
- Updates the `RecordingService` to fetch both `SESSION_END` and `DESKTOP_SESSION_ENDED` events
- Updates `makeRecording` to be compatible with both types of events it can expect to get back as described in `events.proto` (most up to date `DESKTOP_SESSION_ENDED` message is [here](https://github.com/gravitational/teleport/blob/2b550f39ce6d33e4b8bfa3c8c7e37022899beb79/api/types/events/events.proto#L1668-L1705))
- Adds a `Type` column to the Recordings table that displays the appropriate icon for ssh vs desktop sessions
- Opportunistically update the `AuditService`'s `EventQuery.filterBy` parameter to be an `EventCode` type in order to make the API marginally more comprehensible.

Tested against [`isaiah/update-session-end-2`](https://github.com/gravitational/teleport/pull/9839). (Arguably this might best be merged into a feature branch, but I think merging into master is ok because it's totally backwards compatible.)


packages/teleport/src/services/recordings/makeRecording.ts
```diff
@@ -1,35 +1,43 @@
...
+    code === eventCodes.DESKTOP_SESSION_ENDED ? 'desktop' : 'ssh';
   let durationText = '';
   let duration = 0;
   if (session_start && session_stop) {
     duration = moment(session_stop).diff(session_start);
     durationText = moment.duration(duration).humanize();
   }
 
-  let hostname = server_hostname || 'N/A';
+  // Desktop sessions will have desktop_name set, ssh sessions can have server_hostname or neither.
```
ibeckermayer, line 32: I inferred "ssh sessions can have server_hostname or neither" from the logic, though I'm not sure in what situation an ssh session wouldn't have a server_hostname.

packages/teleport/src/services/recordings/recordings.ts
```diff
@@ -12,11 +14,14 @@ export default class RecordingsService {
     const start = params.from.toISOString();
     const end = params.to.toISOString();
 
-    const url = cfg.getClusterEventsRecordingsUrl(clusterId, {
+    const url = cfg.getClusterEventsUrl(clusterId, {
```
alex-kovoy, line 17: lets remove this service and move `fetchRecordings` and `makeRecording` methods to the `audit` service.

alex-kovoy (dismissed):

packages/teleport/src/services/recordings/recordings.ts
alex-kovoy: Actually, lets keep it as is, seems cleaner to keep these abstractions separated.

alex-kovoy: actually, lets keep it. Seems cleaner to separate recordings from events this way.

packages/teleport/src/services/recordings/makeRecording.ts
ibeckermayer: I think so, though I'm not sure precisely where the names are coming from. cc @zmb3

packages/teleport/src/services/audit/audit.ts
```diff
@@ -18,6 +18,7 @@ import api from 'teleport/services/api';
 import cfg from 'teleport/config';
 import makeEvent from './makeEvent';
 import { EventQuery, EventResponse } from './types';
+import { formatters } from 'teleport/services/audit';
```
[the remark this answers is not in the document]
ibeckermayer: I don't see the issue, what does a cross reference mean?

packages/teleport/src/Recordings/RecordingsList.tsx
```diff
@@ -81,6 +91,21 @@ export default function RecordingsList(props: Props) {
   );
 }
 
+const iconCell = (type: RecordingType) => {
```
alex-kovoy, line 94: `iconCell` -> `renderIconCell` as it returns react elements

packages/teleport/src/Recordings/RecordingsList.tsx
```diff
@@ -81,6 +91,21 @@ export default function RecordingsList(props: Props) {
...
+  const iconProps = {
+    p: '1',
+    mr: '3',
+    fontSize: '3',
+  };
+  const icon = type === 'ssh' ? Icons.Cli : Icons.Desktop;
+
+  return (
+    <Cell>
+      <Icon {...iconProps} as={icon} />
```
alex-kovoy, line 104: I think it's more readable when these attributes are close to the component itself. 
`<Icon   p={1} mr={3}  fontSize={3}..`

packages/teleport/src/services/recordings/types.ts
```diff
@@ -10,12 +10,19 @@ export type RecordingsResponse = {
   startKey: string;
 };
 
+export type RecordingType = 'ssh' | 'desktop';
+export type RecordingDescription =
+  | 'play'
+  | 'recording disabled'
+  | 'non-interactive';
+
```
alex-kovoy, line 18: why is it a `type` ? The type is a `string` and these are the values.

packages/teleport/src/services/recordings/makeRecording.ts
```diff
@@ -1,7 +1,50 @@
...
+  sid,
+  desktop_name,
+  session_recording,
+}) {
+  const { duration, durationText } = formatDuration(
+    session_start,
+    session_stop
+  );
+
+  let description: RecordingDescription = 'play';
```
alex-kovoy, line 28: why typing a text?

packages/teleport/src/services/recordings/makeRecording.ts
```diff
@@ -1,7 +1,50 @@
...
+
+  return {
+    duration,
+    durationText,
+    sid,
+    createdDate: time,
+    users: user,
+    hostname: desktop_name,
+    description,
+    recordingType,
```
alex-kovoy, line 43:
```
recordingType: 'desktop'
```

packages/teleport/src/services/recordings/makeRecording.ts
alex-kovoy: @zmb3 consider using similar property names for desktop events. For example `user` could be a `participant` - these things will make it a bit easier for people to consume the events.

packages/teleport/src/services/audit/audit.ts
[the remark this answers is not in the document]
alex-kovoy: it means that 2 files reference each other. Webpack can resolve it, yes. but it's best to avoid it in the first place.

packages/teleport/src/services/recordings/makeRecording.ts
alex-kovoy: Oh, i see why. This is because the same description text is used for 2 event types. In this case i think it's OK to keep it.

ibeckermayer: My editor was complaining at me if I didn't do it like this 🤷

packages/teleport/src/services/recordings/types.ts
ibeckermayer: I typed it because we are relying on this elsewhere to determine whether we show the "play" button or not ([link](https://github.com/gravitational/webapps/blob/master/packages/teleport/src/Recordings/RecordingsList.tsx#L85-L91)). Awkward though, I'll use a constant instead.

packages/teleport/src/Recordings/RecordingsList.tsx
```diff
@@ -81,8 +95,18 @@ export default function RecordingsList(props: Props) {
...
+  return (
+    <Cell>
+      <Icon p={1} mr={3} fontSize={3} as={icon} />
+    </Cell>
+  );
+};
+
 const renderPlayCell = ({ description, sid }: Recording, clusterId: string) => {
-  if (description !== 'play') {
+  if (description !== PlayDescription) {
```
alex-kovoy, line 109: sorry this does not seem quite right when description text is used as a condition. What if we add a new property `playable` and use it instead? 

Plus, remove this type `PlayDescription` and use a regular `const` instead (if the same description text is used in several places)

ibeckermayer: Sure. Since we're here, what do you think about just removing non-playable sessions from the table ([ref](https://github.com/gravitational/teleport/pull/9839#issuecomment-1016767764))?

alex-kovoy: that was product/ux decision

alex-kovoy (approved):

PR #568, opened 2022-01-25
Replace momentjs with date-fns

### LINKED ISSUE #467: [infra] Replace momentjs npm package with date-fns
Lets replace momentjs (that adds ~200kb to our bundle) with date-fns. 

Some random links to get an idea for the change:
https://date-fns.org/
https://blog.bitsrc.io/date-fns-vs-momentjs-9833f7463751
https://medium.com/@k2u4yt/momentjs-vs-date-fns-6bddc7bfa21e

rudream:
## Purpose

This PR resolves #467 

Remove momentjs package and migrate to date-fns


packages/shared/services/loc/loc.ts
```diff
@@ -46,13 +44,15 @@ export function displayDate(date: Date) {
...
+    return format(date, cfg.dateTimeFormat);
   } catch (err) {
     logger.error('displayDateTime()', err);
     return 'undefined';
   }
 }
+
+function dateToUtc(date: Date) {
+  return new Date(date.getTime() + date.getTimezoneOffset() * 60 * 1000);
+}
```
kimlisa, line 58: hm, curious what is `60 * 1000` for?

packages/teleport/src/services/clusters/makeCluster.ts
```diff
@@ -37,7 +37,7 @@ export default function makeCluster(json): Cluster {
     'proxyVersion',
   ]);
 
-  const connectedText = displayDateTime(lastConnected);
+  const connectedText = displayDateTime(new Date(lastConnected));
```
kimlisa, line 40:
```suggestion
  const lastConnectedDate = new Date(lastConnected)
  const connectedText = displayDateTime(lastConnectedDate);
```
then on line 44 of this file, remove the date constructor

packages/teleport/src/services/recordings/makeRecording.ts
```diff
@@ -85,13 +85,17 @@ function makeSshRecording({
   } as Recording;
 }
 
-function formatDuration(start: string, stop: string) {
+function formatDuration(startDate: string, stopDate: string) {
```
kimlisa, line 88: i would switch the naming, b/c `startDate` implies it's in date format?

packages/teleport/src/services/audit/makeEvent.ts
```diff
@@ -219,8 +218,9 @@ export const formatters: Formatters = {
       }
 
       if (event.session_start && event.session_stop) {
-        const duration = moment(event.session_stop).diff(event.session_start);
-        const durationText = moment.duration(duration).humanize();
+        const start = new Date(event.session_start);
+        const end = new Date(event.session_stop);
```
kimlisa, line 222: should this be converted in `makeRecording`?

packages/teleport/src/services/ssh/makeSession.ts
```diff
@@ -14,15 +14,14 @@ See the License for the specific language governing permissions and
...
+import { formatDistanceStrict } from 'date-fns';
 import { map } from 'lodash';
 import { Session, Participant } from './types';
 
 export default function makeSession(json): Session {
   const clusterId = json.cluster_name;
   const created = new Date(json.created);
-  const duration = moment(new Date()).diff(created);
-  const durationText = moment.duration(duration).humanize();
+  const durationText = formatDistanceStrict(new Date(), new Date(created));
```
kimlisa, line 24: `created` is already in date format? line 23

packages/teleport/src/services/nodes/makeNodeBashCmd.ts
```diff
@@ -14,13 +14,12 @@ See the License for the specific language governing permissions and
...
 
-import moment from 'moment';
+import { formatDistanceStrict } from 'date-fns';
 import cfg from 'teleport/config';
 import { NodeToken, BashCommand } from './types';
 
 export default function makeNodeBashCmd(token: NodeToken): BashCommand {
-  const duration = moment(new Date()).diff(token.expiry);
-  const expires = moment.duration(duration).humanize();
+  const expires = formatDistanceStrict(new Date(), new Date(token.expiry));
```
kimlisa, line 22: confused on this one, the typing implies that `token.expiry` should already come as date format, but we are formatting it again. Isn't `token: Nodetoken` a  json response? If it is, i'd remove the typing and rename it to `json`

packages/teleport/src/Users/UserTokenLink/UserTokenLink.tsx
```diff
@@ -33,9 +33,7 @@ export default function UserTokenLink({
   asInvite = false,
 }: Props) {
   const tokenUrl = cfg.getUserResetTokenRoute(token.value, asInvite);
-
-  const duration = moment(Date.now()).diff(token.expires);
-  const expiresText = moment.duration(duration).humanize();
+  const expiresText = formatDistanceStrict(Date.now(), new Date(token.expires));
```
kimlisa, line 36: `token.expires` should already come as date formatted?

packages/shared/services/loc/loc.ts
hatched: You can pass the ms since Epoch to the constructor of `Date` to get a new `Date` instance at a certain time.

In this case, `getTime` returns the time in ms since Epoch but `getTimezoneOffset` returns the time in minutes. So to convert minutes to ms you multiply it by 60,000. 

```typescript
const date = new Date()
> undefined
new Date(date.getTime() + date.getTimezoneOffset() * 60 * 1000);
> Date Tue Jan 25 2022 21:55:29 GMT-0600 (Central Standard Time)
```

packages/teleport/src/components/EventRangePicker/utils.ts
```diff
@@ -1,25 +1,16 @@
-import moment from 'moment';
+import { startOfDay, endOfDay, subDays } from 'date-fns';
 
-export function getRangeOptions(): EventRange[] {
+export function getRangeOptions() {
```
hatched, line 3: This still appears to return the type `EventRange[]`, accidental removal?

alex-kovoy (approved):

packages/shared/services/loc/loc.ts
rudream: I break it up into `* 60 * 1000` instead of just `* 60000` because that makes it a bit more clear that we're converting to seconds and then to milliseconds.

packages/teleport/src/services/audit/makeEvent.ts
rudream: no because `makeRecording` is only used for the `Recordings` tab, not audit log. Everything else in `makeEvent` deals with raw response data.

PR #572, opened 2022-01-28
desktop session recording

ibeckermayer:
Opening up a review worthy feature branch for the Desktop Session player. This is the webapps counterpart to https://github.com/gravitational/teleport/pull/9583.


packages/teleport/src/lib/tdp/client.ts
```diff
@@ -93,14 +101,42 @@ export default class Client extends EventEmitter {
...
+  handleMouseMove(buffer: ArrayBuffer) {
+    this.handleError(
+      new Error(
+        `recieved unsupported message type ${this.codec._decodeMessageType(
+          buffer
+        )}`
+      )
+    );
+  }
+
```
ibeckermayer, line 133: These are overridden in `PlayerClient` (which inherits `Client`).

packages/teleport/src/components/TdpClientCanvas/TdpClientCanvas.tsx
```diff
@@ -98,19 +105,16 @@ export default function TdpClientCanvas(props: Props) {
...
       canvas.onmouseup = onmouseup;
       const onwheel = (e: WheelEvent) => {
-        e.preventDefault();
         onMouseWheelScroll(tdpCli, e);
       };
       canvas.onwheel = onwheel;
 
       // Key controls.
       const onkeydown = (e: KeyboardEvent) => {
-        e.preventDefault();
```
ibeckermayer, line 108: moved `e.preventDefault()` to `useTdpClientCanvas.tsx` (which holds the default callbacks for `DesktopSession`s), because we don't want to prevent default with the `DesktopPlayer`.

packages/teleport/src/components/TdpClientCanvas/TdpClientCanvas.tsx
```diff
@@ -135,13 +139,14 @@ export default function TdpClientCanvas(props: Props) {
 
 export type Props = {
   tdpCli: TdpClient | null;
-  tdpCliOnImageFragment: (
-    ctx: CanvasRenderingContext2D,
-    data: ImageFragment
-  ) => void;
+  tdpCliOnPngFrame: (ctx: CanvasRenderingContext2D, pngFrame: PngFrame) => void;
```
ibeckermayer, line 142: Updating the naming to more closely correspond with [the TDP spec](https://github.com/gravitational/teleport/blob/master/rfd/0037-desktop-access-protocol.md).

packages/teleport/src/lib/tdp/codec.ts
```diff
@@ -328,26 +353,34 @@ export default class Codec {
     return message;
   }
 
-  // decodeRegion decodes the region from a PNG_FRAME tdp message.
-  decodeRegion(buffer: ArrayBuffer): Region {
+  // decodePngFrame decodes a raw tdp PNG frame message and returns it as a PngFrame
+  // | message type (2) | left uint32 | top uint32 | right uint32 | bottom uint32 | data []byte |
+  // https://github.com/gravitational/teleport/blob/master/rfd/0037-desktop-access-protocol.md#2---png-frame
+  decodePngFrame(buffer: ArrayBuffer, onload: (PngFrame) => any): PngFrame {
```
ibeckermayer, line 359: Refactored in order to create a pattern of having `decodeTdpMessage(buffer: ArrayBuffer): TdpMessage` (another example is `decodeClientScreenSpec(buffer: ArrayBuffer): ClientScreenSpec`.

However in this particular case, I needed to pass an `onload` function in since the `onload` must be set before `data.src`, otherwise there's a race condition (see [here](https://github.com/gravitational/webapps/pull/572/files#diff-f7aac69b83399d0d34d5e42582b6a8e1c04375ef9205491f486f417191a93b9bR136-R139) for how it's used).

packages/teleport/src/Player/DesktopPlayer.tsx
```diff
@@ -0,0 +1,124 @@
...
+      // TODO
+      return;
+    },
+    toggle: () => {
+      // emits PlayerClientEvent.TOGGLE_PLAY_PAUSE
+      playerClient.togglePlayPause();
+    },
+  });
+
+  playerClient.on(PlayerClientEvent.TOGGLE_PLAY_PAUSE, () => {
```
zmb3, line 72: 1. Won't this subscribe to the play/pause event on every render?
2. Shouldn't there be an unsubscribe when the component is unmounted?

packages/teleport/src/Player/DesktopPlayer.tsx
```diff
@@ -0,0 +1,124 @@
...
+const useDesktopPlayer = ({
+  sid,
+  clusterId,
+}: {
+  sid: string;
+  clusterId: string;
+}) => {
+  const playerClient = new PlayerClient(
+    cfg.api.desktopPlaybackWsAddr
+      .replace(':fqdm', getHostName())
```
zmb3, line 91: Is this a typo? Shouldn't it be `fqdn`?

packages/teleport/src/components/TdpClientCanvas/TdpClientCanvas.tsx
zmb3: Why not? What do we need keyboard input for in the player? Play/pause controls?

packages/teleport/src/config.ts
```diff
@@ -65,7 +65,7 @@ const cfg = {
     consoleNodes: '/web/cluster/:clusterId/console/nodes',
     consoleConnect: '/web/cluster/:clusterId/console/node/:serverId/:login',
     consoleSession: '/web/cluster/:clusterId/console/session/:sid',
-    player: '/web/cluster/:clusterId/session/:sid',
+    player: '/web/cluster/:clusterId/session/:sid/:recordingType', // recordingType is one of 'ssh' | 'desktop'
```
zmb3, line 68: Recording type seems like it would make more sense as a query param than part of the path, but I don't _really_ care if this is how we typically do things.

packages/teleport/src/lib/tdp/codec.ts
```diff
@@ -301,21 +322,15 @@ export default class Codec {
...
     const messageType = new DataView(buffer).getUint8(0);
-    if (
-      messageType === MessageType.PNG_FRAME ||
-      messageType === MessageType.CLIPBOARD_DATA ||
-      messageType === MessageType.ERROR
-    ) {
-      return messageType;
-    } else {
-      // We don't expect to need to decode any other value on the client side
+    if (messageType > MessageType.ERROR) {
```
zmb3, line 330: This is okay for now but seems fragile as we add messages.

I might go with a switch and be explicit about the types you decode. This way it's obvious when one is missing.

packages/teleport/src/lib/tdp/codec.ts
```diff
@@ -328,26 +343,34 @@ export default class Codec {
...
-  toBase64(buffer: ArrayBuffer) {
+  _toBase64(buffer: ArrayBuffer) {
     const binary = String.fromCharCode.apply(null, new Uint8Array(buffer, 17));
     return btoa(binary);
   }
 
   // decodePng creates a data:image uri from the png data part of a PNG_FRAME tdp message.
-  decodePng(buffer: ArrayBuffer): string {
-    return `data:image/png;base64,${this.toBase64(buffer)}`;
+  _decodePng(buffer: ArrayBuffer): string {
```
zmb3, line 373: Is there a better name for this? You're not actually decoding PNG data. If anything, you're _encoding_ base64 data.

packages/teleport/src/lib/tdp/playerClient.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  handleMouseButton(buffer: ArrayBuffer) {
+    this.logger.warn('TODO: handleMouseButton');
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  handleMouseMove(buffer: ArrayBuffer) {
+    this.logger.warn('TODO: handleMouseMove');
```
zmb3, line 51: Why do you want to log a warning here? Wouldn't it be better to just not have a mouse move handler at all?

packages/teleport/src/Player/DesktopPlayer.tsx
ibeckermayer: Yes and yes, good catch.

packages/teleport/src/components/TdpClientCanvas/TdpClientCanvas.tsx
ibeckermayer: We don't _need_ it necessarily (currently we don't have keyboard controls defined, though we might do so in the future). 

The original purpose of those preventDefault's on the canvas was to ensure that what a user was doing on the remote machine wasn't affecting the browser. For example a user intending to press Windows+T on the remote machine (corresponding to cmd+T on Mac) shouldn't find themselves with a new browser tab open. However that's no longer relevant.

packages/teleport/src/lib/tdp/playerClient.ts
ibeckermayer: No need to log a warning. However we do want to override the [default behavior](https://github.com/gravitational/webapps/blob/4c8ef4ae479833cf29bcd1c294ac1cbe746799d5/packages/teleport/src/lib/tdp/client.ts#L114-L132) from the `Client` parent class.

packages/teleport/src/lib/tdp/codec.ts
ibeckermayer: True, added a `TODO`.

packages/teleport/src/config.ts
ibeckermayer: True, I was just going with what I know but its better to do it that way, especially since we plan to add the time parameter as well.

ibeckermayer:
FTR I've added issues for a couple of known deficiencies:
- https://github.com/gravitational/webapps/issues/574
- https://github.com/gravitational/webapps/issues/575

packages/teleport/src/DesktopSession/DesktopSession.tsx
```diff
@@ -124,16 +126,18 @@ export function DesktopSession(props: State) {
           flex: 1, // ensures the canvas fills available screen space
         }}
         tdpCli={tdpClient}
-        tdpCliOnImageFragment={onImageFragment}
+        tdpCliOnPngFrame={onPngFrame}
         tdpCliOnTdpError={onTdpError}
         tdpCliOnWsClose={onWsClose}
         tdpCliOnWsOpen={onWsOpen}
+        tdpCliOnClientScreenSpec={onClientScreenSpec}
         onKeyDown={onKeyDown}
```
rudream, line 134: We can save space and clean this up a little by replacing line 134-140 with `{...props}` here and not destructuring them on line 47-53

packages/teleport/src/Player/DesktopPlayer.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+  return (
+    <StyledPlayer>
+      <TdpClientCanvas
+        tdpCli={playerClient}
+        tdpCliOnPngFrame={tdpCliOnPngFrame}
+        tdpCliOnClientScreenSpec={tdpCliOnClientScreenSpec}
+        onContextMenu={() => true}
+        tdpCliOnTdpError={doNothing} // TODO: we can show the error
+        tdpCliOnWsClose={doNothing}
```
rudream, line 33: Not sure how I feel about this approach using `doNothing`. Why not make these props optional for `TdpClientCanvas` and instead add a flag like `isRecording`, and then handle it accordingly within the `TdpClientCanvas` component logic?

ibeckermayer: Good idea to make those optional. I decided to just wrap all the logic that depends on the now optional handlers in if statements so that the `TdpClientCanvas` remains fully configurable and agnostic to specific uses.

packages/teleport/src/DesktopSession/DesktopSession.tsx
ibeckermayer: That would make it more difficult to read imo.

packages/teleport/src/config.ts
```diff
@@ -65,7 +66,7 @@ const cfg = {
     consoleNodes: '/web/cluster/:clusterId/console/nodes',
     consoleConnect: '/web/cluster/:clusterId/console/node/:serverId/:login',
     consoleSession: '/web/cluster/:clusterId/console/session/:sid',
-    player: '/web/cluster/:clusterId/session/:sid',
+    player: '/web/cluster/:clusterId/session/:sid', // ?recordingType=ssh|desktop
```
kimlisa, line 69: i think you can do something like below: https://github.com/remix-run/react-router/blob/v3/docs/guides/RouteMatching.md#path-syntax
```suggestion
    player: '/web/cluster/:clusterId/session/:sid(?recordingType=:recordingType)'
```

packages/teleport/src/config.ts
```diff
@@ -264,8 +267,11 @@ const cfg = {
     return generatePath(cfg.routes.appLauncher, { ...params });
   },
 
-  getPlayerRoute(params: UrlPlayerParams) {
-    return generatePath(cfg.routes.player, { ...params });
+  getPlayerRoute(params: UrlPlayerParams, search: UrlPlayerSearch) {
```
kimlisa, line 270: why can't `recordingType` be part of `UrlPlayerParam`?

packages/teleport/src/Player/Player.tsx
```diff
@@ -16,17 +16,26 @@ limitations under the License.
...
+import { getUrlParameter } from 'teleport/services/history';
 
 export default function Player() {
   const { sid, clusterId } = useParams<UrlPlayerParams>();
+  const recordingType = getUrlParameter('recordingType', useLocation().search);
+  if (recordingType !== 'ssh' && recordingType !== 'desktop') {
+    throw new Error(
+      `Invalid recording type: ${recordingType}, should be 'ssh' or 'desktop'`
+    );
+  }
```
kimlisa, line 37: i'm not sure if we should be throwing... i rarely see this, wouldn't it be better if we show an error banner like we do everywhere else?

packages/teleport/src/Audit/EventList/EventTypeCell.tsx
```diff
@@ -127,7 +127,15 @@ export default function renderTypeCell(event: Event, clusterId: string) {
...
             title="Open Session Player"
-            href={cfg.getPlayerRoute({ clusterId, sid: event.raw.sid })}
+            href={cfg.getPlayerRoute(
+              {
+                clusterId,
+                sid: event.raw.sid,
+              },
+              {
+                recordingType: 'ssh',
+              }
```
kimlisa, line 137: seems awkward, is there a reason why we can't combine `recordingType` with the first param object?

packages/teleport/src/DesktopSession/useTdpClientCanvas.tsx
```diff
@@ -126,18 +126,27 @@ export default function useTdpClientCanvas(props: Props) {
     }
   };
 
+  // Not used for DesktopSessions (but needed for playback)
+  const onClientScreenSpec = () => {};
+
+  // Block browser context menu so as not to obscure the context menu
+  // on the remote machine.
+  const onContextMenu = () => false;
```
kimlisa, line 134: I'm not sure I follow why we need these, and why it needs to be defined here

packages/teleport/src/Player/Player.tsx
kimlisa: https://gravitational.slack.com/archives/C02DQ1C2BMW/p1643753870653229

packages/teleport/src/config.ts
ibeckermayer: That guide is for an older version of React Router (v3), we are using v5.1.1. In the case of your suggestion, the parentheses and "?" are interpreted as regex symbols. I've done some more research, and afaict React Router doesn't have much if any support for dealing with query parameters.

https://stackoverflow.com/a/35604855/6277051
https://learnwithparam.com/blog/how-to-handle-query-params-in-react-router/

packages/teleport/src/config.ts
ibeckermayer: [Related](https://github.com/gravitational/webapps/pull/572#discussion_r797713609)

It looks like it could be, as in I can do

```typescript
cfg.routes.player: '/web/cluster/:clusterId/session/:sid', // ?recordingType=ssh|desktop

export interface UrlPlayerParams {
  clusterId: string;
  sid: string;
  recordingType: RecordingType;
}

getPlayerRoute(params: UrlPlayerParams) {
    return generatePath(cfg.routes.player, { ...params }) + `?recordingType=${params.recordingType}`;
  },
```

and `generatePath` will just ignore the additional `recordingType` that it doesn't see in `cfg.routes.player`. 

However afaik throughout the rest of the code we are using `*Params` to mean "params` in the react router sense of the word, namely path parameters, and so I decided to make it clear that `recordingType` is not one of those, it's distinct (a query parameter, or "search" since that's the nomenclature used by [`URLSearchParams`](https://github.com/gravitational/webapps/blob/isaiah/desktop-session-recording/packages/teleport/src/services/history/history.ts#L115-L119))

packages/teleport/src/Audit/EventList/EventTypeCell.tsx
ibeckermayer: https://github.com/gravitational/webapps/pull/572#discussion_r797725636

packages/teleport/src/config.ts
hatched: Just for a reference if you're interested, my go to for defining and typing query params is to use [use-query-params](https://www.npmjs.com/package/use-query-params)

packages/teleport/src/DesktopSession/useTdpClientCanvas.tsx
ibeckermayer: `onClientScreenSpec` is a holdover from before I took Yassine's advice and made that an optional prop.

`onContextMenu` is needed so that the browser's context menu doesn't pop up when the user right clicks on the canvas (`TdpClientCanvas`) -- in that case, only the _Windows_ context menu should show up and shouldn't be blocked by the browser's menu.

packages/teleport/src/Player/DesktopPlayer.tsx
```diff
@@ -0,0 +1,118 @@
...
+      <ProgressBarDesktop playerClient={playerClient} />
+    </StyledPlayer>
+  );
+};
+
+export const ProgressBarDesktop = (props: { playerClient: PlayerClient }) => {
+  const { playerClient } = props;
+
+  const [state, setState] = useState({
+    max: 500, // TODO: total time of the recording in ms
```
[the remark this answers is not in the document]
ibeckermayer: Those need to be set to the actual parameters of the recording, and updated periodically so that the progress bar updates as the recording progresses.

Implemented in the follow-on PR: https://github.com/gravitational/webapps/pull/577

packages/teleport/src/Player/DesktopPlayer.tsx
```diff
@@ -0,0 +1,118 @@
...
+    time: '-:--', // TODO: the human readable time the playback is at
+    isPlaying: true, // determines whether play or pause symbol is shown
+    move: () => {
+      // TODO
+      return;
+    },
+    toggle: () => {
+      // emits PlayerClientEvent.TOGGLE_PLAY_PAUSE
+      playerClient.togglePlayPause();
+    },
```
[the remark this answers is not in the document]
ibeckermayer: We could, but then we also need `playerClient` down below for `playerClient.addListener`

packages/teleport/src/components/TdpClientCanvas/TdpClientCanvas.tsx
```diff
@@ -49,83 +49,115 @@ export default function TdpClientCanvas(props: Props) {
...
-        canvas.removeEventListener('keyup', onkeyup);
-        canvas.removeEventListener('wheel', onwheel);
+        if (onContextMenu)
+          canvas.removeEventListener('contextmenu', oncontextmenu);
+        if (onMouseMove) canvas.removeEventListener('mousemove', onmousemove);
+        if (onMouseDown) canvas.removeEventListener('mousedown', onmousedown);
+        if (onMouseUp) canvas.removeEventListener('mouseup', onmouseup);
+        if (onKeyDown) canvas.removeEventListener('keydown', onkeydown);
+        if (onKeyUp) canvas.removeEventListener('keyup', onkeyup);
+        if (onMouseWheelScroll) canvas.removeEventListener('wheel', onwheel);
```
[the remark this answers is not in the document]
ibeckermayer: Oh yeah, you're correct. I prefixed all of these with `_`, and moved `const _onkeyup = (e: KeyboardEvent) => ...` out of the `if`'s context. 

So now the `_onkeyup` in `if (onKeyUp) canvas.removeEventListener('keyup', _onkeyup);` corresponds to the one I define above, and it's clear that it's not just the default `onkeyup` listener (which I believe is what those `removeEeventListener`'s were removing before this fix).

packages/teleport/src/config.ts
ibeckermayer: @hatched thanks, looks nice enough but my first impression is that its a bit overpowered for this application. We're not using them for any sort of complex state management.

packages/teleport/src/config.ts
kimlisa: is `recordingType` really a query type though? to me query seems more like "search for something", but we aren't searching

packages/teleport/src/Player/DesktopPlayer.tsx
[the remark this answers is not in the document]
kimlisa: i'm not sure what you mean

packages/teleport/src/config.ts
ibeckermayer: The lines are a bit blurry with a SPA I suppose. I was thinking of the path parameters as "distinct path to the backend resource" and the query parameters as "other". I originally had what you suggested, but changed it owing to [Zac's suggestion](https://github.com/gravitational/webapps/pull/572#discussion_r795799523).

packages/teleport/src/Player/DesktopPlayer.tsx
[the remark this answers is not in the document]
ibeckermayer: Its easiest for me to just take in the entire `playerClient` as a prop, since I need it for multiple uses in here. It would need to be `<ProgressBar playerClient={playerClient} toggle={playerClient.togglePlayPause} />`, but that seems like an unnecessary additional prop.

packages/teleport/src/config.ts
kimlisa: i see, i'm curious about @hatched opinions as well

kimlisa: i also maybe think these should be two distinct routes

packages/teleport/src/Player/DesktopPlayer.tsx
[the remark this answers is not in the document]
ibeckermayer: Never mind I'm wrong, I was mixing up `ProgressBarDesktop` and `ProgressBar`.

packages/teleport/src/config.ts
zmb3: My reasoning for suggesting the query param..

`/web/cluster/:clusterId/session/:sid/` is unique and refers to a single resource. The session is either a desktop session or an SSH session. It can never be both, meaning we could never have `/web/cluster/foo/session/123/desktop` and `/web/cluster/foo/session/123/ssh`.

The path should describe exactly **what** I want to see. `/web/cluster/:clusterId/session/:sid/` uniquely describes the session I want to view.

The query params are _paramaters_ to the query that can adjust **how** I want to see it. Other examples of using query params in this way are APIs that take a `?format=json` or `?sort=desc` param. Again, these aren't specifying what resource, but changing how it is presented or displayed.

kimlisa: @zmb3 that makes sense, thanks for the explanation!

packages/teleport/src/Player/Player.tsx
```diff
@@ -35,22 +41,40 @@ export default function Player() {
...
-      </Flex>
+      {recordingType !== 'ssh' && recordingType !== 'desktop' ? (
+        <Box textAlign="center" m={10}>
+          <Danger>
+            {' '}
+            `Invalid recording type: {recordingType}, should be 'ssh' or
+            'desktop'`{' '}
+          </Danger>
+        </Box>
+      ) : (
```
kimlisa, line 52: i think this is easier to read:
```ts
  if (recordingType !== 'ssh' && recordingType !== 'desktop') {
    return (
      <StyledPlayer>
        <Box textAlign="center" m={10}>
          <Danger>
            `Invalid recording type: {recordingType}, should be 'ssh' or
            'desktop'`
          </Danger>
        </Box>
      </StyledPlayer>
    );
  }

  return (
    <StyledPlayer>
      <Flex bg={colors.primary.light} height="38px">
        ....etc
```

packages/teleport/src/Player/Player.tsx
```diff
@@ -16,17 +16,23 @@ limitations under the License.
...
 import session from 'teleport/services/session';
 import { colors } from 'teleport/Console/colors';
 import { UrlPlayerParams } from 'teleport/config';
+import { getUrlParameter } from 'teleport/services/history';
+import { Danger } from 'design/Alert';
 
 export default function Player() {
   const { sid, clusterId } = useParams<UrlPlayerParams>();
+
+  const recordingType = getUrlParameter('recordingType', useLocation().search);
```
kimlisa, line 34:
```suggestion
  const recordingType = getUrlParameter('recordingType', useLocation().search) as RecordingType;
```

packages/teleport/src/Player/Player.story.tsx
```diff
@@ -40,3 +40,20 @@ export const Player = () => {
     </Router>
   );
 };
+
```
kimlisa, line 43: lets fix the `Player` story, it's broken b/c it expects a `recordingType` in url, we also should probably make another story to see rendering of both types of players

packages/teleport/src/components/TdpClientCanvas/TdpClientCanvas.tsx
```diff
@@ -134,23 +166,28 @@ export default function TdpClientCanvas(props: Props) {
...
     canvas: HTMLCanvasElement,
     e: MouseEvent
   ) => void;
-  onMouseDown: (cli: TdpClient, e: MouseEvent) => void;
-  onMouseUp: (cli: TdpClient, e: MouseEvent) => void;
-  onMouseWheelScroll: (cli: TdpClient, e: WheelEvent) => void;
+  onMouseDown?: (cli: TdpClient, e: MouseEvent) => void;
+  onMouseUp?: (cli: TdpClient, e: MouseEvent) => void;
+  onMouseWheelScroll?: (cli: TdpClient, e: WheelEvent) => void;
+  onContextMenu?: () => void;
```
kimlisa, line 191: shouldn't this be 
```suggestion
  onContextMenu?: () => boolean;
```

packages/teleport/src/lib/tdp/client.ts
```diff
@@ -77,30 +79,56 @@ export default class Client extends EventEmitter {
...
     try {
       if (messageType === MessageType.PNG_FRAME) {
-        this.processFrame(buffer);
+        this.handlePngFrame(buffer);
+      } else if (messageType === MessageType.CLIENT_SCREEN_SPEC) {
+        this.handleClientScreenSpec(buffer);
+      } else if (messageType === MessageType.MOUSE_BUTTON) {
+        this.handleMouseButton(buffer);
+      } else if (messageType === MessageType.MOUSE_MOVE) {
+        this.handleMouseMove(buffer);
```
kimlisa, line 91: nit: switch case will be more readable
```ts
switch (messageType) {
  case MessageType.PNG_FRAME:
  ....
}
```

packages/teleport/src/lib/tdp/client.ts
```diff
@@ -77,30 +79,56 @@ export default class Client extends EventEmitter {
...
       } else if (messageType === MessageType.ERROR) {
         this.handleError(new Error(this.codec.decodeErrorMessage(buffer)));
       } else {
-        this.handleError(
-          new Error(`recieved unsupported message type ${messageType}`)
-        );
+        this.logger.warn(`received unsupported message type ${messageType}`);
```
kimlisa, line 95: why did we decide not to `this.handleError`? won't it be confusing if nothing happens? same as the handleXXX funcs below (not entirely sure how it gets invoked)

ibeckermayer: It's a DX +1 while giving nothing important up UX-wise:

Those cases should never happen, and only would show up if we screwed something up on the backend. Since its not super important that the `client[Player]` stop if an unexpected TDP message is received, I decided to just log it instead since it makes the development of new features easier (Zac asked for this change because he's developing the Clipboard feature).

packages/teleport/src/Player/Player.story.tsx
ibeckermayer: Updated. FYI the story surfaces this known issue: https://github.com/gravitational/webapps/issues/574

packages/teleport/src/config.ts
```diff
@@ -264,8 +267,11 @@ const cfg = {
...
+    return (
+      generatePath(cfg.routes.player, { ...params }) +
+      `?recordingType=${search.recordingType}`
+    );
```
kimlisa, line 274:
```suggestion
    return `${generatePath(cfg.routes.player, { ...params })}?recordingType=${search.recordingType}`
```

kimlisa (approved):
couple of things:

1) We should follow a consistent way of defining TODO comments like we do in teleport `TODO(isaiah) blah blah`, we don't want to start influencing bad habits to new devs :)
2) I trust that every change you made in our old components, you tested out the story book (it shouldn't break like it did for our Player), and that it was tested manually

PR #577, opened 2022-02-01
moving progress bar

ibeckermayer:
The primary changes are
- The overriding of `client.ts::processMessage` in `playerClient.ts:: processMessage` to parse an [`"end"` message](https://github.com/gravitational/teleport/pull/10088) or else parse a [full `DesktopRecording` event](https://github.com/gravitational/teleport/pull/10052) json and ultimately alert the `DesktopPlayer` of the new time to show on the progress bar via the `PlayerClientEvent.UPDATE_CURRENT_TIME` and `PlayerClientEvent.SESSION_END` events.
- The logic in `DesktopPlayer` for making the progress bar respond to those events.

### DIFF
--- a/packages/teleport/package.json
+++ b/packages/teleport/package.json
@@ -19,6 +19,7 @@
   "dependencies": {
     "@gravitational/design": "1.0.0",
     "@gravitational/shared": "1.0.0",
+    "base64-arraybuffer": "^1.0.2",
     "xterm": "^4.15.0",
     "xterm-addon-fit": "^0.5.0"
   },
--- a/packages/teleport/src/Player/DesktopPlayer.tsx
+++ b/packages/teleport/src/Player/DesktopPlayer.tsx
@@ -23,8 +23,6 @@ export const DesktopPlayer = ({
       clusterId,
     });
 
-  const doNothing = () => {};
-
   return (
     <StyledPlayer>
       <TdpClientCanvas
@@ -32,15 +30,6 @@ export const DesktopPlayer = ({
         tdpCliOnPngFrame={tdpCliOnPngFrame}
         tdpCliOnClientScreenSpec={tdpCliOnClientScreenSpec}
         onContextMenu={() => true}
-        tdpCliOnTdpError={doNothing} // TODO: we can show the error
-        tdpCliOnWsClose={doNothing}
-        tdpCliOnWsOpen={doNothing}
-        onKeyDown={doNothing}
-        onKeyUp={doNothing}
-        onMouseMove={doNothing}
-        onMouseDown={doNothing}
-        onMouseUp={doNothing}
-        onMouseWheelScroll={doNothing}
         // overflow: 'hidden' is needed to prevent the canvas from outgrowing the container due to some weird css flex idiosyncracy.
         // See https://gaurav5430.medium.com/css-flex-positioning-gotchas-child-expands-to-more-than-the-width-allowed-by-the-parent-799c37428dd6.
         style={{ display: 'flex', flexGrow: 1, overflow: 'hidden' }}
--- a/packages/teleport/src/Player/DesktopPlayer.tsx
+++ b/packages/teleport/src/Player/DesktopPlayer.tsx
@@ -14,14 +14,11 @@ export const DesktopPlayer = ({
   sid: string;
   clusterId: string;
 }) => {
-  const {
-    playerClient,
-    tdpCliOnPngFrame,
-    tdpCliOnClientScreenSpec,
-  } = useDesktopPlayer({
-    sid,
-    clusterId,
-  });
+  const { playerClient, tdpCliOnPngFrame, tdpCliOnClientScreenSpec } =
+    useDesktopPlayer({
+      sid,
+      clusterId,
+    });
 
   let doNothing = () => {};
 
@@ -88,7 +85,7 @@ const useDesktopPlayer = ({
 }) => {
   const playerClient = new PlayerClient(
     cfg.api.desktopPlaybackWsAddr
-      .replace(':fqdm', getHostName())
+      .replace(':fqdn', getHostName())
       .replace(':clusterId', clusterId)
       .replace(':sid', sid)
       .replace(':token', getAccessToken())
--- a/packages/teleport/src/Player/DesktopPlayer.tsx
+++ b/packages/teleport/src/Player/DesktopPlayer.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import styled from 'styled-components';
 import ProgressBar from './ProgressBar';
 import { PlayerClient, PlayerClientEvent } from 'teleport/lib/tdp';
@@ -20,7 +20,7 @@ export const DesktopPlayer = ({
       clusterId,
     });
 
-  let doNothing = () => {};
+  const doNothing = () => {};
 
   return (
     <StyledPlayer>
@@ -66,12 +66,20 @@ export const ProgressBarDesktop = (props: { playerClient: PlayerClient }) => {
     },
   });
 
-  playerClient.on(PlayerClientEvent.TOGGLE_PLAY_PAUSE, () => {
-    setState({
-      ...state,
-      isPlaying: !state.isPlaying,
+  useEffect(() => {
+    playerClient.addListener(PlayerClientEvent.TOGGLE_PLAY_PAUSE, () => {
+      // setState({...state, isPlaying: !state.isPlaying}) doesn't work because
+      // the listener is added when state == initialState, and that initialState
+      // value is effectively hardcoded into its logic.
+      setState(prevState => {
+        return { ...prevState, isPlaying: !prevState.isPlaying };
+      });
     });
-  });
+
+    return () => {
+      playerClient.nuke();
+    };
+  }, [playerClient]);
 
   return <ProgressBar {...state} />;
 };
--- a/packages/teleport/src/Player/DesktopPlayer.tsx
+++ b/packages/teleport/src/Player/DesktopPlayer.tsx
@@ -76,9 +76,7 @@ export const ProgressBarDesktop = (props: { playerClient: PlayerClient }) => {
       });
     });
 
-    return () => {
-      playerClient.nuke();
-    };
+    return playerClient.nuke();
   }, [playerClient]);
 
   return <ProgressBar {...state} />;
--- a/packages/teleport/src/Player/DesktopPlayer.tsx
+++ b/packages/teleport/src/Player/DesktopPlayer.tsx
@@ -6,13 +6,16 @@ import { PngFrame, ClientScreenSpec } from 'teleport/lib/tdp/codec';
 import cfg from 'teleport/config';
 import { getAccessToken, getHostName } from 'teleport/services/api';
 import TdpClientCanvas from 'teleport/components/TdpClientCanvas';
+import { throttle } from 'lodash';
 
 export const DesktopPlayer = ({
   sid,
   clusterId,
+  durationMs,
 }: {
   sid: string;
   clusterId: string;
+  durationMs: number;
 }) => {
   const { playerClient, tdpCliOnPngFrame, tdpCliOnClientScreenSpec } =
     useDesktopPlayer({
@@ -42,19 +45,28 @@ export const DesktopPlayer = ({
         // See https://gaurav5430.medium.com/css-flex-positioning-gotchas-child-expands-to-more-than-the-width-allowed-by-the-parent-799c37428dd6.
         style={{ display: 'flex', flexGrow: 1, overflow: 'hidden' }}
       />
-      <ProgressBarDesktop playerClient={playerClient} />
+      <ProgressBarDesktop playerClient={playerClient} durationMs={durationMs} />
     </StyledPlayer>
   );
 };
 
-export const ProgressBarDesktop = (props: { playerClient: PlayerClient }) => {
-  const { playerClient } = props;
+export const ProgressBarDesktop = (props: {
+  playerClient: PlayerClient;
+  durationMs: number;
+}) => {
+  const { playerClient, durationMs } = props;
+
+  const toHuman = (currentMs: number) => {
+    const date = new Date(null);
+    date.setMilliseconds(currentMs);
+    return date.toISOString().substr(14, 5);
+  };
 
   const [state, setState] = useState({
-    max: 500, // TODO: total time of the recording in ms
-    min: 0, // TODO: the recording always starts at 0 ms
-    current: 0, // TODO: the current time the playback is at
-    time: '-:--', // TODO: the human readable time the playback is at
+    max: durationMs,
+    min: 0,
+    current: 0, // the recording always starts at 0 ms
+    time: toHuman(0),
     isPlaying: true, // determines whether play or pause symbol is shown
     move: () => {
       // TODO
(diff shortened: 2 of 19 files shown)


packages/teleport/src/Player/DesktopPlayer.tsx
```diff
@@ -65,7 +77,39 @@ export const ProgressBarDesktop = (props: { playerClient: PlayerClient }) => {
...
+        setState(prevState => {
+          return {
+            ...prevState,
+            current: currentTimeMs,
+            time: toHuman(currentTimeMs),
+          };
+        });
+      },
+      // Magic number to throttle progress bar updates so that the playback is smoother.
+      50
```
hatched, line 91: What do you think about bumping it to 60ms? once per frame?

packages/teleport/src/Player/DesktopPlayer.tsx
```diff
@@ -65,7 +77,39 @@ export const ProgressBarDesktop = (props: { playerClient: PlayerClient }) => {
...
+    );
+
+    playerClient.addListener(
+      PlayerClientEvent.UPDATE_CURRENT_TIME,
+      currentTimeMs => throttledUpdateCurrentTime(currentTimeMs)
+    );
+
+    playerClient.addListener(PlayerClientEvent.SESSION_END, () => {
+      throttledUpdateCurrentTime.cancel();
+      // TODO: This logic can cause the progress bar to jump abruptly to the end, if the session was sitting without
```
hatched, line 101: Usually I create issues and then link them in the code along side TODO's so that they don't get lost.

packages/teleport/src/lib/tdp/playerClient.ts
```diff
@@ -34,6 +39,21 @@ export class PlayerClient extends Client {
...
 
+  // Overrides Client implementation.
+  processMessage(buffer: ArrayBuffer) {
+    const raw = this.textDecoder.decode(buffer);
+    if (raw === PlayerClientEvent.SESSION_END) {
+      this.emit(PlayerClientEvent.SESSION_END);
+      return;
+    }
+
+    const json = JSON.parse(raw);
```
hatched, line 50: Could put a try/catch around this will throw if `raw` is incorrectly formatted.

packages/teleport/src/Player/Player.tsx
```diff
@@ -30,9 +30,21 @@ import { getUrlParameter } from 'teleport/services/history';
...
+  );
+
   if (recordingType !== 'ssh' && recordingType !== 'desktop') {
     throw new Error(
-      `Invalid recording type: ${recordingType}, should be 'ssh' or 'desktop'`
+      `Invalid query parameter recordingType: ${recordingType}, should be 'ssh' or 'desktop'.`
+    );
+  }
+  if (isNaN(durationMs) || !Number.isInteger(durationMs)) {
+    throw new Error(
```
hatched, line 43: Typically I prefer to have components gracefully fail and handle these errors internally instead of throwing. I see above that this component is already throwing so you might want to do this as a follow-up.

ibeckermayer: Good call, I fixed in that in the PR for `isaiah/desktop-session-recording` and will integrate this with that change.

packages/teleport/src/Player/DesktopPlayer.tsx
ibeckermayer: Correct me if I'm wrong, but I don't think the frame rate is that straight forward in the browser (hence the need for [`requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame)).

packages/teleport/src/lib/tdp/playerClient.ts
ibeckermayer: We could, although I'm not sure what we'd want to do if we caught it besides throw it.

packages/teleport/src/Player/DesktopPlayer.tsx
hatched: You're right, ignore me 😄

ibeckermayer: Well maybe it would be smoother if we synchronized it using `requestAnimationFrame`.

packages/teleport/src/Player/DesktopPlayer.tsx
```diff
@@ -31,19 +34,28 @@ export const DesktopPlayer = ({
...
+  playerClient: PlayerClient;
+  durationMs: number;
+}) => {
+  const { playerClient, durationMs } = props;
+
+  const toHuman = (currentMs: number) => {
+    const date = new Date(null);
+    date.setMilliseconds(currentMs);
+    return date.toISOString().substr(14, 5);
+  };
```
kimlisa, line 52: couple things:

1) my editor says `substr` is deprecated
2) `const date = new Date(currentMs)` is same as `date.setMilliseconds(currentMs)`
2) isn't `substr(14,5)` fragile? i looked up [toISOString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString) and it says string can be 24 or 27 characters long. Instead i think this should be the equivalent, we have a new func called [dateToUtc](https://github.com/gravitational/webapps/blob/master/packages/shared/services/loc/loc.ts#L56), yassine added a new library called `date-fns` which you can pull out `format`:

```
format(dateInUTC, 'mm:ss') 
```

packages/teleport/package.json
```diff
@@ -19,6 +19,7 @@
   "dependencies": {
     "@gravitational/design": "1.0.0",
     "@gravitational/shared": "1.0.0",
+    "base64-arraybuffer": "^1.0.2",
```
kimlisa, line 22: are we sure we need this? can you take something out of here: https://github.com/gravitational/webapps/blob/master/packages/shared/utils/base64url.ts 

i feel like we have multiple ways of decode/encoding base64

ibeckermayer: I don't think that works because the strings are "standard base64" encoded, not "base64 url" encoded.

kimlisa: yes but if you look at the code, it looks like we can just take out the two first lines of code?
```ts
export function base64urlToBuffer(baseurl64String: Base64urlString): ArrayBuffer {
  // Base64url to Base64
  const padding = '=='.slice(0, (4 - (baseurl64String.length % 4)) % 4);
  const base64String = baseurl64String.replace(/-/g, '+').replace(/_/g, '/') + padding;

  // Base64 to binary string
  const str = atob(base64String);

  // Binary string to buffer
  const buffer = new ArrayBuffer(str.length);
  const byteView = new Uint8Array(buffer);
  for (let i = 0; i < str.length; i++) {
    byteView[i] = str.charCodeAt(i);
  }
  return buffer;
}
```

packages/teleport/src/Player/DesktopPlayer.tsx
ibeckermayer: > looked up [toISOString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString) and it says string can be 24 or 27 characters long.

Thanks, I was just aping a stack overflow suggestion, but it looks like its broken. Fixed using your suggestion.

packages/teleport/src/Player/DesktopPlayer.tsx
ibeckermayer: Updated this issue to include that suggestion: https://github.com/gravitational/webapps/issues/579

packages/teleport/src/Player/Player.tsx
```diff
@@ -36,23 +36,42 @@ export default function Player() {
...
   ) as RecordingType;
+  const durationMs = parseFloat(
+    getUrlParameter('durationMs', useLocation().search)
+  );
+
+  const validRecordingType =
+    recordingType === 'ssh' || recordingType === 'desktop';
+  const validDurationMs =
+    recordingType !== 'desktop' || // durationMs only needed for desktop recordings
+    (!isNaN(durationMs) && Number.isInteger(durationMs));
```
kimlisa, line 47: i think `isNan` and `Number.isInteger` seems to be doing the same thing? also, we are `parsing for float` but then check if its an integer. couple of questions:

1) do we expect an integer? b/c there is `parseInt`
2) are we aware that parseXXX parses beginning numbers but ignores the rest ie `parseXXX("23something") = 23`? if not, i suggest taking look at `Number(string|number)`

packages/teleport/src/Recordings/RecordingsList.tsx
```diff
@@ -112,7 +112,13 @@ const renderPlayCell = (
     );
   }
 
-  const url = cfg.getPlayerRoute({ clusterId, sid }, { recordingType });
+  const url = cfg.getPlayerRoute(
+    { clusterId, sid },
+    {
+      recordingType,
+      durationMs: recordingType === 'desktop' ? duration : undefined,
```
kimlisa, line 119: when we want to explicitly mean nothing, i think it's better to use `null`

packages/teleport/src/config.ts
```diff
@@ -268,9 +268,15 @@ const cfg = {
...
   getPlayerRoute(params: UrlPlayerParams, search: UrlPlayerSearch) {
-    return `${generatePath(cfg.routes.player, { ...params })}?recordingType=${
-      search.recordingType
-    }`;
+    let route = `${generatePath(cfg.routes.player, {
+      ...params,
+    })}?recordingType=${search.recordingType}`;
+
+    if (search.durationMs) {
+      route += `&durationMs=${search.durationMs}`;
```
kimlisa, line 276:
```suggestion
      route = `${route}&durationMs=${search.durationMs}`;
```

packages/teleport/src/config.ts
```diff
@@ -268,9 +268,15 @@ const cfg = {
   },
 
   getPlayerRoute(params: UrlPlayerParams, search: UrlPlayerSearch) {
-    return `${generatePath(cfg.routes.player, { ...params })}?recordingType=${
-      search.recordingType
-    }`;
+    let route = `${generatePath(cfg.routes.player, {
+      ...params,
+    })}?recordingType=${search.recordingType}`;
```
kimlisa, line 273: rant, prettier and its 80 col limit
```suggestion
   let route = generatePath(cfg.routes.player, {...params})
   route = `${route}?recordingType=${search.recordingType}`;
```

packages/teleport/src/lib/tdp/playerClient.ts
```diff
@@ -34,6 +39,19 @@ export class PlayerClient extends Client {
...
   }
 
+  // Overrides Client implementation.
+  processMessage(buffer: ArrayBuffer) {
+    const json = JSON.parse(this.textDecoder.decode(buffer));
+    if (json.message === 'end') {
+      this.emit(PlayerClientEvent.SESSION_END);
+      return;
+    }
+    let ms = json.ms;
```
kimlisa, line 49:
```suggestion
    const ms = json.ms;
```

packages/teleport/src/Recordings/RecordingsList.tsx
ibeckermayer: I think `undefined` is more appropriate here, because `durationMs` is optional and my intention is to say "don't define this thing".

packages/shared/utils/base64.ts
```diff
@@ -17,14 +17,13 @@
 export type Base64urlString = string;
 
 // Sourced from https://github.com/github/webauthn-json
-export function base64urlToBuffer(
-  baseurl64String: Base64urlString
-): ArrayBuffer {
-  // Base64url to Base64
+// Converts Base64url to Base64
+function base64urlToBase64(baseurl64String: Base64urlString): string {
```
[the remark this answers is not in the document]
ibeckermayer: Is it worth writing our own test file? It seems like a lot of additional work to reinvent the wheel. Why not just use this optimized and tested library?

https://github.com/niklasvh/base64-arraybuffer

packages/teleport/src/Player/DesktopPlayer.tsx
```diff
@@ -6,13 +6,18 @@ import { PngFrame, ClientScreenSpec } from 'teleport/lib/tdp/codec';
 import cfg from 'teleport/config';
 import { getAccessToken, getHostName } from 'teleport/services/api';
 import TdpClientCanvas from 'teleport/components/TdpClientCanvas';
+import { throttle } from 'lodash';
+import { dateToUtc } from 'shared/services/loc';
+import { format } from 'date-fns';
```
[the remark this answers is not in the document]
ibeckermayer: Thanks, I always neglect this. Is it documented somewhere in Slab?

packages/teleport/src/Player/Player.tsx
ibeckermayer: I'm just trying to check that it's a number with no alphabetic or other characters and no decimal place. I think `Number.isInteger(Number(durationMs))` gives me what I want.

packages/shared/utils/base64.ts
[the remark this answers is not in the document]
kimlisa: makes sense, but we already have our own various implementation of `encode(buffer)` and i think it's because it was faster the way we do it, and we were developing it around the same time so now we have like two versions. We should've also tested it but we didn't (our bad :sweat_smile: ). 

We are bringing in this library just b/c we need the `decode` as far as i can tell so it feels weird to use a library now. We can get inspiration from their test files and incorporate it to our own (that's what i would've done). Or if we really want to use this library, then we use this and clean up our implementation of it. But in my experience here, something as small as this, we just copy the source code and credit the person.

packages/teleport/src/Player/DesktopPlayer.tsx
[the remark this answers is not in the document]
kimlisa: no, i just been advising as i go. we probably should though, do we have a doc somewhere that i'm not aware of?

packages/shared/utils/base64.ts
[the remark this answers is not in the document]
ibeckermayer: I went ahead and confirmed which are the fastest by running some tests on jsbench.me:
- The `base64-arraybuffer` library's implementation of base64 --> arraybuffer is ~50% faster: https://jsbench.me/vkkz8ly9os/1
- The fromCharCode (our current implementation) of arraybuffer --> Base64 is ~50% faster: https://jsbench.me/krkz8mf04o/1

I'll just add the source and test files from `base64-arraybuffer`, swap out our better implementation for `encode`, and use their test suite.

packages/teleport/src/Player/Player.tsx
```diff
@@ -31,26 +31,47 @@ import { RecordingType } from 'teleport/services/recordings';
...
+    loc.search
   ) as RecordingType;
+  const durationMs = Number(getUrlParameter('durationMs', loc.search));
+
+  const validRecordingType =
+    recordingType === 'ssh' || recordingType === 'desktop';
+  // True if (recordingType !== 'desktop') because
+  // durationMs is only needed for desktop recordings.
+  const validDurationMs =
+    recordingType !== 'desktop' || Number.isInteger(durationMs);
```
kimlisa, line 47: based on the comment `durationMs is only needed for desktop recordings`, shouldn't the condition then be:
```suggestion
    recordingType == 'desktop' && Number.isInteger(durationMs);
```
then we don't need the comment b/c the condition itself is self documenting?

i also think `validDurationMs` is misleading (or error prone depending on how you look at it) b/c it returns true for `recordingType !== 'desktop'` but doesn't check if its a valid integer.

kimlisa (approved):
after addressing one last comment

packages/teleport/src/Player/Player.tsx
ibeckermayer: We want it to short circuit if `recordingType !== 'desktop'`. In that case, we're not using `durationMs` and so we don't care whether `Number.isInteger(durationMs) === true`.

ibeckermayer: Still terrible design there, see latest commit.

packages/teleport/src/Player/Player.tsx
```diff
@@ -31,26 +31,44 @@ import { RecordingType } from 'teleport/services/recordings';
 
 export default function Player() {
   const { sid, clusterId } = useParams<UrlPlayerParams>();
+  const loc = useLocation();
```
kimlisa, line 34:
```suggestion
  const {search} = useLocation();
```

packages/teleport/src/Player/Player.tsx
```diff
@@ -31,26 +31,44 @@ import { RecordingType } from 'teleport/services/recordings';
...
   const recordingType = getUrlParameter(
     'recordingType',
-    useLocation().search
+    loc.search
   ) as RecordingType;
+  const durationMs = Number(getUrlParameter('durationMs', loc.search));
+
+  const validRecordingType =
+    recordingType === 'ssh' || recordingType === 'desktop';
+  const validDurationMs = durationMs > 0 && Number.isInteger(durationMs);
```
kimlisa, line 44:
```suggestion
  const validDurationMs = Number.isInteger(durationMs) && durationMs > 0;
```

kimlisa (approved):

PR #578, opened 2022-02-02
Fix typo in makeCluster

rudream:
## Purpose

Fixes a typo that was causing a console error and invalid `connectedText`


hatched (approved):
Thanks

kimlisa (changes requested):
lets add a test for this file, we normally do testing but this is probably an old one

rudream:
@kimlisa We don't do any testing for `makeXXX` files, the only way a test here would help is if we mocked the raw back-end response, which we don't do in any of our tests afaik.

kimlisa (approved):
just one item in mocked array is fine

Issue #566, opened 2022-01-24
`session.connect` event not displayed correctly in UI


atburke:
The `session.connect` event introduced in gravitational/teleport#9370 is displayed as "Unknown" in the audit log in the web UI. The type and description should be fixed to correctly describe the event.

![Screen Shot 2022-01-24 at 15 22 07](https://user-images.githubusercontent.com/31974658/150881771-efa5c4cf-0642-40e2-847b-03b5cb54263b.png)

PR #583, opened 2022-02-04
Add session connect event

kimlisa:
resolves https://github.com/gravitational/webapps/issues/566

before style fix:
image

after style fix:
image


zmb3:
The vertical alignment seems off on your screenshot. Is something wrong here?

kimlisa:
@zmb3 i noticed it too, currently looking into it, we did refactor our table few weeks ago so it must've been from that

zmb3 (approved):

PR #591, opened 2022-02-09
modifyResponse: Add optional space before /> to regex

ravicious:
## Purpose

6308951f548 changed the regexes used for CSRF & bearer tokens. The previous regexes were matching anything between the end of the `name` attribute and `/>`. The new regex assumes there's no space between the end of the `content` attribute and `/>`, but inspecting the `source` argument shows that there's indeed a space (at least on my machine).

https://github.com/gravitational/webapps/blob/6308951f54871a41a3d8516dad1855608a6182c9/packages/build/devserver/modifyResponse.js#L63-L66

This PR makes it so that the regex accepts an optional space before the end of the tag, just as the regex used to replace the tag in `target`:

https://github.com/gravitational/webapps/blob/6308951f54871a41a3d8516dad1855608a6182c9/packages/build/devserver/modifyResponse.js#L67-L74

### DIFF
--- a/packages/build/devserver/modifyResponse.js
+++ b/packages/build/devserver/modifyResponse.js
@@ -48,7 +48,7 @@ module.exports = function modifyIndexHtmlMiddleware(compiler) {
 
 function injectCsrf(source, target) {
   var value = source.match(
-    new RegExp(/<meta name="grv_csrf_token" content="[a-zA-Z0-9_.-=]*"\/>/)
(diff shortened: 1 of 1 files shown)


alex-kovoy:
There was an accidental commit of the wrong webassets to the Teleport master  that had old HTML template that broke the behavior. It's been addressed now https://github.com/gravitational/teleport/pull/10235 so this one can be closed.

alex-kovoy:
actually, this PR allows the devServer to work with both versions of HTML template. Nice.

PR #592, opened 2022-02-10
Allow the automatic toggle to be visible when adding new nodes in OSS version.

hatched:
Requires associated Teleport PR's to land before this change is functional.

Fixes: https://github.com/gravitational/teleport/issues/10000


kimlisa (approved):
lgtm, i'm sure you've done this but double checking before we merge lets:

1) test it actually works in both open source and enterprise
2) merge backend changes first
3) then merge frontend changes
4) update webassets

fyi, if we backport, and then we update webassets, we have to manually test that this updated webassets also works before merging the updated webassets PR (b/c backports are assumed to be always working). The backport double checking is true for any backports actually.

PR #601, opened 2022-02-15
Add more tests to Teleterm

gzdunek:
This PR adds more tests to parts that shouldn't change so much with design changes.

### DIFF
--- a/packages/teleterm/src/ui/DocumentCluster/ClusterResources/renderLabelCell.tsx
+++ b/packages/teleterm/src/ui/DocumentCluster/ClusterResources/renderLabelCell.tsx
@@ -1,0 +1,16 @@
+import React from 'react';
+import { makeLabelTag } from 'teleport/components/formatters';
+import { Cell } from 'design/DataTable';
+import { Label as SingleLabel } from 'design';
+import { Label } from 'teleport/types';
+
+export function renderLabelCell<T extends { labelsList: Label[] }>(props: T) {
+  const labels = props.labelsList.map(makeLabelTag);
+  const $labels = labels.map(label => (
+    <SingleLabel mb="1" mr="1" key={label} kind="secondary">
+      {label}
+    </SingleLabel>
+  ));
+
+  return <Cell>{$labels}</Cell>;
+}
--- a/packages/teleterm/src/ui/TabHost/TabHost.test.tsx
+++ b/packages/teleterm/src/ui/TabHost/TabHost.test.tsx
@@ -47,6 +47,7 @@ function getTestSetup() {
     unsubscribe: jest.fn(),
     findRootClusterByResource: jest.fn(),
     findCluster: jest.fn(),
+    findGateway: jest.fn()
   };
 
   const utils = render(
--- a/packages/teleterm/src/ui/TabHost/TabHost.test.tsx
+++ b/packages/teleterm/src/ui/TabHost/TabHost.test.tsx
@@ -1,0 +1,147 @@
+import { fireEvent, render } from 'design/utils/testing';
+import React from 'react';
+import { TabHost } from 'teleterm/ui/TabHost/TabHost';
+import { MockAppContextProvider } from 'teleterm/ui/fixtures/MockAppContextProvider';
+import { Document, DocumentsService } from 'teleterm/ui/services/docs';
+import { KeyboardShortcutsService } from 'teleterm/ui/services/keyboardShortcuts';
+import {
+  MainProcessClient,
+  TabContextMenuOptions,
+} from 'teleterm/mainProcess/types';
+import { getMockDocuments } from 'teleterm/ui/fixtures/mocks';
+import { ClustersService } from 'teleterm/ui/services/clusters';
+
+function getTestSetup() {
+  const mockedDocuments = getMockDocuments();
+
+  const keyboardShortcutsService: Partial<KeyboardShortcutsService> = {
+    subscribeToEvents() {},
+    unsubscribeFromEvents() {},
+  };
+
+  const mainProcessClient: Partial<MainProcessClient> = {
+    openTabContextMenu: jest.fn(),
+  };
+
+  const docsService: Partial<DocumentsService> = {
+    getDocuments(): Document[] {
+      return mockedDocuments;
+    },
+    getActive() {
+      return mockedDocuments[0];
+    },
+    close: jest.fn(),
+    open: jest.fn(),
+    closeOthers: jest.fn(),
+    closeToRight: jest.fn(),
+    openNewTerminal: jest.fn(),
+    swapPosition: jest.fn(),
+    duplicatePtyAndActivate: jest.fn(),
+    useState() {
+      return null;
+    },
+  };
+
+  const clustersService: Partial<ClustersService> = {
+    subscribe: jest.fn(),
+    unsubscribe: jest.fn(),
+    findRootClusterByResource: jest.fn(),
+    findCluster: jest.fn(),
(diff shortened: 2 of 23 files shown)


packages/teleterm/src/ui/TabHost/TabHost.test.tsx
```diff
@@ -0,0 +1,147 @@
...
+import { Document, DocumentsService } from 'teleterm/ui/services/docs';
+import { KeyboardShortcutsService } from 'teleterm/ui/services/keyboardShortcuts';
+import {
+  MainProcessClient,
+  TabContextMenuOptions,
+} from 'teleterm/mainProcess/types';
+import { getMockDocuments } from 'teleterm/ui/fixtures/mocks';
+import { ClustersService } from 'teleterm/ui/services/clusters';
+
+function getTestSetup() {
```
ravicious, line 14: Something is causing TabHost.test.tsx to output a lot of warning related to `DocumentTerminal`:

<details>
<summary>DocumentTerminal warning</summary>

```
  console.error
    Warning: An update to DocumentTerminal inside a test was not wrapped in act(...).
    
    When testing, code that causes React state updates should be wrapped into act(...):
    
    act(() => {
      /* fire events that update state */
    });
    /* assert on the output */
    
    This ensures that you're testing the behavior the user would see in the browser. Learn more at https://fb.me/react-wrap-tests-with-act
        in DocumentTerminal (created by DocumentTerminalContainer)
        in DocumentTerminalContainer (created by MemoizedDocument)
        in MemoizedDocument (created by TabHost)
        in div (created by styled.div)
        in styled.div (created by TabHost)
        in TabHost
        in AppContextProvider (created by MockAppContextProvider)
        in DndProvider (created by MockAppContextProvider)
        in MockAppContextProvider
        in StyleSheetManager (created by ThemeProvider)
        in ThemeProvider (created by ThemeProvider)
        in ThemeProvider (created by Providers)
        in Providers

      43 |       return [data, null] as [R, Error];
      44 |     } catch (err) {
    > 45 |       setState({
         |       ^
      46 |         ...state,
      47 |         status: 'error',
      48 |         statusText: err.message,

      at printWarning (node_modules/react-dom/cjs/react-dom.development.js:88:30)
      at error (node_modules/react-dom/cjs/react-dom.development.js:60:5)
      at warnIfNotCurrentlyActingUpdatesInDEV (node_modules/react-dom/cjs/react-dom.development.js:23284:7)
      at setState (node_modules/react-dom/cjs/react-dom.development.js:15656:9)
      at run (packages/teleterm/src/ui/useAsync.ts:45:7)
```
</details>

I didn't have much time to dive into this, but do you know what could cause this?

gzdunek: Right, they are caused by running async actions in `DocumentTerminal` (there are also errors about missing canvas API in test env). I changed the documents mock so it won't include terminal documents, testing of each document type will be done separately.

packages/teleterm/src/ui/TabHost/TabHost.test.tsx
```diff
@@ -0,0 +1,148 @@
...
+test('should swap tabs', () => {
+  const { getByTitle, docsService } = getTestSetup();
+  const documents = docsService.getDocuments();
+  const $firstTab = getByTitle(documents[1].title);
+  const $thirdTab = getByTitle(documents[3].title);
+
+  fireEvent.dragStart($thirdTab);
+  fireEvent.drop($firstTab);
+
+  expect(docsService.swapPosition).toHaveBeenCalledWith(2, 0); // not '3, 1' because doc.home is omitted
```
ravicious, line 147: This is just a quirk in our tests, because in a real app `doc.home` wouldn't be there in the first place, but `getMockDocuments` defines it, right?

packages/teleterm/src/ui/TabHost/TabHost.test.tsx
```diff
@@ -0,0 +1,148 @@
...
+  );
+
+  return {
+    ...utils,
+    docsService,
+    mainProcessClient,
+  };
+}
+
+test('should render documents without home document', () => {
```
ravicious, line 77: Do we use any convention when it comes to how tests are supposed to be named? I quickly grepped through the project and it seems that tests starting with "should" are uncommon and most of them start with a more direct form.

```
$ rg "^test" | wc -l
     288
$ rg "^test\('should" | wc -l
      14
```

Perhaps it would make sense to rephrase them? Just dropping the "should" should accomplish that, I think.

packages/teleterm/src/ui/TabHost/TabHost.test.tsx
ravicious: This makes me wonder if we should have a global fixture with documents like we do in `teleterm/ui/fixtures/mocks`. As far as I understand, here below is another tests which does something extraordinary because of how the fixtures are set up. I left another comment under it.

https://github.com/gravitational/webapps/blob/72a31a8c51b5fd5c14efb1f95f808819d788bdc2/packages/teleterm/src/ui/TabHost/TabHost.test.tsx#L147

I'm a fan of crafting the setup for each test by hand, even if that means a lot of boilerplate code. Though I understand that the setup for TabHost tests is quite complicated and we need to mock a lot of stuff. Maybe adding `mockedDocuments` as an arg to `getTestSetup` would allow us to set up each test so that it includes only what's absolutely necessary? If "should open context menu" operates on only a single document, it could have only that document in its mocked documents, for example.

Though again, I'm not bent on having it this way, just trying to understand if this is something that we usually do and if it's helping us in certain scenarios.

ravicious (approved):
I have some minor comments about the names of the tests and how mocks are set up, but other than that it looks good to me.

packages/teleterm/src/ui/TabHost/TabHost.test.tsx
gzdunek: I've tried to address your comment, global `getMockedDocuments()` is removed, I added local functions to create mocks, and whenever possible I'm taking from them only what is necessary.

packages/teleterm/src/ui/TabHost/TabHost.test.tsx
gzdunek: `DocumentService` adds `doc.home` on the first place internally, so `getMockDocuments()` reflects that behaviour.

packages/teleterm/src/ui/TabHost/TabHost.test.tsx
gzdunek: Thanks, done

packages/teleterm/src/ui/TabHost/TabHost.test.tsx
ravicious: Ah okay, makes sense!

packages/teleterm/src/ui/DocumentCluster/ClusterResources/renderLabelCell.tsx
```diff
@@ -0,0 +1,16 @@
...
+export function renderLabelCell<T extends { labelsList: Label[] }>(props: T) {
+  const labels = props.labelsList.map(makeLabelTag);
+  const $labels = labels.map(label => (
+    <SingleLabel mb="1" mr="1" key={label} kind="secondary">
+      {label}
+    </SingleLabel>
+  ));
+
+  return <Cell>{$labels}</Cell>;
+}
```
ravicious, line 16: Good stuff. 👍

ravicious (approved):

PR #605, opened 2022-02-17
Disable autocomplete on the SSH login input

zmb3:
This prevents password managers like 1password from attempting to
autocomplete the SSH login, which is not a standard password field.

Fixes gravitational/teleport#10356


zmb3:
I'm actually not able to reproduce the original behavior, so it would be good to get confirmation that this fixes the issue prior to merging. 

cc @benarent

benarent:
@zmb3 is there any way to get a build of this so I can test it? 

image

packages/shared/components/MenuSshLogin/MenuSshLogin.tsx
```diff
@@ -120,6 +120,7 @@ export const LoginItemList = ({ logins, onClick, onKeyPress }) => {
         type="text"
         autoFocus
         placeholder="Enter login name..."
+        autocomplete="off"
```
kimlisa, line 123: i was able to reproduce this (strangely only pop ups for me in console nodes list, but the change still applies b/c it uses same component)
```suggestion
        autoComplete="off"
```
with react, attributes need to be camel cased for it to work: https://reactjs.org/docs/dom-elements.html

kimlisa (approved):
after the minor change

PR #609, opened 2022-02-17
Add chrome as unsupported for U2F checks

kimlisa:
** requires backporting to `v8`

### Description
u2f api no longer works for chrome ([since this month](https://www.yubico.com/blog/google-chrome-u2f-api-decommission-what-the-change-means-for-your-users-and-how-to-prepare/)), and when users try to log in or 'create account', the error is obscure or hidden (shows up in console log) so it appears to do nothing. Firefox still supports u2f (could not find when they are ending it). 

This PR checks if users are on chrome, to give a better error message. Same error message is shown to safari users.

image

### DIFF
--- a/packages/teleport/src/components/FormNewCredentials/FormNewCredentials.tsx
+++ b/packages/teleport/src/components/FormNewCredentials/FormNewCredentials.tsx
@@ -210,11 +210,11 @@ export type Props = {
 
 function ErrorMessage({ message = '' }) {
   // quick fix: check if error text has U2F substring
-  const browserSupported = !message.includes('does not support U2F');
-  const showU2fErrorLink = browserSupported && message.includes('U2F');
+  const notSupportedErr = !message.includes('is not supported');
+  const showU2fErrorLink = notSupportedErr && message.includes('U2F');
 
   return (
-    <Alerts.Danger>
+    <Alerts.Danger style={{ wordBreak: 'break-word' }}>
       <div>
         {message}
         {showU2fErrorLink && (
--- a/packages/teleport/src/services/auth/auth.ts
+++ b/packages/teleport/src/services/auth/auth.ts
@@ -35,7 +35,7 @@ const auth = {
         'The U2F API for hardware keys is not supported by your browser. \
         Please notify your system administrator to update cluster settings \
         to use WebAuthn as the second factor protocol. In the meantime, \
-        you can use Firefox to log in.'
+        you can use Firefox to use this hardware key.'
       );
     }
   },
--- a/packages/teleport/src/services/auth/auth.ts
+++ b/packages/teleport/src/services/auth/auth.ts
@@ -30,13 +30,14 @@ import { DeviceType } from './types';
 
 const auth = {
   u2fBrowserSupported() {
-    if (window['u2f']) {
-      return null;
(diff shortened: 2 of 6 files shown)


xinding33:
@ibeckermayer makes a good point about not showing this warning to Firefox users. Let's make this case dependent and remove the warning for those users.

In the case where a user is using an unsupported browser, I do think that we should tell them to contact their administrator instead of _just_ telling them they can use Firefox. This will create internal pressure for customers to switch to WebAuthn.

To make sure we don't block users from using the product (and address the fact that Safari _technically_ didn't deprecate support), we can change the text to:

The U2F API for hardware keys is not supported by your browser. Please notify your system administrator to update cluster settings to use WebAuthn as the second factor protocol. In the meantime, you can use Firefox to log in.

packages/teleport/src/components/FormNewCredentials/FormNewCredentials.tsx
```diff
@@ -102,6 +102,14 @@ export default function FormNewCredentials(props: Props) {
               <Text typography="h2" mb={3} textAlign="center" color="light">
                 {title}
               </Text>
+              {attempt.status !== 'failed' && mfaType?.value === 'u2f' && (
+                <Alerts.Warning style={{ wordBreak: 'break-word' }}>
+                  The U2F API for hardware keys has been deprecated by your
+                  browser. Please notify your system administrator to update
+                  cluster settings to use WebAuthn as the second factor
+                  protocol.
+                </Alerts.Warning>
```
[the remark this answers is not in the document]
kimlisa: no longer applies

packages/teleport/src/services/auth/auth.ts
```diff
@@ -30,13 +30,14 @@ import { DeviceType } from './types';
...
 const auth = {
   u2fBrowserSupported() {
-    if (window['u2f']) {
-      return null;
+    if (!window['u2f'] || navigator.userAgent.includes('Chrome')) {
+      return new Error(
+        'The U2F API for hardware keys is not supported by your browser. \
+        Please notify your system administrator to update cluster settings \
+        to use WebAuthn as the second factor protocol. In the meantime, \
+        you can use Firefox to log in.'
```
xinding33, line 38: Looks good to me!

xinding33 (approved):
Warning/error message and behavior looks good to me!

packages/teleport/src/services/auth/auth.ts
ibeckermayer: nit: `In the meantime, you can use Firefox` or `In the meantime, you can use Firefox to use this hardware key`, since it's possible for this message to show up when the user is logged in (if they're first registering a key).

ibeckermayer (approved):
approved with a nitpick: https://github.com/gravitational/webapps/pull/609#discussion_r810350401

packages/teleport/src/services/auth/auth.ts
kimlisa: good catch, the language also does not make sense when "creating" account as well. How about this:

`In the meantime, Firefox can be used as an alternative.`

kimlisa: got a second opinion from zac, will go with your suggestion

PR #613, opened 2022-02-18
desktop per session mfa

ibeckermayer:
Currently only has webauthn implemented:

`useWebAuthn` was abstracted out into a hook that takes a `EventEmitterWebAuthnSender` as an argument:

```typescript
class EventEmitterWebAuthnSender extends EventEmitter {
  constructor() {
    super();
  }

  // eslint-disable-next-line @typescript-eslint/no-unused-vars
  sendWebAuthn(data: WebauthnAssertionResponse) {
    throw new Error('Not implemented');
  }
}
```
`tdp.Client` and `tty.Tty` now both extend `EventEmitterWebAuthnSender` rather than `EventEmitter`.

U2F per session mfa will not be supported for Desktop Access, as we are planning to sunset U2F support (https://github.com/gravitational/teleport/issues/10375).


packages/teleterm/webpack.renderer.dev.config.js
```diff
@@ -12,7 +12,7 @@ devCfg.devServer = {
     directory: path.join(__dirname, 'build/app/dist/renderer'),
     serveIndex: false,
   },
-  allowedHosts: 'auto',
+  allowedHosts: 'all',
```
ibeckermayer, line 15: This was helpful for testing webauthn stuff, since webauthn doesn't seem to like localhost and so I needed to use proxy.0.0.0.0.nip.io as the configured domain instead.

packages/teleport/src/DesktopSession/DesktopSession.tsx
```diff
@@ -56,7 +57,7 @@ export function DesktopSession(props: State) {
   if (fetchAttempt.status === 'failed') {
     alertText = fetchAttempt.statusText;
   } else if (tdpConnection.status === 'failed') {
-    alertText = tdpConnection.status;
+    alertText = tdpConnection.statusText;
```
hatched, line 60: it looks like `statusText` can be optional. Did you want to provide a default value in the event that it's undefined?

packages/teleport/src/lib/EventEmitterWebAuthnSender.ts
```diff
@@ -0,0 +1,15 @@
+import { EventEmitter } from 'events';
+import { WebauthnAssertionResponse } from 'teleport/services/auth';
+
+class EventEmitterWebAuthnSender extends EventEmitter {
+  constructor() {
+    super();
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
hatched, line 9: Curious why you've added this instead of just removing the arg?

packages/teleport/src/lib/tdp/client.ts
```diff
@@ -149,6 +154,24 @@ export default class Client extends EventEmitter {
     );
   }
 
+  handleMfaChallenge(buffer: ArrayBuffer) {
+    const mfaJson = this.codec.decodeMfaJson(buffer);
```
hatched, line 158: If this can throw an error as outlined in the comment on line 162 then shouldn't this be wrapped in a try/catch so that it can fail gracefully?

packages/teleport/src/lib/tdp/client.ts
```diff
@@ -149,6 +154,24 @@ export default class Client extends EventEmitter {
...
+    if (mfaJson.mfaType == 'n') {
+      this.emit(TermEventEnum.WEBAUTHN_CHALLENGE, mfaJson.jsonString);
+    } else {
+      // mfaJson.mfaType === 'u', or else decodeMfaJson would have thrown an error.
```
hatched, line 162: Maybe you should explicitly check for u as it's one of the options in the type and then have a failure conditional for any other values.

packages/teleport/src/lib/tdp/codec.ts
```diff
@@ -369,6 +397,18 @@ export default class Codec {
     return this._decodeStringMessage(buffer);
   }
 
+  // decodeMfaChallenge decodes a raw tdp MFA challenge message and returns it as a string (of a json).
+  // | message type (10) | mfa_type byte | message_length uint32 | json []byte
+  decodeMfaJson(buffer: ArrayBuffer): MfaJson {
+    let dv = new DataView(buffer);
+    const mfaType = String.fromCharCode(dv.getUint8(1));
+    if (mfaType !== 'n' && mfaType !== 'u') {
+      throw new Error(`invalid mfa type ${mfaType}, should be "n" or "u"`);
```
hatched, line 406: I'm not a fan of throwing as it'll break the app and as mentioned previously there isn't anything 'catching' this. Maybe it should instead return an error that's null otherwise.

packages/teleport/src/lib/tdp/client.ts
ibeckermayer: It's handled by the `try catch` now, as noted int he comment, there's no possibility of receiving anything else since the codec will throw an error if it's not 'n' or 'u'.

packages/teleport/src/lib/tdp/codec.ts
ibeckermayer: I don't understand why returning an error would make more sense. We want the app to "break" if this section breaks, in the sense that we want to alert the user that they won't be able to access the desktop session they requested.

packages/teleport/src/lib/tdp/codec.ts
```diff
@@ -344,20 +352,40 @@ export default class Codec {
     return buffer;
   }
 
+  // | message type (10) | mfa_type byte | message_length uint32 | json []byte
+  encodeMfaJson(mfaJson: MfaJson): Message {
+    const dataUtf8array = this.encoder.encode(mfaJson.jsonString);
+
+    const bufLen = 1 + 1 + 4 + dataUtf8array.length;
```
kimlisa, line 359: we should start consting these magic numbers

kimlisa (approved):

PR #621, opened 2022-02-24
If running in enterprise version then provide the enterprise download links

hatched:
- Change the file download link depending on the enterprise status of the deployment.
- Consolidate the links into a separate component.

Fixes: https://github.com/gravitational/teleport/issues/10341

### DIFF
--- a/packages/teleport/src/components/DownloadLinks/DownloadLinks.tsx
+++ b/packages/teleport/src/components/DownloadLinks/DownloadLinks.tsx
@@ -1,0 +1,38 @@
+import React from 'react';
+import { Box, Link } from 'design';
+import getDownloadLink from 'teleport/services/links';
+import useTeleport from 'teleport/useTeleport';
+
+export default function DownloadLinks({ version }: Props) {
+  const ctx = useTeleport();
+  const isEnterprise = ctx.isEnterprise;
+
+  return (
+    <Box>
+      <Link
+        href={getDownloadLink('mac', version, isEnterprise)}
+        target="_blank"
+        mr="2"
+      >
+        MacOS
+      </Link>
+      <Link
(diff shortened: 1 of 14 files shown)


zmb3 (approved):

packages/teleport/src/components/DownloadLinks/DownloadLinks.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
```
kimlisa, line 1: missing license header

kimlisa (approved):

packages/teleport/src/components/DownloadLinks/DownloadLinks.tsx
hatched: Thanks, added.

packages/teleport/src/components/DownloadLinks/DownloadLinks.tsx
```diff
@@ -0,0 +1,54 @@
...
+ * limitations under the License.
+ */
+
+import React from 'react';
+import { Box, Link } from 'design';
+import getDownloadLink from 'teleport/services/links';
+import useTeleport from 'teleport/useTeleport';
+
+export default function DownloadLinks({ version }: Props) {
+  const ctx = useTeleport();
```
rudream, line 23: We only care about `isEnterprise` from the context, so it is better to just add it as a prop to this component rather than retrieving the entire context. This will also prevent unit tests from breaking due to the unretrievable context. 

eg. for `AddDatabase` you would just get the `isEnterprise` flag from the ctx in `useDatabases` and then pass it down via `Databases` -> `AddDatabase` -> `DownloadLinks`

rudream (approved):
LGTM

PR #622, opened 2022-02-24
Edit api response for getXXX (resources)

kimlisa:
### Description
Making changes in response to PR https://github.com/gravitational/teleport/pull/10598, where the api responses are different for our resource getters. Adds extra fields (startKey and totalCount) that are currently not used but will be in future work.

### Counterpart PRs
https://github.com/gravitational/webapps.e/pull/122

### DIFF
--- a/packages/teleport/src/Main/Main.story.tsx
+++ b/packages/teleport/src/Main/Main.story.tsx
@@ -64,12 +64,12 @@ function useMainStory() {
     ctx.auditService.fetchEvents = () =>
       Promise.resolve({ events, startKey: '' });
     ctx.clusterService.fetchClusters = () => Promise.resolve(clusters);
-    ctx.nodeService.fetchNodes = () => Promise.resolve(nodes);
+    ctx.nodeService.fetchNodes = () => Promise.resolve({ nodes });
     ctx.sshService.fetchSessions = () => Promise.resolve(sessions);
-    ctx.appService.fetchApps = () => Promise.resolve(apps);
-    ctx.kubeService.fetchKubernetes = () => Promise.resolve(kubes);
-    ctx.databaseService.fetchDatabases = () => Promise.resolve(databases);
-    ctx.desktopService.fetchDesktops = () => Promise.resolve(desktops);
+    ctx.appService.fetchApps = () => Promise.resolve({ apps });
+    ctx.kubeService.fetchKubernetes = () => Promise.resolve({ kubes });
+    ctx.databaseService.fetchDatabases = () => Promise.resolve({ databases });
+    ctx.desktopService.fetchDesktops = () => Promise.resolve({ desktops });
     ctx.storeUser.setState(userContext);
     getFeatures().forEach(f => f.register(ctx));
     return ctx;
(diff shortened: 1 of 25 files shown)


packages/teleport/src/services/apps/types.ts
```diff
@@ -28,6 +28,12 @@ export interface App {
   awsConsole: boolean;
 }
 
+export interface AppsResponse {
```
ibeckermayer, line 31: Perhaps we should choose between `type` and `interface` and stick with it. Personally I prefer `type`, because interfaces allow you to do
```typescript
interface Window {
  title: string
}

interface Window {
  ts: TypeScriptAPI
}

const src = 'const a = "Hello World"';
window.ts.transpileModule(src, {});
``` 
which is confusing and needless.

packages/teleport/src/services/nodes/types.ts
```diff
@@ -32,3 +32,9 @@ export interface BashCommand {
   text: string;
   expires: string;
 }
+
+export interface NodesResponse {
```
ibeckermayer, line 36: See `interface` vs `type` comment above.

ibeckermayer (approved):

packages/teleport/src/services/apps/types.ts
hatched: I'd also prefer using `type` unless you REALLY need the extra features provided by interfaces

packages/teleport/src/Main/Main.story.tsx
```diff
@@ -64,12 +64,12 @@ function useMainStory() {
     ctx.auditService.fetchEvents = () =>
       Promise.resolve({ events, startKey: '' });
     ctx.clusterService.fetchClusters = () => Promise.resolve(clusters);
-    ctx.nodeService.fetchNodes = () => Promise.resolve(nodes);
+    ctx.nodeService.fetchNodes = () => Promise.resolve({ nodes });
     ctx.sshService.fetchSessions = () => Promise.resolve(sessions);
```
hatched, line 68: Why were `clusters` and `sessions` left out of this structure change?

packages/teleport/src/services/apps/makeApps.ts
```diff
@@ -29,8 +29,11 @@ export default function makeApp(json: any): App {
     awsConsole = false,
   } = json;
```
hatched, line 30: I understand that this wasn't done in this PR but the value here isn't `JSON` so the name is confusing. If you wouldn't mind updating it to be a bit more appropriate.

hatched (approved):

packages/teleport/src/services/apps/types.ts
kimlisa: lol, i forgot i preferred type too, i was mindlessly copying and pasting the nearest type :laughing:

packages/teleport/src/Main/Main.story.tsx
kimlisa: they are not affected by the change that is coming (yet)

packages/teleport/src/services/apps/makeApps.ts
kimlisa: this is the pattern we follow every where else so i am going to leave as is. i might be biased now since i'm so used to it, but to me it makes sense b/c json means `unformatted response from backend`

hatched: ahh ok, but yeah that's not really clear to anyone coming from the outside as JSON is formatted, and a string :D

Issue #626, opened 2022-02-25
Clipboard sharing can be broken by using keyboard shortcuts in-browser


ibeckermayer:
Clipboard sharing from the client (browser side) to the server (windows desktop side) is currently implemented by checking the client's clipboard in an `onmouseenter` listener. This system always works so long as the user is using their mouse to navigate around their browser and desktop, such as if they are in a desktop access tab, then use the mouse to click on another browser tab, copy something, and then navigate back to the window and right click + "paste". The fact that their mouse was forced to leave the canvas to click on another tab, and come back in in order to right click + "paste".

However there are ways to break this system by using keyboard shortcuts. For example, if the user is in a desktop access tab with a cursor in a text window (such as the Notepad app), then uses cmd + shift + ] to navigate to another tab (the macos shortcut in chromium to switch tabs, which isn't deactivated by `e.preventDefault()`), copies something there, and uses cmd + shift + [ to navigate back, the browser doesn't see that as an `onmouseenter` event, and so clipboards are never synced.

Another way to break it is to:
- Be in a desktop access tab in a with a cursor in a text window.
- Click to another tab and copy something.
- Click back to the desktop access tab, but keep the mouse hovering over the browser's tab element.
- Try to paste with ctrl + V.

Because the mouse never re-entered the canvas, `onmouseenter` never fired and the clipboards were never synced. The user will be able to enter text into the text window they were in (in the windows desktop), because that doesn't require the mouse entering the canvas, however their clipboard behavior will be off.

PR #628, opened 2022-02-25
Fix clipboard sync

ibeckermayer:
fixes https://github.com/gravitational/webapps/issues/626


kimlisa (approved):
i'm just curious, if we have `onfocus` listener, do we still need the `onmouse` listener?

ibeckermayer:
@kimlisa yes, for the edge cases where the user wants to copy something in the browser's URL bar, or something in the UI's top bar.

PR #631, opened 2022-02-28
Fix clipboard-having-an-initial-value bug

ibeckermayer:
A less-than-ideal fix for https://github.com/gravitational/teleport/issues/10649. These changes can be reverted once we implement https://github.com/gravitational/teleport/issues/8583, which will send us a signal that the user has fully authenticated, at which point we can begin sharing the clipboard without silent failures.


gzdunek (approved):

hatched (approved):
This looks good but I think it's worth adding some comments to the code as to why this is necessary, or even just a link to the original issue as it's a bit of a cryptic problem to solve.

ibeckermayer:
@hatched I added this comment https://github.com/gravitational/teleport/issues/8583#issuecomment-1054554688 on #8583 that links back to this description, which itself links to https://github.com/gravitational/teleport/issues/10649, which I think should be enough context to recall what's going on.

Issue #587, opened 2022-02-07
Maintain aspect ratio in playback recordings


ibeckermayer:
Update the Desktop Player to maintain the original aspect ratio of the desktop recording, in order to prevent situations like below:

![image](https://user-images.githubusercontent.com/13578537/152882625-17cab435-9005-430e-93f3-af161a1b75af.png)

PR #635, opened 2022-03-01
Maintain aspect ratio on Desktop Playback

ibeckermayer:
Previously, desktop sessions recorded at one aspect ratio and played at another could look insane (see the screenshot in https://github.com/gravitational/webapps/issues/587). These changes maintain the aspect ratio during playback.

#### Playback aspect ratio > recorded aspect ratio
Screen Shot 2022-02-28 at 20 50 00

#### Playback aspect ratio < recorded aspect ratio
Screen Shot 2022-02-28 at 20 49 26

#### RFD
Due to slight differences between the size available to the desktop session during recording, and the size available to the player, a session recorded and then played back in a browser window of precisely the same size may look something like this:

Screen Shot 2022-02-28 at 20 50 49

Note the black channels up and down the left and right size of the window, which are created by the same logic that creates the desirable, necessary black-space in the `Playback aspect ratio > recorded aspect ratio` screen shot above. I'm not ecstatic about how that looks, but am not sure precisely what to do about it (if anything).

closes https://github.com/gravitational/webapps/issues/587

### DIFF
--- a/packages/teleport/src/Player/DesktopPlayer.tsx
+++ b/packages/teleport/src/Player/DesktopPlayer.tsx
@@ -1,17 +1,14 @@
-import React, { useState, useEffect } from 'react';
+import React from 'react';
 
 import styled from 'styled-components';
-import { throttle } from 'lodash';
-import { dateToUtc } from 'shared/services/loc';
-import { format } from 'date-fns';
 
 import cfg from 'teleport/config';
-import { PlayerClient, PlayerClientEvent } from 'teleport/lib/tdp';
+import { PlayerClient } from 'teleport/lib/tdp';
 import { PngFrame, ClientScreenSpec } from 'teleport/lib/tdp/codec';
 import { getAccessToken, getHostName } from 'teleport/services/api';
 import TdpClientCanvas from 'teleport/components/TdpClientCanvas';
 
-import ProgressBar from './ProgressBar';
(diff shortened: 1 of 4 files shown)


hatched (approved):

packages/teleport/src/Player/DesktopPlayer.tsx
```diff
@@ -141,20 +94,40 @@ const useDesktopPlayer = ({
     canvas: HTMLCanvasElement,
     spec: ClientScreenSpec
   ) => {
+    const styledPlayer = canvas.parentElement;
```
zmb3, line 97: Do we need to run this logic on every PNG frame? Or can we instead run it once when we initialize the player and (optionally) run it in a window resize handler?

ibeckermayer: It's only being run when receive the `client screen spec` tdp message, so since we currently don't support a dynamic resize, it just happens once at the beginning. Once we do implement resize, this will also account for that.

zmb3: Oh! Github's UI confuses me, I looked up and it looked like this was inside `tdpCliOnPngFrame`, but it's not. My eyes always glance over the light blue bars. Sorry!

zmb3 (approved):

Issue #575, opened 2022-01-31
Handle errors in `DesktopPlayer`


ibeckermayer:
If we receive a TDP error during playback, we should show it as it would've been displayed to the user, but with the playback progress bar still on the screen so that it's more clear that it's part of the playback and not a playback error.

Issue #574, opened 2022-01-31
Polish `Player` edge case


ibeckermayer:
With our [`Player.tsx`](https://github.com/gravitational/webapps/blob/a25c9094d5e5f5ec305cc829b0398b91b08152d6/packages/teleport/src/Player/Player.tsx) now accounting for both an SSH player and a Desktop player, there is an edge case that needs to be accounted for:

- If the player url has a valid session ID for a `recordingType='ssh'` session but the `recordingType` param is set to `'desktop'` instead, the system fails silently (should show "Recording for this session is not available." but instead just shows a black screen).
  - **UPDATE:** the above is actually just a subset of the case where `DesktopPlayer` is shown (`recordingType=desktop`) but the recording is not available.

PR #638, opened 2022-03-02
desktop playback error handling

ibeckermayer:
Adds UI support for error messages during desktop session playback. There are three cases where we want to display an error message:

#### 1. There is a backend error during playback

In this case, I've crafted a new json message that may be sent back by the playback player (i.e. [here](https://github.com/gravitational/teleport/blob/270fd9c37c832102251f787215b5bc5d10e8a447/lib/web/desktop/playback.go#L217-L219)) of the form
```json
{
    "message": "error",
    "errorText": "some error text"
}
```

which is handled by the `PlayerClient` [here](https://github.com/gravitational/webapps/blob/250e948712575723063c5986b3ff2b8a8cdb7864/packages/teleport/src/lib/tdp/playerClient.ts#L49-L50) and [here](https://github.com/gravitational/webapps/blob/250e948712575723063c5986b3ff2b8a8cdb7864/packages/teleport/src/Player/DesktopPlayer.tsx#L144-L152).

Screen Shot 2022-03-02 at 17 59 03 (2)

#### 2. A TDP error happened during the session that's being played

Since I needed to create this error display regardless, I figured I may as well add support for displaying TDP error messages if the session ended with one. This was relatively simple to slot in, since the tdp `Client` which `PlayerClient` inherits [already knows how to handle them](https://github.com/gravitational/webapps/blob/250e948712575723063c5986b3ff2b8a8cdb7864/packages/teleport/src/lib/tdp/client.ts#L104-L106).

To do this, I [told our server to record errors](https://github.com/gravitational/teleport/blob/270fd9c37c832102251f787215b5bc5d10e8a447/lib/srv/desktop/windows_server.go#L875) and then [defined](https://github.com/gravitational/webapps/blob/250e948712575723063c5986b3ff2b8a8cdb7864/packages/teleport/src/Player/DesktopPlayer.tsx#L176-L181) the relevant [`TdpClientCanvas` `prop`](https://github.com/gravitational/webapps/blob/250e948712575723063c5986b3ff2b8a8cdb7864/packages/teleport/src/components/TdpClientCanvas/TdpClientCanvas.tsx#L128-L136) to handle them.

Screen Shot 2022-03-02 at 18 06 35 (2)

#### network error

If the websocket disconnects for some unknown reason, [we just display this error](https://github.com/gravitational/webapps/blob/250e948712575723063c5986b3ff2b8a8cdb7864/packages/teleport/src/Player/DesktopPlayer.tsx#L160-L174)

Screen Shot 2022-03-02 at 18 08 26 (2)

closes https://github.com/gravitational/webapps/issues/574
closes https://github.com/gravitational/webapps/issues/575
corresponds with https://github.com/gravitational/teleport/pull/10765


packages/teleport/src/Player/DesktopPlayer.tsx
```diff
@@ -113,14 +132,61 @@ const useDesktopPlayer = ({
...
+      playerClient.addListener(PlayerClientEvent.SESSION_END, () => {
+        setAttempt({ status: '' });
+      });
+
+      playerClient.addListener(
+        PlayerClientEvent.PLAYBACK_ERROR,
+        (err: Error) => {
+          setAttempt({
+            status: 'failed',
+            statusText: 'playback error: ' + err.message,
```
zmb3, line 149: "playback error" isn't quite clear enough to me. Suggestion:

```suggestion
            statusText: `There was an error while playing this session: ${err.message}`,
```

packages/teleport/src/Player/DesktopPlayer.tsx
```diff
@@ -113,14 +132,61 @@ const useDesktopPlayer = ({
...
+          });
+        }
+      );
+
+      return () => {
+        playerClient.nuke();
```
zmb3, line 155: Can this be called more than once?

packages/teleport/src/Player/DesktopPlayer.tsx
```diff
@@ -25,39 +27,50 @@ export const DesktopPlayer = ({
...
   });
 
+  const displayCanvas = attempt.status === 'success' || attempt.status === '';
+  const displayProgressBar = attempt.status !== 'processing';
+
   return (
     <>
       <StyledPlayer>
-        {!showCanvas && (
+        {attempt.status === 'processing' && (
```
zmb3, line 44: Should you make constants or a type or something that defines the valid statuses so we don't just have magic strings scattered around?

zmb3 (approved):

ravicious (approved):

PR #641, opened 2022-03-03
Simplify manual app joining process

mcbattirola:
This PR implements the suggestion made by @xinding33 in https://github.com/gravitational/teleport/issues/10340 for adding apps.

In the manual steps of adding an app, instead of asking the user to log in to generate a token, we use the token API to generate the token (just like with the automatic join script), removing steps 2 and 3.

The old 4 step process is kept as a backup and is shown in case the token generation fails. 

Screenshot from 2022-03-03 15-04-32

**It is pretty much identical to #619 but for adding apps instead of nodes.**

Here are the main changes in code:

- new storybook entries to represent the new possible states of the `AddApp` component
- component `Manually` now receives a token and attempt to print the steps with the token or the old manual steps when the attempt failed
- bash command creation functions moved to its own file, `packages/teleport/src/services/commands.ts`

I followed @kimlisa's [advice](https://github.com/gravitational/webapps/pull/619#discussion_r816152558)  and moved the bash creation methods outside of the hook. I was not sure where to place those methods, ended up moving them to `services` following the example of `packages/teleport/src/services/links.ts` that is somehow similar (a file for methods that don't necessarily interact with an external API).

Please let me know your thoughts


packages/teleport/src/Apps/AddApp/useAddApp.ts
```diff
@@ -14,9 +14,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import { useState } from 'react';
-import TeleportContext from 'teleport/teleportContext';
+import { useState, useEffect } from 'react';
 import useAttempt from 'shared/hooks/useAttemptNext';
+import { createAppBashCommand } from 'teleport/services/commands';
```
kimlisa, line 19: our service layer is for containing api calls and formatting the values that return from it (the `makeXXX` funcs).
utility funcs like this, if only used in one place for now, it can be defined closest to where it is used (defined in this file)

packages/teleport/src/Apps/AddApp/useAddApp.ts
```diff
@@ -27,12 +28,19 @@ export default function useAddApp(ctx: TeleportContext) {
   const [automatic, setAutomatic] = useState(isEnterprise);
   const [cmd, setCmd] = useState('');
   const [expires, setExpires] = useState('');
+  const [token, setToken] = useState('');
+
+  useEffect(() => {
```
kimlisa, line 33: automatic for apps works slightly different

how automatic was working before:
  - validate fields
  - fetch token

now it's doing:
  - fetch token
  - validate fields
  - fetch token

i think the proper way to handle this is on manual render, fetch token (so put this block in manual component)

mcbattirola: Thanks for pointing that out!

packages/teleport/src/Apps/AddApp/useAddApp.ts
mcbattirola: Thanks, I removed `packages/teleport/src/services/commands.ts` and moved `createNodeBashCommand` to the files they are used (but not inside the custom hook itself)

packages/teleport/src/Apps/AddApp/useAddApp.ts
```diff
@@ -56,4 +58,43 @@ export default function useAddApp(ctx: TeleportContext) {
...
+
+  const text = `sudo bash -c "$(curl -fsSL '${bashUrl}')"`;
+
+  return {
+    text,
+    expires,
+  };
+}
+
+export function createNodeBashCommand(node: NodeToken): BashCommand {
```
kimlisa: don't think you meant to include this here

kimlisa: ^ nvm lol i was going by commits

packages/teleport/src/Apps/AddApp/Manually/Manually.tsx
```diff
@@ -14,72 +14,71 @@
...
+  attempt,
 }: Props) {
   const { hostname, port } = window.document.location;
   const host = `${hostname}:${port || '443'}`;
   let tshLoginCmd = `tsh login --proxy=${host}`;
 
+  useEffect(() => {
+    createToken();
+  }, []);
+
```
kimlisa, line 49: er actually, this causes a problem with our automatic tab b/c the token is shared between the two tabs. the auto tab will create a script with empty name and url (which users have to fill in prior to getting the script).

we should probably create the command after validation if token is already available

kimlisa: what we can do is:

- create the token as before (move it back to the custom hook)
- in auto: after validation, create and render the command

mcbattirola: Hi Lisa, thanks for noticing this!

So this is what I did:
- moved useEffect back to hook
- added a new state variable `showCmd` to `Automatic` that start false and is set to true only when the form is valid
- only shows the script (and change the texts) when should `showCmd` is true (instead of when `cmd` exists, the way it used to be fone)

Commit: https://github.com/gravitational/webapps/pull/641/commits/fad2aa3e66e262f524b827b064d4772c09f4a150

packages/teleport/src/Apps/AddApp/Automatically/Automatically.tsx
```diff
@@ -34,13 +34,15 @@ export default function Automatically(props: Props) {
...
   const [uri, setUri] = React.useState('');
+  const [showCmd, setShowCmd] = React.useState(false);
 
   function handleCreate(validator: Validator) {
     if (!validator.validate()) {
       return;
     }
 
     props.onCreate(name, uri);
+    setShowCmd(true);
```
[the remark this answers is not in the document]
mcbattirola: Hmmmm, I see, so `Generate Scripts` just uses the token that is already fetched and builds the command with the app name and URL, while `Regenerate` will fetch a new token, right?

I like it, but we still have one problem, the `cmd` is generated in the hook when, and it uses the app name and URL.

We can make a method in the hook, something like, `setCmdParams(appName, appUri)` that updates the `cmd` without fetching a new token. What do you think?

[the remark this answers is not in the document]
kimlisa: hm, i think it's better, that we take `createAppBashCommand` out of the hook (so hook is only about fetching token), and move it closer to where it's really needed which is in `Automatic`

would this simplify things?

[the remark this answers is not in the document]
mcbattirola: Hmmm yes, I think it would 

We can make `Automatically` receives the token and generate the command when the form is submitted.

This way the hook only worries about fetching and storing the token and it's expiry date.

https://github.com/gravitational/webapps/pull/641/commits/a79399831d59b3e30e2c6dbaf3c407697ae30917 (there is a lot of changes, but mostly it is moving command methods and its tests to other files)

We still need a `useEffect` to listen to new token generations and create a new command:

```jsx
React.useEffect(() => {
    if (!name || !uri) {
      return;
    }

    const cmd = createAppBashCommand(token, name, uri);
    setCmd(cmd);
  }, [token]);
```

(added on a following commit, https://github.com/gravitational/webapps/pull/641/commits/e6875c1e2c3691b1c260539ccc10153e2b4c882c)

But I think this is much better

Please let me know what you think

packages/teleport/src/Apps/AddApp/Automatically/Automatically.tsx
```diff
@@ -28,27 +28,53 @@ import Validation, { Validator } from 'shared/components/Validation';
...
+  const [showCmd, setShowCmd] = React.useState(false);
 
-  function handleCreate(validator: Validator) {
+  React.useEffect(() => {
+    if (!name || !uri) {
+      return;
+    }
+
+    const cmd = createAppBashCommand(token, name, uri);
+    setCmd(cmd);
```
kimlisa, line 47:
```suggestion
    if (name && uri) {
      const cmd = createAppBashCommand(token, name, uri);
      setCmd(cmd);
    }
```

packages/teleport/src/Apps/AddApp/Automatically/Automatically.tsx
```diff
@@ -28,27 +28,53 @@ import Validation, { Validator } from 'shared/components/Validation';
...
+import cfg from 'teleport/config';
 
 export default function Automatically(props: Props) {
-  const { cmd, onClose, attempt, expires } = props;
+  const { onClose, attempt, expires, token } = props;
 
   const [name, setName] = React.useState('');
   const [uri, setUri] = React.useState('');
+  const [cmd, setCmd] = React.useState('');
+  const [showCmd, setShowCmd] = React.useState(false);
```
kimlisa, line 39: don't think we need this boolean anymore? can we rely on the falsey/truthy value of `cmd`?

packages/teleport/src/Apps/AddApp/Manually/Manually.tsx
```diff
@@ -88,10 +83,76 @@ export default function Manually({
...
+  </Box>
+);
+
 type Props = {
   onClose(): void;
   isEnterprise: boolean;
   version: string;
   user: string;
   isAuthTypeLocal: boolean;
+  token: string;
```
kimlisa, line 154:
```suggestion
  token: State['token'];
```

kimlisa (approved):

PR #642, opened 2022-03-03
Fix Table Regression Bugs

### LINKED ISSUE #629: Table refactor regressions and other minor issues
#### Table regressions
Most issues had to do with searching, checkout previous versions of teleport for comparison

audit log:
- [ ] cannot search by date as seen in column

session recording:
- [ ] cannot search by duration

active sessions:
- [ ] looks like we added a search bar, but it barely works

node list:
- [ ] searching by keyword "tunnel", when there are tunnels, does not work

recovery table:
- [ ] table `device successfully enrolled`, when hovering, the row without the `remove` btn does not highlight entire row

#### Others
Just some other minor stuff i noticed

trusted clusters
- [ ] no tooltip when add button is disabled like others
 
support links:
- [ ] the troubleshooting links does not go to the proper page anymore, need to be updated

rudream:
## Purpose

This PR resolves #629 

Fixes some table regressions and other improvements

## Implementation

- Added `searchableProps` to Table config to allow searching through the pieces of data that are part of the object but that aren't indexed by the table columns. 

- Added search functionality to arrays of objects and arrays of strings, this is useful in the case of searching through Active Sessions because you are now able to search by participants. 

- Added search functionality for booleans, eg. you can now search `tunnel` and get rows that have `tunnel: true`. 

### Searching through Active Sessions by duration and by users
searchingsessions2

### DIFF
--- a/packages/design/src/DataTable/types.ts
+++ b/packages/design/src/DataTable/types.ts
@@ -7,6 +7,9 @@ export type TableProps<T> = {
   pagination?: PaginationConfig;
   isSearchable?: boolean;
   searchableProps?: Extract<keyof T, string>[];
+  // customSearchMatchers contains custom functions to run when search matching.
+  // 'targetValue' prop will have to be uppercased for proper matching since
+  // the root matcher will uppercase the searchValue.
   customSearchMatchers?: MatchCallback<T>[];
   initialSort?: InitialSort<T>;
   fetching?: FetchingConfig;
--- a/packages/design/src/DataTable/types.ts
+++ b/packages/design/src/DataTable/types.ts
@@ -1,3 +1,5 @@
+import { MatchCallback } from 'design/utils/match';
+
 export type TableProps<T> = {
   data: T[];
   columns: TableColumn<T>[];
@@ -5,6 +7,7 @@ export type TableProps<T> = {
   pagination?: PaginationConfig;
   isSearchable?: boolean;
   searchableProps?: Extract<keyof T, string>[];
+  customSearchMatchers?: MatchCallback<T>[];
   initialSort?: InitialSort<T>;
   fetching?: FetchingConfig;
   showFirst?: (data: T[]) => T;
--- a/packages/design/src/DataTable/types.ts
+++ b/packages/design/src/DataTable/types.ts
@@ -4,6 +4,7 @@ export type TableProps<T> = {
   emptyText: string;
   pagination?: PaginationConfig;
   isSearchable?: boolean;
+  searchableProps?: Extract<keyof T, string>[];
   initialSort?: InitialSort<T>;
   fetching?: FetchingConfig;
   showFirst?: (data: T[]) => T;
--- a/packages/design/src/DataTable/useTable.ts
+++ b/packages/design/src/DataTable/useTable.ts
@@ -173,39 +173,35 @@ function searchAndFilterCb<T>(
   searchValue: string,
   propName: keyof T & string
 ) {
-  if (propName.toLocaleLowerCase().includes('date')) {
-    return displayDate(targetValue).includes(searchValue);
-  }
-  if (propName.toLocaleLowerCase().includes('time')) {
-    return displayDateTime(targetValue).includes(searchValue);
-  }
-  if (typeof targetValue === 'boolean') {
-    return (
-      propName.toLocaleLowerCase().includes(searchValue.toLocaleLowerCase()) &&
-      targetValue
-    );
+  const propNameLowerCased = propName.toLocaleLowerCase();
+  const searchValueLowerCased = searchValue.toLocaleLowerCase();
+
+  let targetValueArr = targetValue;
+
+  if (!Array.isArray(targetValue)) {
+    targetValueArr = [targetValue];
   }
 
-  // For searching through elements of an array
-  if (Array.isArray(targetValue)) {
-    return targetValue.some(item => {
-      if (typeof item === 'object') {
-        for (const key in item) {
-          if (
-            item[key]
-              .toLocaleLowerCase()
-              .includes(searchValue.toLocaleLowerCase())
-          ) {
-            return true;
-          }
+  return targetValueArr.some(item => {
+    if (typeof item === 'object' && item !== null && !Array.isArray(item)) {
+      for (const key in item) {
+        if (item[key].toLocaleLowerCase().includes(searchValueLowerCased)) {
+          return true;
         }
-      } else {
-        return item
-          .toLocaleLowerCase()
-          .includes(searchValue.toLocaleLowerCase());
       }
-    });
-  }
+    } else if (propNameLowerCased.includes('date')) {
+      return displayDate(item).includes(searchValueLowerCased);
+    } else if (propNameLowerCased.includes('time')) {
+      return displayDateTime(item).includes(searchValueLowerCased);
+    } else if (typeof item === 'boolean') {
+      return propNameLowerCased.includes(searchValueLowerCased) && item;
+    } else {
+      return item
+        .toString()
+        .toLocaleLowerCase()
+        .includes(searchValueLowerCased);
+    }
+  });
 }
 
 export type State<T> = Omit<
--- a/packages/design/src/DataTable/useTable.ts
+++ b/packages/design/src/DataTable/useTable.ts
@@ -1,5 +1,5 @@
 import { useEffect, useState } from 'react';
-import isMatch from 'design/utils/match';
+import isMatch, { MatchCallback } from 'design/utils/match';
 import { displayDate, displayDateTime } from 'shared/services/loc';
 import paginateData from './Pager/paginateData';
 import { TableProps, TableColumn } from './types';
@@ -10,6 +10,7 @@ export default function useTable<T>({
   pagination,
   showFirst,
   searchableProps,
+  customSearchMatchers = [],
   ...props
 }: TableProps<T>) {
   const [state, setState] = useState(() => {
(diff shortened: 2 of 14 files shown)


packages/design/src/DataTable/useTable.ts
```diff
@@ -172,14 +173,39 @@ function searchAndFilterCb<T>(
...
+  }
+  if (typeof targetValue === 'boolean') {
+    return (
+      propName.toLocaleLowerCase().includes(searchValue.toLocaleLowerCase()) &&
+      targetValue
+    );
+  }
+
+  // For searching through elements of an array
+  if (Array.isArray(targetValue)) {
```
kimlisa, line 190: this assumes the items are of type `string`, what happens if its other than string?

packages/design/src/DataTable/useTable.ts
```diff
@@ -172,14 +173,39 @@ function searchAndFilterCb<T>(
...
 ) {
-  if (propName === 'tags') {
-    return targetValue.some(item => {
-      return item.toLocaleLowerCase().includes(searchValue.toLocaleLowerCase());
-    });
-  }
   if (propName.toLocaleLowerCase().includes('date')) {
     return displayDate(targetValue).includes(searchValue);
   }
+  if (propName.toLocaleLowerCase().includes('time')) {
```
kimlisa, line 179: `propName.toLocaleLowerCase()` is getting called multiple times needlessly

packages/design/src/DataTable/useTable.ts
```diff
@@ -172,14 +173,39 @@ function searchAndFilterCb<T>(
...
+    return displayDateTime(targetValue).includes(searchValue);
+  }
+  if (typeof targetValue === 'boolean') {
+    return (
+      propName.toLocaleLowerCase().includes(searchValue.toLocaleLowerCase()) &&
+      targetValue
+    );
+  }
```
kimlisa, line 187: i'm not sure how i feel about this and others b/c we are depending on the propName to include a certain keyword, why did we go away from sending in custom functions?

packages/design/src/DataTable/useTable.ts
```diff
@@ -172,14 +173,39 @@ function searchAndFilterCb<T>(
...
+
+  // For searching through elements of an array
+  if (Array.isArray(targetValue)) {
+    return targetValue.some(item => {
+      if (typeof item === 'object') {
```
hatched, line 192: You'll want to be more robust on this check as...

```
>  typeof null === "object"
<- true
```

packages/design/src/DataTable/useTable.ts
```diff
@@ -172,14 +173,39 @@ function searchAndFilterCb<T>(
...
+          if (
+            item[key]
+              .toLocaleLowerCase()
+              .includes(searchValue.toLocaleLowerCase())
+          ) {
+            return true;
+          }
+        }
+      } else {
+        return item
```
hatched, line 203: This assumes it's a `string`, it appears that it could be possible that it's not as you're only validating the type of the value if it's _not_ an `array`. You might want to combine these checks so that you're not handling array values differently. You could do this by simply placing the `targetValue` into an array if it isn't one and then looping over each value with the same validation checks.

hatched (changes requested):

packages/design/src/DataTable/useTable.ts
rudream: The custom search functions led to too much repeated code

rudream:
### Changes

Updated `Troubleshooting` link to docs

image

packages/design/src/DataTable/useTable.ts
```diff
@@ -39,13 +40,32 @@ export default function useTable<T>({
...
+        if (isMatched) {
+          return true;
+        }
+      }
+
+      // No match found.
+      return false;
+    };
+  });
+
```
kimlisa, line 60: lol i know i wrote this, but it's unnecessary pre-optimization (define the function once) lets define it regularly

```
function searchAndFilterCb(targetValue: any, searchValue: string, propName: keyof T & string) {
     for (const matcher of customSearchMatchers) {
        const isMatched = matcher(targetValue, searchValue, propName);
        if (isMatched) {
          return true;
        }
      }

      // No match found.
      return false;
}
```

packages/teleport/src/components/NodeList/NodeList.tsx
```diff
@@ -96,6 +105,23 @@ export const renderAddressCell = ({ addr, tunnel }: Node) => (
...
+  if (
+    propName === 'tunnel' &&
+    targetValue &&
+    propName.includes(searchValue.toLocaleLowerCase())
+  ) {
+    console.log('reached here');
+    return true;
+  }
+
+  return false;
```
kimlisa, line 122:
```suggestion
  return propName === 'tunnel' && targetValue && propName.includes(searchValue.toLocaleLowerCase())
```

packages/design/src/utils/match/matchers.ts
```diff
@@ -0,0 +1,20 @@
+import { displayDate, displayDateTime } from 'shared/services/loc';
+import { MatchCallback } from './match';
+
+export const dateMatcher =
+  <T>(datePropNames: (keyof T & string)[]): MatchCallback<T> =>
+  (targetValue, searchValue, propName) => {
+    if (datePropNames.includes(propName)) {
+      return displayDate(targetValue).toLocaleLowerCase().includes(searchValue);
+    }
+  };
```
kimlisa, line 10: my brain cramped reading this lol, i would save shortcut return statements for when its short and easy to read. I also tend to favor using keyword `function` whenever i can b/c it clearly states it's a function and when it concerns typing, it's easier to reason:

```suggestion
export function dateMatcher<T>(
  datePropNames: (keyof T & string)[]
): MatchCallback<T> {
  return (targetValue, searchValue, propName) => {
    if (datePropNames.includes(propName)) {
      return displayDate(targetValue).toLocaleLowerCase().includes(searchValue);
    }
  };
}
```

packages/teleport/src/Sessions/SessionList/SessionList.tsx
```diff
@@ -98,3 +111,21 @@ type Props = {
...
 };
+
+function participantMatcher(
+  targetValue: any,
+  searchValue: string,
+  propName: keyof Session & string
+) {
+  if (propName === 'parties') {
+    return targetValue.some((participant: Participant) => {
+      if (participant.remoteAddr.toLocaleUpperCase().includes(searchValue)) {
```
kimlisa, line 122: on second thought, idk if we should be relying on the root searchValue transofrmation b/c someone can change it and then everything will break right? Lets also transform the search value in our matchers, and then in the root match only transform when about to match?

packages/teleport/src/Sessions/SessionList/SessionList.tsx
```diff
@@ -98,3 +111,21 @@ type Props = {
...
+        return true;
+      }
+
+      return participant.user.toLocaleUpperCase().includes(searchValue);
+    });
+  }
+  // No match.
+  return false;
```
kimlisa, line 130: do we really need this? i think it returns undefined if no return statement and that's a falsey right?

kimlisa (changes requested):

packages/teleport/src/Sessions/SessionList/SessionList.tsx
rudream: correct, and the `.some` would actually return `false`

packages/design/src/DataTable/types.ts
```diff
@@ -1,9 +1,13 @@
+import { MatchCallback } from 'design/utils/match';
+
 export type TableProps<T> = {
   data: T[];
   columns: TableColumn<T>[];
   emptyText: string;
   pagination?: PaginationConfig;
   isSearchable?: boolean;
+  searchableProps?: Extract<keyof T, string>[];
+  customSearchMatchers?: MatchCallback<T>[];
```
kimlisa, line 10: lets add back the comment explaining that searchValue comes transformed to lower case

packages/design/src/utils/match/index.ts
```diff
@@ -0,0 +1,6 @@
+import isMatch, { MatchCallback } from './match';
+import { dateMatcher, dateTimeMatcher } from './matchers';
+
+export default isMatch;
+export { dateMatcher, dateTimeMatcher };
+export type { MatchCallback };
```
kimlisa, line 6: what is `export type { MatchCallback };`?

```suggestion
export { dateMatcher, dateTimeMatcher, MatchCallback };
```

packages/teleport/src/components/NodeList/NodeList.tsx
```diff
@@ -96,6 +105,14 @@ export const renderAddressCell = ({ addr, tunnel }: Node) => (
   <Cell>{tunnel ? renderTunnel() : addr}</Cell>
 );
 
+function tunnelMatcher(
+  targetValue: any,
+  searchValue: string,
+  propName: keyof Node & string
+) {
+  return propName === 'tunnel' && targetValue && propName.includes(searchValue);
```
kimlisa, line 113: makes more sense to me even though they do same thing

```suggestion
  return propName === 'tunnel' && targetValue && searchValue.includes(propName);
```

kimlisa (approved):

packages/design/src/utils/match/index.ts
rudream: Re-exporting the type this way causes this error:

```
export 'MatchCallback' (reexported as 'MatchCallback') was not found in './match' (possible exports: default)
design/src/utils/match/index.ts
```

We can use [type-only exports](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#-type-only-imports-and-export) to fix this

packages/teleport/src/components/NodeList/NodeList.tsx
kimlisa: nvm, i am wrong

packages/design/src/utils/match/index.ts
kimlisa: ah got it thanks

hatched (approved):
Thanks for the updates! 👍

PR #644, opened 2022-03-03
Update default roles template

stevenGravy:
Set v5 as the default and change the default resource rules to more limited RBAC then everything.  Addresses https://github.com/gravitational/teleport/issues/10807

### DIFF
--- a/packages/teleport/src/Roles/templates/role.yaml
+++ b/packages/teleport/src/Roles/templates/role.yaml
@@ -16,10 +16,20 @@ spec:
     node_labels:
       '*': '*'
 
-    # RBAC rules for various resources within a cluster.
+    # RBAC rules for various resources within a cluster. This
+    # example provides access to the Audit Log and replaying a user's own sessions.
     rules:
-    - resources: ['*']
-      verbs: ['*']
+    - resources:
+      - event
+      verbs:
+      - list
+      - read
+    - resources:
+      - session
+      verbs:
+      - read
+      - list
+      where: contains(session.participants, user.metadata.name)
 
   # The 'deny' section can have settings that override their 'allow' counterparts
   # It uses the same format as the 'allow' section
@@ -30,4 +40,4 @@ spec:
       # Limits user credentials to 8 hours. After the time to live (TTL) expires,
       # users must re-login
       max_session_ttl: 8h0m0s
-version: v3
+version: v5
--- a/packages/teleport/src/Roles/templates/role.yaml
+++ b/packages/teleport/src/Roles/templates/role.yaml
@@ -42,6 +42,7 @@ spec:
       '*': '*'
     # Windows logins a user is allowed to use for desktop sessions.
     windows_desktop_logins:
+    - '{{internal.windows_logins}}'
     - developer
 
     # RBAC rules for various resources within a cluster. This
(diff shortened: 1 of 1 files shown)


stevenGravy:
This should be targeted for v9 since this is part of the Testing problems

packages/teleport/src/Roles/templates/role.yaml
```diff
@@ -6,20 +6,55 @@ spec:
   # This example defines an administrative role. It maps to Kubernetes "admin"
   # group and allows SSH to every node.
   allow:
+    # List of Kubernetes cluster users can access the k8s API
+    kubernetes_labels:
+      '*': '*'
     # This role is mapped to Kubernetes 'admin' group.
     kubernetes_groups: [admin]
-
+
```
zmb3, line 14: Extra whitespace here.

packages/teleport/src/Roles/templates/role.yaml
```diff
@@ -6,20 +6,55 @@ spec:
...
+    db_users:
+    - dba-admin
+    
+    # List of windows desktop access labels that users can open desktop sessions to
+    windows_desktop_labels:
+      '*': '*'
+    # Windows logins a user is allowed to use for desktop sessions.
+    windows_desktop_logins:
+    - '{{internal.windows_logins}}'
+    - Administrator
```
zmb3, line 42: If we didn't make `root` a default Unix login, then we probably shouldn't make `Administrator` a default Windows login.

stevenGravy: I changed to use developer for that and other examples

packages/teleport/src/Roles/templates/role.yaml
```diff
@@ -3,15 +3,15 @@ metadata:
...
+  # with typical developer credentials.
   allow:
     # List of Kubernetes cluster users can access the k8s API
     kubernetes_labels:
       '*': '*'
-    # This role is mapped to Kubernetes 'admin' group.
-    kubernetes_groups: [admin]
-    
+    # This role is mapped to Kubernetes 'developer' group.
+    kubernetes_groups: [developer]
```
benarent, line 13:
```suggestion
    kubernetes_groups:
    - '{{internal.kubernetes_groups}}'
    - developer
 ```

File: packages/teleport/src/Roles/templates/role.yaml [hunk 4]
```diff
@@ -3,15 +3,15 @@ metadata:
...
+
```
benarent, line 14:
```suggestion
    kubernetes_users:
    - '{{internal.kubernetes_users}}'
    - 'dev'
```

packages/teleport/src/Roles/templates/role.yaml
```diff
@@ -31,15 +31,14 @@ spec:
...
     db_users:
-    - dba-admin
+    - developer
     
     # List of windows desktop access labels that users can open desktop sessions to
     windows_desktop_labels:
       '*': '*'
     # Windows logins a user is allowed to use for desktop sessions.
     windows_desktop_logins:
-    - '{{internal.windows_logins}}'
```
benarent: I would keep {{internal}} it's a good time to educe the user on what it is.

stevenGravy: done for all applicable

zmb3 (approved):

benarent (approved):

benarent:
Can you also backport to the teleport-9 branch,

PR #648, opened 2022-03-04
smooth out progress bar

ibeckermayer:
Adds an interval for smoothing out the progress bar.

### DIFF
--- a/packages/teleport/src/Player/ProgressBar/ProgressBarDesktop.tsx
+++ b/packages/teleport/src/Player/ProgressBar/ProgressBarDesktop.tsx
@@ -116,7 +116,6 @@ export const ProgressBarDesktop = (props: {
       });
 
       return () => {
-        throttledUpdateCurrentTime.cancel();
         playerClient.nuke();
         stopProgress();
       };
--- a/packages/teleport/src/Player/ProgressBar/ProgressBarDesktop.tsx
+++ b/packages/teleport/src/Player/ProgressBar/ProgressBarDesktop.tsx
@@ -33,18 +33,24 @@ export const ProgressBarDesktop = (props: {
     isPlaying: true, // determines whether play or pause symbol is shown
   });
 
+  // updateCurrentTime is a helper function to update the state variable.
+  // It should be used within a setState, like
+  // setState(prevState => {
+  //   return updateCurrentTime(prevState, newTime)
+  // })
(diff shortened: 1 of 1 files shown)


packages/teleport/src/Player/ProgressBar/ProgressBarDesktop.tsx
```diff
@@ -34,54 +35,94 @@ export const ProgressBarDesktop = (props: {
...
-      playerClient.addListener(TdpClientEvent.TDP_ERROR, () => {
-        progressToEnd();
+      playerClient.addListener(TdpClientEvent.WS_CLOSE, () => {
+        stopProgress();
       });
 
       return () => {
         throttledUpdateCurrentTime.cancel();
         playerClient.nuke();
+        stopProgress();
```
kimlisa, line 125: line 123 no longer needed?

packages/teleport/src/Player/ProgressBar/ProgressBarDesktop.tsx
```diff
@@ -34,54 +35,94 @@ export const ProgressBarDesktop = (props: {
...
+                current: nextTimeMs,
+                time: toHuman(nextTimeMs),
+              };
+            } else {
+              stopProgress();
+              return {
+                ...prevState,
+                current: durationMs,
+                time: toHuman(durationMs),
+              };
```
kimlisa, line 70: can we re-use `updateCurrentTime` here?

kimlisa (approved):

packages/teleport/src/Player/ProgressBar/ProgressBarDesktop.tsx
ibeckermayer: It's still needed, that function is still called when we get events from the server (carrying TDP messages)

hatched (approved):

packages/teleport/src/Player/ProgressBar/ProgressBarDesktop.tsx
kimlisa: even when stopProgress calls it already? :thinking:

ibeckermayer: Oh no, I forgot that I'd added that. Thanks.

PR #650, opened 2022-03-07
Implement new version of command bar

ravicious:
# Todo

* [x] Implement picking the suggestions.
* [x] Implement hostname autocomplete for `tsh ssh`.
* [x] Implement launching the commands.
* [ ] Support `--cluster` arg in `tsh ssh` autocomplete.
* [x] Add tests.
* [ ] Remove behavior related to `goBack`.
* [x] Make sure regexes for SSH logins and hosts cover valid characters.
* [ ] Use local cluster URI for SSH login/node autocomplete.
* [ ] Implement autocomplete for `tsh proxy db`.

Possibly for later (not this PR)

* Take cursor index into account.
* Handle env vars before the command.
   * `FOO=bar tsh ssh` should still offer autocomplete work.
   * Perhaps this can be done when we implement a more sophisticated solution.

---

This PR so far implements only showing the suggestions in the UI, not actually picking them. This is what I'm going to work on next.

The biggest change compared to the previous command palette is that it was essentially a multi-stage UI which kept track of the "progress" through `picker` in `QuickInputService` state.

With autocomplete, the only state we need to keep track of is the input value. So I rewrote QuickInputService so that on each input change it just parses the whole input and then returns 

* autocomplete suggestions
* what picker to use
   * This will be important once I actually get to implementing picking the suggestions.

For autocomplete, this feels like a natural approach and this model will allow us to write a more sophisticated version in the next release.

Due to how everything was designed, I was able to reuse lots of code responsible for UI. `QuickInputService` and `quickPickers` are the two things that changed the most.

Best reviewed commit-by-commit.

### DIFF
--- a/packages/teleterm/src/ui/DocumentTerminal/useDocumentTerminal.ts
+++ b/packages/teleterm/src/ui/DocumentTerminal/useDocumentTerminal.ts
@@ -58,9 +58,19 @@ async function initState(
     docsService.update(doc.uri, { cwd, title: cwd });
   };
 
+  const removeInitCommand = () => {
+    if (doc.kind !== 'doc.terminal_shell') {
+      return;
+    }
+    // The initCommand has to be launched only once, not every time we recreate the document from
+    // the state.
+    docsService.update(doc.uri, { initCommand: undefined });
+  };
+
   ptyProcess.onOpen(() => {
     docsService.update(doc.uri, { status: 'connected' });
     refreshTitle();
+    removeInitCommand();
   });
 
   ptyProcess.onExit(() => {
@@ -92,6 +102,7 @@ function createCmd(doc: Doc): PtyCommand {
   return {
     kind: 'pty.shell',
     cwd: doc.cwd,
+    initCommand: doc.initCommand,
   };
 }
 
--- a/packages/teleterm/src/ui/QuickInput/useQuickInput.ts
+++ b/packages/teleterm/src/ui/QuickInput/useQuickInput.ts
@@ -19,15 +19,14 @@ import { useAppContext } from 'teleterm/ui/appContextProvider';
 import { useKeyboardShortcuts } from 'teleterm/ui/services/keyboardShortcuts';
 
 export default function useQuickInput() {
-  const { quickInputService: serviceQuickInput, workspacesService } =
-    useAppContext();
+  const { quickInputService, workspacesService } = useAppContext();
   workspacesService.useState();
   const documentsService =
     workspacesService.getActiveWorkspaceDocumentService();
-  const { visible, inputValue } = serviceQuickInput.useState();
+  const { visible, inputValue } = quickInputService.useState();
   const [activeSuggestion, setActiveSuggestion] = React.useState(0);
   const autocompleteResult = React.useMemo(
-    () => serviceQuickInput.getAutocompleteResult(inputValue),
+    () => quickInputService.getAutocompleteResult(inputValue),
     [inputValue]
   );
   const hasSuggestions =
@@ -35,7 +34,7 @@ export default function useQuickInput() {
 
   const onFocus = (e: any) => {
     if (e.relatedTarget) {
-      serviceQuickInput.lastFocused = new WeakRef(e.relatedTarget);
+      quickInputService.lastFocused = new WeakRef(e.relatedTarget);
     }
   };
 
@@ -55,7 +54,7 @@ export default function useQuickInput() {
     const suggestion = autocompleteResult.suggestions[index];
 
     setActiveSuggestion(index);
-    serviceQuickInput.pickSuggestion(
+    quickInputService.pickSuggestion(
       autocompleteResult.targetToken,
       suggestion
     );
@@ -67,15 +66,15 @@ export default function useQuickInput() {
     // If there are suggestions to show, the first onBack call should always just close the
     // suggestions and the second call should actually go back.
     if (visible && hasSuggestions) {
-      serviceQuickInput.hide();
+      quickInputService.hide();
     } else {
-      serviceQuickInput.goBack();
+      quickInputService.goBack();
     }
   };
 
   useKeyboardShortcuts({
     'focus-global-search': () => {
-      serviceQuickInput.show();
+      quickInputService.show();
     },
   });
 
@@ -100,9 +99,9 @@ export default function useQuickInput() {
     onBack,
     onEnter,
     onActiveSuggestion,
-    onInputChange: serviceQuickInput.setInputValue,
-    onHide: serviceQuickInput.hide,
-    onShow: serviceQuickInput.show,
+    onInputChange: quickInputService.setInputValue,
+    onHide: quickInputService.hide,
+    onShow: quickInputService.show,
   };
 }
 
--- a/packages/teleterm/src/ui/QuickInput/useQuickInput.ts
+++ b/packages/teleterm/src/ui/QuickInput/useQuickInput.ts
@@ -28,7 +28,6 @@ export default function useQuickInput() {
   );
   const hasSuggestions =
     autocompleteResult.kind === 'autocomplete.partial-match';
-  const picker = autocompleteResult.picker;
 
   const onFocus = (e: any) => {
     if (e.relatedTarget) {
@@ -47,8 +46,14 @@ export default function useQuickInput() {
     if (!hasSuggestions) {
       return;
     }
+
+    const suggestion = autocompleteResult.suggestions[index];
+
     setActiveSuggestion(index);
-    picker.onPick(autocompleteResult.suggestions[index]);
+    serviceQuickInput.pickSuggestion(
+      autocompleteResult.targetToken,
+      suggestion
+    );
   };
 
   const onBack = () => {
@@ -62,9 +67,17 @@ export default function useQuickInput() {
     },
   });
 
+  // Reset active suggestion when the suggestion list changes.
+  // We extract just the tokens and stringify the list to avoid stringifying big objects.
+  // See https://github.com/facebook/react/issues/14476#issuecomment-471199055
   useEffect(() => {
     setActiveSuggestion(0);
-  }, [picker]);
+  }, [
+    hasSuggestions &&
+      JSON.stringify(
+        autocompleteResult.suggestions.map(suggestion => suggestion.token)
+      ),
+  ]);
 
   return {
     visible,
--- a/packages/teleterm/src/ui/QuickInput/useQuickInput.ts
+++ b/packages/teleterm/src/ui/QuickInput/useQuickInput.ts
@@ -17,9 +17,18 @@ limitations under the License.
 import React, { useEffect } from 'react';
 import { useAppContext } from 'teleterm/ui/appContextProvider';
 import { useKeyboardShortcuts } from 'teleterm/ui/services/keyboardShortcuts';
+import {
+  AutocompleteResult,
+  AutocompletePartialMatch,
+} from 'teleterm/ui/services/quickInput/types';
 
 export default function useQuickInput() {
-  const { quickInputService, workspacesService } = useAppContext();
+  const {
+    quickInputService,
+    workspacesService,
+    clustersService,
+    commandLauncher,
+  } = useAppContext();
   workspacesService.useState();
   const documentsService =
     workspacesService.getActiveWorkspaceDocumentService();
@@ -47,10 +56,41 @@ export default function useQuickInput() {
 
   const onEnter = (index?: number) => {
     if (!hasSuggestions || !visible) {
-      documentsService.openNewTerminal(inputValue);
+      executeCommand(autocompleteResult);
       return;
     }
 
+    // Passing `autocompleteResult` directly to narrow down AutocompleteResult type to
+    // AutocompletePartialMatch.
+    pickSuggestion(autocompleteResult, index);
+  };
+
+  const executeCommand = (autocompleteResult: AutocompleteResult) => {
+    const { command } = autocompleteResult;
+
+    switch (command.kind) {
+      case 'command.unknown': {
+        documentsService.openNewTerminal(inputValue);
+        break;
+      }
+      case 'command.tsh-ssh': {
+        const { localClusterUri } = workspacesService.getActiveWorkspace();
+
+        commandLauncher.executeCommand('tsh-ssh', {
+          loginHost: command.loginHost,
+          localClusterUri,
+        });
+        break;
+      }
+    }
+
+    quickInputService.clearInputValueAndHide();
+  };
+
+  const pickSuggestion = (
+    autocompleteResult: AutocompletePartialMatch,
+    index?: number
+  ) => {
     const suggestion = autocompleteResult.suggestions[index];
 
     setActiveSuggestion(index);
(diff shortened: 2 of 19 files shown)


packages/teleterm/src/ui/commandLauncher.ts
```diff
@@ -117,6 +117,18 @@ const commands = {
...
+    displayName: 'tsh ssh',
+    description: 'Run shell or execute a command on a remote SSH node',
+    run() {},
+  },
+  'autocomplete.tsh-proxy-db': {
+    displayName: 'tsh proxy db',
+    description:
+      'Start local TLS proxy for database connections when using Teleport',
+    run() {},
+  },
```
ravicious, line 131: Not sure if this is the best place for these commands to live, especially since their `run` functions will always be noops I think. But since for now we only have two, I think we might keep them here.

packages/teleterm/src/ui/services/quickInput/quickInputService.ts
```diff
@@ -45,28 +49,34 @@ export class QuickInputService extends Store<State> {
...
+    );
   }
 
   state: State = {
-    picker: null,
     inputValue: '',
     visible: false,
   };
 
+  // TODO: There's no "back" in the new command bar. We can probably just remove this method and the
```
ravicious, line 77: I'll get rid of all the `TODO`s that don't have my username next to it before I merged this PR.

packages/teleterm/src/ui/services/quickInput/quickPickers.ts
```diff
@@ -101,6 +104,10 @@ export class QuickLoginPicker extends ClusterPicker {
     return items;
   }
 
+  getAutocompleteResult() {
+    return null;
+  }
+
```
ravicious, line 110: I left some of the old pickers in the code, they are not doing any harm ATM and we might just remove them in the future.

packages/teleterm/src/ui/services/quickInput/quickPickers.ts
```diff
@@ -130,14 +141,218 @@ export class QuickServerPicker extends ClusterPicker {
...
+      .filter(cmd => {
+        const completeMatchRegex = new RegExp(`^${cmd.displayName}\\b`);
+        return (
+          cmd.displayName.startsWith(input) ||
+          // `completeMatchRegex` handles situations where the `input` akin to "tsh ssh foo".
+          // In that case, "tsh ssh" is the matching command, even though
+          // `cmd.displayName` ("tsh ssh") doesn't start with `input` ("tsh ssh foo").
+          completeMatchRegex.test(input)
+        );
+      });
```
ravicious, line 196: Technically the var definition above could be extracted to a separate function, but we're not repeating this job anywhere and I like being able to read code top-to-bottom. 😶

packages/teleterm/src/ui/services/quickInput/types.ts
```diff
@@ -34,4 +36,17 @@ export type Item =
...
+  kind: T;
+  picker: QuickInputPicker;
+};
+
+export type AutocompletePartialMatch =
+  AutocompleteBase<'autocomplete.partial-match'> & { listItems: Item[] };
+
+export type AutocompleteNoMatch = AutocompleteBase<'autocomplete.no-match'>;
+
+export type AutocompleteResult = AutocompletePartialMatch | AutocompleteNoMatch;
```
ravicious, line 52: So initially this had like four or five different possible values. I changed my approach twice and what we're left with is clearly just a binary choice which we could represent trough just `null | AutocompletePartialMatch`. However, I'm not 100% sure how actually picking the suggestions will change this, but I'll be fine with changing it in the future.

Now that I think about it though, we'll always have to return the picker anyway. But then `listItems` can be just `null | Item[]`.

gzdunek (commented):
Looks good! I only left a few minor comments.

packages/teleterm/src/ui/services/quickInput/quickInputService.ts
```diff
@@ -89,6 +98,13 @@ export class QuickInputService extends Store<State> {
     });
   };
 
+  getAutocompleteResult = (input: string): AutocompleteResult => {
+    const autocompleteResult =
+      this.quickCommandPicker.getAutocompleteResult(input);
+
+    return autocompleteResult;
```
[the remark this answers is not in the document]
ravicious: Thanks, I was doing some logging in this function and forgot to clean it up.

packages/teleterm/src/ui/services/quickInput/quickPickers.ts
```diff
@@ -130,14 +141,218 @@ export class QuickServerPicker extends ClusterPicker {
...
+    return null;
+  }
 }
 
 export class QuickCommandPicker implements QuickInputPicker {
+  launcher: CommandLauncher;
+  pickerRegistry: Map<string, QuickInputPicker>;
+
+  constructor(launcher: CommandLauncher) {
+    this.launcher = launcher;
```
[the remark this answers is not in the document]
gzdunek: I meant [Parameter Properties ](https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties)

Your code is equal to this:
```typescript
export class QuickCommandPicker implements QuickInputPicker {
  constructor(private launcher: CommandLauncher) {}
```

packages/teleterm/src/ui/services/quickInput/types.ts
```diff
@@ -23,9 +23,11 @@ export type ItemNewCluster = Base<
...
 
+export type ItemSshLogin = Base<'item.ssh-login', string>;
+
 export type QuickInputPicker = {
-  onFilter(value: string): Item[];
   onPick(item: Item): void;
+  getAutocompleteResult(input: string): AutocompleteResult;
 };
 
 export type Item =
```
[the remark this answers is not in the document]
ravicious: I think in the context of autocomplete it's fine, maybe we could change it to 'Suggestion' or sth later on.

Is it okay if I keep it this way for now? I'm thinking that perhaps it's another one of those things that we'll change once I add support for actually picking suggestions.

packages/teleterm/src/ui/services/quickInput/quickPickers.ts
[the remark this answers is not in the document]
ravicious: Ah, that's cool, I'll update it. I think there's a couple of other places which could use this.

packages/teleterm/src/ui/services/quickInput/quickPickers.ts
```diff
@@ -130,14 +141,218 @@ export class QuickServerPicker extends ClusterPicker {
...
   }
+
+  getAutocompleteResult() {
+    return null;
+  }
 }
 
 export class QuickCommandPicker implements QuickInputPicker {
+  launcher: CommandLauncher;
+  pickerRegistry: Map<string, QuickInputPicker>;
```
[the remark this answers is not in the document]
ravicious: They don't, I'll add the private modifier and use parameter properties here.

packages/teleterm/src/ui/services/quickInput/quickPickers.ts
```diff
@@ -130,43 +141,209 @@ export class QuickServerPicker extends ClusterPicker {
...
 }
 
-export class QuickCmdPicker implements QuickInputPicker {
-  launcher: CommandLauncher;
+export class QuickCommandPicker implements QuickInputPicker {
+  private pickerRegistry: Map<string, QuickInputPicker>;
 
-  constructor(launcher: CommandLauncher) {
+  constructor(private launcher: CommandLauncher) {
     this.launcher = launcher;
```
gzdunek, line 154: I guess you forgot to remove these assignments 😄 (also in other places)

ravicious: Omg, sorry, I totally missed the fact that they're unnecessary. I thought it just gets rid of the class-level declarations but apparently I can also skip code in the constructor.

gzdunek (approved):

packages/teleterm/src/ui/services/quickInput/types.ts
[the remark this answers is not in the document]
ravicious: I'm going to change it to `Suggestion` in the upcoming commit tomorrow.

ravicious:
@gzdunek Added three new commits which implement actually picking the suggestions.

Next I'll add autocompleting the hostname.

packages/teleterm/src/ui/services/quickInput/quickPickers.ts
```diff
@@ -126,8 +138,8 @@ export class QuickCommandPicker implements QuickInputPicker {
 }
 
 export class QuickTshSshPicker implements QuickInputPicker {
-  // TODO: Make sure this regex is okay and covers valid usernames.
-  loginRegex = /^\s+([a-zA-Z0-9_-]*\b)?$/;
+  // Taken from https://unix.stackexchange.com/a/435120
+  sshLoginRegex = /^[a-z_]([a-z0-9_-]{0,31}|[a-z0-9_-]{0,30}\$)$/;
```
gzdunek, line 142: Let's make it `private`

packages/teleterm/src/ui/services/quickInput/quickInputService.ts
```diff
@@ -17,56 +17,55 @@ limitations under the License.
...
   inputValue: string;
   visible: boolean;
 };
 
 export class QuickInputService extends Store<State> {
-  quickCmdPicker: pickers.QuickCmdPicker;
-  quickLoginPicker: pickers.QuickLoginPicker;
-  quickDbPicker: pickers.QuickDbPicker;
-  quickServerPicker: pickers.QuickServerPicker;
+  quickCommandPicker: pickers.QuickCommandPicker;
```
gzdunek, line 30: `private` (same for line 31)

gzdunek (approved):
Looks very solid!

packages/teleterm/src/ui/services/quickInput/quickInputService.ts
ravicious: `lastFocused` on line 31 is used by `useQuickInput`, though I suspect we'll get rid of this property soon anyway as it's related to the `goBack` behavior which doesn't exist for this implementation of the command bar.

gzdunek (approved):

ravicious:
@gzdunek For now I made it so all commands are launched in a local shell, the upcoming commits will make it so that `tsh ssh` actually uses the correct document type.

packages/teleterm/src/ui/DocumentTerminal/useDocumentTerminal.ts
```diff
@@ -58,9 +58,19 @@ async function initState(
...
 
+  const removeInitCommand = () => {
+    if (doc.kind !== 'doc.terminal_shell') {
+      return;
+    }
+    // The initCommand has to be launched only once, not every time we recreate the document from
+    // the state.
+    docsService.update(doc.uri, { initCommand: undefined });
+  };
+
```
ravicious, line 69: It seems like a hack to me, but given the other functions in this hook I think we're fine for now. Once we add more stuff here it should probably get reorganized.

ravicious: Also, I understand now why you defined that one function inside a component or hook rather than outside. 🙊

ravicious: Ah, forgot to add that since launching a command means just adding a new document and then the side effect happen when the document gets rendered, this seems like the only way in which we can implement "Launch this command, but only once and not each time this doc is recreated".

I guess we could also have a boolean like `hasInitCommandBeenLaunched`, but for now we wouldn't even need that.

packages/teleterm/src/ui/QuickInput/useQuickInput.ts
```diff
@@ -46,8 +46,8 @@ export default function useQuickInput() {
     setActiveSuggestion(index);
   };
 
-  const onPickSuggestion = (index: number) => {
-    if (!hasSuggestions) {
+  const onEnter = (index?: number) => {
```
ravicious, line 49: Renamed it from `onPickSuggestion` to `onEnter`, because it's no longer about picking a suggestion, but kinda doing the next step in the command bar.

I wanted to have a name that isn't really specific to a key that's being used to perform the action. So initially I thought about `onProceed`, but then I figured out I may as well use `onEnter`. Let me know if you have a better idea for the name.

packages/teleterm/src/ui/DocumentTerminal/useDocumentTerminal.ts
gzdunek: Yeah, it looks like a hack, but it's fine for now

gzdunek (approved):

packages/teleterm/src/ui/services/quickInput/types.ts
```diff
@@ -31,16 +31,35 @@ export type AutocompleteToken = {
...
 export type AutocompleteResult = AutocompletePartialMatch | AutocompleteNoMatch;
+
+type CommandBase<T> = {
+  kind: T;
+};
+
+export type AutocompleteUnknownCommand = CommandBase<'command.unknown'>;
+
+export type AutocompleteTshSshCommand = CommandBase<'command.tsh-ssh'> & {
+  loginHost: string;
```
ravicious, line 60: This version doesn't support passing arbitrary options yet, I think I'll work on that today.

packages/teleterm/src/ui/commandLauncher.ts
```diff
@@ -15,15 +15,81 @@ limitations under the License.
...
+        const { rootClusterId, leafClusterId } =
+          routing.parseClusterUri(clusterUri).params;
+        serverUri = routing.getServerUri({
+          serverId: host,
+          rootClusterId,
+          leafClusterId,
+        });
+        serverHostname = host;
+      }
+      // TODO(ravicious): Handle failure due to incorrect host name.
```
ravicious, line 85: We haven't talked much about error handling for the command bar in general. Right now if the pty process exits, we just close the tab. I think we should just not close it if the exit code is a non-success code.

gzdunek: Agree

packages/teleterm/src/ui/services/workspacesService/workspacesService.ts
```diff
@@ -75,22 +76,30 @@ export class WorkspacesService extends ImmutableStore<WorkspacesState> {
   getWorkspaceDocumentService(
     clusterUri: string
   ): DocumentsService | undefined {
-    if (!this.documentsServicesCache.has(clusterUri)) {
+    // If the passed clusterUri is a leaf cluster URI, create a root cluster URI from it.
+    // Otherwise just use clusterUri.
+    const { leafClusterId, rootClusterId } =
```
gzdunek, line 81: I'm afraid that it creates inconsistency. In other methods `clusterUri` parameter means `rootClusterUri` and only this method would be different.
What would you say for using `routing.parseClusterUri` in `commandLauncher`? I can see it is used there anyway.

gzdunek (approved):

packages/teleterm/src/ui/services/workspacesService/workspacesService.ts
ravicious: Alright, I'll change it then and close my other PR. Let's do the parsing in `commandLauncher` and see if we're going to run into more places which would like to use leaf cluster URI as well.

ravicious: I like it even better, because I renamed the `clusterUri` arg that the command launcher gets to `localClusterUri`, which signifies that the URI can be a root or a leaf cluster URI. 👍

PR #651, opened 2022-03-07
Add keyboard support to `Connections` popover

gzdunek:
`KeyboardArrowsNavigation` has been implemented using `context` to allow iterating over the items which are placed on different DOM levels. It will be useful for example in `Identity` popup where on a single list we will have `FilterableList` items and Logout item.

```
| <searchbox>
| Cluster 1
| Cluster 2              <--- FilterableList
| Cluster 3
-----------
Logout                   <---- separate item
```


ravicious (approved):
Interesting approach with `addItem` in that hook, I think it's the first time I'm seeing something like this!

PR #653, opened 2022-03-07
Simplify manual db join

mcbattirola:
edit: https://github.com/gravitational/teleport/pull/10914 was merged, but with a different endpoint than what is used in this PR currently.

I decided to make a single endpoint to request join tokens with a given role (instead of an endpoint specifically for db/node/app). Because of that, I made a PR to update the current way join tokens are generated from the front end on #672. When #672 is merged, I'll update this one to follow the same pattern. Till then, this pr is blocked

edit2: #672, I've updated this PR to use the new join token service
______

This PR implements the suggestion made by @xinding33 in https://github.com/gravitational/teleport/issues/10340 for adding DBs.

In the manual steps of adding an app, instead of asking the user to log in to generate a token, we use the token API to generate the token (just like with the automatic join script), removing steps 2 and 3.

The old 5 step process is kept as a backup and is shown in case the token generation fails.

Note that it is not possible to generate a join token for a DB via webapi (only app and node). Here is the PR that adds it on Teleport: https://github.com/gravitational/teleport/pull/10914

It is pretty much identical to https://github.com/gravitational/webapps/pull/619 and https://github.com/gravitational/webapps/pull/641 but for adding **DBs** instead of nodes/apps.

Screenshot from 2022-03-07 15-02-29

main changes:

- added a new method to fetch tokens on `services/databases`
- added new hook `useAddDatabase`
- added new steps to `AddDatabase.tsx`
- tests


packages/teleport/src/Databases/AddDatabase/AddDatabase.test.tsx
```diff
@@ -16,23 +16,32 @@
...
+    ${'rds postgres'}         | ${'failed'}  | ${'teleport db start --token=[generated-join-token] --auth-server=localhost:443 --name=[db-name] --protocol=postgres --uri=[uri] --aws-region=[region]'}
+    ${'rds postgres'}         | ${'success'} | ${'teleport db start --token=some-token --auth-server=localhost:443 --name=[db-name] --protocol=postgres --uri=[uri] --aws-region=[region]'}
+    ${'cloud sql postgres'}   | ${'failed'}  | ${'teleport db start --token=[generated-join-token] --auth-server=localhost:443 --name=[db-name] --protocol=postgres --uri=[uri] --ca-cert=[instance-ca-filepath] --gcp-project-id=[project-id] --gcp-instance-id=[instance-id]'}
+    ${'cloud sql postgres'}   | ${'success'} | ${'teleport db start --token=some-token --auth-server=localhost:443 --name=[db-name] --protocol=postgres --uri=[uri] --ca-cert=[instance-ca-filepath] --gcp-project-id=[project-id] --gcp-instance-id=[instance-id]'}
+    ${'redshift'}             | ${'failed'}  | ${'teleport db start --token=[generated-join-token] --auth-server=localhost:443 --name=[db-name] --protocol=postgres --uri=[uri] --aws-region=[region] --aws-redshift-cluster-id=[cluster-id]'}
```
kimlisa, line 37: lol i think this is overkill, especially since we have a snapshot somewhere where we test for correct rendering. i think how the test was before was fine. The only thing that changes is the token value (all the success tests pretty much tested that our generateCmd formats things as expected) :shrug:

packages/teleport/src/Databases/AddDatabase/AddDatabase.test.tsx
```diff
@@ -46,28 +55,61 @@ describe('correct database add command generated with given input', () => {
...
 
 test('correct tsh login command generated with sso authType', () => {
-  render(<AddDialog {...props} authType="sso" />);
+  render(<AddDatabase {...props} authType="sso" />);
   const output = 'tsh login --proxy=localhost:443';
 
   expect(screen.queryByText(output)).not.toBeNull();
 });
 
+test('show correct steps when token attempt succeeded', () => {
```
kimlisa, line 71: this is also redundant with the snapshot testing from earlier

packages/teleport/src/Databases/AddDatabase/AddDatabase.test.tsx
```diff
@@ -46,28 +55,61 @@ describe('correct database add command generated with given input', () => {
...
+  );
+
+  expect(screen.queryByText('Step 1')).not.toBeNull();
+  expect(screen.queryByText('Step 2')).not.toBeNull();
+  expect(screen.queryByText('Step 3')).not.toBeNull();
+  expect(screen.queryByText('Step 4')).toBeNull();
+  expect(screen.queryByText('Step 5')).toBeNull();
+});
+
+test('show correct steps when token attempt failed', () => {
```
kimlisa, line 83: this too is also redundant, the only thing i see different is that this test tests for `authType="sso"` render, i would add a storybook and snapshot that render to test this (much simpler, and more visual)

packages/teleport/src/Databases/AddDatabase/AddDatabase.test.tsx
```diff
@@ -46,28 +55,61 @@ describe('correct database add command generated with given input', () => {
...
+  expect(screen.queryByText('Step 1')).not.toBeNull();
+  expect(screen.queryByText('Step 2')).not.toBeNull();
+  expect(screen.queryByText('Step 3')).not.toBeNull();
+  expect(screen.queryByText('Step 4')).not.toBeNull();
+  expect(screen.queryByText('Step 5')).not.toBeNull();
+});
+
 test('render instructions dialog for adding database', () => {
-  render(<AddDialog {...props} />);
+  render(<AddDatabase {...props} />);
```
kimlisa, line 96: we probably don't need this test anymore since we added snapshots with our storybook (it's pretty much the same thing just different props right?)

kimlisa (commented):
i'm sure you do this already but making sure, when we snapshot test, we look at the result to make sure it is what we intended (correct rendering like texts or elements)

packages/teleport/src/Databases/AddDatabase/AddDatabase.test.tsx
mcbattirola: Thanks, added a new entry and snapshot test in https://github.com/gravitational/webapps/pull/653/commits/3d4500c7c514c36f95505d23216b97bf541bebcf

packages/teleport/src/Databases/AddDatabase/AddDatabase.tsx
```diff
@@ -73,81 +90,167 @@ export default function AddDatabase({
...
-          <Box mt={2}>
-            <Select
-              value={selectedDbOption}
-              onChange={(o: Option<DatabaseInfo>) => setSelectedDbOption(o)}
-              options={dbOptions}
-              isSearchable={true}
-              maxMenuHeight={220}
-            />
-          </Box>
+      {attempt.status === 'processing' ? (
```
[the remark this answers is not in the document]
mcbattirola: Awesome, this is much better indeed
Thanks

packages/teleport/src/Databases/AddDatabase/AddDatabase.test.tsx
```diff
@@ -46,28 +47,32 @@ describe('correct database add command generated with given input', () => {
 });
 
 test('correct tsh login command generated with local authType', () => {
-  render(<AddDialog {...props} />);
+  render(<AddDatabase {...props} />);
```
kimlisa, line 50: we don't need this and the below tests since we have snapshot test for it

kimlisa (approved):
after addressing minor comment

mcbattirola:
awaiting for https://github.com/gravitational/teleport/pull/10914 to be merged before I merge this one

kimlisa (approved):

PR #655, opened 2022-03-08
Updates Alert to use break-word

ibeckermayer:
Opportunistically updates `DesktopSession` to have an extra margin and simplifies it's `if` clause.

Corresponds to https://github.com/gravitational/webapps.e/pull/126 and should not be merged until https://github.com/gravitational/webapps.e/pull/126 is merged and this branch's `e` ref is updated.


kimlisa (approved):
thanks for doing this! this word wrap has been bugging me for a long time. could you also remove the break word prop here please: https://github.com/gravitational/webapps/blob/master/packages%2Fteleport%2Fsrc%2Fcomponents%2FFormNewCredentials%2FFormNewCredentials.tsx#L217

PR #656, opened 2022-03-08
Support the 'unknown' audit event

zmb3:
This event is emitted as a fallback if we encounter an event that
Teleport does not know how to interpret.

See: gravitational/teleport#10665


xacrimon (approved):

kimlisa (approved):
:tada:

PR #668, opened 2022-03-11
Add clusters picker

gzdunek:
A few notes:
1. I had to change `KeyboardArrowsNavigation` behaviour a bit: by default none of the items should be selected (`activeIndex: -1`) and when user starts searching then the first item should be selected (`activeIndex: 0`). I was wondering if we don't want to use `KeyboardArrowsNavigation` inside of `FilterableList`, but finally I decided not to do it and add a callback to `FilterableList` to have more flexibility (any thoughts about it?).
2. I reverted the change with providing `clusterUri` to the `DocumentService` (it was used only by `createClusterDocument()`). That document has to be created not only for root, but also for leaf clusters. 
3. I extracted logic for creating a new tab to the `useNewTabOpener` hook.

### DIFF
--- a/packages/teleterm/src/ui/TopBar/Clusters/ClustersFilterableList/ClustersFilterableList.tsx
+++ b/packages/teleterm/src/ui/TopBar/Clusters/ClustersFilterableList/ClustersFilterableList.tsx
@@ -1,0 +1,37 @@
+import React from 'react';
+import { FilterableList } from 'teleterm/ui/components/FilterableList';
+import { ClusterItem } from './ClusterItem';
+import { Box } from 'design';
+import { useKeyboardArrowsNavigationStateUpdate } from 'teleterm/ui/components/KeyboardArrowsNavigation';
+import { Cluster } from 'teleterm/services/tshd/types';
+
+interface ClustersFilterableListProps {
+  items: Cluster[];
+  selectedItem: Cluster;
+
+  onSelectItem(clusterUri: string): void;
+}
+
+export function ClustersFilterableList(props: ClustersFilterableListProps) {
+  const { setActiveIndex } = useKeyboardArrowsNavigationStateUpdate();
+  return (
+    <Box width="260px">
+      <FilterableList<Cluster>
+        items={props.items}
+        filterBy="name"
+        onFilterChange={value =>
+          value.length ? setActiveIndex(0) : setActiveIndex(-1)
+        }
+        placeholder="Search Leaf Cluster"
+        Node={({ item, index }) => (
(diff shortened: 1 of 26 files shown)


packages/teleterm/src/ui/TopBar/Clusters/ClustersFilterableList/ClustersFilterableList.tsx
```diff
@@ -0,0 +1,37 @@
...
+export function ClustersFilterableList(props: ClustersFilterableListProps) {
+  const { setActiveIndex } = useKeyboardArrowsNavigationStateUpdate();
+  return (
+    <Box width="260px">
+      <FilterableList<Cluster>
+        items={props.items}
+        filterBy="name"
+        onFilterChange={value =>
+          value.length ? setActiveIndex(0) : setActiveIndex(-1)
+        }
```
ravicious, line 24: >  I was wondering if we don't want to use `KeyboardArrowsNavigation` inside of `FilterableList`, but finally I decided not to do it and add a callback to `FilterableList` to have more flexibility (any thoughts about it?).

You mean this callback here? I understand this callback as "Reset the active index if the filter changes". The implementation right here seems quite imperative, ideally I suppose the active index would reset itself when the filter changes, but these two things know nothing about each other, right? I think there has to be a way to handle it in a more declarative fashion but nothing comes to my mind atm.

Using `KeyboardArrowsNavigation` would help to hide this logic and as far as I understand, `ClusterItem` would no longer have to care about calling `useKeyboardArrowsNavigation`, that wolud be handled by each individual node in `FilterableList`.

My gut tells me that this would be the way to go, `FilterableList` then acts as something which, well, given items provides a filterable list which you can further restrict with filter and navigate with keyboard. But, for now we use it only in two places, so I'd say let's wait for the third time we use it. If we see that it'd make sense to roll keyboard navigation under `FilterableList` by then, then let's do it.

---

BTW, in the command bar I made it so that the index resets only if the items in the list change, because if the list remains the same after the filter changes then I don't think there's a reason to change the index. I think we can keep it for now though and see what the rest of the team thinks of that.

ravicious (approved):

packages/teleterm/src/ui/TopBar/Clusters/ClustersFilterableList/ClustersFilterableList.tsx
gzdunek: I agree with your opinion, let's move `KeyboardArrowsNavigation` to `FilterableList` when there will be one more usage.

PR #671, opened 2022-03-14
Allow the automatic toggle to be visible when adding new nodes in OSS version. (#592)

hatched:
Backport #592


packages/teleport/src/Nodes/AddNode/useAddNode.ts
```diff
@@ -49,14 +49,14 @@ export default function useAddNode(ctx: TeleportContext) {
 
   return {
     createJoinToken,
+    isEnterprise,
```
kimlisa, line 52: is this flag still needed? (just looking at the backport PR it was removed)

kimlisa (approved):
after minor comment

packages/teleport/src/Nodes/AddNode/useAddNode.ts
hatched: Good catch, but yeah in this fork the `Manually` component still requires the `isEnterprise` flag whereas in `master` it doesn't so it wasn't a direct backport.

PR #672, opened 2022-03-14
Refactor join tokens generation to use new endpoint

mcbattirola:
This PR refactors the join token API calls to use a new endpoint.
It also adds a new service for join tokens, removing the methods from nodes service.

As part of https://github.com/gravitational/teleport/issues/10340, I'm adding a new endpoint to generate join tokens for role DB from the webapi.

I noticed we have an endpoint to generate join tokens for roles app and node (at the same time), so I refactored it to be a single endpoint that receives the roles in the request body.

Here's what a request in the new endpoint looks like:

```http
POST /v1/webapi/token
{
  "roles": ["Node"]
}
```

Instead of the old one:

```http
POST /v1/webapi/nodes/token
```

Here's the PR on core that add this endpoint: https://github.com/gravitational/teleport/pull/10914

The only behavior change that this PR implements is that now, when `useAddApp` and `useAddNode` are started, a token with the correct role (`App` or `Node`, respectively) will be generated, instead tokens with both roles.


packages/teleport/src/services/joinToken/joinToken.ts
```diff
@@ -0,0 +1,28 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import api from 'teleport/services/api';
+import cfg from 'teleport/config';
+import makeJoinToken from './makeJoinToken';
+import { JoinToken } from './types';
+
+const service = {
```
kimlisa, line 22: we are moving away from global service, and prefer using classes to get one instance of it (makes testing easier too), look at other services where we define it as a class

packages/teleport/src/services/joinToken/types.ts
```diff
@@ -0,0 +1,26 @@
...
+export type JoinToken = {
+  id: string;
+  expiry: Date;
+};
+
+export enum NodeRoles {
+  App = 'App',
+  Node = 'Node',
+  Database = 'Db',
+}
```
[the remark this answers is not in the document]
kimlisa: also not sure `NodeRoles` is a good name b/c `node` means... type Node? Maybe TokenType? or TokenRoles? or maybe just `Roles` b/c its defined in `joinToken`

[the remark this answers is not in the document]
mcbattirola: Agreed, thanks

packages/teleport/src/services/joinToken/joinToken.ts
mcbattirola: Thanks, I was wondering why some services are classes and others don't

packages/teleport/src/services/joinToken/joinToken.ts
```diff
@@ -0,0 +1,28 @@
...
+limitations under the License.
+*/
+
+import api from 'teleport/services/api';
+import cfg from 'teleport/config';
+import makeJoinToken from './makeJoinToken';
+import { JoinToken, Roles } from './types';
+
+class JoinTokenService {
+  fetchJoinToken(roles: Roles[]): Promise<JoinToken> {
```
hatched, line 23: You've defined `roles` as an array but I only ever see it being used with a single value. I'm assuming this is just for some future proofing? What do you think about allowing it to be....
```suggestion
  fetchJoinToken(roles: Roles[] | Roles): Promise<JoinToken> {
```

hatched (approved):

packages/teleport/src/services/joinToken/joinToken.ts
mcbattirola: Hmmm the main reason I made it an array is that the backend accepts multiple roles. Even if we don't need more than one role now, I think it is better to keep this way in case we want to generate a token with multiple roles in the future.

```diff
-  fetchJoinToken(roles: Roles[]): Promise<JoinToken> {
+  fetchJoinToken(roles: Roles[] | Roles): Promise<JoinToken> {
```

I think this would make things harder to read (we also would have to change the method's body to test `roles`' type), not sure if it is worth it to remove a `[]`

hatched: Sure no problem, implementers choice 😁

packages/teleport/src/services/joinToken/makeJoinToken.ts
```diff
@@ -14,12 +14,9 @@ See the License for the specific language governing permissions and
...
 */
 
-import { at } from 'lodash';
-import { NodeToken } from './types';
-
-export default function makeToken(json): NodeToken {
-  const [id, expiry] = at(json, ['id', 'expiry']);
+import { JoinToken } from './types';
 
+export default function makeToken({ id, expiry }): JoinToken {
```
kimlisa, line 19: @hatched actually made me realize something we do at other places and forgot in other places, just as a safeguard if object happens to be `undefined`: 

[checks for undefined
](https://github.com/gravitational/webapps/blob/master/packages/teleport/src/services/apps/makeApps.ts#L21)[tests for undefined](https://github.com/gravitational/webapps/blob/master/packages/teleport/src/services/apps/apps.test.ts#L46) (or null same thing in this case)
```suggestion
export default function makeToken({ id, expiry } = {}): JoinToken {
```

hatched: If you wanted, to fix that error you can provide default values in the functions call signature...

```js
function foo ({bar, baz} = {bar: 'zero', baz: 0}) {
  console.log(bar, baz);
}
foo();
// --> zero 0
```

kimlisa: yeah normally we check if something is undefined before giving it to something that may error out, the way I would do it is:

```
function foo ({id, expiry} = {}) {
  return {
      id,
      expiry: expiry ? new Date(expiry) : null, // then let the root caller handle the response values
  }
}
```

the lodash `at`, i think did something similar (which explains the possible usage of it in the past)

mcbattirola: Thanks, incorporated these feedbacks on https://github.com/gravitational/webapps/pull/672/commits/9adcc81bc66a3e2e4fe036d1bbfa4e4e86bd07f9

mcbattirola:
> a empty `{}` is the same thing (except it'll be undefined instead of null), so i'd remove the default values

An empty object would fail type check, we have to provide a default object that has the `id` and `expiry` fields:

```bash
yarn tsc
Initializer provides no value for this binding element and the binding element has no default value.
```

kimlisa (approved):

PR #674, opened 2022-03-16
Teleterm styles update

gzdunek:
Updates the style of teleterm to match the designs.

Best to review commit by commit


packages/teleterm/src/ui/services/workspacesService/documentsService/documentsService.ts
```diff
@@ -157,11 +157,12 @@ export class DocumentsService {
...
-    const nextUri = this.getNextUri(uri);
-    const docs = this.getState().documents.filter(d => d.uri !== uri);
     this.setState(draft => {
-      draft.documents = docs;
-      draft.location = nextUri;
+      if (draft.location === uri) {
+        draft.location = this.getNextUri(uri);
+      }
+
+      draft.documents = this.getState().documents.filter(d => d.uri !== uri);
```
ravicious, line 165: What problem does this solve? I have trouble imagining what was wrong with the previous code.

packages/teleterm/src/ui/TabHost/ClusterConnectPanel/ClusterConnectPanel.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { ButtonPrimary, Flex, Text } from 'design';
+import clusterPng from './clusters.png';
+import Image from 'design/Image';
+import { useAppContext } from 'teleterm/ui/appContextProvider';
+
+export function ClusterConnectPanel() {
+  const ctx = useAppContext();
+
+  function handleConnect() {
+    ctx.commandLauncher.executeCommand('cluster-connect', {});
```
ravicious, line 11: In `~/Library/Application\ Support/Electron/`, I temporarily renamed `app-state.json` and `tsh` to something else to have a clean app. After providing the cluster URI, I see this in the console:

> Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
>    in ClusterAdd (created by ClusterConnect)

Could you investigate that? Perhaps we don't need to handle it right now but it'd be worth at least having some faint idea on what causes this, maybe it's been here for months.

Anyway, after I provide my credentials, there's an error:

```
TypeError: Cannot read property 'findClusterDocument' of undefined
    at Object.run (app.js:26878)
    at CommandLauncher.executeCommand (app.js:26949)
    at Object.defaultHandler [as onSuccess] (app.js:26846)
    at onSuccess (app.js:14929)
    at app.js:15999
    at commitHookEffectListMount (vendor.js:56681)
    at commitPassiveHookEffects (vendor.js:56719)
    at HTMLUnknownElement.callCallback (vendor.js:37138)
    at Object.invokeGuardedCallbackDev (vendor.js:37187)
    at invokeGuardedCallback (vendor.js:37242)
```

This seems to be a bigger problem which we should fix before the demo.

ravicious (commented):
There's an error after connecting to a cluster from a fresh state. Other than that it looks good.

packages/teleterm/src/ui/services/workspacesService/documentsService/documentsService.ts
gzdunek: In the previous implementation `location` was changed when any of documents was closed and that resulted in changing the active tab when you closed inactive tab. We want to change the `location` only when the currently active tab is closed.

packages/teleterm/src/ui/TabHost/ClusterConnectPanel/ClusterConnectPanel.tsx
gzdunek: I pushed a fix for the second issue. The first one is more tricky, but not critical, I will add it to our master ticket.

ravicious (approved):

PR #677, opened 2022-03-17
Add styling to `QuickInput` and add empty state to pickers 

gzdunek:


packages/teleterm/src/ui/TopBar/Identity/EmptyIdentityList.tsx
```diff
@@ -0,0 +1,22 @@
...
+  const ctx = useAppContext();
+
+  function handleConnect() {
+    ctx.commandLauncher.executeCommand('cluster-connect', {});
+  }
+
+  return (
+    <Flex m="auto" flexDirection="column" alignItems="center">
+      <Text typography="h6" bold mb={2}>
+        Connect with a Cluster
```
ravicious, line 15:
```suggestion
        Connect to a Cluster
```

I grepped the repo and it seems we use "connect to" rather than "connect with". `packages/teleterm/src/ui/TabHost/ClusterConnectPanel/ClusterConnectPanel.tsx` also says "connect with".

"Connect with X" is fine IMHO when X is another program/tool, like "Connect with Psql" in `packages/teleterm/src/ui/DocumentGateway/Postgres/Postgres.tsx`.

packages/teleterm/src/ui/QuickInput/QuickInputList/QuickInputList.tsx
```diff
@@ -22,17 +22,25 @@ import * as types from 'teleterm/ui/services/quickInput/types';
...
+
   const $items = items.map((r, index) => {
     const Cmpt = ComponentMap[r.kind] || UnknownItem;
+    const isActive = index === activeItem;
     return (
       <StyledItem
         data-attr={index}
-        $active={index === activeItem}
+        ref={isActive ? activeItemRef : null}
+        $active={isActive}
```
ravicious, line 43: Fixes gravitational/webapps.e#133 (mentioning just to have this commit referenced in that issue).

ravicious (approved):

packages/teleterm/src/ui/TopBar/Identity/EmptyIdentityList.tsx
gzdunek: Done, thanks for catching this

PR #678, opened 2022-03-17
Display cluster name for each connection & add autoconnect 

gzdunek:
Also fixes a bug when connecting to server in a leaf cluster

### DIFF
--- a/packages/teleterm/src/ui/DocumentGateway/useDocumentGateway.ts
+++ b/packages/teleterm/src/ui/DocumentGateway/useDocumentGateway.ts
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import React from 'react';
+import React, { useEffect } from 'react';
 import { useAppContext } from 'teleterm/ui/appContextProvider';
 import * as types from 'teleterm/ui/services/workspacesService';
 import useAsync from 'teleterm/ui/useAsync';
@@ -25,6 +25,7 @@ export default function useGateway(doc: types.DocumentGateway) {
   const workspaceDocumentsService = useWorkspaceDocumentsService();
   const gateway = ctx.clustersService.findGateway(doc.gatewayUri);
   const connected = !!gateway;
+  const cluster = ctx.clustersService.findRootClusterByResource(doc.targetUri);
 
   const [connectAttempt, createGateway, setConnectAttempt] = useAsync(
     async () => {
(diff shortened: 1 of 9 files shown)


packages/teleterm/src/ui/DocumentGateway/useDocumentGateway.ts
```diff
@@ -75,6 +73,12 @@ export default function useGateway(doc: types.DocumentGateway) {
     }
   }, [disconnectAttempt.status]);
 
+  useEffect(() => {
+    if (cluster.connected) {
+      createGateway();
+    }
+  }, [cluster.connected]);
+
```
ravicious, line 81: What kind of scenario is this + the change in the `reconnect` above are going to handle?

packages/teleterm/src/ui/ServerConnect/useServerConnect.ts
```diff
@@ -23,8 +23,10 @@ export default function useServerConnect({ serverUri, onClose }: Props) {
   const logins = cluster?.loggedInUser?.sshLoginsList || [];
 
   const connect = (login: string) => {
+    const rootCluster =
+      ctx.clustersService.findRootClusterByResource(serverUri);
     const documentsService = ctx.workspacesService.getWorkspaceDocumentService(
-      cluster.uri
+      rootCluster.uri
```
ravicious, line 29: Yeah, I'm still kinda conflicted whether we should convert a resource URI to a root cluster URI this way. The root cluster URI seems to be baked into every other URI, so theoretically we should be able to get it just from operations on that string alone, without having to use `clustersService`. My main concern is that in other places it'll require pulling in that `clustersService` which can unnecessarily complicate tests.

We might think about something like `routing.extractRootClusterUri` in the future. For now I added just `ensureRootClusterUri` there which takes any kind of cluster URI and converts it to a root cluster URI.

ravicious (approved):

packages/teleterm/src/ui/ServerConnect/useServerConnect.ts
gzdunek: Good point, I will add it to our master ticket

packages/teleterm/src/ui/DocumentGateway/useDocumentGateway.ts
gzdunek: When you click on a connection it opens new tab with text: Click to reconnect, even when you are connected to the cluster (and in `useReconnect` is just changes document status).
After this change, clicking on a connection will try to make the _connection_ automatically and open a tab.

PR #679, opened 2022-03-18
Take `localClusterUri` into account in `QuickInput`

gzdunek:


packages/teleterm/src/ui/services/quickInput/quickPickers.ts
```diff
@@ -268,10 +268,13 @@ export class QuickSshLoginPicker implements QuickInputPicker {
...
 
   private filterSshLogins(input: string): SuggestionSshLogin[] {
-    // TODO(ravicious): Use local cluster URI.
     // TODO(ravicious): Handle the `--cluster` tsh ssh flag.
-    const rootClusterUri = this.workspacesService.getRootClusterUri();
-    const cluster = this.clustersService.findCluster(rootClusterUri);
+    const localClusterUri =
+      this.workspacesService.getActiveWorkspace()?.localClusterUri;
+    if (!localClusterUri) {
+      return [];
```
ravicious, line 275: Good catch, I thought of this scenario once but then it escaped my head.

packages/teleterm/src/ui/QuickInput/useQuickInput.ts
```diff
@@ -36,7 +36,7 @@ export default function useQuickInput() {
   const [activeSuggestion, setActiveSuggestion] = React.useState(0);
   const autocompleteResult = React.useMemo(
     () => quickInputService.getAutocompleteResult(inputValue),
-    [inputValue]
+    [inputValue, workspacesService.getActiveWorkspace()?.localClusterUri]
```
ravicious, line 39: Let's just add a comment here explaining why it's in the dependency list to make sure we're highlighting a potential code smell.

ravicious (approved):

PR #686, opened 2022-03-21
Update Empty state to new layout

hatched:
Updates the layout when there are no entities of a certain type to the new designs. 

Screen Shot 2022-03-21 at 10 27 57 AM

Fixes: https://github.com/gravitational/teleport/issues/10339

#### Note
There is a jarring flash of the loading indicator when switching between views because the state always enters the `processing` state. We should look into resolving this with an update to the loading indicator design in a follow-up.


packages/teleport/src/Desktops/Desktops.tsx
```diff
@@ -64,6 +71,26 @@ export function Desktops(props: State) {
...
+
+const emptyStateInfo: EmptyStateInfo = {
+  title: 'Add your first Windows desktop to Teleport',
+  byline:
+    'Teleport Desktop Access provides graphical desktop access to remote Windows hosts.',
+  docsURL: DOC_URL,
+  entityType: 'desktop',
+  readOnly: {
+    title: 'No Desktops Found',
+    resource: 'applications',
```
zmb3, line 94: Should this be `applications`?

hatched: Good catch, it should actually be `desktops` :)

packages/teleport/src/Databases/Databases.tsx
```diff
@@ -51,18 +51,22 @@ export function Databases(props: State) {
     authType,
   } = props;
 
-  const isEmpty = attempt.status === 'success' && databases.length === 0;
+  const isEmpty =
+    (attempt.status === 'success' || attempt.status === 'processing') &&
+    databases.length === 0;
```
kimlisa, line 56: i think this will show both the indicator and the empty state if it's processing?

packages/teleport/src/Desktops/Desktops.tsx
```diff
@@ -53,7 +60,7 @@ export function Desktops(props: State) {
         </Box>
       )}
       {attempt.status === 'failed' && <Danger>{attempt.statusText}</Danger>}
-      {attempt.status === 'success' && (
+      {!isEmpty && (
```
kimlisa, line 63: the `isEmpty` flag will be true when it is still processing, so while processing it will probably show the table as well (in its empty state), so you'd probably want a `hasDesktops` flag

packages/teleport/src/Databases/Databases.tsx
```diff
@@ -51,18 +51,22 @@ export function Databases(props: State) {
...
   return (
     <FeatureBox>
       <FeatureHeader alignItems="center" justifyContent="space-between">
         <FeatureHeaderTitle>Databases</FeatureHeaderTitle>
-        <ButtonAdd
-          isLeafCluster={isLeafCluster}
-          canCreate={canCreate}
-          onClick={showAddDialog}
-        />
+        {!isEmpty && attempt.status !== 'processing' && (
```
kimlisa, line 63: use the `hasDatabases` flag instead? (and where ever else applicable)

packages/teleport/src/Desktops/Desktops.tsx
```diff
@@ -64,6 +71,26 @@ export function Desktops(props: State) {
           />
         </>
       )}
+      {isEmpty && (
+        <Empty
+          clusterId={clusterId}
+          canCreate={canCreate && !isLeafCluster}
+          onClick={() => window.open(DOC_URL)}
```
kimlisa, line 78: i think it would be confusing (or maybe see it as a bug?) if the user clicks on `Add Desktops` only to lead to the same link as `View Documentation`? your call, but we have to make sure we set `noreferrer` on all links

packages/teleport/src/components/Empty/Empty.tsx
```diff
@@ -16,29 +16,36 @@ limitations under the License.
...
 import Image from 'design/Image';
-import * as Icons from 'design/Icon';
-import empty from './assets';
+
+import application from './assets/appplication.png';
+import database from './assets/database.png';
+import desktop from './assets/desktop.png';
+import stack from './assets/stack.png';
+
+function getAccentImage(entityType: string): string {
```
kimlisa, line 28: i think it would be better if `entityType` was a literal type, and use the same type for `readOnly.resource`?

kimlisa:
there is also odd jumping of the empty states and it seems to happens when the image renders later? the text renders first which makes the height of the container small, then image decides to join in, which causes the jumping of the height of the container. Can we maybe set a height to the image container so that we have a "placeholder" so the rendering process is smoother?

packages/teleport/src/components/Empty/Empty.tsx
hatched: I added a literal type but I wasn't able to use it for `readOnly.resource` because that is pluralized. I looked into pluralizing the `entityType` value but then I ran into "kubernetes" 😁

packages/teleport/src/Desktops/Desktops.tsx
hatched: Yeah, I was following the pattern set by the k8s page, but you're right that this could be interpreted as a bug. I'll remove the "Add x" buttons from both pages for now and leave the "View Documentation" button.

packages/teleport/src/components/Empty/Empty.tsx
```diff
@@ -16,29 +16,44 @@ limitations under the License.
...
 import Image from 'design/Image';
-import * as Icons from 'design/Icon';
-import empty from './assets';
+
+import application from './assets/appplication.png';
+import database from './assets/database.png';
+import desktop from './assets/desktop.png';
+import stack from './assets/stack.png';
+
+type ResourseTypes =
```
kimlisa, line 28:
```suggestion
type ResourceTypes =
```

packages/teleport/src/components/Empty/Empty.tsx
```diff
@@ -16,29 +16,44 @@ limitations under the License.
...
+  | 'application'
+  | 'database'
+  | 'desktop'
+  | 'kubernetes'
+  | 'server'
+  | 'stack';
```
kimlisa, line 34: do we need stack included?

packages/teleport/src/Databases/Databases.tsx
```diff
@@ -58,11 +58,13 @@ export function Databases(props: State) {
     <FeatureBox>
       <FeatureHeader alignItems="center" justifyContent="space-between">
         <FeatureHeaderTitle>Databases</FeatureHeaderTitle>
-        <ButtonAdd
-          isLeafCluster={isLeafCluster}
-          canCreate={canCreate}
-          onClick={showAddDialog}
-        />
+        {!hasDatabases && attempt.status !== 'processing' && (
```
kimlisa, line 61: hm, i need clarification on when we want to show the button? i thought it was if user has database (successful and non-empty state). Do we want to show the button when it errors out too?

packages/teleport/src/components/Empty/Empty.tsx
hatched: Well that's embarrassing 😄

packages/teleport/src/Desktops/Desktops.tsx
```diff
@@ -53,7 +61,7 @@ export function Desktops(props: State) {
         </Box>
       )}
       {attempt.status === 'failed' && <Danger>{attempt.statusText}</Danger>}
-      {attempt.status === 'success' && (
+      {hasDesktops && (
```
kimlisa, line 64: should desktop also have a `view documentation` button similar to our kube view? (since they both have no "add" directions yet)

packages/teleport/src/Nodes/Nodes.tsx
```diff
@@ -68,16 +68,19 @@ export function Nodes(props: State) {
...
-            isLeafCluster={isLeafCluster}
-            canCreate={canCreate}
-            onClick={showAddNode}
-          />
-        </Flex>
+        {hasNodes && (
+          <Flex alignItems="center">
+            {hasNodes && (
+              <QuickLaunch width="280px" onPress={onSshEnter} mr={3} />
+            )}
```
kimlisa, line 75:
```suggestion
              <QuickLaunch width="280px" onPress={onSshEnter} mr={3} />
```

packages/teleport/src/Desktops/Desktops.tsx
hatched: Sure, added

kimlisa (approved):

packages/teleport/src/Nodes/Nodes.tsx
```diff
@@ -108,14 +108,11 @@ export function Nodes(props: State) {
...
 const emptyStateInfo: EmptyStateInfo = {
-  title: 'ADD YOUR FIRST SERVER',
-  description: (
-    <Text>
-      Instant SSH access for engineers with native security and compliance.
-    </Text>
-  ),
-  buttonText: 'ADD SERVER',
-  videoLink: 'https://www.youtube.com/watch?v=tUXYtwP-Kvw',
+  title: 'Add your first Linux server to Teleport',
```
rudream, line 111: `Add your first Linux server to Teleport` -> `Add your first SSH Server to Teleport` 

It doesn't have to be Linux afaik (you can add a MacBook as a node for example)

packages/teleport/src/components/Empty/Empty.tsx
```diff
@@ -16,29 +16,43 @@ limitations under the License.
...
 import Image from 'design/Image';
-import * as Icons from 'design/Icon';
-import empty from './assets';
+
+import application from './assets/appplication.png';
+import database from './assets/database.png';
+import desktop from './assets/desktop.png';
+import stack from './assets/stack.png';
+
+type ResourceTypes =
```
rudream, line 28: nit: `ResourceTypes` -> `ResourceType`

PR #687, opened 2022-03-22
Show username when possible in identity list

gzdunek:
Closes https://github.com/gravitational/webapps.e/issues/142


ravicious (approved):
Looks good! It'll definitely benefit from gravitational/webapps.e#145 as right now it makes the picker show even less information since more of the cluster name is getting truncated. ;)

PR #690, opened 2022-03-22
Add IAM method to web ui

mcbattirola:
This PR adds a new tab `AWS` on the `ADD SERVER` modal that helps users use the IAM method.

https://github.com/gravitational/teleport/pull/11339 (approved but not merged yet) changes the join token endpoint to allow requesting join tokens with the IAM method. It also adds the option to use the automatic install script to join with an iam token:

```
sudo bash -c "$(curl -fsSL https://localhost:8080/scripts/7a906f128ad7f7c854c1868b620d4e4d/install-node.sh?method=iam)"
```

Main changes:

- added a new IAM token to `useAddToken` hook
- added new tab to ADD SERVER dialog
- changed `automatic` bool for a `method` var to control the currently selected tab
- `fetchJoinToken` now has parameters methods and rules (defaulting to `token` and empty rules)

Screenshot from 2022-04-11 16-39-11

Screenshot from 2022-04-11 16-39-01


packages/teleport/src/Nodes/AddNode/Iam/Iam.tsx
```diff
@@ -0,0 +1,267 @@
...
+export default function Iam({
+  token,
+  expiry,
+  attempt,
+  onGenerate,
+  isEnterprise,
+  version,
+  onClose,
+}: Props) {
+  const { hostname, port } = window.document.location;
```
hatched, line 30: This should probably use the `react-router` `useLocation` as it covers a few edge cases and allows it to easier be mocked for tests.

packages/teleport/src/Nodes/AddNode/Iam/Iam.tsx
```diff
@@ -0,0 +1,267 @@
...
+  }
+
+  function handleRemove(index: number) {
+    const newRules = [...rules];
+    newRules.splice(index, 1);
+    setRules(newRules);
+  }
+
+  function handleGenerate(validator: Validator) {
+    if (!validator.validate()) {
```
hatched, line 54: What's going on here? `validator` doesn't appear to actually be used after this check.

packages/teleport/src/Nodes/AddNode/Iam/Iam.tsx
```diff
@@ -0,0 +1,267 @@
...
+  token,
+  expiry,
+  attempt,
+  onGenerate,
+  isEnterprise,
+  version,
+  onClose,
+}: Props) {
+  const { hostname, port } = window.document.location;
+  const host = `${hostname}:${port || '443'}`;
```
hatched, line 31: Do we want to have this default like this? Or should we instead rely on the protocol?

packages/teleport/src/Nodes/AddNode/Iam/Iam.tsx
```diff
@@ -0,0 +1,267 @@
...
+  text-decoration: none;
+`;
+
+const ButtonRemoveRule = styled(ButtonLink)`
+  position: absolute;
+  top: 12px;
+  right: 12px;
+`;
+
+export const AWS_ACC_ID_REGEXP = /^\d{12}$/;
```
hatched, line 229: Can you add a plaintext description for this regex. I find it helps future developers who maybe aren't super familiar with regex to know what it's doing and what it's _supposed_ to be doing in the event of a bug.

packages/teleport/src/Nodes/AddNode/Iam/Iam.tsx
```diff
@@ -0,0 +1,267 @@
...
+  // no need for an AWS account if an ARN is set
+  if (!value) {
+    return { valid: true };
+  }
+
+  const isValidId = value.match(AWS_ACC_ID_REGEXP);
+  if (!isValidId) {
+    return {
+      valid: false,
+      message: 'AWS account must have 12 digits',
```
hatched, line 247: Just going from the regex this error is a bit better...
```suggestion
      message: 'AWS account must be 12 digits',
```

packages/teleport/src/services/joinToken/types.ts
```diff
@@ -20,3 +20,10 @@ export type JoinToken = {
 };
 
 export type Roles = 'App' | 'Node' | 'Db';
+
+export type Method = 'token' | 'ec2' | 'iam';
+
+export type Rule = {
+  awsAccount?: string;
```
hatched, line 27: 🤷 
```suggestion
  awsAccountID?: string;
```

packages/teleport/src/Nodes/AddNode/Iam/Iam.tsx
```diff
@@ -0,0 +1,267 @@
...
+                </RuleBox>
+              ))}
+            </Box>
+            <Box
+              onClick={() =>
+                setRules([
+                  ...rules,
+                  { awsAccount: '', awsArn: '', isCollapsed: false },
+                ])
+              }
```
kimlisa, line 158: i think it will be cleaner and leaner if we add a helper func `addRule` then:
```suggestion
              onClick={addRule}
```

packages/teleport/src/Nodes/AddNode/Iam/Iam.tsx
```diff
@@ -0,0 +1,267 @@
...
+                      </Box>
+                      <FieldInput
+                        mb={2}
+                        label="AWS ARN"
+                        onChange={e =>
+                          setRuleAtIndex(index, {
+                            ...rules[index],
+                            awsArn: e.target.value,
+                          })
+                        }
```
kimlisa, line 143: when i see `setXXX` i think of hook :thinking: , how about `updateRule` instead?
also, i'd make the code a bit leaner by moving the logic to the helper func so like `updateRule(index, e?)`

packages/teleport/src/Nodes/AddNode/Iam/Iam.tsx
```diff
@@ -0,0 +1,267 @@
...
+              {rules.map((rule, index) => (
+                <RuleBox
+                  key={index}
+                  bg="primary.lighter"
+                  borderRadius="2"
+                  mt={1}
+                  p="3"
+                  position="relative"
+                >
+                  <Text typography="h5">{`Rule #${index + 1}`}</Text>
```
kimlisa, line 102: i suggest `h6`, i think smaller or equal text size for inner content flows better visually

packages/teleport/src/Nodes/AddNode/Iam/Iam.tsx
```diff
@@ -0,0 +1,267 @@
...
+`;
+
+export const AWS_ACC_ID_REGEXP = /^\d{12}$/;
+const requiredAwsAccount = (value, rule: Rule) => () => {
+  if (!rule.awsAccount && !rule.awsArn) {
+    return {
+      valid: false,
+      message: 'Rule cannot be empty',
+    };
+  }
```
kimlisa, line 236: i think we need to re-think how to do this b/c I was initially confused (b/c i'm one of those who doesn't really read error unless i have to :), so i initially thought the first input box had to be required, and then i thought it was a bug when that "requirement" went away when I filled the second box. I would maybe ask rishi the best way to design the `either or, or both` case. Otherwise it might be better to just highlight the whole box as red, and put like `either account ID or ans is required` next to the rule title `Rule #X`

packages/teleport/src/Nodes/AddNode/useAddNode.ts
```diff
@@ -20,26 +20,28 @@ import useAttempt from 'shared/hooks/useAttemptNext';
...
   const version = ctx.storeUser.state.cluster.authVersion;
   const user = ctx.storeUser.state.username;
   const isAuthTypeLocal = !ctx.storeUser.isSso();
-  const [automatic, setAutomatic] = useState(true);
+  const [method, setMethod] = useState<JoinMethod>('iam');
   const [script, setScript] = useState('');
   const [expiry, setExpiry] = useState('');
   const [token, setToken] = useState('');
+  const [iamJoinToken, setIamJoinToken] = useState('');
+  const [iamExpiry, setIamExpiry] = useState('');
```
kimlisa, line 36: hm, do we really need these two extra states?

packages/teleport/src/Nodes/AddNode/Iam/Iam.tsx
mcbattirola: `validator.validate()` will run all methods in the form's `FieldInput` `rule` fields and display and show the error message in case it fails.

That's why we can just return in this case, the validator will take care of the rest. The same is used in multiple places

packages/teleport/src/Nodes/AddNode/Iam/Iam.tsx
mcbattirola: Do you mean to use port 80 in case the protocol is `HTTP`? I'm not sure, can you run Teleport without a cert?

packages/teleport/src/Nodes/AddNode/Iam/Iam.tsx
mcbattirola: Thanks!

packages/teleport/src/services/joinToken/types.ts
mcbattirola: Thanks, I changed it to `awsAccountId`, with lower case 'd' since it seems this is how we do in the codebase.

packages/teleport/src/Nodes/AddNode/Iam/Iam.tsx
```diff
@@ -0,0 +1,267 @@
...
+                  { awsAccount: '', awsArn: '', isCollapsed: false },
+                ])
+              }
+            >
+              <ButtonAddRule>+ Add new rule</ButtonAddRule>
+            </Box>
+            {token && (
+              <>
+                <Text mt="3" mb="2" bold typography="h5">
+                  Script Generated
```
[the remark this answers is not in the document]
mcbattirola: I followed the design there, but I agree that this is not consistent with other wizards. I left a comment on Figma about this and will reach Rishi. Thanks for the feedback.

packages/teleport/src/Nodes/AddNode/Iam/Iam.tsx
mcbattirola: I agree, I'll reach Rishi and ask for his help to improve the UX there, thanks

packages/teleport/src/Nodes/AddNode/Iam/Iam.tsx
mcbattirola: Hmmm, `location` doesn't seem to have hostname and port, am I missing something?

https://v5.reactrouter.com/web/api/location

packages/teleport/src/Nodes/AddNode/useAddNode.ts
mcbattirola: Hmmm, we don't *need* it. 

Actually, I initially reused the token and expiry that we had already, but I didn't want the users to lose their IAM token if they changed tabs (since going to another tab would trigger a request to a join token with a different method).

I think this is better UX but no strong feelings there, let me know if you think I should change it.

packages/teleport/src/Nodes/AddNode/Iam/Iam.tsx
hatched: Hmm sorry I may have steered you wrong...I was sure that those values were in there but I guess not!

packages/teleport/src/Nodes/AddNode/Iam/Iam.tsx
hatched: Can you add a comment to this effect in the code as it's not clear. Be as detailed as you need to be :)

mcbattirola: Added a comment there, thanks

mcbattirola:
Update: I talked to Rishi and he will go over the things I commented on and check with @xinding33. I'll probably update this PR with some changes by tomorrow

mcbattirola:
Hey folks, the following changes were made in the design of this wizard:

- we now use the automatic script instead of `teleport start`
- we will use only one rule in the web UI. The user can do more complex rules following the documentation (should add this information on a copy somewhere?)
- the token now expires in a thousand years. This is informed to the user in a text below the script

packages/teleport/src/Nodes/AddNode/Iam/Iam.tsx
mcbattirola: (marking as resolved since this is not a problem with a single rule)

packages/teleport/src/Nodes/AddNode/Iam/Iam.tsx
mcbattirola: Update:

The team agreed on going for a simple flow here, making the AWS Account ID mandatory and arn optional.

If the user wants to do something different, they can read the docs; the UI will aim for solving the simplest use case to get users starting.

mcbattirola:
Not crazy about the long labels in the inputs. I think we should add a new prop to these inputs like `labelTip` or something, that will show after the label but not bold. Thoughts?

packages/shared/components/FieldInput/FieldInput.tsx
```diff
@@ -30,15 +31,22 @@ export default function FieldInput({
...
   const { valid, message } = useRule(rule(value));
   const hasError = !valid;
   const labelText = hasError ? message : label;
   return (
     <Box mb="4" {...styles}>
-      {label && <LabelInput hasError={hasError}>{labelText}</LabelInput>}
+      {label && (
+        <LabelInput hasError={hasError}>
+          {labelText}
+          {labelTip && <LabelTip text={labelTip} />}
```
hatched, line 45: Can you add a story & test case for this new syntax plz.

packages/teleport/src/Nodes/AddNode/Iam/Iam.tsx
```diff
@@ -0,0 +1,150 @@
...
+                  labelTip="nodes must match this AWS Account ID to join your Teleport cluster"
+                  autoFocus
+                  name="awsAccountId"
+                  onChange={e =>
+                    setRule({ ...rule, awsAccountId: e.target.value })
+                  }
+                  rule={requiredAwsAccountId}
+                  placeholder="111111111111"
+                  value={rule.awsAccountId}
+                  onKeyPress={e =>
```
hatched, line 75: I think that these fields and the submit button should both be in a `form` so that you can rely on the native submit handling of a form. You can then attach a single `onsubmit` event handler to handle the submits.

hatched (approved):
Looks good, thanks for sticking with this one! Just a couple comments about expanding the test coverage.

packages/teleport/src/services/joinToken/types.ts
```diff
@@ -20,3 +20,10 @@ export type JoinToken = {
 };
 
 export type Roles = 'App' | 'Node' | 'Db';
+
+export type Method = 'token' | 'ec2' | 'iam';
+
+export type Rule = {
+  awsAccountId?: string;
```
kimlisa, line 27: maybe? b/c it's required now?

```suggestion
  awsAccountId: string;
```

kimlisa (commented):
note that there is a update to e-ref file, probably didn't mean to include that

packages/teleport/src/Nodes/AddNode/useAddNode.ts
```diff
@@ -48,31 +50,61 @@ export default function useAddNode(ctx: TeleportContext) {
...
+    return run(() =>
+      ctx.joinTokenService
+        .fetchJoinToken(['Node'], 'iam', [rules])
+        .then(iamToken => {
+          const expires = formatDistanceStrict(
+            new Date(),
+            new Date(iamToken.expiry)
+          );
+          setIamExpiry(expires);
+          setIamJoinToken(iamToken.id);
```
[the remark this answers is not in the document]
mcbattirola: Yes, I think this makes a lot of sense. I added it here https://github.com/gravitational/webapps/pull/690/commits/6d1e8c87463c354d16768d6b1c3963cc1e3d66fa

Please let me know what you think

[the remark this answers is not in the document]
mcbattirola: Also maybe it is a good idea in the future to refactor all those wizards into common components?

```jsx
<Wizard>
  <Step title="Install Teleport">
    ...
  </Step>
  <Step title="Run Script">
    ...
  </Step>
</Wizard>
```

mcbattirola:
> note that there is a update to e-ref file, probably didn't mean to include that

So I linted `input.jsx` (https://github.com/gravitational/webapps/pull/690/commits/937bcc38014e25394ce9ce886710529ae0c846a2) which caused some snapshots on `webapps.e` to need updating... I opened a PR to update those snapshots [here](https://github.com/gravitational/webapps.e/pull/222), but maybe I should just revert the lining to save this trouble

packages/teleport/src/Nodes/AddNode/Iam/Iam.tsx
```diff
@@ -0,0 +1,147 @@
...
+                placeholder="arn:aws:sts::111111111111:assumed-role/teleport-node-role/i-*"
+                value={rule.awsArn}
+              />
+            </Box>
+            <Box>
+              <Text bold as="span">
+                Step 3
+              </Text>{' '}
+              - Generate and run script
+              <ButtonPrimary
```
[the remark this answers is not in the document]
mcbattirola: Update: talked to @xinding33, he believes this button should be part of the step so it make sense to have it placed where it is. He thinks we are ok to ship

packages/teleport/src/Nodes/AddNode/useAddNode.ts
```diff
@@ -14,65 +14,62 @@ See the License for the specific language governing permissions and
...
 
   function createJoinToken() {
     return run(() =>
-      ctx.joinTokenService.fetchJoinToken(['Node']).then(token => {
-        const cmd = createNodeBashCommand(token);
-        setExpiry(cmd.expires);
-        setScript(cmd.text);
-        setToken(token.id);
+      ctx.joinTokenService.fetchJoinToken(['Node'], 'token').then(token => {
+        setToken(token);
```
kimlisa, line 36: shorthand
```suggestion
      ctx.joinTokenService.fetchJoinToken(['Node'], 'token').then(setToken);
```

packages/teleport/src/Nodes/AddNode/useAddNode.ts
```diff
@@ -14,65 +14,62 @@ See the License for the specific language governing permissions and
...
     );
   }
 
+  function createIamJoinToken(rules: Rule) {
+    return run(() =>
+      ctx.joinTokenService
+        .fetchJoinToken(['Node'], 'iam', [rules])
+        .then(iamToken => {
+          setIamJoinToken(iamToken);
+        })
```
kimlisa, line 47: shorthand
```suggestion
        .then(setIamJoinToken)
```

packages/teleport/src/Nodes/AddNode/useAddNode.ts
```diff
@@ -14,65 +14,62 @@ See the License for the specific language governing permissions and
...
-  const [script, setScript] = useState('');
-  const [expiry, setExpiry] = useState('');
-  const [token, setToken] = useState('');
-
-  useEffect(() => {
-    createJoinToken();
-  }, []);
+  const [method, setMethod] = useState<JoinMethod>('iam');
+  const [token, setToken] = useState<JoinToken>(null);
+  const [iamJoinToken, setIamJoinToken] = useState<JoinToken>(null);
```
kimlisa, line 31: nit: i think setting it to nothing is equivalent to null (or i think more accurately undefined), unless for some reason we want it to specifically be null?
```suggestion
  const [token, setToken] = useState<JoinToken>();
  const [iamJoinToken, setIamJoinToken] = useState<JoinToken>();
```

packages/teleport/src/Nodes/AddNode/Automatically/Automatically.tsx
```diff
@@ -14,15 +14,29 @@
...
 export default function Automatically(props: Props) {
-  const { script, expiry, createJoinToken, attempt } = props;
+  const { createJoinToken, attempt, onClose, joinToken } = props;
 
-  if (attempt.status === 'processing') {
+  useEffect(() => {
+    if (!joinToken) {
+      createJoinToken();
+    }
+  }, [joinToken]);
```
kimlisa, line 37: hm, do we need to state `joinToken` as a dependency?

kimlisa (approved):
after addressing minor comments. note that for the minor comments, there were a few other places where my comments also applied that I didn't comment on (ie: shorthand usage, using null as initial state etc)

also thanks for the refactor! looks much better :)

kimlisa:
> > note that there is a update to e-ref file, probably didn't mean to include that
> 
> So I linted `input.jsx` ([937bcc3](https://github.com/gravitational/webapps/commit/937bcc38014e25394ce9ce886710529ae0c846a2)) which caused some snapshots on `webapps.e` to need updating... I opened a PR to update those snapshots [here](https://github.com/gravitational/webapps.e/pull/222), but maybe I should just revert the linting to save this trouble

oh ha! i _think_ that might make backporting quite painful, so i'd just unlint it (up to you)

packages/teleport/src/Nodes/AddNode/useAddNode.ts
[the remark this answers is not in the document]
kimlisa: yeah that might be a good idea

packages/teleport/src/Nodes/AddNode/Automatically/Automatically.tsx
mcbattirola: Not really. Some lint rules (that we don't use) would complain if we didn't add `joinToken` here. But I think in this case it is ok to remove it

mcbattirola:
> > > note that there is a update to e-ref file, probably didn't mean to include that
> > 
> > 
> > So I linted `input.jsx` ([937bcc3](https://github.com/gravitational/webapps/commit/937bcc38014e25394ce9ce886710529ae0c846a2)) which caused some snapshots on `webapps.e` to need updating... I opened a PR to update those snapshots [here](https://github.com/gravitational/webapps.e/pull/222), but maybe I should just revert the linting to save this trouble
> 
> oh ha! i _think_ that might make backporting quite painful, so i'd just unlint it (up to you)

Ok, removed the lint change on https://github.com/gravitational/webapps/pull/690/commits/c6367172202ccf2c267735f6e15323f1b19cef79 😅
This updates the snapshots back and remove the webapps.e change

closed https://github.com/gravitational/webapps.e/pull/222

PR #691, opened 2022-03-23
Rework identity popover

gzdunek:
Closes https://github.com/gravitational/webapps.e/issues/145


packages/teleterm/src/ui/TopBar/Identity/useIdentity.ts
```diff
@@ -16,18 +16,7 @@ export function useIdentity() {
   }
 
   async function logout(clusterUri: string): Promise<void> {
```
ravicious, line 18: Does it need to be `async` still?

packages/teleterm/src/ui/ClusterLogout/useClusterLogout.ts
```diff
@@ -0,0 +1,45 @@
+import { useAppContext } from '../appContextProvider';
+import useAsync from '../useAsync';
+import { useEffect } from 'react';
+
+export function useClusterLogout({ clusterUri, onClose, clusterTitle }: Props) {
+  const ctx = useAppContext();
+  const [{ status, statusText }, removeCluster] = useAsync(async () => {
+    //TODO gzdunek: logout and removeCluster should be combined into a single acton in tshd
+    await ctx.clustersService.logout(clusterUri);
+    await ctx.clustersService.removeCluster(clusterUri);
```
ravicious, line 10: I was going to ask "What about error handling?", but I see it already works quite nicely. 👍

packages/teleterm/src/ui/TopBar/Identity/useIdentity.ts
```diff
@@ -15,20 +15,8 @@ export function useIdentity() {
...
-    await ctx.clustersService.logout(rootClusterUri);
-    const [firstConnectedWorkspace] =
-      ctx.workspacesService.getConnectedWorkspacesClustersUri();
-    if (firstConnectedWorkspace) {
-      await ctx.workspacesService.setActiveWorkspace(firstConnectedWorkspace);
-    } else {
-      await ctx.workspacesService.setActiveWorkspace(null);
-    }
+  async function logout(clusterUri: string): Promise<void> {
+    ctx.commandLauncher.executeCommand('cluster-logout', {clusterUri})
```
ravicious, line 19: We have to be on the lookout when it comes to `ctx.commandLauncher`, I have a feeling that it's becoming a place to dump any business logic that doesn't really fit anywhere else. 😅

Maybe we'll get some better ideas in the future.

ravicious (approved):
Some minor copy changes are needed I think, but it looks good otherwise.

`<Identity>` still doesn't show enough information for my liking, but I'll let the team judge that.

It just barely has space to hold my info for asteroid.earth when expanded and still clips it when collapsed. Given how much space is available for it, I don't see why it couldn't show the whole thing.

identity component

packages/teleterm/src/ui/ClusterLogout/ClusterLogout.tsx
```diff
@@ -41,41 +42,41 @@ export function ClusterRemove({
...
+        maxWidth: '400px',
         width: '100%',
       })}
     >
-      <DialogHeader>
-        <Text typography="h4" css={{ whiteSpace: 'nowrap' }}>
-          Remove Cluster {clusterTitle}
+      <DialogHeader justifyContent="space-between" mb={0}>
+        <Text typography="h5" bold style={{ whiteSpace: 'nowrap' }}>
+          Logout from Cluster {clusterTitle}
```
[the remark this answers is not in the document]
gzdunek: For me, the lowercase "cluster" looks fine, changed to this version

packages/teleterm/src/ui/TopBar/Identity/useIdentity.ts
gzdunek: No, removed

packages/teleterm/src/ui/TopBar/Identity/useIdentity.ts
gzdunek: Maybe we should switch to more specialised services for these actions? (in the future)

gzdunek:
> Some minor copy changes are needed I think, but it looks good otherwise.
> 
> `<Identity>` still doesn't show enough information for my liking, but I'll let the team judge that.
> 
> It just barely has space to hold my info for asteroid.earth when expanded and still clips it when collapsed. Given how much space is available for it, I don't see why it couldn't show the whole thing.
> 
> identity component

I think you're right, there is no reason to restrict the width of these elements 👍

packages/teleterm/src/ui/TopBar/Identity/useIdentity.ts
ravicious: Yeah, idk, it looks like most services in our code are those long-lived objects that are stored in the app context.

But if you look at the commands in the command launcher, they need to just orchestrate the work of a single or a couple of services (with the exception of `tsh-ssh` which definitely could be refactored to keep all of that logic elsewhere 🙊). The way I think about it is that those commands play a similar role as the controller layer in MVC.

I think having a layer like that is useful, maybe it just shouldn't be tied to the concept of a command launcher (since we don't really have a command launcher) and instead of operating on strings to identify the commands it could just export a couple of functions.

packages/teleterm/src/ui/TopBar/Identity/IdentityList/IdentityListItem.tsx
```diff
@@ -32,33 +34,44 @@ export function IdentityListItem(props: IdentityListItemProps) {
     <ListItem
       onClick={props.onSelect}
       isActive={isActive}
-      onMouseEnter={() => setIsHovered(true)}
-      onMouseLeave={() => setIsHovered(false)}
+      ref={ref}
+      style={{ maxWidth: newWidth && newWidth + 'px' }}
+      onMouseEnter={() => {
+        // we set maxWidth to list item element, because otherwise it becomes wider on hover (new element is added to it)
+        setMaxWidth(ref.current.clientWidth);
```
ravicious, line 41: I strongly dislike solutions like this, but I think there's no way around it because the Popover element itself is not responsive and just uses `top` and `left` to position itself – I suppose the popover size wasn't meant to be changed while it is opened.

So I suspect there's no pure CSS solution to this. 😢

gzdunek: I spend a lot of time trying to find CSS only solution, but I think it is not possible. As we don't want to set any restrictions on popover width there is no way to block changing it's width when we add a new element in row (which is logout button). Otherwise, popover width will jump on longer items hover.

I agree, JS solutions are problematic, however in this case it is very light and simple :)

PR #692, opened 2022-03-23
Use connection dropdown

gzdunek:
Closes https://github.com/gravitational/webapps.e/issues/146


packages/shared/components/MenuSshLogin/MenuSshLogin.tsx
```diff
@@ -100,7 +101,7 @@ export const LoginItemList = ({ logins, onClick, onKeyPress }) => {
         key={key}
         px="2"
         mx="2"
-        as={NavLink}
+        as={url ? NavLink : null}
```
gzdunek, line 104: In teleterm we don't have classical routing, so rendering `NavLink` causes errors (that it should be used in context of `Router`)

packages/shared/components/MenuSshLogin/MenuSshLogin.tsx
```diff
@@ -100,7 +100,7 @@ export const LoginItemList = ({ logins, onClick, onKeyPress }) => {
...
```
ravicious, line 104: Could we change it so that instead of rendering a `div` when there's no URL it renders a `button` styled as a link instead? This actually provides keyboard support OOTB and helps us avoid having onClicks on divs.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Servers/Servers.tsx
```diff
@@ -60,19 +65,24 @@ function ServerList(props: State) {
...
+  onConnect: (login: string) => void
 ) => {
   return (
     <Cell align="right">
-      <ButtonBorder
-        size="small"
-        onClick={() => {
-          connect(serverUri);
+      <MenuSshLogin
+        menuListCss={menuListCss}
```
ravicious, line 74: I've been here for just two months, so I suppose @kimlisa might have something more to say. But long-term, injecting arbitrary CSS seems to me like a wrong way of modifying the appearance of a shared component.

Ideally I suppose everything should work OOTB after we change the theme in Teleterm. However, I noticed that the way we use the theme is color-oriented rather than purpose-oriented.

https://github.com/gravitational/webapps/blob/0b65709668092d338ed96909b01052ec6bfcc11c/packages/shared/components/MenuSshLogin/MenuSshLogin.tsx#L130-L137

https://github.com/gravitational/webapps/blob/0b65709668092d338ed96909b01052ec6bfcc11c/packages/design/src/Menu/MenuList.jsx#L32-L33

I might be wrong, but I think this makes it harder to reuse components, because `theme.colors.light` might mean different things in different contexts. I saw [a good thread about that](https://twitter.com/buildsghost/status/1505072846564839424) a couple of days ago which goes into detail about this issue.

OTOH, taking care of that is probably out of scope for this PR.

One solution which comes to my mind is having a special copy of the theme in Teleterm which will override the Teleterm theme in places like this one here where we really need to bend a shared component to make it look a certain way. This means that in this context `theme.colors.light` could be something entirely else than it is in the general Teleterm theme. What do you think about this idea?

gzdunek: Thanks for detailed comment, I agree! 
I chose the way with css styling to apply more complicated changes like: different border radius, no separator between items, rounded corners of hovered item.
However, since we have agreed that color customization is sufficient, styling using a provider is simple.

packages/shared/components/MenuSshLogin/MenuSshLogin.tsx
```diff
@@ -156,7 +156,7 @@ const Input = styled.input(
   border: 1px solid ${theme.colors.subtle};
   border-radius: 4px;
   box-sizing: border-box;
-  color: #263238;
+  color: ${theme.colors.blueGrey[900]};
```
ravicious, line 159: I wanted to check if the change from `<React.Fragment>` to `<Flex flexDirection="column">` doesn't change anything in the web UI, but when I try to use the web UI on this branch I get "`theme.colors.blueGrey` is undefined" after clicking on Connect.

Could you make sure that this change and the `<React.Fragment>` change don't cause problems for the web UI?

ravicious (commented):
Ran into some problems when testing how this change impacts the web UI.

packages/shared/components/MenuSshLogin/MenuSshLogin.tsx
gzdunek: I tested that change in the web UI as well...
If only I had noticed that I am testing on `localhost:8080`, not `localhost:3080` 🙄🙄

Fixed

gzdunek:
Force-push to re-trigger builds

PR #697, opened 2022-03-25
Enable `babel-plugin-styled-components` in production and tests

gzdunek:
Since this plugin was only enabled in the development, [css prop](https://styled-components.com/docs/api#css-prop) worked locally but didn't work in the bundled app.

The plugin also improves bundle size a bit (`2.36 MiB` vs `2.38 MiB` previously).

### DIFF
--- a/packages/build/.babelrc.js
+++ b/packages/build/.babelrc.js
@@ -42,10 +42,13 @@ module.exports = {
     development: {
       plugins: [
         ...plugins,
-        'babel-plugin-styled-components',
+        ['babel-plugin-styled-components', { displayName: true, ssr: false }],
       ],
     },
   },
   presets: makePresents(),
-  plugins,
+  plugins: [
+    ...plugins,
+    ['babel-plugin-styled-components', { displayName: false, ssr: false }],
+  ],
 };


packages/build/.babelrc.js
```diff
@@ -42,10 +42,13 @@ module.exports = {
     development: {
       plugins: [
         ...plugins,
-        'babel-plugin-styled-components',
+        ['babel-plugin-styled-components', { displayName: true, ssr: false }],
```
ravicious, line 45: Damn, `displayName` is amazing for Teleterm where adding React DevTools is a bit of a pain. 🙏

> This option enhances the attached CSS class name on each component with richer output to help identify your components in the DOM without React DevTools. In your page source you'll see: `<button class="Button-asdf123 asdf123" />` instead of just `<button class="asdf123" />`.

ravicious (approved):
Tested this with the web UI by running the following command from `packages/teleterm` and it looks like it doesn't break anything.

```
yarn g-start --config $(pwd)/webpack.prod.config.js --target=https://localhost:3080/web
```

packages/build/.babelrc.js
gzdunek: True, but I just want to add that there is no change for generated classnames.
This plugin generated `displayName` by default (so `{ displayName: true }` in dev mode doesn't change anything - I added it for informative purposes).

gzdunek:
> Tested this with the web UI by running the following command from `packages/teleterm` and it looks like it doesn't break anything.

Probably you mean `packages/teleport` :)

kimlisa (approved):

PR #698, opened 2022-03-25
Limit navigation capabilities to reduce attack surface

ravicious:
This takes care of TEL-Q122-19 from the recent security audit.

At the moment we don't create new windows nor navigate away from the rendered app, so we can just block everything. Obviously when adding support for multiple windows we'll have to change that and use Node's URL parser, preferably.

To test this, I made it so that the table with servers renders a regular link next to each hostname. Shift + click tries to open the link in a new window.

Testing the `<webview>` tag required adding `webviewTag: true` to `webPreferences` of the `BrowserWindow`.

https://github.com/gravitational/webapps/blob/f8cfdf3839e9efa7faf9067e57c0442baab360c9/packages/teleterm/src/mainProcess/mainProcess.ts#L55-L59

### DIFF
--- a/packages/teleterm/src/main.ts
+++ b/packages/teleterm/src/main.ts
@@ -47,6 +47,34 @@ app.whenReady().then(() => {
   mainProcess.createWindow();
 });
 
+// Limit navigation capabilities to reduce the attack surface.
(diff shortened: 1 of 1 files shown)


gzdunek (approved):
Good job!

PR #702, opened 2022-03-29
Improve error handling

gzdunek:
In some places informing the user about an error was not sufficient.
I added:
- standard, red labels over the tables in `ClusterResources` (servers, dbs, kubes, apps)
- notifications in other places, where we don't have an option to show static error label (closing db connection, synchronising root cluster) 

Notification component: 
Screenshot 2022-03-29 at 11 49 00
(IMO this looks quite nice for someone with no design experience 🙂)

I also let the app render when the init function fails - this should make it easier to track why the app sometimes stops at `Loading`. 
Closes: https://github.com/gravitational/webapps.e/issues/151

### DIFF
--- a/packages/teleterm/src/ui/DocumentCluster/ClusterResources/Applications/Applications.tsx
+++ b/packages/teleterm/src/ui/DocumentCluster/ClusterResources/Applications/Applications.tsx
@@ -33,6 +33,7 @@ import Table, { Cell } from 'design/DataTable';
 import * as types from 'teleterm/ui/services/clusters/types';
 import { useApps, State } from './useApps';
 import { renderLabelCell } from '../renderLabelCell';
+import { Danger } from 'design/Alert';
 
 export default function Container() {
   const state = useApps();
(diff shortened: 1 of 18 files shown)


packages/teleterm/src/ui/DocumentCluster/ClusterResources/Applications/Applications.tsx
```diff
@@ -33,52 +33,58 @@ import Table, { Cell } from 'design/DataTable';
...
-        {
-          altKey: 'launch-btn',
-          render: renderConnectButton,
-        },
-      ]}
-      emptyText="No Applications Found"
-      pagination={{ pageSize: 100, pagerPosition: 'bottom' }}
-    />
+    <>
+      {syncStatus.status === 'failed' && (
```
ravicious, line 48: How can I test this? In what scenario is `syncStatus` set `failed`? I tried to either disconnect my network connection or shut down the cluster, but in both cases the app crashes with

```
2 UNKNOWN: dial tcp 127.0.0.1:3023: connect: connection refused
```

packages/teleterm/src/ui/services/clusters/clustersService.ts
```diff
@@ -271,10 +284,18 @@ export class ClustersService extends ImmutableStore<ClustersServiceState> {
...
+      await this.client.removeGateway(gatewayUri);
+      this.setState(draft => {
+        draft.gateways.delete(gatewayUri);
+      });
+    } catch (error) {
+      this.notificationsService.notifyError({
+        title: 'Could not close DB connection',
+        description: error.message,
+      });
+      throw error;
```
ravicious, line 297: Same question as for cluster resources, how can I trigger this path when using the app?

gzdunek:
I don't think you can easily trigger these errors. I wanted to add that handling for _unknown_ situations, like when using a cellular connection with poor coverage, I think there will be higher chance for a particular request to fail.  
For resources we already have this `syncStatus` data, so for me it's good to show it.

During development I was just throwing an error in the request:
```typescript
await this.client.removeGateway(gatewayUri).then(() => {
  throw new Error('Network failure');
});
```

gzdunek:
Btw, we were accessing `serversSyncStatus` like an object, but it is a `Map`, so I fixed that.

ravicious:
> I don't think you can easily trigger these errors. (…) For resources we already have this `syncStatus` data, so for me it's good to show it.

Yeah, it's probably better to add it anyway.

I wish the app didn't crash if the network is offline or the given cluster goes offline, but I suppose that's fine for the preview release. I'll add these two things to the tech debt backlog.

gzdunek:
Pushed small visual changes to the `Notification` component, content can now take full width.

ravicious (approved):

packages/teleterm/src/ui/components/Notifcations/Notification.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+const textCss = css`
+  line-height: 20px;
+`;
+
+const shortTextCss = css`
+  ${textCss}
+  display: -webkit-box;
+  -webkit-box-orient: vertical;
+  -webkit-line-clamp: 3;
```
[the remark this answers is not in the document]
gzdunek: Unfortunately, to use `-webkit-line-clamp` we have to use `-webkit-box` https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp.
I wanted to show more than one line when text is overflowing and there is no other way 😕 
It is not a standard feature, but IMO as long as we know the target browser we should be safe.

packages/teleterm/src/ui/services/clusters/clustersService.ts
```diff
@@ -271,10 +284,18 @@ export class ClustersService extends ImmutableStore<ClustersServiceState> {
...
-      draft.gateways.delete(gatewayUri);
-    });
+    try {
+      await this.client.removeGateway(gatewayUri);
+      this.setState(draft => {
+        draft.gateways.delete(gatewayUri);
+      });
+    } catch (error) {
+      this.notificationsService.notifyError({
+        title: 'Could not close DB connection',
```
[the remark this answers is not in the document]
gzdunek: Done

ravicious (approved):

packages/teleterm/src/ui/services/notifications/notificationsService.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  getNotifications(): NotificationItem[] {
+    return this.state;
+  }
+
+  useState(): NotificationItem[] {
+    return useStore(this).state;
+  }
+
+  private notify(options: Omit<NotificationItem, 'id'>): string {
```
ravicious, line 38: Oh, I didn't know about it! Thanks.

PR #704, opened 2022-03-30
Submit modals' forms on `Enter`

gzdunek:
I moved submit functions from `onClick` to `onSubmit` - it seems to be more correct usage of `form` element (they works in `onClick` too).

Closes: https://github.com/gravitational/webapps.e/issues/152


ravicious (approved):
Perfect, PEFRECT solution! 😗👌

The only thing I like more than changes like this one is removing onClick from divs.

PR #709, opened 2022-03-30
Add Redis, MariaDB and Microsoft SQL Server to DB wizard

mcbattirola:
This PR adds missing DB types in the dropdown of `ADD DATABASE`

- Changed `MySQL` label on screen to `MySQL/MariaDB` since they use the same protocol and there is no difference in the command generated. Let me know your thoughts on that.

- Added extra parameters needed for `sqlserver` protocol

Should be backported to v9

Closes gravitational/teleport#10757

Screenshot from 2022-03-30 15-59-59


packages/teleport/src/Databases/AddDatabase/AddDatabase.tsx
```diff
@@ -298,15 +298,25 @@ const generateDbStartCmd = (
...
 
+  let cmd = baseCommand;
+  switch (protocol) {
+    case 'sqlserver':
+      cmd =
+        `${baseCommand} --ad-keytab-file=/path/to/teleport.keytab ` +
+        `--ad-domain=EXAMPLE.COM ` +
+        `--ad-spn=MSSQLSvc/sqlserver.example.com:1433`;
+      break;
+  }
```
kimlisa, line 309: i would make it simpler 
```
let baseCommand = `...`

if protocol === 'sqlserver' {
  baseCommand = `...`
}
```

or
```ts
  let cmd;
  switch (protocol) {
    case 'sqlserver':
      cmd =
        `${baseCommand} --ad-keytab-file=/path/to/teleport.keytab ` +
        `--ad-domain=EXAMPLE.COM ` +
        `--ad-spn=MSSQLSvc/sqlserver.example.com:1433`;
      break;
    default:
      com = `...`
  }
```

kimlisa (approved):
after minor comment

packages/teleport/src/Databases/AddDatabase/AddDatabase.tsx
mcbattirola: Agreed, thanks!

PR #711, opened 2022-03-31
Remove u2f components and logic

kimlisa:
#### Description
This PR removes u2f related components and logic in response to [u2f being aliased to webauthn](https://github.com/gravitational/teleport/issues/10375#issuecomment-1050216853). This is part of passwordless work.

#### Before Merge
- [x] Update e-ref after counterpart PR gets merged to master https://github.com/gravitational/webapps.e/pull/214 (required to get CI to pass)

#### Testing
- [x] ran tests and tsc with the `e` change
- [x] manually test paths changed with a u2f device (did not test teleterm)


packages/shared/services/types.ts
```diff
@@ -16,18 +16,12 @@ limitations under the License.
...
-  | 'optional'
-  | 'on'
-  | 'webauthn';
+export type Auth2faType = 'otp' | 'off' | 'optional' | 'on' | 'webauthn';
 
 // PreferredMfaType is used to determine which MFA option
 // is preferred when more than one option can be available
 // and only one should be preferred.
-export type PreferredMfaType = 'webauthn' | 'u2f' | '';
+export type PreferredMfaType = 'webauthn' | '';
```
ibeckermayer, line 24: Is this needed now that there's only one type? Or are there two types, one of which is represented by `''`? Seems like it should either be able to be removed, or at least a comment should clarify what `''` means.

packages/teleport/src/lib/tdp/codec.ts
```diff
@@ -63,8 +79,8 @@ export type ClipboardData = {
 
 // | message type (10) | mfa_type byte | message_length uint32 | json []byte
 export type MfaJson = {
-  // TODO(isaiah) make this a type that ensures it's the same as MessageTypeEnum.U2F_CHALLENGE and MessageTypeEnum.WEBAUTHN_CHALLENGE
-  mfaType: 'u' | 'n';
```
ibeckermayer, line 67: Technically `'u'` is still a part of the [TDP spec](https://github.com/gravitational/teleport/blob/master/rfd/0037-desktop-access-protocol.md#10---mfa), so we should leave this as is until that's changed.

packages/teleport/src/lib/tdp/codec.ts
```diff
@@ -403,8 +419,8 @@ export default class Codec {
   decodeMfaJson(buffer: ArrayBuffer): MfaJson {
     let dv = new DataView(buffer);
     const mfaType = String.fromCharCode(dv.getUint8(1));
-    if (mfaType !== 'n' && mfaType !== 'u') {
```
ibeckermayer, line 406: See above comment

ibeckermayer (changes requested):

packages/shared/services/types.ts
kimlisa: i added a TODO to remove it later, it will be in next PR b/c there are several places it's affected (it [doesn't do anything](https://github.com/gravitational/webapps/pull/711/commits/f7b1a313530a10524af5268ee2d80896617ad75d#diff-3120151bac13eea22702b220d01a4b2f88201a6ca9f4114ce0d2953986b2e72dR51) atm)

packages/teleport/src/lib/tdp/codec.ts
kimlisa: reverted the file

packages/teleport/src/lib/tdp/codec.ts
kimlisa: reverted

packages/teleterm/src/ui/ClusterConnect/ClusterLogin/FormLogin/FormLogin.tsx
```diff
@@ -170,12 +170,6 @@ export default function LoginForm(props: Props) {
                         mb={0}
                       />
                     )}
-                    {mfaType.value === 'u2f' && isProcessing && (
-                      <Text typography="body2" mb={1}>
-                        Insert your hardware key and press the button on the
-                        key.
-                      </Text>
-                    )}
```
ravicious, line 178: I think the Teleterm code might require some more changes since that's the only place that asks users to press the button on the hardware key, but with webauthn that is still required, isn't it?

I'll try to take a look in an hour but if I don't manage to then I'll submit a patch on Monday.

kimlisa: the webauthn relies on the browsers dialog. for u2f, this prompt was needed b/c chrome didn't have a browser dialog (but firefox did)

ibeckermayer (approved):

packages/teleterm/src/ui/ClusterConnect/ClusterLogin/FormLogin/FormLogin.tsx
ravicious: Thanks! I also didn't realize that in Teleterm there's a `PromptHardwareKey` component for that.

gzdunek (approved):

zmb3 (commented):
+163 −2,331

Love it.

zmb3 (approved):

ravicious (commented):
I have a bunch of questions which I think stem mostly from me not understanding our release timelines and the timeline of sunsetting u2f.

* Is this change going to land in 9.1 or 10?
* I tested this on a bunch of clusters with different configurations, unfortunately I don't have a v9 cluster running which would not have u2f aliasing enabled (gravitational/teleport@f8b7b330ad9). As far as I understand, the app is still supposed to work with v9 clusters that use u2f?
* The web UI and Teleterm crash when trying to connect to my v8 cluster in the cloud. Teleterm is supposed to come out in 9.2, so I think it should still work with the previous major version. That probably means it should still work with u2f?
   * Teleterm fetches auth preferences straight from `/web/config.js`, so perhaps we should add aliasing there in the Go code.
* `packages/teleport/src/services/mfa/makeMfaDevice.ts` has one line related to u2f, is it supposed to be left there for now?

zmb3:
> Is this change going to land in 9.1 or 10?

In 10. This won't be backported to branch/v9.

ravicious:
Ah, thanks Zac, I guess this resolves my third point as well: we'll backport to 9.2 only the original PR with Teleterm without these changes here.

packages/teleterm/src/ui/ClusterConnect/ClusterLogin/FormLogin/FormLogin.tsx
kimlisa: @ravicious i just played with teleterm, and i now understood your concern, did not know how teleterm worked at all :sweat_smile: thanks for the check!

hatched (approved):

kimlisa:
> packages/teleport/src/services/mfa/makeMfaDevice.ts has one line related to u2f, is it supposed to be left there for now?

previous devices registered as `u2f` still comes back as type `u2f`

PR #712, opened 2022-03-31
Prevent crash when network or cluster is offline

gzdunek:
Closes https://github.com/gravitational/webapps.e/issues/204

### DIFF
--- a/packages/teleterm/src/ui/services/clusters/clustersService.ts
+++ b/packages/teleterm/src/ui/services/clusters/clustersService.ts
@@ -61,15 +61,10 @@ export class ClustersService extends ImmutableStore<ClustersServiceState> {
 
   async syncRootCluster(clusterUri: string) {
     try {
-      await this.syncClusterInfo(clusterUri);
-
-      this.syncLeafClusters(clusterUri);
-      this.syncKubes(clusterUri);
-      this.syncApps(clusterUri);
-      this.syncDbs(clusterUri);
-      this.syncServers(clusterUri);
-      this.syncKubes(clusterUri);
-      this.syncGateways();
+      await Promise.all([
+        this.syncClusterInfo(clusterUri),
+        this.syncLeafClusters(clusterUri),
+      ]);
     } catch (e) {
       this.notificationsService.notifyError({
         title: `Could not synchronize cluster ${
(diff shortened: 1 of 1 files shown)


packages/teleterm/src/ui/services/clusters/clustersService.ts
```diff
@@ -61,24 +61,24 @@ export class ClustersService extends ImmutableStore<ClustersServiceState> {
...
         description: e.message,
       });
-      throw e;
     }
+    this.syncKubes(clusterUri);
+    this.syncApps(clusterUri);
+    this.syncDbs(clusterUri);
+    this.syncServers(clusterUri);
+    this.syncKubes(clusterUri);
+    this.syncGateways();
```
gzdunek, line 81: `syncGateways` doesn't depend on `clusterUri`, so it could be invoked once (not for each cluster), but I guess it was done this way to make sure that it is always up to date.

ravicious, line 81: Are these outside of the `try` block because we're ignoring errors from them anyway?

packages/teleterm/src/ui/services/clusters/clustersService.ts
```diff
@@ -114,10 +122,17 @@ export class ClustersService extends ImmutableStore<ClustersServiceState> {
...
-      draft.gateways = new Map(gws.map(g => [g.uri, g]));
-    });
+    try {
+      const gws = await this.client.listGateways();
+      this.setState(draft => {
+        draft.gateways = new Map(gws.map(g => [g.uri, g]));
+      });
+    } catch (error) {
+      this.notificationsService.notifyError({
+        title: 'Could not fetch databases',
```
ravicious, line 132: This name is okay for now but we should change it once we start using gateways for something else than just databases.

Though, with the recent changes to `tsh proxy db`, Zac mentioned how we can consider to stop using "gateway" as a concept and instead just execute `tsh proxy db` underneath. The implementation of it is still unclear to me, but I should know more after I work on the changes to proxies that Roman introduced.

ravicious (approved):
Tested this by disconnecting from the internet and it works quite nicely. 👍

In the future I can imagine that we'd want to perhaps clear the error notifications automatically once we manage to sync successfully, so that the user doesn't have to remove them manually one-by-one.

packages/teleterm/src/ui/services/clusters/clustersService.ts
gzdunek: They are outside because each method has its own error handling. Thanks to this we can show a notification from `try catch` and a label above each table.

packages/teleterm/src/ui/services/clusters/clustersService.ts
gzdunek: I understand, I added it here because this method is used in two places, so I wanted to avoid duplication of error messages.

PR #713, opened 2022-03-31
Use dark background for app loader

ravicious:
Just a quick workaround to avoid sudden flash of white background.


gzdunek (commented):
I think we should move `appContext.int()` and 'Loading' to `AppInitializer` (which I removed in my PR https://github.com/gravitational/webapps/pull/712/files 🙃).
It will solve that problem right away as it is placed inside `<ThemeProvider />`.

gzdunek:
Closing in favor of https://github.com/gravitational/webapps/pull/722

PR #716, opened 2022-04-01
Use teleterm/logger in runtimeSettings

ravicious:
While trying to fix the issue with tsh connection on startup, I tried to use the Logger that was already imported in `runtimeSettings`. It turned out that attempting to use it would always crash the app, as the logger from shared/libs/logger attempts to call `window`, which doesn't exist in the context of Electron main process.

So instead we just use the one from `teleterm/logger` which calls `console` directly.


gzdunek (approved):
Thanks for catching this!

PR #718, opened 2022-04-04
Resolve shell env

gzdunek:
The implementation is mostly based on `vsocde` solution (I've put link in the code).

I'm not particularly happy with the error handling. Reading env is not critical to the terminal work, so we should just show the notification and continue without env. Unfortunately, there is no way to trigger a notification from the main process and the only place to do so is the `useDocumentTerminal`.  I think we should somehow allow showing notifications from the main process.  

Closes https://github.com/gravitational/webapps.e/issues/194


packages/teleterm/src/services/pty/resolveShellEnv.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  const env = {
+    ...process.env,
+    ELECTRON_RUN_AS_NODE: '1',
+    ELECTRON_NO_ATTACH_CONSOLE: '1',
+  };
+
+  return new Promise<typeof process.env>((resolve, reject) => {
+    const command = `'${process.execPath}' -p '"${mark}" + JSON.stringify(process.env) + "${mark}"'`;
+    const shellArgs = shell === '/bin/tcsh' ? ['-ic'] : ['-ilc'];
```
gzdunek, line 87: Does it have to be `interactive`?

packages/teleterm/src/services/pty/ptyService.ts
```diff
@@ -55,18 +56,27 @@ export default function createPtyService(
...
+): Promise<PtyOptions> {
   const env = {
+    ...process.env,
     TELEPORT_HOME: settings.tshd.homeDir,
   };
 
+  const localShellEnv = {
+    ...env,
+    ...(await resolveShellEnvCached(settings.defaultShell)),
+  };
```
ravicious, line 71: I'm not sure if this should be this way but I'm running out of time for today, I'll check it tomorrow morning + I have a solution for setting the correct cwd instead of opening the shell in `/`.

packages/teleterm/src/services/pty/resolveShellEnv.ts
ravicious: Matter of fact, we should perhaps link to this stackexchange post in case anyone has the same question in the future. ;)

packages/teleterm/src/services/pty/ptyService.ts
ravicious: Okay, so I think the priority should be this:

1. TELEPORT_HOME
2. resolveShellEnvCached
3. process.env

What was clear to you but took a second for me is that obviously we want `resolveShellEnvCached` to override whatever path is set in `process.env`, because that's the problem we're solving here – launching the app from outside of the terminal doesn't set path correctly.

Now, I wonder what we should do about `TELEPORT_HOME`. The current code is going to override it with whatever comes from user's env. Regular Teleport users probably won't have this env var set anyway, but powerusers might in which case they might not expect Teleterm to use it and it'll mess with their tsh directory. OTOH if always override `TELEPORT_HOME` with `settings.tshd.homeDir`, users won't be able to easily change that. But for the preview release that should be okay? If someone really needs that they can probably symlink `~/Library/Application\ Support/teleterm/tsh` to whatever they want. 🤔

Now that I think about it, we probably don't even need to include stuff from `process.env` when launching a shell. `resolveShellEnvCached` already gives us all the env vars we need. That being said, it's hard for me to say if this is going to have any unforeseen consequences, so I cannot really recommend that we do it.

ravicious:
The way to set correct cwd is this:

```diff
diff --git a/packages/teleterm/src/services/pty/ptyProcess.ts b/packages/teleterm/src/services/pty/ptyProcess.ts
index 43384017..a58b0eca 100644
--- a/packages/teleterm/src/services/pty/ptyProcess.ts
+++ b/packages/teleterm/src/services/pty/ptyProcess.ts
@@ -43,7 +43,9 @@ class PtyProcess extends EventEmitter {
       cols,
       rows,
       name: 'xterm-color',
-      cwd: this.options.cwd || process.cwd(),
+      // HOME should be always defined. But just in case it isn't let's use the cwd from process.
+      // https://unix.stackexchange.com/questions/123858
+      cwd: this.options.cwd || process.env['HOME'] || process.cwd(),
       env: {
         ...process.env,
         ...this.options.env,
```

ravicious:
It might be a good idea to cut a new demo build after merging this, I imagine this is a huge issue if you don't know that starting the app from the terminal fixes this.

packages/teleterm/src/services/pty/resolveShellEnv.ts
```diff
@@ -0,0 +1,139 @@
...
+ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+ */
+
+//Based on https://github.com/microsoft/vscode/blob/main/src/vs/platform/shell/node/shellEnv.ts
```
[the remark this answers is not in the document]
gzdunek: Done

packages/teleterm/src/services/pty/resolveShellEnv.ts
```diff
@@ -0,0 +1,139 @@
...
+  abortSignal: AbortSignal,
+): Promise<typeof process.env> {
+  const mark = unique().replace(/-/g, '').substring(0, 12);
+  const regex = new RegExp(mark + '(.*)' + mark);
+
+  const env = {
+    ...process.env,
+    ELECTRON_RUN_AS_NODE: '1',
+    ELECTRON_NO_ATTACH_CONSOLE: '1',
+  };
```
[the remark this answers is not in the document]
gzdunek: Good point, done

packages/teleterm/src/services/pty/ptyService.ts
gzdunek: I don't have any strong opinion on envs priority, but it seems ok to override `TELEPORT_HOME` with `settings.tshd.homeDir`.
Regarding `process.env` from what I see all of its properties are overridden by the resolved shell, but I'm not sure if we should remove it, I think I'd leave it as it is.

gzdunek:
@ravicious what about using our resolved env here?
```cwd: this.options.cwd || this.options.env['HOME'] || process.cwd()``

**Review by ravicious — COMMENTED** [review]
I think there's just one small change we should introduce and then it's ready to go.

---

> what about using our resolved env here?
> `cwd: this.options.cwd || this.options.env['HOME'] || process.cwd()`

Good point, idk how I missed that!

File: packages/teleterm/src/services/pty/ptyService.ts [hunk 5]
```diff
@@ -55,24 +56,37 @@ export default function createPtyService(
...
 
-function buildOptions(settings: RuntimeSettings, cmd: PtyCommand): PtyOptions {
+async function buildOptions(
+  settings: RuntimeSettings,
+  cmd: PtyCommand
+): Promise<PtyOptions> {
   const env = {
+    ...process.env,
     TELEPORT_HOME: settings.tshd.homeDir,
   };
```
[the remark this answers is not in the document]
gzdunek: I was wondering if we need to pass that resolved env for a remote session, but you're right, it is better to always have correct env. Fix pushed.

ravicious (approved):

PR #720, opened 2022-04-05
Simplify the db connection tab

ravicious:
As described in gravitational/webapps.e#177, we want to replace the db
tab with just two sections:

* "Connect with CLI" which will show the command to use in terminal
* "Connect with GUI" which links to our documentation

After gravitational/teleport#11720 gets merged, the "Connect with CLI"
section will be massively simplified: it'll be basically just something like:

    psql postgres://localhost:12345

Moreover, @smallinsky suggested that tsh should be responsible for creating
those CLI connection commands. We should also do this in the future as
it'll let us support new protocols as soon as they land in tsh, without
us having to touch Teleterm codebase, for the most part.


packages/teleterm/src/ui/DocumentGateway/DocumentGateway.tsx
```diff
@@ -71,20 +71,55 @@ export function DocumentGateway(props: State) {
...
+      break;
     case 'mysql':
-      return (
-        <MySql gateway={gateway} title={doc.title} disconnect={disconnect} />
-      );
-    default:
-      return <Box> {`Unknown protocol type ${gateway.protocol}`}</Box>;
+      cliConnectionString = useMySql(gateway);
+      break;
   }
```
ravicious, line 86: As mentioned in the PR description, this will get removed at some point before the preview release in favor of tsh doing this for Teleterm.

packages/teleterm/src/ui/DocumentGateway/Mongo/useMongo.ts
```diff
@@ -16,15 +16,7 @@ limitations under the License.
...
-export type MongoProps = {
-  title: string;
-  disconnect(): void;
-  gateway: Gateway;
-};
-
-export function useMongo(props: MongoProps) {
-  const { gateway, disconnect, title } = props;
-
+export function useMongo(gateway: Gateway) {
```
ravicious, line 19: Technically these `useMongo` type functions no longer need to live in protocol-specific folder, but we're going to remove them anyway later on. So I just kept them where they are.

packages/teleterm/src/ui/DocumentGateway/DocumentGateway.tsx
```diff
@@ -71,20 +71,55 @@ export function DocumentGateway(props: State) {
     );
   }
 
+  let cliConnectionString: string;
```
gzdunek, line 74: In such cases I like to create a new function, so I can avoid mutations:
```typescript
function getCliConnectionString(gateway: Gateway) {
  switch (gateway.protocol as GatewayProtocol) {
    case 'mongodb':
      return useMongo(gateway);
    case 'postgres':
      return usePostgres(gateway);
    case 'mysql':
      return useMySql(gateway);
  }
}
``` 

And theoretically, we shouldn't use `use...` names as they're reserved for hooks and we can't use hooks conditionally (like in switch-case statements), but I'm fine with leaving it if we are going to remove them anyway.

gzdunek (approved):

ravicious:
Request from Roman:

> for now, can we just add Redis and SQL Server commands to the frontend? it’s just a couple if simple switch cases
for Redis the command would be `redis-cli -h localhost -p <port>`
for MSSQL: `mssql-cli -S localhost,<port> -U <user> -P <random string>` (password is required by MSSQL client IIRC but doesn’t matter since we’re connecting to local proxy)

ravicious:
Actually, I'm going to make a separate PR with that Redis and MSSQL support. It doesn't make sense to cram in more stuff here.

PR #721, opened 2022-04-05
Add documents reopen dialog

gzdunek:
Adds dialog window instead of a notification.
Screenshot 2022-04-05 at 14 17 09

It also makes usage of `openClusterConnectDialog()` more clear - all flows are now fully wrapped in `Promise` (previously `resolve() was invoked only on `onSuccess`).

Closes: https://github.com/gravitational/webapps.e/issues/178


ravicious:
I tried to test this, but it seems that no matter which option I pick, Teleterm restores the tabs.

gzdunek:
What a shame 🙄, fix pushed

packages/teleterm/src/ui/ClusterConnect/ClusterLogin/useClusterLogin.ts
```diff
@@ -84,7 +84,6 @@ export default function useClusterLogin(props: Props) {
     }
 
     if (loginAttempt.status === 'success') {
-      onClose();
```
ravicious, line 87: Was this call meant to be removed from here?

packages/teleterm/src/ui/services/workspacesService/workspacesService.ts
```diff
@@ -134,17 +134,26 @@ export class WorkspacesService extends ImmutableStore<WorkspacesState> {
...
-      if (clusterUri && !isConnected) {
-        this.modalsService.openClusterConnectDialog(clusterUri, () => {
-          setWorkspace();
-          resolve();
-        });
-      } else {
+    const cluster = this.clustersService.findCluster(clusterUri);
+    if (!cluster) {
+      return Promise.resolve();
+    }
```
ravicious, line 140: Which scenario does this code cover? It seems like it's going to behave differently now if the cluster wasn't found?

ravicious (approved):
The only thing that's surprising for me is that we ask this question each time you open one of your existing clusters, rather than just once at the startup – is that the intended behavior?

packages/teleterm/src/ui/services/workspacesService/workspacesService.ts
gzdunek: One or two times I had a situation where cluster wasn't found and our code tried to operate on that `undefined` cluster which resulted in error (we were showing login dialog and that failed with error `cannot read name (clusterName) of undefined`). 
The worst thing is that I can't remember why it happened :( 
I added a notification and logging, so we will be able to track this problem.

gzdunek:
IMO we should try to reopen tabs for each workspace, not only for the first one on startup

PR #722, opened 2022-04-05
App initialization improvements 

gzdunek:
Thanks to `BrowserWindow.backgroundColor` we won't have white splash before _Loading_.

### DIFF
--- a/packages/teleterm/src/mainProcess/mainProcess.ts
+++ b/packages/teleterm/src/mainProcess/mainProcess.ts
@@ -16,6 +16,7 @@ import {
   subscribeToConfigServiceEvents,
 } from '../services/config';
 import { subscribeToTabContextMenuEvent } from './contextMenus/tabContextMenu';
+import theme from 'teleterm/ui/ThemeProvider/theme';
 
 type Options = {
   settings: RuntimeSettings;
@@ -52,6 +53,7 @@ export default class MainProcess {
       height,
       title: 'Teleport Terminal',
       icon: getAssetPath('icon.png'),
+      backgroundColor: theme.colors.primary.main,
       webPreferences: {
         contextIsolation: true,
         nodeIntegration: false,
--- a/packages/teleterm/src/ui/AppInitializer.tsx
+++ b/packages/teleterm/src/ui/AppInitializer.tsx
@@ -1,18 +1,23 @@
-import React, { ReactNode, useEffect } from 'react';
+import React, { FC, useEffect } from 'react';
 import { useAppContext } from 'teleterm/ui/appContextProvider';
+import useAsync from 'teleterm/ui/useAsync';
+import styled from 'styled-components';
 
-interface AppInitializerProps {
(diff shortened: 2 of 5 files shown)


packages/teleterm/src/mainProcess/mainProcess.ts
```diff
@@ -52,6 +52,7 @@ export default class MainProcess {
       height,
       title: 'Teleport Terminal',
       icon: getAssetPath('icon.png'),
+      backgroundColor: '#01172C',
```
ravicious, line 55: What do you think about making `getPlatform` compatible with Electron and then doing this?

```suggestion
      backgroundColor: theme.colors.primary.main,
```


`packages/design/src/theme/utils/getPlatform.js`:

```typescript
export default function getPlatform() {
  if (typeof window !== 'undefined') {
    const userAgent = window.navigator.userAgent;

    return {
      isWin: userAgent.indexOf('Windows') >= 0,
      isMac: userAgent.indexOf('Macintosh') >= 0,
      isLinux: userAgent.indexOf('Linux') >= 0,
    };
  } else {
    const platform = process.platform;
    const isWin = platform === 'win32';
    const isMac = platform === 'darwin';

    return {
      isWin,
      isMac,
      isLinux: !isWin && !isMac,
    };
  }
}
```

packages/teleterm/src/ui/AppInitializer.tsx
```diff
@@ -1,18 +1,23 @@
...
 import { useAppContext } from 'teleterm/ui/appContextProvider';
+import useAsync from 'teleterm/ui/useAsync';
+import styled from 'styled-components';
 
-interface AppInitializerProps {
-  children?: ReactNode;
-}
-
-export function AppInitializer(props: AppInitializerProps) {
+export const AppInitializer: FC = props => {
```
ravicious, line 6: What does moving this logic here gives us? From what I can tell, the loading message appears later than before.

ravicious (approved):

packages/teleterm/src/mainProcess/mainProcess.ts
gzdunek: Good advice, it will make our life easier when updating the theme.

packages/teleterm/src/ui/AppInitializer.tsx
gzdunek: IMO there was no benefit of doing it in `boot.tsx` (maybe it was a bit faster, but I think it can be counted in milliseconds). By moving it to the `AppInitializer` we have correct styling for `Loading` out of the box, provided by global style from `styled-components` (and in `boot.tsx` we couldn't even use it). 
Additionally, in your PR https://github.com/gravitational/webapps/pull/713/files you had to manually import `theme`.  It has a drawback of not being responsive to changes, so when we would add a white theme we would have to import another file (but we will still have to handle it for `BrowserWindow.backgroundColor`).

PR #724, opened 2022-04-05
Passwordless wiring

kimlisa:
#### Description
Updates api request bodies to pass on passwordless requests. Does not change any functionality, just wiring (still waiting on final design)

Note: `teleterm` passwordless work will be done separately

#### Related PR
To enable passwordless registration (for invites and resets), this [PR](https://github.com/gravitational/teleport/pull/11748) needs to be merged

### DIFF
--- a/packages/teleport/src/services/mfa/types.ts
+++ b/packages/teleport/src/services/mfa/types.ts
@@ -1,3 +1,19 @@
+/**
+ * Copyright 2022 Gravitational, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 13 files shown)


ravicious (approved):

packages/teleport/src/services/mfa/types.ts
```diff
@@ -15,4 +31,8 @@ export type AddNewTotpDeviceRequest = {
 export type AddNewHardwareDeviceRequest = {
   tokenId: string;
   deviceName: string;
+  deviceUsage?: DeviceUsage;
 };
+
+export type DeviceType = 'totp' | 'webauthn';
+export type DeviceUsage = 'passwordless' | 'mfa';
```
rudream, line 38: This name is a bit confusing imo, why not something like `AuthOption`?

rudream (approved):

packages/teleport/src/services/mfa/types.ts
kimlisa: `DeviceUsage` is sim to how it is defined in the back, it describes how a device will be used (a device can be used as passwordless, or as a multi factor e.g password + device)

PR #725, opened 2022-04-07
Store app size and position

gzdunek:
Closes https://github.com/gravitational/webapps.e/issues/166

It also solves the problem of not saving the application state when it is closed immediately after an action.

### DIFF
--- a/packages/teleterm/src/mainProcess/windowsManager.ts
+++ b/packages/teleterm/src/mainProcess/windowsManager.ts
@@ -1,0 +1,98 @@
+import { BrowserWindow, Rectangle, screen } from 'electron';
+import { getAssetPath } from 'teleterm/mainProcess/runtimeSettings';
+import path from 'path';
+import { FileStorage } from 'teleterm/services/fileStorage';
+import { RuntimeSettings } from 'teleterm/mainProcess/types';
+import theme from 'teleterm/ui/ThemeProvider/theme';
+
+type WindowState = Rectangle;
+
+export class WindowsManager {
+  private storageKey = 'windowState';
+
+  constructor(
+    private fileStorage: FileStorage,
+    private settings: RuntimeSettings
+  ) {}
+
+  createWindow(): void {
+    const windowState = this.getWindowState();
+    const window = new BrowserWindow({
+      x: windowState.x,
+      y: windowState.y,
+      width: windowState.width,
+      height: windowState.height,
+      backgroundColor: theme.colors.primary.main,
+      minWidth: 400,
+      minHeight: 300,
+      title: 'Teleport Terminal',
+      icon: getAssetPath('icon.png'),
+      webPreferences: {
+        contextIsolation: true,
+        nodeIntegration: false,
+        preload: path.join(__dirname, 'preload.js'),
+      },
+    });
+
+    window.once('close', () => {
+      this.saveWindowState(window);
+    });
+
+    if (this.settings.dev) {
+      window.loadURL('https://localhost:8080');
+    } else {
+      window.loadFile(path.join(__dirname, '../renderer/index.html'));
+    }
+  }
+
+  private saveWindowState(window: BrowserWindow): void {
+    const windowState: WindowState = {
+      ...window.getNormalBounds(),
+    };
+
+    this.fileStorage.put(this.storageKey, windowState);
+  }
+
+  private getWindowState(): WindowState {
+    const windowState = this.fileStorage.get<WindowState>(this.storageKey);
+    const getDefaults = () => ({
+      height: 720,
+      width: 1280,
+      x: undefined,
+      y: undefined,
+    });
+
+    if (!windowState) {
+      return getDefaults();
+    }
+
+    const getPositionAndSize = () => {
+      const displayBounds = screen.getDisplayNearestPoint({
+        x: windowState.x,
+        y: windowState.y,
+      }).bounds;
+
+      const isWindowWithinDisplayBounds =
+        windowState.x >= displayBounds.x &&
+        windowState.y >= displayBounds.y &&
+        windowState.x + windowState.width <=
+          displayBounds.x + displayBounds.width &&
+        windowState.y + windowState.height <=
+          displayBounds.y + displayBounds.height;
+
+      if (isWindowWithinDisplayBounds) {
+        return {
+          x: windowState.x,
+          y: windowState.y,
+          width: windowState.width,
+          height: windowState.height,
+        };
+      }
+    };
+
+    return {
+      ...getDefaults(),
+      ...getPositionAndSize(),
+    };
+  }
+}
--- a/packages/teleterm/src/main.ts
+++ b/packages/teleterm/src/main.ts
@@ -6,8 +6,13 @@ import createLoggerService from 'teleterm/services/logger';
 import Logger from 'teleterm/logger';
 import * as types from 'teleterm/types';
 import { ConfigServiceImpl } from 'teleterm/services/config';
+import { createFileStorage } from 'teleterm/services/fileStorage';
+import path from 'path';
 
 const settings = getRuntimeSettings();
+const fileStorage = createFileStorage({
+  filePath: path.join(settings.userDataDir, 'app_state.json'),
+});
 const logger = initMainLogger(settings);
 const configService = new ConfigServiceImpl();
 
@@ -17,13 +22,19 @@ process.on('uncaughtException', error => {
 });
 
 // init main process
-const mainProcess = MainProcess.create({ settings, logger, configService });
+const mainProcess = MainProcess.create({
+  settings,
+  logger,
+  configService,
+  fileStorage,
+});
 
 // node-pty is not yet context aware
 app.allowRendererProcessReuse = false;
 app.commandLine.appendSwitch('ignore-certificate-errors', 'true');
 
 app.on('will-quit', () => {
+  fileStorage.putAllSync();
   globalShortcut.unregisterAll();
   mainProcess.dispose();
 });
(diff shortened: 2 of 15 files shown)


packages/teleterm/src/mainProcess/windowsManager.ts
```diff
@@ -0,0 +1,96 @@
+import { BrowserWindow, Rectangle, screen } from 'electron';
+import { getAssetPath } from 'teleterm/mainProcess/runtimeSettings';
+import path from 'path';
+import { FileStorage } from 'teleterm/services/fileStorage';
+import { RuntimeSettings } from 'teleterm/mainProcess/types';
+import theme from 'teleterm/ui/ThemeProvider/theme';
+
+type WindowState = Rectangle;
```
gzdunek, line 8: I wanted to store maximised/fullscreen info as well, but I couldn't force it to work reliably when using multiple displays.
When I close the app on the second screen I would to see it on the same screen when it is opened. Unfortunately, it opens on the first screen when the app was closed in maximised/full screen state.

But THB, I'm not sure it's that important. Apple native apps/Chrome don't work that way. When you close a maximized app, it will open at its normal size.

packages/teleterm/src/mainProcess/windowsManager.ts
```diff
@@ -0,0 +1,96 @@
...
+        x: windowState.x,
+        y: windowState.y,
+      }).bounds;
+      if (
+        windowState.x >= displayBounds.x &&
+        windowState.y >= displayBounds.y &&
+        windowState.x + windowState.width <=
+          displayBounds.x + displayBounds.width &&
+        windowState.y + windowState.height <=
+          displayBounds.y + displayBounds.height
```
ravicious, line 80: What does this check do? I think it might be worthwhile to put the result into a named variable.

packages/teleterm/src/mainProcess/windowsManager.ts
```diff
@@ -0,0 +1,96 @@
...
+    if (this.settings.dev) {
+      window.loadURL('https://localhost:8080');
+    } else {
+      window.loadFile(path.join(__dirname, '../renderer/index.html'));
+    }
+  }
+
+  private saveWindowState(window: BrowserWindow): void {
+    const windowState: WindowState = {
+      ...window.getNormalBounds(),
```
ravicious, line 50: > Note: whatever the current state of the window : maximized, minimized or in fullscreen, this function always returns the position and size of the window in normal state. In normal state, getBounds and getNormalBounds returns the same [Rectangle](https://www.electronjs.org/docs/latest/api/structures/rectangle).

Did you experiment with using `getBounds` here? I don't have time to check it properly, but when I used `getBounds` here, the app was able to restore the window to the correct display when it was in a simple fullscreen mode and in normal fullscreen mode. In normal fullscreen mode it restored the app to just simple fullscreen, but I think that's okay for now.

gzdunek: TBH, `getBounds` solves one problem (window is restored maximised), but it creates another (if you try to unmaxmize the window by using double click it became very small - in our case 400x300 - the minimal size). For me, the most correct way is to save dimensions from `normalBounds`, because to create other states (`fullScreen/maxmized`) we don't need any previous information.

I've just spend some time experimenting with `getBounds()`. I wanted to save both info, then use `getBounds()` to restore correct screen, and `getNormalBound()` to restore correct size and placement, but it still doesn't work for fullscreen mode :(

ravicious (approved):
For the record, here's my version with some added logging:

<details>
<summary>browser windows with logging</summary>

```diff
diff --git a/packages/teleterm/src/mainProcess/windowsManager.ts b/packages/teleterm/src/mainProcess/windowsManager.ts
index c641b879..ba238f61 100644
--- a/packages/teleterm/src/mainProcess/windowsManager.ts
+++ b/packages/teleterm/src/mainProcess/windowsManager.ts
@@ -5,7 +5,23 @@ import { FileStorage } from 'teleterm/services/fileStorage';
 import { RuntimeSettings } from 'teleterm/mainProcess/types';
 import theme from 'teleterm/ui/ThemeProvider/theme';
 
-type WindowState = Rectangle;
+type WindowState = Rectangle & {
+  mode: WindowMode;
+  maxBounds: Rectangle;
+};
+
+let counter = 0;
+
+function getCounter() {
+  counter += 1;
+  return counter;
+}
+
+enum WindowMode {
+  Normal = 'Normal',
+  Maximized = 'Maximized',
+  Fullscreen = 'Fullscreen',
+}
 
 export class WindowsManager {
   private storageKey = 'windowState';
@@ -17,12 +33,14 @@ export class WindowsManager {
 
   createWindow(): void {
     const windowState = this.getWindowState();
+    console.log('Restored window state', windowState);
     const window = new BrowserWindow({
       x: windowState.x,
       y: windowState.y,
       width: windowState.width,
       height: windowState.height,
       backgroundColor: theme.colors.primary.main,
+      maximizable: true,
       minWidth: 400,
       minHeight: 300,
       title: 'Teleport Terminal',
@@ -34,9 +52,81 @@ export class WindowsManager {
       },
     });
 
-    window.once('close', () => {
+    // window.once('close', () => {
+    //   this.saveWindowState(window);
+    // });
+
+    const getWindowMode = () => {
+      if (window.isFullScreen()) {
+        return WindowMode.Fullscreen;
+      }
+      if (window.isMaximized()) {
+        return WindowMode.Maximized;
+      }
+      return WindowMode.Normal;
+    };
+    window.on('resized', () => {
+      console.log('resized', {
+        counter: getCounter(),
+        bounds: window.getBounds(),
+        normalBounds: window.getNormalBounds(),
+        mode: getWindowMode(),
+      });
+      this.saveWindowState(window);
+    });
+    window.on('moved', () => {
+      console.log('moved', {
+        counter: getCounter(),
+        bounds: window.getBounds(),
+        normalBounds: window.getNormalBounds(),
+        mode: getWindowMode(),
+      });
+      this.saveWindowState(window);
+    });
+    window.on('maximize', () => {
+      console.log('maximize', {
+        counter: getCounter(),
+        bounds: window.getBounds(),
+        normalBounds: window.getNormalBounds(),
+        mode: getWindowMode(),
+      });
+      this.saveWindowState(window);
+    });
+    window.on('unmaximize', () => {
+      console.log('unmaximize', {
+        counter: getCounter(),
+        bounds: window.getBounds(),
+        normalBounds: window.getNormalBounds(),
+        mode: getWindowMode(),
+      });
       this.saveWindowState(window);
     });
+    window.on('enter-full-screen', () => {
+      console.log('enter-full-screen', {
+        counter: getCounter(),
+        bounds: window.getBounds(),
+        normalBounds: window.getNormalBounds(),
+        mode: getWindowMode(),
+      });
+      this.saveWindowState(window);
+    });
+    window.on('leave-full-screen', () => {
+      console.log('leave-full-screen', {
+        counter: getCounter(),
+        bounds: window.getBounds(),
+        normalBounds: window.getNormalBounds(),
+        mode: getWindowMode(),
+      });
+      this.saveWindowState(window);
+    });
+
+    if (windowState.mode === WindowMode.Fullscreen) {
+      window.setFullScreen(true);
+    }
+
+    if (windowState.mode === WindowMode.Maximized) {
+      window.maximize();
+    }
 
     if (this.settings.dev) {
       window.loadURL('https://localhost:8080');
@@ -46,10 +136,23 @@ export class WindowsManager {
   }
 
   private saveWindowState(window: BrowserWindow): void {
+    const getWindowMode = () => {
+      if (window.isFullScreen()) {
+        return WindowMode.Fullscreen;
+      }
+      if (window.isMaximized()) {
+        return WindowMode.Maximized;
+      }
+      return WindowMode.Normal;
+    };
     const windowState: WindowState = {
       ...window.getNormalBounds(),
+      mode: getWindowMode(),
+      maxBounds: window.getBounds(),
     };
 
+    console.log('Saved window state', windowState);
+
     this.fileStorage.put(this.storageKey, windowState);
   }
 
@@ -60,6 +163,13 @@ export class WindowsManager {
       width: 1280,
       x: undefined,
       y: undefined,
+      mode: WindowMode.Normal,
+      maxBounds: {
+        x: undefined,
+        y: undefined,
+        width: undefined,
+        height: undefined,
+      },
     });
 
     if (!windowState) {
@@ -67,19 +177,23 @@ export class WindowsManager {
     }
 
     const getPositionAndSize = () => {
-      const displayBounds = screen.getDisplayNearestPoint({
-        x: windowState.x,
-        y: windowState.y,
+      const displayBounds = screen.getDisplayMatching({
+        x: windowState.maxBounds.x,
+        y: windowState.maxBounds.y,
+        width: windowState.maxBounds.width,
+        height: windowState.maxBounds.height,
       }).bounds;
 
       const isWindowWithinDisplayBounds =
-        windowState.x >= displayBounds.x &&
-        windowState.y >= displayBounds.y &&
-        windowState.x + windowState.width <=
+        windowState.maxBounds.x >= displayBounds.x &&
+        windowState.maxBounds.y >= displayBounds.y &&
+        windowState.maxBounds.x + windowState.maxBounds.width <=
           displayBounds.x + displayBounds.width &&
-        windowState.y + windowState.height <=
+        windowState.maxBounds.y + windowState.maxBounds.height <=
           displayBounds.y + displayBounds.height;
 
+      console.log('isWindowWithinDisplayBounds', isWindowWithinDisplayBounds);
+
       if (isWindowWithinDisplayBounds) {
         return {
           x: windowState.x,
@@ -90,9 +204,16 @@ export class WindowsManager {
       }
     };
 
+    const getWindowMode = () => {
+      return {
+        mode: windowState.mode,
+      };
+    };
+
     return {
       ...getDefaults(),
       ...getPositionAndSize(),
+      ...getWindowMode(),
     };
   }
 }
```
</details>

Let's merge it as it is and maybe Electron will have a solution ready in the future. Could you write down the issues with the current solution in a separate ticket though?

PR #726, opened 2022-04-08
Use DB CLI commands provided by tsh daemon

ravicious:
This PR makes it so that Teleterm uses the DB CLI commands from the tsh daemon, implemented in gravitational/teleport#11835. tsh needs to be compiled with those changes for this code to work.

Field names in proto files were reused, because we don't need to worry about backwards compatibility at this point.

* Part of gravitational/webapps.e#177
* Fixes gravitational/webapps.e#173
* Fixes gravitational/webapps.e#217
* Fixes gravitational/webapps.e#219


gzdunek (approved):
LGTM!

PR #727, opened 2022-04-12
Simplify terminal UI

gzdunek:
Closes https://github.com/gravitational/webapps.e/issues/209 
I added `className` prop in a few places, so we can use `css` prop on React components (we have to pass down `className` to the element we want to style).


packages/teleterm/src/ui/TopBar/TopBar.tsx
```diff
@@ -33,24 +45,11 @@ const Grid = styled.div`
...
-  align-items: center;
-  height: 100%;
-`;
-
-const JustifyRight = styled.div`
-  display: grid;
-  justify-self: end;
+  flex: 1;
+  margin-left: 22px;
+  margin-right: 10px;
```
ravicious, line 51: Could you change these values so that they're multiples of 4? It'll be easier to adjust the layout later if we keep this rule.

```suggestion
  margin-left: 24px;
  margin-right: 12px;
```

ravicious (approved):
I think it'd be better if the command bar stayed in the same place when switching between clusters, but I suppose Rishi would need to redesign the UI or specify how it should behave when changing dimensions and content that it displays.

gzdunek:
I reverted previous layout with small changes, so there should be no jumping. `CentralContainer`is centered relative to the application (slightly shifted to the left, but it looks natural).

ravicious:
Great job! It looks much better now.

PR #731, opened 2022-04-14
Fix path to packaged assets in Teleterm

ravicious:
When packaged, the main window opens this file:

    window.loadFile(path.join(__dirname, '../renderer/index.html'));

The assets are under ../renderer/assets, but since Webpack used absolute path for it, Electron tried to load them from /assets instead.

I'm not sure if that's the best way to modify our Webpack config, let me know if you have a better idea.

Fixes gravitational/webapps.e#153.

### DIFF
--- a/packages/teleterm/webpack.renderer.extend.js
+++ b/packages/teleterm/webpack.renderer.extend.js
@@ -5,7 +5,7 @@ const resolvepath = require('@gravitational/build/webpack/resolvepath');
 
 function extend(cfg) {
   cfg.entry = { app: ['./src/ui/boot'] };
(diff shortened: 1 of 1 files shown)


gzdunek (commented):
Seems like the `cfg.output.publicPath = 'auto'` does the trick as well https://webpack.js.org/guides/public-path/#automatic-publicpath

ravicious:
Ah, that's much better, I'll change it then.

ravicious:
The assets have double slash in them, but that should be okay for now.

gzdunek (approved):

PR #732, opened 2022-04-14
Use dropdown for the db connect button

ravicious:
This PR is part of the effort to simplify the db tab (gravitational/webapps.e#177).

This one replaces the modal with the same dropdown that we use for servers. For now the list of suggestions is empty. In an upcoming PR I'll add fetching available usernames from the backend.

Best reviewed commit-by-commit.


packages/shared/components/MenuLogin/types.ts
```diff
@@ -19,7 +19,7 @@ export type LoginItem = {
   login: string;
 };
 
-export type MenuSshLoginProps = {
+export type MenuLoginProps = {
   onOpen: () => LoginItem[];
```
gzdunek, line 23: Would you like to rename this prop to sth like `getLoginItems`? IMO `onOpen` is really confusing

packages/shared/components/MenuLogin/index.ts
```diff
@@ -14,6 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import MenuSshLogin from './MenuSshLogin';
-export default MenuSshLogin;
+import MenuLogin from './MenuLogin';
+export default MenuLogin;
```
gzdunek, line 18: Some time ago we decided to prefer named export over default, this is a great opportunity to change it here

gzdunek (approved):
Great job!

packages/shared/components/MenuLogin/types.ts
ravicious: Ah, good call, I was confused by this as well but I completely forgot to change it.

gzdunek (approved):

PR #733, opened 2022-04-14
DocumentGateway: Replace the Copy button with a Run button

ravicious:
Sasha asked about this change on Slack. For now I just copied the relevant pieces from `TextSelectCopy` implementation.

### DIFF
--- a/packages/teleterm/src/ui/DocumentGateway/DocumentGateway.tsx
+++ b/packages/teleterm/src/ui/DocumentGateway/DocumentGateway.tsx
@@ -19,6 +19,7 @@ import { Text, Flex, Box, ButtonPrimary, ButtonSecondary, Link } from 'design';
 import TextSelectCopy from 'teleport/components/TextSelectCopy';
 import Document from 'teleterm/ui/Document';
 import * as Alerts from 'design/Alert';
+import { useTheme } from 'styled-components';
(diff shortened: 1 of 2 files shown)


packages/teleterm/src/ui/DocumentGateway/DocumentGateway.tsx
```diff
@@ -101,3 +104,48 @@ export function DocumentGateway(props: State) {
...
+      mb={4}
+    >
+      <Flex
+        mr="2"
+        style={{
+          overflow: 'auto',
+          whiteSpace: 'pre',
+          wordBreak: 'break-all',
+          fontSize: '12px',
+          fontFamily: useTheme().fonts.mono,
```
gzdunek, line 131: Minor thing, you can use css prop - it has access to theme (and for me, syntax of `css` is nicer than of `style` :) )
```
css={{
          overflow: auto,
          white-space: pre,
          word-break: break-all,
          fonts-size: 12px,
          font-family: ${props => props.theme.fonts.mono},
        }}
```

gzdunek (approved):

packages/teleterm/src/ui/DocumentGateway/DocumentGateway.tsx
ravicious: That's indeed much better, thanks!

PR #734, opened 2022-04-14
Fix contrast issues

gzdunek:
Closes: https://github.com/gravitational/webapps.e/issues/140
Closes: https://github.com/gravitational/webapps.e/issues/228

### DIFF
--- a/packages/teleterm/src/ui/Tabs/TabItem.tsx
+++ b/packages/teleterm/src/ui/Tabs/TabItem.tsx
@@ -48,28 +48,32 @@ export function TabItem(props: TabItemProps) {
   };
 
   return (
-    <>
-      <StyledTabItem
-        onClick={onClick}
-        onContextMenu={onContextMenu}
-        ref={ref}
-        active={active}
-        dragging={isDragging}
-        title={name}
-        canDrag={canDrag}
-      >
-        <StyledTabButton>
-          <Text color="inherit" fontWeight={700} fontSize="12px">
-            {name}
-          </Text>
-        </StyledTabButton>
-        {onClose && (
-          <ButtonIcon size={0} mr={1} title="Close" onClick={handleClose}>
-            <CloseIcon fontSize="16px" />
-          </ButtonIcon>
-        )}
-      </StyledTabItem>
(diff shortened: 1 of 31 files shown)


ravicious (commented):
The new colors look great!

However, some of the changes broke something about adjusting xterm size. Now when I resize the window, xterm doesn't get readjusted to the new size and just cuts off a bit of the prompt.

image

ravicious:
Also, just discovered that if the notification title is too short, the button appears right after the text rather than being aligned to the right. This is something we can fix later, as right now I don't think we have any notification where this would be a problem.

image

gzdunek:
xterm cutting - fixed
notification layout - fixed

> I think I saw something in the console log about some element not being able to resize something because "the element appears to be hidden" or sth like that? I wasn't able to reproduce it though, maybe it was a red herring.

It works this way on purpose - we wan't to resize the visible terminal only (you will see this notification when you have at least two terminals). Resizing the terminal which is not visible doesn't work properly.

zmb3:
Great testing and review @ravicious! Thanks for catching those issues and thanks @gzdunek for fixing them so quickly!

packages/teleterm/src/ui/Tabs/TabItem.tsx
```diff
@@ -107,16 +112,16 @@ const StyledTabItem = styled.div(({ theme, active, dragging, canDrag }) => {
   return styles;
 });
 
-const StyledTabButton = styled.button`
-  display: flex;
+const Title = styled(Text)`
```
ravicious, line 115: What was the reason behind changing this to a div from a button? 🥺

ravicious (approved):
The bugs are now fixed, but I'd be grateful if you could take care of that button though. ;f

packages/teleterm/src/ui/Tabs/TabItem.tsx
gzdunek: This element ins't clickable, so there is no need for it to be a `button`.
We should think how to change the parent `div` (which has handlers) to be an interactive element.

PR #738, opened 2022-04-14
Use the mac package download link instead of the tarball.

hatched:
Fixes: https://github.com/gravitational/teleport/issues/11738


russjones:
@hatched Thanks for fixing this, this one is going to be huge. Potential customers won't get a malware prompt when trying Teleport out for the first time!

zmb3 (approved):

PR #739, opened 2022-04-15
Add Serverside Pagination, Filtering, and Sorting

rudream:
## Purpose

Adds serverside pagination, filtering, and sorting for Nodes, Apps, Databases, Kubernetes, and Desktops.

## Implementation

Since the backend response only includes the `startKey` of the next page, we need a way to figure out what page we are on and how to go back to the previous page. In order to keep track of past pages, an array of `startKeys` is created and updated as more and more `startKeys` are received as the user navigates through pages. The `startKeys` array contains all the `startKeys` from the first page up until the `startKey` of the next page.

### Simple Search
simplesearch

### Advanced Search
advancedsearch

### Tooltip

image

### Error (invalid predicate language search)

image


packages/design/src/DataTable/Table.tsx
```diff
@@ -56,30 +59,51 @@ export function Table<T>({
...
-    data.map((item, rowIdx) => {
-      const cells = columns.map((column, columnIdx) => {
-        const $cell = column.render ? (
-          column.render(item)
-        ) : (
-          <TextCell data={item[column.key]} />
-        );
+    if (fetching?.fetchStatus === 'loading') {
+      return <LoadingIndicator colSpan={columns.length} />;
+    } else {
```
zmb3, line 64: I would drop the else here since you've already returned in the if-case.

packages/design/src/DataTable/Cells.tsx
```diff
@@ -2,26 +2,60 @@ import React from 'react';
...
 import { displayDate } from 'shared/services/loc';
-import { SortDir } from './types';
+import { ServersideConfig, SortDir, TableColumn } from './types';
 
 export const Cell = props => <td children={props.children} {...props} />;
 
-export const SortHeaderCell = ({ dir, text, onClick }: SortHeaderCellProps) => {
+export function SortHeaderCell<T>({
+  column,
+  serverside,
```
zmb3, line 11: `serverside` is a confusing name here. Can we think of a better name?

packages/design/src/DataTable/Cells.tsx
```diff
@@ -2,26 +2,60 @@ import React from 'react';
...
+  onClick,
+}: SortHeaderCellProps<T>) {
+  function handleClick() {
+    if (serverside?.sort) {
+      serverside.setSort({
+        dir: serverside.sort.dir === 'ASC' ? 'DESC' : 'ASC',
+        fieldName: column.key,
+      });
+    } else {
+      onClick();
```
zmb3, line 23: What's going on here? This component either accepts a custom `onClick` or does sorting, but not both?

packages/design/src/DataTable/Table.tsx
```diff
@@ -56,30 +59,51 @@ export function Table<T>({
...
+      data.map((item, rowIdx) => {
+        const cells = columns.map((column, columnIdx) => {
+          const $cell = column.render ? (
```
zmb3, line 67: Does the `$` prefix on the variable name mean something special?

packages/design/src/DataTable/Table.tsx
```diff
@@ -56,30 +59,51 @@ export function Table<T>({
...
+        pagination={state.pagination}
+        fetching={fetching}
+        serverside={serverside}
+      />
+    );
+  }
+
   if (state.pagination) {
     return (
       <PagedTable
```
zmb3, line 109: Do we need two separate table components? The props look almost identical.

packages/design/src/DataTable/types.ts
```diff
@@ -31,10 +33,20 @@ export type PaginationConfig = {
...
 export type FetchingConfig = {
-  onFetchMore: () => void;
+  onFetchNext?: () => void;
+  onFetchPrev?: () => void;
+  onFetchMore?: () => void;
   fetchStatus: FetchStatus;
 };
 
+export type ServersideConfig = {
+  // 'from and 'to' are the indices of the first and last items on the page, respectively, for displaying: "SHOWING from - to of totalItemCount"
```
zmb3, line 43: What is this comment supposed to apply to? I see a mention of 'from' and 'to' but there's nothing here named 'from' or 'to'.

packages/teleport/src/Apps/useApps.ts
```diff
@@ -15,42 +15,128 @@
...
+        });
+        setFetchStatus(res.startKey ? '' : 'disabled');
+        setStartKeys([...startKeys, res.startKey]);
+      })
+      .catch((err: Error) => {
+        setAttempt({ status: 'failed', statusText: err.message });
+      });
+  };
+
+  const fetchPrev = () => {
```
zmb3, line 117: These fetch funcs are duplicated for each type of resource - I wonder if there's a way we can share more code rather than duplicate.

packages/design/src/DataTable/Table.tsx
rudream: It doesn't do anything special, it's just a naming convention to indicate that the variable contains a DOM node

packages/design/src/DataTable/types.ts
rudream: Leftover from a previous implementation that I forgot to delete

packages/design/src/DataTable/Cells.tsx
rudream: Yes, the custom `onClick` is used when we want client-side sorting, but if the Table is `serverside` then we want the sorting to be handled by the back. I will refactor this to try to make it a little more clear.

packages/design/src/DataTable/Table.tsx
rudream: The `PagedTable` uses the built-in `Pager` component while `ServersideTable` uses the `serversideSearchPanel` that is passed in. Even though they share many props, the `PagedTable` component already has a good amount of logic, and trying to merge the two tables into one component with more logic within the component to determine what to render would make it hard to read and follow. We tried a similar approach before (merging the basic, paginated, and searchable tables into one), but Alexey and I agreed that it was better to create a separate component for each type of Table. Let me know what you think

packages/design/src/DataTable/Cells.tsx
rudream: hm, how about `serversideProps`?

zmb3: Better, but still a bit confusing. These props are used for filtering and sorting right? Can we somehow include that in the name?

packages/design/src/DataTable/Table.tsx
zmb3: Fair, let's not do anything too disruptive then.

packages/teleport/src/components/ServersideSearchPanel/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,71 @@
+import React, { createRef } from 'react';
```
kimlisa, line 1: why is this file in js? and we are using class?

packages/teleport/src/components/NodeList/NodeList.tsx
```diff
@@ -18,51 +18,84 @@ import React from 'react';
...
+        data={nodes}
+        pagination={{
+          pageSize,
+        }}
+        fetching={{
+          onFetchNext: fetchNext,
+          onFetchPrev: fetchPrev,
+          fetchStatus,
+        }}
+        serversideProps={{
```
kimlisa, line 81: need better naming, something with pagination maybe

packages/design/src/DataTable/Cells.tsx
```diff
@@ -2,11 +2,41 @@ import React from 'react';
...
+  dir,
+  text,
+  onClick,
+}: SortHeaderCellProps<T>) {
+  function handleServersideClick() {
+    serversideProps.setSort({
+      dir: serversideProps.sort?.dir === 'ASC' ? 'DESC' : 'ASC',
+      fieldName: column.key,
+    });
+  }
```
kimlisa, line 21: should this come from the caller? and send it in as `onClick`

kimlisa (commented):
- Before merge, please add license headers to most files
- Address reviews after merging and backporting to v9

kimlisa (approved):

PR #740, opened 2022-04-15
Resolve issues on logout

gzdunek:
Closes https://github.com/gravitational/webapps.e/issues/226

### DIFF
--- a/packages/teleterm/src/ui/ClusterLogout/useClusterLogout.ts
+++ b/packages/teleterm/src/ui/ClusterLogout/useClusterLogout.ts
@@ -18,6 +18,7 @@ export function useClusterLogout({ clusterUri, onClose, clusterTitle }: Props) {
         await ctx.workspacesService.setActiveWorkspace(null);
       }
     }
+    ctx.workspacesService.removeWorkspace(clusterUri);
   });
 
   useEffect(() => {
--- a/packages/teleterm/src/ui/TabHost/TabHost.tsx
+++ b/packages/teleterm/src/ui/TabHost/TabHost.tsx
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import React, { useMemo } from 'react';
+import React from 'react';
 import styled from 'styled-components';
 import { Flex } from 'design';
 import { useAppContext } from 'teleterm/ui/appContextProvider';
@@ -24,31 +24,36 @@ import { useTabShortcuts } from './useTabShortcuts';
 import { DocumentsRenderer } from 'teleterm/ui/Documents';
 import { useNewTabOpener } from './useNewTabOpener';
 import { ClusterConnectPanel } from './ClusterConnectPanel/ClusterConnectPanel';
+import AppContext from 'teleterm/ui/appContext';
 
 export function TabHostContainer() {
   const ctx = useAppContext();
   ctx.workspacesService.useState();
-
   const isRootClusterSelected = !!ctx.workspacesService.getRootClusterUri();
 
-  return useMemo(() => {
-    if (isRootClusterSelected) {
-      return <TabHost />;
-    }
-    return <ClusterConnectPanel />;
-  }, [isRootClusterSelected]);
+  if (isRootClusterSelected) {
+    return <TabHost ctx={ctx} />;
+  }
+  return <ClusterConnectPanel />;
 }
 
-export function TabHost() {
-  const ctx = useAppContext();
+export function TabHost({ ctx }: { ctx: AppContext }) {
   const documentsService =
     ctx.workspacesService.getActiveWorkspaceDocumentService();
-  const activeDocument = documentsService.getActive();
-  const { openClusterTab } = useNewTabOpener();
-  ctx.workspacesService.useState();
+  const activeDocument = documentsService?.getActive();
 
-  // enable keyboard shortcuts
-  useTabShortcuts();
+  // TODO(gzdunek): make workspace refactor - it'd be helpful to have a single object that fully represents a workspace
+  const { openClusterTab } = useNewTabOpener({
+    documentsService,
+    localClusterUri:
+      ctx.workspacesService.getActiveWorkspace()?.localClusterUri,
+  });
+
+  useTabShortcuts({
+    documentsService,
+    localClusterUri:
+      ctx.workspacesService.getActiveWorkspace()?.localClusterUri,
+  });
 
   function handleTabClick(doc: types.Document) {
     documentsService.open(doc.uri);
--- a/packages/teleterm/src/ui/TabHost/useNewTabOpener.ts
+++ b/packages/teleterm/src/ui/TabHost/useNewTabOpener.ts
@@ -1,15 +1,13 @@
-import { useAppContext } from 'teleterm/ui/appContextProvider';
 import { useCallback } from 'react';
+import { DocumentsService } from 'teleterm/ui/services/workspacesService';
 
-export function useNewTabOpener() {
-  const ctx = useAppContext();
-
-  const documentsService =
-    ctx.workspacesService.getActiveWorkspaceDocumentService();
-
-  const localClusterUri =
-    ctx.workspacesService.getActiveWorkspace()?.localClusterUri;
-
+export function useNewTabOpener({
+  documentsService,
+  localClusterUri,
+}: {
+  documentsService: DocumentsService;
+  localClusterUri: string;
+}) {
   const openClusterTab = useCallback(() => {
     if (localClusterUri) {
       const clusterDocument = documentsService.createClusterDocument({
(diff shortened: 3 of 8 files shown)


packages/teleterm/src/ui/TabHost/TabHost.tsx
```diff
@@ -28,22 +28,19 @@ import { ClusterConnectPanel } from './ClusterConnectPanel/ClusterConnectPanel';
...
+  }
+  return <ClusterConnectPanel />;
 }
 
 export function TabHost() {
   const ctx = useAppContext();
   const documentsService =
     ctx.workspacesService.getActiveWorkspaceDocumentService();
-  const activeDocument = documentsService.getActive();
+  const activeDocument = documentsService?.getActive();
```
gzdunek, line 43: Sometimes (very rarely for me) there was the error `cannot read property getActive of undefined` after logout. Theoretically, it should not be possible, because condition in `TabHostContainer` protects us from it - we can access `activeDocumentService` only when `rootClusterUri` is not null. However, the error was occurring. The reason for this is the order in which the hooks are called. React runs `useEffect` starting from children. It means that `TabHost` added itself to the `workspaceService` subscribers before `TabHostContainer`, so when `rootClusterUri` became `null` child component was notified before the parent and tried to re-render. 
I got rid of unnecessary subscriptions. I don't know if we should take to do something with it globally (like calling callbacks in reverse order, but that could potentially cause problems elsewhere).

packages/teleterm/src/ui/TopBar/Clusters/useClusters.ts
```diff
@@ -16,12 +16,12 @@ export function useClusters() {
...
   const localClusterUri =
     workspacesService.getActiveWorkspace()?.localClusterUri;
-  const items = rootClusterUri
+  const items = (rootClusterUri
     ? [
         clustersService.findCluster(rootClusterUri),
         ...findLeaves(rootClusterUri),
       ]
-    : [];
+    : []).filter(Boolean);
```
gzdunek, line 24: `clustersService.findCluster(rootClusterUri)` is `undefined` after logout

packages/teleterm/src/ui/ClusterLogout/useClusterLogout.ts
```diff
@@ -18,6 +18,7 @@ export function useClusterLogout({ clusterUri, onClose, clusterTitle }: Props) {
         await ctx.workspacesService.setActiveWorkspace(null);
       }
     }
+    ctx.workspacesService.removeWorkspace(clusterUri);
```
gzdunek, line 21: Removing a workspace has nothing to do with these bugs, but we should make the clean-up

ravicious: Does this take care of gravitational/webapps.e#154? I suppose only in part, because it'll remove the docs but probably won't take care of the other stuff (I don't actually know if there's any other stuff to take care of, perhaps any open connections?).

ravicious (commented):
Looks good, but I have some questions and suggestions.

packages/teleterm/src/ui/TopBar/Clusters/useClusters.ts
gzdunek: Yeah, much better now, thanks!

packages/teleterm/src/ui/ClusterLogout/useClusterLogout.ts
gzdunek: This PR had to resolve this issue: `Cluster with URI /clusters/teleport-local does not exist`, but I didn't commit it :/ (fix is now in `workspacesService` line 141).
Unfortunately, the line you mentioned won't resolve any issues from the linked issue. The problem is in `StatePersitenceService` and its messy logic (it doesn't remove the workspace from the `app_state.json`)

packages/teleterm/src/ui/TabHost/useTabShortcuts.ts
```diff
@@ -25,7 +25,6 @@ import { useNewTabOpener } from 'teleterm/ui/TabHost/useNewTabOpener';
 
 export function useTabShortcuts() {
   const { workspacesService } = useAppContext();
-  workspacesService.useState();
```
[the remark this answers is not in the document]
gzdunek: Ah, it looks like I didn't commit all the changes I wanted. 
I wanted to remove `workspacesService.useState()` from `TabHost` as well and leave it only in `TabHostContainer`.

I agree with you, we should be more explicit, I added whole `ctx` object to the `TabHost`, `useTabShortcuts`, `useNewTabOpener` (some of these need more than only `documentsServive`).

packages/teleterm/src/ui/TabHost/TabHost.tsx
```diff
@@ -28,22 +28,19 @@ import { ClusterConnectPanel } from './ClusterConnectPanel/ClusterConnectPanel';
...
   ctx.workspacesService.useState();
-
   const isRootClusterSelected = !!ctx.workspacesService.getRootClusterUri();
 
-  return useMemo(() => {
-    if (isRootClusterSelected) {
-      return <TabHost />;
-    }
-    return <ClusterConnectPanel />;
-  }, [isRootClusterSelected]);
```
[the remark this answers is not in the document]
gzdunek: No, `useMemo` wasn't useful at all, because we were listening for the same state updates in both parent and child anyway. After the changes, we will subscribe only in the parent (which will re-render child) to make things simpler.

packages/teleterm/src/ui/TabHost/useTabShortcuts.ts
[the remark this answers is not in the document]
ravicious: I suggested passing `documentService` around since it was my impression that since `useTabShortcuts` is no longer subscribed to `workspacesService`, we could remove the dependency on `workspacesService` altogether and just pass `documentService` from outside.

If this was not done, we could run into problems where, for example, we take the logic from `useTabShortcuts` that depends on getting the active `documentService` and we move it outside of `TabHost`. Since `useTabShortcuts` would no longer be called again when the active `documentService` changes, this could lead to bugs.

You explained that it is not enough to pass `documentService`, which is fine. But now what does passing `ctx` give us? In theory it can still be given `ctx` by a component that doesn't listen to `workspacesService` updates, so we might as well call `useAppContext` there.

---

It seems like the only other piece of data that's needed is `localClusterUri`. In the future, it'd be helpful to have a single object that fully represents a workspace, that is it has access to both `localClusterUri` and the document service for that workspace. Then we could just pass this one single thing around.

packages/teleterm/src/ui/services/workspacesService/workspacesService.ts
```diff
@@ -137,6 +137,12 @@ export class WorkspacesService extends ImmutableStore<WorkspacesState> {
       });
     };
 
+    // empty cluster URI - no cluster selected
+    if (!clusterUri) {
+      setWorkspace();
```
ravicious, line 142: Why do we need to call `setWorkspace` if we know that there's no `clusterUri`? Just to set `rootClusterUri` to `undefined`?

If so then I think we should just do it here, that would be more readable. Right now when reading this function you have to jump up to see what `setWorkspace` does, only to realize that you can actually ignore almost all of its body except one line.

packages/teleterm/src/ui/TabHost/useTabShortcuts.ts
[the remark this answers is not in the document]
gzdunek: Totally agree, for now I added `{documentService, localClusterUri}` object as a param (but we should refactor `workspace/documentsService`).

packages/teleterm/src/ui/services/workspacesService/workspacesService.ts
gzdunek: I agree, it is more readable now.

packages/teleterm/src/ui/services/workspacesService/workspacesService.ts
```diff
@@ -137,6 +136,20 @@ export class WorkspacesService extends ImmutableStore<WorkspacesState> {
...
+      return Promise.resolve();
+    }
```
ravicious, line 143: It looks like you might have forgot to remove this bit, no?

packages/teleterm/src/ui/TabHost/useNewTabOpener.ts
```diff
@@ -1,25 +1,20 @@
...
-  const documentsService =
-    ctx.workspacesService.getActiveWorkspaceDocumentService();
-
-  const localClusterUri =
-    ctx.workspacesService.getActiveWorkspace()?.localClusterUri;
-
+export function useNewTabOpener(params: {
+  documentsService: DocumentsService;
+  localClusterUri: string;
+}) {
```
ravicious, line 7: I'd be for not using the name `params` here at all and destructuring those args instead, but that can be fixed in the future.

ravicious (approved):
Something went wrong with the recent changes and there are some compilation errors, but once you deal with that I think it's ready to go.

packages/teleterm/src/ui/services/workspacesService/workspacesService.ts
gzdunek: Yes, thanks

packages/teleterm/src/ui/TabHost/useNewTabOpener.ts
gzdunek: Done now :)

PR #750, opened 2022-04-20
Simplify k8s join

mcbattirola:
This pr adds a `Add Kubernetes` button to the Kubernetes page.

We will use the helm chart to help users add a k8s cluster.

Design: https://www.figma.com/file/Gpjs9vjhzUKF1GDbeG9JGE/Application-Design-System?node-id=1856%3A8594
Note that the design is using a side drawer, that are choose not to implement yet.
Also the implementation changes some of the design based on the comments left by @xinding33 (mainly add the script generated as a step 3).

Screenshot from 2022-04-20 10-06-19

Screenshot from 2022-04-20 10-07-48

Closes gravitational/teleport#11465


packages/teleport/src/Apps/AddApp/Automatically/Automatically.tsx
```diff
@@ -25,8 +25,8 @@ import {
   ButtonPrimary,
 } from 'design';
 import Validation, { Validator } from 'shared/components/Validation';
-import FieldInput from 'shared/components/FieldInput';
 import { DialogContent, DialogFooter } from 'design/Dialog';
+import FieldInput from 'shared/components/FieldInput';
```
kimlisa, line 29: i don't think we pay close attention to this, but when we do: https://gravitational.slack.com/archives/C02DQ1C2BMW/p1643999171617869

packages/teleport/src/Kubes/Kubes.tsx
```diff
@@ -67,14 +70,8 @@ export function Kubes(props: State) {
...
-          <ButtonPrimary
-            as="a"
-            width="240px"
-            target="_blank"
-            href={DOC_URL}
-            rel="noreferrer"
-          >
-            View documentation
+          <ButtonPrimary width="240px" onClick={() => setShowAddKube(true)}>
+            Add Kubernetes
```
kimlisa, line 74: i think we need to add a `ButtonAdd` equivalent that checks if user has perm to create a token (like we do in apps, nodes, etc), maybe even bring it out to component/ level to share it since the only thing that seems to differ is the resource name: `ResourceButtonAdd`? or something like that

packages/teleport/src/Kubes/AddKube/__snapshots__/AddKube.story.test.tsx.snap
```diff
@@ -0,0 +1,9 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`render failed 1`] = `null`;
```
kimlisa, line 3: the snapshot results don't look correct (i would always check the snapshot result to make sure it's what we wanted to freeze, sometimes we target the incorrect element)

packages/teleport/src/Kubes/AddKube/AddKube.story.tsx
```diff
@@ -0,0 +1,57 @@
...
+  />
+);
+
+const props = {
+  onClose() {
+    return null;
+  },
+  createToken() {
+    return Promise.resolve(null);
+  },
```
kimlisa, line 50: shorthand 
```suggestion
  onClose: () => null,
  createToken: () => Promise.resolve(null),
```

packages/teleport/src/Kubes/AddKube/AddKube.story.tsx
```diff
@@ -0,0 +1,57 @@
...
+  attempt: {
+    status: 'success',
+    statusText: '',
+  } as any,
```
kimlisa, line 54: i would import the type and avoid using "as any"

packages/teleport/src/Kubes/AddKube/AddKube.tsx
```diff
@@ -0,0 +1,248 @@
...
+                      labelTip="Name shown to Teleport users connecting to the cluster."
+                      value={clusterName}
+                      placeholder="my-cluster"
+                      width="100%"
+                      mr="3"
+                      onChange={e => setClusterName(e.target.value)}
+                    />
+                    <ButtonPrimary
+                      block
+                      mt="1"
```
kimlisa, line 171: the btn is too close for comfort to the input box above
```suggestion
                      mt="2"
```

packages/teleport/src/Kubes/AddKube/AddKube.tsx
```diff
@@ -0,0 +1,248 @@
...
+            Install Teleport Agent in your cluster via Helm to easily connect
+            your Kubernetes cluster with Teleport. For all the available values
+            of the helm chart see{' '}
+            <Link
+              href="https://goteleport.com/docs/kubernetes-access/helm/reference/teleport-kube-agent/"
+              target="_blank"
+            >
+              the documentation
+            </Link>
+            {' .'}
```
kimlisa, line 116:
```suggestion
            {'.'}
```

packages/teleport/src/Kubes/AddKube/AddKube.tsx
```diff
@@ -0,0 +1,248 @@
...
+      namespace,
+      clusterName,
+      host,
+      token.id,
+      version
+    );
+    setCmd(generatedCmd);
+  }, [token]);
+
+  function handleGenerate(e: React.SyntheticEvent, validator: Validator) {
```
kimlisa, line 74:
```suggestion
  function handleOnSubmit(e: React.SyntheticEvent, validator: Validator) {
```

packages/teleport/src/Kubes/AddKube/AddKube.tsx
```diff
@@ -0,0 +1,248 @@
...
+    return {
+      valid: false,
+      message: `${fieldName} is required`,
+    };
+  }
+
+  return {
+    valid: true,
+  };
+};
```
kimlisa, line 226: we already have a rule for [required fields](https://github.com/gravitational/webapps/blob/master/packages/shared/components/Validation/rules.js#L18)?

mcbattirola: Ah, that's good to know, thanks!

packages/teleport/src/Kubes/AddKube/__snapshots__/AddKube.story.test.tsx.snap
mcbattirola: Yes, thanks!

packages/teleport/src/Kubes/Kubes.tsx
mcbattirola: I'm glad you noticed that, thanks!

I've added the component in this commit https://github.com/gravitational/webapps/pull/750/commits/ba34e77a26f2d50db108b71cab20330a1da53cb3

And replaced the old buttons form apps and nodes in this one https://github.com/gravitational/webapps/pull/750/commits/f8e98516e191b7894cda9c2425320c52420809fc (Note that the snapshots have changed casing of the resource name but the CSS of the button changes all text to uppercase anyway)

Please let me know your thoughts

packages/teleport/src/Kubes/AddKube/AddKube.tsx
```diff
@@ -0,0 +1,235 @@
...
+        </DialogFooter>
+      </Flex>
+    </Dialog>
+  );
+}
+const generateCmd = (
+  namespace: string,
+  clusterName: string,
+  proxyAddr: string,
+  token: string,
```
hatched, line 219: As you're passing the `token.id` in...

```suggestion
  tokenId: string,
```

hatched (approved):

packages/teleport/src/components/ResourceButtonAdd/index.ts
```diff
@@ -0,0 +1,3 @@
+import ResourceButtonAdd from './ResourceButtonAdd';
```
kimlisa, line 1: i think we should start using `agent` to mean `node,app,db,kube,desktop`, so `AgentButtonAdd`? [Resource](https://github.com/gravitational/webapps/blob/master/packages/teleport/src/services/resources/resource.ts) kinda means something else in the front (which is everything else other than agents which is auth connectors, roles, trusted cluster). I think zac says we often refer to node,app,kube,desktop,db as agents also.

packages/teleport/src/components/ResourceButtonAdd/ResourceButtonAdd.tsx
```diff
@@ -17,17 +17,17 @@ limitations under the License.
...
-export default function ButtonAdd(props: Props) {
-  const { canCreate, isLeafCluster, onClick } = props;
+export default function ResourceButtonAdd(props: Props) {
+  const { canCreate, isLeafCluster, onClick, resource } = props;
   const disabled = isLeafCluster || !canCreate;
 
   let title = '';
   if (!canCreate) {
-    title = 'You do not have access to add an application';
+    title = `You do not have access to add ${resource}`;
```
kimlisa, line 26: lets keep the `an` or `a` b/c the sentence flows better. We can another prop `beginsWithVowel` or something like that, so that when true we use `an`, otherwise `a`

packages/teleport/src/components/ResourceButtonAdd/ResourceButtonAdd.tsx
```diff
@@ -46,4 +46,5 @@ type Props = {
   isLeafCluster: boolean;
   canCreate: boolean;
   onClick?: () => void;
+  resource: string;
```
kimlisa, line 49:
```suggestion
  agentName: string;
```

packages/teleport/src/Kubes/AddKube/AddKube.story.tsx
```diff
@@ -0,0 +1,55 @@
...
+  />
+);
+
+const props: Props & State = {
+  onClose: () => null,
+  createToken: () => Promise.resolve(null),
+  attempt: {
+    status: 'success',
+    statusText: '',
+  } as Attempt,
```
kimlisa, line 52: b/c we are using the component type, we don't need `as Attempt` anymore

kimlisa (approved):

PR #753, opened 2022-04-21
Change app name to `Teleport Connect`

gzdunek:
Closes https://github.com/gravitational/webapps.e/issues/223

### DIFF
--- a/packages/teleterm/package.json
+++ b/packages/teleterm/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@gravitational/teleterm",
   "version": "1.0.0",
-  "description": "Teleport Terminal",
+  "description": "Teleport Connect",
   "main": "build/app/dist/main/main.js",
   "author": {
     "name": "Teleport",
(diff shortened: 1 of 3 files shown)


packages/teleterm/package.json
```diff
@@ -17,7 +17,7 @@
     "build-main": "webpack build --config webpack.main.config.js --progress --mode=production",
     "build-renderer": "webpack build --config webpack.renderer.prod.config.js --progress",
     "build-natives": "electron-builder install-app-deps",
-    "package": "electron-builder build --publish never -c.extraMetadata.name=teleterm"
+    "package": "electron-builder build --publish never -c.extraMetadata.name=teleconnect"
```
gzdunek, line 20: [There was an issue with `productName` on Linux](https://github.com/electron-userland/electron-builder/issues/1567). For now I will change it to `teleconnect`, but we will verify that when we will release the app for that OS.

ravicious (approved):

PR #754, opened 2022-04-21
Show database username suggestions in Teleport Connect

ravicious:
This PR implements one of the last remaining parts of gravitational/webapps.e#177.

* The first three PRs add a variant of `MenuLogin` that works with `useAsync`.
* The fourth commit updates the gRPC files so it's just a bunch of autogenerated code.
* The last one actually shows the user suggestions in the UI.

The PR which adds the necessary changes to the tsh daemon has been already merged (gravitational/teleport#12070). You need to compile tsh with those changes for the usernames to show up.

https://user-images.githubusercontent.com/27113/164047651-376990ae-bbe4-4f3a-a002-ff83bc6d30e0.mov


packages/shared/components/MenuLogin/MenuLogin.story.tsx
```diff
@@ -20,8 +20,24 @@ import { Flex } from 'design';
...
+storiesOf('Shared/MenuLogin', module).add(
+  'MenuLogin in Teleport Connect',
+  () => {
+    return (
+      <MenuLoginTheme>
+        <MenuLoginExamples />
+      </MenuLoginTheme>
+    );
+  }
+);
```
ravicious, line 38: Added those to let other teammates modifying `MenuLogin` know that Teleport Connect uses a special variant of it.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/Databases.tsx
```diff
@@ -60,13 +65,26 @@ function DatabaseList(props: State) {
...
+
   return (
     <Cell align="right">
       <MenuLoginTheme>
-        <MenuLogin
+        <MenuLoginAsync
           placeholder="Enter username…"
-          getLoginItems={() => []}
+          getLoginItemsAttempt={getUsersAttempt}
+          getLoginItems={() => {
```
gzdunek, line 83: I feel like we can simplify it. I see no benefit from passing an attempt to the component (and it increases usage complexity). Why the component can't handle data fetching on its own? 
Current solution triggers the request only once, but we can achieve the same effect by simply wrapping a promise into `memoize` here.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/useDatabaseUsers.ts
```diff
@@ -0,0 +1,23 @@
+import useAsync from 'shared/hooks/useAsync';
+import { useAppContext } from 'teleterm/ui/appContextProvider';
+
+export function useDatabaseUsers(dbUri: string) {
```
gzdunek, line 4: IMO we don't need this file anymore if we just pass a promise to `MenuLoginAsync`.

gzdunek (commented):
I'm wondering if even need both `MenuLogin` and `MenuLoginAsync`. I think `getLoginItems()` can have the type `() => LoginItem[] | Promise<LoginItem[]>` (and if it is a simple object, then just wrap it in `Promise.resolve()`). Additionally, I'd use `useAsync` locally, in that new `MenuLogin`. IMO it would really simplify the code.

ravicious:
Oops, I screwed up how the ref is managed and the storybook no longer renders the last `MenuLogin` open. I'll push a fix soon.

packages/shared/components/MenuLogin/MenuLogin.story.tsx
ravicious: It's a special case just for the storybook, so I feel that's fine. 🤔 The storybook files are completely different entrypoints than the rest of the code.

I could move it under the Teleterm section, but then it would be easy to miss and thus didn't accomplish its original purpose. Would you agree that this makes the "MenuLogin in Teleport Connect" example justified here?

packages/shared/components/MenuLogin/MenuLogin.tsx
```diff
@@ -93,25 +121,13 @@ export class MenuLogin extends React.Component<MenuLoginProps> {
...
-        {login}
-      </StyledMenuItem>
-    );
-  });
+export const LoginItemList = ({
+  getLoginItemsAttempt,
+  onClick,
+  onKeyPress,
+  placeholder,
+}) => {
```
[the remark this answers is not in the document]
ravicious: I think I'll just make it private and then I could skip adding the interface? It's not used anywhere in the project anyway.

packages/shared/components/MenuLogin/MenuLogin.story.tsx
gzdunek: A part of me still feels a little wrong 😄, but since it's a non-production code, let's do it. I can agree that moving it under Teleterm section doesn't make much sense.

packages/shared/components/MenuLogin/MenuLogin.tsx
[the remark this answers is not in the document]
gzdunek: I would still add a type (it can inline) for this component, it will make any future changes easier.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/Databases.tsx
ravicious: > Why the component can't handle data fetching on its own?

My instinct is to push any async stuff to the outermost layer of the app, as far as possible. I was also thinking about a scenario where we'd want to share async data between `MenuLoginAsync` and some other compontent. But even in that case, it should be possible to just pass a promise to `MenuLogin` that was prepared somewhere else. 🤔

And I agree that the usage complexity would be lower then. Because then essentially I'd do

```jsx
<MenuLogin
  getLoginItems={someFunctionReturningPromise}
/>
```

right?

> Current solution triggers the request only once, but we can achieve the same effect by simply wrapping a promise into `memoize` here.

We probably wouldn't even need memoize, if `useAsync` was called from inside `MenuLogin` then we could check the status of the attempt object just as I do know, I think.

I'd need to make to make a function component out of `MenuLogin` to use `useAsync` in it, wouldn't I?

gzdunek: I agree, if there would be a case to share some data, then we could create a special promise for that. 

> Because then essentially I'd do
`<MenuLogin
  getLoginItems={someFunctionReturningPromise}
/>`
right?

Yes

> We probably wouldn't even need memoize, if useAsync was called from inside MenuLogin then we could check the status of the attempt object just as I do know, I think.

Ah right, I forgot that `MenuLogin` isn't destroyed when is closed.

> I'd need to make to make a function component out of MenuLogin to use useAsync in it, wouldn't I?

Yes, you can't use hooks in class components.

ravicious:
Using `async` breaks the Web UI with ["regeneratorRuntime is not defined"](https://stackoverflow.com/questions/33527653/babel-6-regeneratorruntime-is-not-defined). So either we change Babel config or rewrite that one `async` call in `useAsync`, but I'd rather do the former.

r0mant (commented):
This looks super slick 🚀

packages/shared/components/MenuLogin/MenuLogin.tsx
```diff
@@ -124,11 +118,53 @@ export const LoginItemList = ({ logins, onClick, onKeyPress, placeholder }) => {
...
 
+function getLoginItemListContent(
+  getLoginItemsAttempt: Attempt<LoginItem[]>,
+  onClick: (e: React.MouseEvent<HTMLAnchorElement>, login: string) => void
+) {
+  switch (getLoginItemsAttempt.status) {
+    case '':
+    case 'processing':
+      return (
+        <Indicator
```
gzdunek, line 134: I noticed a small delay between opening the menu and displaying the loading spinner which is caused by `Indicator` component. Maybe we would like to set `delay="none"` on it?

gzdunek (commented):
Excellent work!
I'm waiting for the Web UI fix :)

packages/shared/components/MenuLogin/MenuLogin.tsx
ravicious: I think that's how Indicator works on purpose: in good conditions you'll never see a spinner, because the usernames will load that fast. If it takes longer than expected, we show the spinner.

gzdunek: I noticed this because `Indicator` creates its own "background". When the data takes a little longer to load, you will first see only the input row, then the height will change and you will see the spinner. It would look better if the height did not change. 
But it's a really minor thing and we can skip it.

https://user-images.githubusercontent.com/20583051/164685089-4f29b4de-6da2-4e89-9c4f-dcbe63cd1a2a.mov

gzdunek (approved):

packages/shared/hooks/useAsync.ts
```diff
@@ -64,33 +64,37 @@ export function useAsync<R, T extends Function>(cb?: AsyncCb<R, T>) {
...
-        ...state,
-        status: 'error',
-        statusText: err.message,
-        data: null,
-      });
-
-      return [null, err] as [R, Error];
-    }
-  };
+  const run = (...p: Parameters<AsyncCb<R, T>>) =>
```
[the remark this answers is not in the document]
ravicious: I'm used to always wrapping any promise-returning function within a promise context. Otherwise any error thrown inside `cb` will behave like a regular error, whereas when it's wrapped within the promise context it'll cause the promise to get rejected instead.

PR #756, opened 2022-04-22
Replace "Lorem ipsum"

gzdunek:
Closes https://github.com/gravitational/webapps.e/issues/227
image

### DIFF
--- a/packages/teleterm/src/ui/TabHost/ClusterConnectPanel/ClusterConnectPanel.tsx
+++ b/packages/teleterm/src/ui/TabHost/ClusterConnectPanel/ClusterConnectPanel.tsx
@@ -14,18 +14,17 @@ export function ClusterConnectPanel() {
   return (
     <Flex
       m="auto"
-      width="400px"
+      width="370px"
       pb={6}
       flexDirection="column"
       alignItems="center"
     >
       <Image width="120px" src={clusterPng} mb={3} />
       <Text typography="h3" bold mb={2}>
-        Connect to a Cluster
+        Connect a Cluster
       </Text>
       <Text color="text.secondary" mb={3} textAlign="center">
-        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
-        tempor incididunt ut labore et dolore magna aliqua.
+        Connect an existing Teleport cluster <br/> to start using Teleport Connect.
       </Text>
       <ButtonPrimary size="large" onClick={handleConnect}>Connect</ButtonPrimary>
     </Flex>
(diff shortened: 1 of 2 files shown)


packages/teleterm/src/ui/TabHost/ClusterConnectPanel/ClusterConnectPanel.tsx
```diff
@@ -14,18 +14,17 @@ export function ClusterConnectPanel() {
...
     >
       <Image width="120px" src={clusterPng} mb={3} />
       <Text typography="h3" bold mb={2}>
-        Connect to a Cluster
+        Connect a Cluster.
       </Text>
       <Text color="text.secondary" mb={3} textAlign="center">
-        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
-        tempor incididunt ut labore et dolore magna aliqua.
+        Connect an existing Teleport cluster to start using Teleport Connect.
```
ravicious, line 27: Let's put a `<br/>` here after "Connect an existing Teleport cluster" so that "Teleport Connect." doesn't hang alone on a separate line. This might not be the best point to break the text, but I think it looks better.

packages/teleterm/src/ui/TabHost/ClusterConnectPanel/ClusterConnectPanel.tsx
```diff
@@ -14,18 +14,17 @@ export function ClusterConnectPanel() {
...
-      width="400px"
+      width="370px"
       pb={6}
       flexDirection="column"
       alignItems="center"
     >
       <Image width="120px" src={clusterPng} mb={3} />
       <Text typography="h3" bold mb={2}>
-        Connect to a Cluster
+        Connect a Cluster.
```
ravicious, line 24: The dots look a bit weird for me, but I asked Xin about them already.

ravicious (approved):

ravicious:
@gzdunek Could you remove the periods from the end of the titles and then merge this PR? Since Xin hasn't responded yet let's do what we think is best here.

packages/teleterm/src/ui/TabHost/ClusterConnectPanel/ClusterConnectPanel.tsx
gzdunek: Done

packages/teleterm/src/ui/TabHost/ClusterConnectPanel/ClusterConnectPanel.tsx
gzdunek: I think, you're right, removed.

PR #757, opened 2022-04-25
Teleport Connect: Add dropdown for database name

ravicious:
Adds a second dropdown for the db name. Needs tsh compiled with changes from gravitational/teleport#12173.

### DIFF
--- a/packages/shared/components/MenuLogin/MenuLogin.test.tsx
+++ b/packages/shared/components/MenuLogin/MenuLogin.test.tsx
@@ -1,0 +1,47 @@
+import React from 'react';
+import { render, fireEvent, waitFor } from 'design/utils/testing';
+import { MenuLogin } from './MenuLogin';
+
+test('does not accept an empty value when required is set to true', async () => {
+  const onSelect = jest.fn();
+  const { findByText, findByPlaceholderText } = render(
+    <MenuLogin
+      placeholder="MenuLogin input"
+      required={true}
+      getLoginItems={() => []}
+      onSelect={() => onSelect()}
+    />
+  );
+
+  fireEvent.click(await findByText('CONNECT'));
+  await waitFor(async () =>
+    fireEvent.keyPress(await findByPlaceholderText('MenuLogin input'), {
+      key: 'Enter',
+      keyCode: 13,
+    })
+  );
+
+  expect(onSelect).toHaveBeenCalledTimes(0);
+});
+
+test('accepts an empty value when required is set to false', async () => {
+  const onSelect = jest.fn();
+  const { findByText, findByPlaceholderText } = render(
+    <MenuLogin
+      placeholder="MenuLogin input"
+      required={false}
+      getLoginItems={() => []}
+      onSelect={() => onSelect()}
+    />
+  );
+
+  fireEvent.click(await findByText('CONNECT'));
+  await waitFor(async () =>
+    fireEvent.keyPress(await findByPlaceholderText('MenuLogin input'), {
+      key: 'Enter',
+      keyCode: 13,
+    })
+  );
+
+  expect(onSelect).toHaveBeenCalledTimes(1);
+});
--- a/packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/Databases.tsx
+++ b/packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/Databases.tsx
@@ -14,13 +14,14 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import React from 'react';
+import React, { useRef, useState } from 'react';
+import styled from 'styled-components';
 import { useDatabases, State } from './useDatabases';
 import { Table } from 'teleterm/ui/components/Table';
 import { Cell } from 'design/DataTable';
 import { renderLabelCell } from '../renderLabelCell';
 import { Danger } from 'design/Alert';
-import { MenuLogin } from 'shared/components/MenuLogin';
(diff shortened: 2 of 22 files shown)


packages/teleterm/src/ui/TopBar/Connections/ConnectionsFilterableList/ConnectionItem.tsx
```diff
@@ -48,37 +48,43 @@ export function ConnectionItem(props: ConnectionItemProps) {
...
+        css={`
+          flex-shrink: 0;
+        `}
+        connected={props.item.connected}
+      />
       <Flex
         alignItems="center"
         justifyContent="space-between"
         flex="1"
-        width="100%"
```
ravicious, line 56: `flex: 1` and `width: 100%` seem unecessary. Instead of taking 100% width of its parent and letting other elements cut into it (like `ConnectionStatusIndicator` did here), let's use flexbox to its fullest: make `ConnectionStatusIndicator` never shrink and make this div here take all remaining space with `flex: 1`.

packages/teleterm/src/ui/TopBar/Connections/ConnectionsFilterableList/ConnectionItem.tsx
```diff
@@ -48,37 +48,43 @@ export function ConnectionItem(props: ConnectionItemProps) {
...
         minWidth="0"
       >
-        <Flex flexDirection="column">
```
ravicious, line 59: Block elements by default behave like flex with direction set to column, so I realized that this one here is unnecessary.

packages/teleterm/src/ui/TopBar/Connections/ConnectionsFilterableList/ConnectionItem.tsx
```diff
@@ -48,37 +48,43 @@ export function ConnectionItem(props: ConnectionItemProps) {
...
+        <div
+          css={`
+            min-width: 0;
+          `}
+        >
```
ravicious, line 68: There's [a really nice answer on SO](https://stackoverflow.com/a/43809765/742872) about it. Flex children by default have `min-width` set to `auto`, so `text-overflow` will never work on them.

Instead, we have to, in a cascading fashion, set `min-width` on those flex children which have content that should get hidden on overflow. So we have `min-width` on this div here and one above it.

packages/teleterm/src/ui/TopBar/Connections/ConnectionsFilterableList/ConnectionItem.tsx
```diff
@@ -48,37 +48,43 @@ export function ConnectionItem(props: ConnectionItemProps) {
...
           <Text
             typography="body1"
             bold
             color="text.primary"
             title={props.item.title}
             css={`
               line-height: 16px;
-              white-space: normal;
```
ravicious, line 67: `white-space: normal` was causing the UI to behave inconsistently. Document titles or cluster names with hyphens in them were getting split on multiple lines with no ellipsis, while the same elements without hyphens were cut and an ellipsis was added.

packages/shared/components/MenuLogin/MenuLogin.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  fireEvent.click(await findByText('CONNECT'));
+  await waitFor(async () =>
+    fireEvent.keyPress(await findByPlaceholderText('MenuLogin input'), {
+      key: 'Enter',
+      keyCode: 13,
+    })
+  );
+
+  expect(onSelect).toHaveBeenCalledTimes(0);
```
ravicious, line 24: I wasn't sure whether to put `waitFor` on `fireEvent` vs `expect().toHaveBeenCalledTimes(0)`.

My understanding was that in order to verify that something was not at all, `waitFor` would have to wait until its default timeout runs out. To verify if `fireEvent` was called, it only needs to check if the event was fired.
 
Turns out this is not how `waitFor` works at all. If the callback doesn't return a promise, `waitFor` just continuously calls the callback until it stops throwing an error or the timeout runs out.

Anyway, I guess it's more readable to have expectations without `waitFor`.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/Databases.tsx
```diff
@@ -72,33 +75,69 @@ function ConnectButton({
...
+            placeholder="Enter username"
+            getLoginItems={() =>
+              getDatabaseUsers(dbUri, clustersService, notificationsService)
+            }
+            onSelect={(_, user) => {
+              setDbUser(user);
+              dbNameMenuLoginRef.current.open();
+            }}
+            transformOrigin={{
+              vertical: 'top',
```
gzdunek, line 115: I'd extract origins to a common variable

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/useDatabases.ts
```diff
@@ -24,7 +24,7 @@ export function useDatabases() {
   const dbs = clusterContext.getDbs();
   const syncStatus = clusterContext.getSyncStatus().dbs;
 
-  function connect(dbUri: string, user: string): void {
+  function connect(dbUri: string, dbUser: string, dbName: string): void {
```
gzdunek, line 27: When having 2+ params, it's better to convert them to object

gzdunek (approved):
Looks good!

PR #693, opened 2022-03-23
Make connection icon bigger & fix compressed items on list

gzdunek:
Closes https://github.com/gravitational/webapps.e/issues/148 
Closes https://github.com/gravitational/webapps.e/issues/170

PR #763, opened 2022-04-26
[v9] Backport Teleport Connect (2 of 5)

ravicious:
Backport for the following PRs:

* #679
* #678 
* #680 
* #682 
* #681
* #689 
* #685 
* #688 
* #687
* #696 
* #692
* #691 
* #697 
* #693


packages/teleterm/src/ui/DocumentGateway/useDocumentGateway.ts
```diff
@@ -75,6 +73,12 @@ export default function useGateway(doc: types.DocumentGateway) {
     }
   }, [disconnectAttempt.status]);
 
+  useEffect(() => {
+    if (cluster.connected) {
```
kimlisa, line 77: is optional chaining required here? just saw it on line 49

packages/teleterm/src/ui/TopBar/Clusters/ClusterSelector/ClusterSelector.tsx
```diff
@@ -22,7 +22,13 @@ export const ClusterSelector = forwardRef<HTMLDivElement, ClusterSelectorProps>(
         isClusterSelected={!!props.clusterName}
         title={text}
       >
-        <Text css={{ whiteSpace: 'nowrap' }}>{text}</Text>
+        <Text
+          css={`
+            white-space: nowrap;
+          `}
```
kimlisa, line 28: just curious between `css` and `style`, do both achieve the same thing (just syntax is a bit different?)

kimlisa (approved):

packages/teleterm/src/ui/TopBar/Clusters/ClusterSelector/ClusterSelector.tsx
ravicious: @gzdunek Can you answer Lisa's question here? I remember you were changing that. ;)

packages/teleterm/src/ui/DocumentGateway/useDocumentGateway.ts
ravicious: Now that I look at it, it seems that the checks for cluster being null are unnecessary here. 🤔 If someone logs out from the cluster (and thus removes it) we remove any open resources from that cluster as well. We'll look into removing that after the preview release, thanks!

packages/teleterm/src/ui/TopBar/Clusters/ClusterSelector/ClusterSelector.tsx
gzdunek: `css` works the same as `styled(Text)...` so you have access to the theme for example. Also, `css` generates a css class, while `style` adds inline style to the element.

kimlisa: @gzdunek i see, thanks! should we be using `css` from now on? or does it matter?

PR #768, opened 2022-04-26
Use new Teleconnect icon

gzdunek:
image

I removed all other sizes as they can be auto generated https://www.electron.build/icons.html 
Closes https://github.com/gravitational/webapps.e/issues/139
Closes gravitational/webapps.e#167


ravicious (approved):
Looks perfect. I updated the PR description as this also closes gravitational/webapps.e#167.

PR #771, opened 2022-04-26
[v9] Add support for notarization (#770)

zmb3:
Backports #770


ravicious:
FYI, compiling the app on M1 won't work on the `teleport-v9` branch yet, as it needs an updated version of electron-builder, which is in the second batch of my backports IIRC.

zmb3:
Understood, thanks for the heads up!

ravicious:
I've backported 95% of our changes to `teleport-v9`, so it should work just fine now.

PR #772, opened 2022-04-26
Hide command bar when no cluster is selected

gzdunek:
Closes https://github.com/gravitational/webapps.e/issues/244 
image


ravicious (commented):
This breaks the story for QuickInput.

Perhaps we could create another container component in `TopBar` and move the logic with `workspacesService` there? We probably don't want to call `workspacesService.useState()` on `TopBar`s level directly.

gzdunek:
I think we should fix the story. 
`QuickInput` already depends on the active workspace, so it may work incorrectly when it is not there.

PR #788, opened 2022-04-28
Omit title when comparing previous and current documents

gzdunek:
Closes https://github.com/gravitational/webapps.e/issues/252


ravicious (approved):
This change makes it so that if two documents have different titles but the rest of the fields is the same, they'll be treated as equal. Which is what we need as far as I understand. Each document has other fields than title characterizing it.

And there seems to be no situation where we'd want to treat two documents as different based on the title alone. We should make sure to keep that true going forward.

PR #792, opened 2022-04-29
Update author and owner in package.json to be Gravitational

reedloden:
Our legal name is "Gravitational, Inc." and our GitHub organization is "gravitational".

Fixes gravitational/webapps.e#255.


gzdunek (approved):
Looks good, thanks!

PR #798, opened 2022-04-29
Fix pagination bug

rudream:
### Purpose

This PR resolves https://github.com/gravitational/teleport/issues/12289

### Implementation

For certain types of resources, such as apps, the `startKeys` that the back-end uses doesn't match what is being used at the front. This change makes it so that the request for the first page uses a blank `startKey`.


kimlisa (approved):

zmb3 (approved):
You're too fast!

PR #800, opened 2022-05-02
Indicate whether session.network events were allowed or blocked

zmb3:
We only have a single event type for all `session.network` events, whether the action was allowed or denied. This makes it hard for an administrator or auditor to detect malicious activity, because they have to look at the event JSON to get this info.

Fix this by updating the description in the audit log to clearly indicate whether the action was allowed or denied.

Fixes TEL-Q421-2
Fixes gravitational/teleport.e#331

### DIFF
--- a/packages/teleport/src/Audit/__snapshots__/Audit.story.test.tsx.snap
+++ b/packages/teleport/src/Audit/__snapshots__/Audit.story.test.tsx.snap
@@ -3271,7 +3271,7 @@ exports[`list of all events 1`] = `
         <td
           style="word-break: break-word;"
         >
-          [ALLOW] Program [bash] successfully opened a connection [10.217.136.161 &lt;-&gt; 190.58.129.4:3000] within a session [44c6cea8-362f-11ea-83aa-125400432324]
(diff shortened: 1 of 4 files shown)


packages/teleport/src/Audit/__snapshots__/Audit.story.test.tsx.snap
```diff
@@ -3271,7 +3271,7 @@ exports[`list of all events 1`] = `
         <td
           style="word-break: break-word;"
         >
-          Program [bash] opened a connection [10.217.136.161 &lt;-&gt; 190.58.129.4:3000] within a session [44c6cea8-362f-11ea-83aa-125400432324]
+          [ALLOW] Program [bash] successfully opened a connection [10.217.136.161 &lt;-&gt; 190.58.129.4:3000] within a session [44c6cea8-362f-11ea-83aa-125400432324]
```
reedloden, line 3274: Possible to test the "[DENY]" part as well?

reedloden (approved):
This seems reasonable to me (and a great UI improvement!), though I defer to somebody like @kimlisa who knows this codebase better to make sure nothing has been missed.

Thanks for adding the additional test case!

kimlisa (approved):

PR #812, opened 2022-05-05
useDocumentGateway: Pin shell to correct cluster

ravicious:
Closes gravitational/webapps.e#269.

Commit b5d30b1fa fixed some reconnect issues, but it also introduced a
locally scoped `cluster` variable. This made it seem like it's the cluster
to which the given gateway belongs, but it was actually the root cluster.

This commit renames the old `cluster` variable to `rootCluster`. It also
makes sure that `workspaceDocumentsService.openNewTerminal` receives
correct root & leaf cluster IDs.


ravicious:
Also, @gzdunek do you know when `rootCluster` can be nil? Lisa asked about this here: https://github.com/gravitational/webapps/pull/763#discussion_r859477626

It seems to me that there should be no situation in which the root cluster is nil, since if you remove (log out of) a root cluster, we remove all its resources as well?

gzdunek:
AFAIK there should be no situation when `rootCluster` is nil

gzdunek (approved):

PR #815, opened 2022-05-05
Create StepSlider Component

kimlisa:
#### Description
Create a component that does sliding and height increase animations. Plan to use it for login and register flows.

#### Demo
TODO


packages/teleport/src/components/StepSlider/StepSlider.story.tsx
```diff
@@ -0,0 +1,319 @@
...
+export const SingleStaticFlow = () => {
+  return (
+    <Card bg="primary.light" my="5" mx="auto" width={464}>
+      <Text typography="h3" pt={5} textAlign="center" color="light">
+        Static Title
+      </Text>
+      <StepSlider<typeof singleFlow>
+        flows={singleFlow}
+        currFlow={'default'}
+        render={(slider, ref, parentStyle) => (
```
hatched, line 35: The `render` method here doesn't feel very "reacty". Why would we do this instead of having the `StepSlider` return itself wrapped in a `Box` with the `ref` and `parentStyle` set?

packages/teleport/src/components/StepSlider/StepSlider.tsx
```diff
@@ -0,0 +1,258 @@
...
+          preMountState.current.step = step + 1;
+          setPreMount(true);
+          setClassName('next');
+        }}
+        onPrev={() => {
+          preMountState.current.step = step - 1;
+          setPreMount(true);
+          setClassName('prev');
+        }}
+        onSwitchFlow={(flow, isBack = false) => {
```
hatched, line 86: `isBack` feels like a bit of a bandaid

If switching flows always resets that flows step to 0, why would the animation be anything other than `next`?

packages/teleport/src/components/StepSlider/StepSlider.tsx
```diff
@@ -0,0 +1,258 @@
...
+          }
+          setClassName('prev');
+        }}
+        {...rest}
+      />
+    );
+  }
+
+  let $content;
+  const Step = flows[currFlow][step];
```
hatched, line 103: In order to support the flow being switched from the outside by updating the `currFlow` prop I THINK we need a `useEffect` for `currFlow` so that we can appropriately animate it back to 0?

hatched (commented):
Thanks @kimlisa this looks great! Let me know if you'd like to discuss any of my comments.

packages/teleport/src/components/StepSlider/StepSlider.story.tsx
```diff
@@ -0,0 +1,319 @@
...
+import { Text, Card, ButtonPrimary, ButtonLink, Box } from 'design';
+
+export default {
+  title: 'Teleport/StepSlider',
+};
+
+const singleFlow = { default: [Body1, Body2] };
+export const SingleStaticFlow = () => {
+  return (
+    <Card bg="primary.light" my="5" mx="auto" width={464}>
```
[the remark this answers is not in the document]
kimlisa: i think it's fine b/c the design comps I used don't do anything other than to look a certain way, and the tests i used strictly looks for texts

packages/teleport/src/components/StepSlider/StepSlider.tsx
kimlisa: it was to provide a different animation if user wants to switch back to another flow:
e.g. if we had:
  - primary flow
  - secondary flow

User will start at primary:
- user clicks secondary flow (applies `prev` animation, on second thought this should be `next` animation instead)
- user is now on secondary flow and wants to go "back" to the primary flow, so `isBack` is set to true, and we apply `prev` animation

packages/teleport/src/components/StepSlider/StepSlider.story.tsx
kimlisa: you're right, i don't remember why i did it this way :thinking:

packages/teleport/src/components/StepSlider/StepSlider.tsx
kimlisa: hm, I wasn't sure what you meant. Are you saying currently, switching flow from outside would not work?

packages/teleport/src/components/StepSlider/StepSlider.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+  // step defines the current step we are in the current flow.
+  const [step, setStep] = useState(0);
+  const [height, setHeight] = useState(0);
+  const [className, setClassName] = useState<'next' | 'prev' | ''>('');
+
+  // preMount is used to invisibly render the next view so we
+  // can get its height. This height is needed in advance
+  // so the height can be animated along with the transform transitions.
+  const [preMount, setPreMount] = useState(false);
```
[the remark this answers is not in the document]
kimlisa: i wasn't sure what you meant, like pre-mount both the possible next and possible prev components?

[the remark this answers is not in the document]
hatched: Yep exactly. They have to be rendered anyways to animate into them so I'm not sure if the performance savings is worth the code complexity.

packages/teleport/src/components/StepSlider/StepSlider.tsx
hatched: Well no, it'll work, but I think if you're on flow 1 step 3, then externally change it to flow 2, it'll have to go from flow 1 step 3 to flow 2 step 0. And I don't think this will correctly apply the "prev" animation.

kimlisa: oh i think i get what you mean, like this:

flow1: 1, 2, 3, 4a
flow2:             4b

So on step 3, we can go `4a` or `4b`, and be able to go back to step 3?

packages/teleport/src/components/StepSlider/StepSlider.tsx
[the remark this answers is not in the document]
kimlisa: hm, i didn't create it like this for performance savings but if that is what it appears yay lol. i did try to keep it mounted and i ran into issues so i'm going to leave it as is until i or someone has more time to work on it

hatched (approved):
Thanks for the updates!

PR #816, opened 2022-05-06
Refactor Teleport Login Flow

kimlisa:
#### Description
Part of the passwordless work: users can now add a pwdless device through their dashboard and login with it

- display primary and secondary logins on different cards
- use sliding transitions
- add a passwordless device through dashboard

Note: Will be easier to review by commit (reduce the noise produced by tests/snapshots)

#### Testing
- [x] Add pwdless device through dashboard and login
- [x] Primary pwdless login
- [x] Primary username/pwd login
- [x] Primary sso login
- [x] Cloud login works as before
- [x] Test all mfa options and login work as before

#### Required to merge
- [x] merge counter part PR https://github.com/gravitational/webapps.e/pull/272
- [x] merge base PR https://github.com/gravitational/webapps/pull/815
- [x] update e-ref

#### Demo when primary is user + pass
All other variations can be found on storybook under `Teleport/FormLogin`

login-slider


packages/teleport/src/config.ts
```diff
@@ -192,6 +197,23 @@ const cfg = {
...
   },
 
+  isPasswordlessEnabled() {
+    return cfg.auth.allowPasswordless;
+  },
+
+  getPrimaryAuthType(): PrimaryAuthType {
+    if (
+      cfg.auth.allowPasswordless &&
+      cfg.auth.localConnectorName === 'passwordless'
```
ravicious, line 209: Why does passwordless need to be controlled through two separate variables instead of one?

packages/teleport/src/Account/ManageDevices/AddDevice/AddDevice.tsx
```diff
@@ -200,6 +207,19 @@ export function AddDevice({
...
                 />
               )}
+              {mfaOption.value === 'webauthn' && (
+                <FieldSelect
+                  width="50%"
+                  label="Allow Passwordless Login?"
+                  value={usageOption}
+                  options={deviceUsageOpts}
+                  onChange={(o: DeviceusageOpt) => {
+                    validator.reset();
```
ravicious, line 217: What's the reason behind calling `validator.reset()` here?

packages/teleport/src/Login/Login.test.tsx
```diff
@@ -35,80 +35,26 @@ test('basic rendering', () => {
...
 
-  // test login pathways
-  await waitFor(() => fireEvent.click(getByText(/login/i)));
+  // test login and redirect
+  await waitFor(() => fireEvent.click(getByText('Sign In')));
   expect(auth.login).toHaveBeenCalledWith('username', '123', '');
   expect(history.push).toHaveBeenCalledWith('http://localhost/web', true);
 });
 
-test('login with password and otp', async () => {
```
ravicious, line 59: Hmm, I don't get why we're removing these tests, aren't those login methods still supported?

ravicious (approved):
I left some questions, but it looks good in general. I want to click through it but I don't want that to block the PR.

Do I need a cluster that's running Teleport off master to test this?

packages/shared/services/types.ts
```diff
@@ -20,6 +20,12 @@ export type Auth2faType = 'otp' | 'off' | 'optional' | 'on' | 'webauthn';
 
 export type AuthType = 'local' | 'github' | 'oidc' | 'saml';
 
+// PrimaryAuthType defines types where if:
+//  - local: preferred authn is with username and password
+//  - pwdless: preferred authn is passwordless
+//  - sso: preferred authn is either with github, oidc or saml provider
+export type PrimaryAuthType = 'local' | 'pwdless' | 'sso';
```
gzdunek, line 27: In some places we use full word `passwordless` and in others `pwdless`. Can we stick to one version (I prefer full words).

gzdunek (approved):

kimlisa:
> Do I need a cluster that's running Teleport off master to test this?

@ravicious yes

packages/shared/components/FieldInput/FieldInput.tsx
```diff
@@ -36,6 +37,36 @@ export default function FieldInput({
...
+    // autoFocusOnTransitionEnd focus's the input element after transition property name
+    // defined by 'transitionPropertyName' has ended. This prevents auto focusing during
+    // transitioning which causes transition to be jumpy caused by trying to bring focused
+    // element into view. This also prevents prematurely showing the browser password
+    // manager icons and tooltips while transitioing.
+    function autoFocusOnTransitionEnd(e) {
+      if (e.propertyName !== transitionPropertyName) return;
+      inputRef.current.focus();
+      // Since we only need to auto focus one time, the listener is no longer needed.
+      window.removeEventListener('transitionend', autoFocusOnTransitionEnd);
```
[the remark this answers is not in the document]
kimlisa: that would've been ideal, but the `transitionend` event can fire from other transition types eg: `opacity`, `background-color` etc

packages/teleport/src/config.ts
kimlisa: i actually don't need to, given below config, if the `connector_name` is set to `passwordless`, disabling passwodless will throw an error as `invalid connector name passwordless disabled`

```
auth_service:
  enabled: yes
  cluster_name: "im-a-cluster-name"
  authentication:
    type: local
    passwordless: true            # enables/disables passwordless
    connector_name: passwordless  # sets passwordless as the default
```

packages/teleport/src/Account/ManageDevices/AddDevice/AddDevice.tsx
kimlisa: ah, it was a copy paste left over, removed (but in general, it is to clear any validation errors when users switches options, otherwise the validation error carries over to the switched input boxes that uses the same validator)

packages/teleport/src/Login/Login.test.tsx
kimlisa: there were similar tests found in [FormLogin.tsx](https://github.com/gravitational/webapps/blob/master/packages/teleport/src/components/FormLogin/FormLogin.test.tsx). I didn't think it was necessary to repeat the test. For this test file though, I kept one test to test the `redirect` path which is the same for login user + mfa combos.

packages/shared/components/FieldInput/FieldInput.tsx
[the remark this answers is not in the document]
ravicious: Ah, sorry, this whole time I misunderstood what the `transitionend` listener is for. I see now why we can't use `once`.

ravicious:
Okay, I managed to run it and whoa this really is magic. :D 

One thing that bothers me though: the "Other sign-in options" and "Back" `<a>` tags don't have a `href` attribute and the `StyledPaswordlessBtn` component is a regular div. This means that the login form can no longer be submitted using only the keyboard. I imagine that might be a huge issue for some users and it reduces the accessibility of the form quite a bit.

Could you address this in a separate PR? I did a quick proof of concept by changing `StyledPaswordlessBtn` and `StyledOptionBtn` to regular buttons and I was able to submit the form using only the keyboard. The styles on those buttons need to be updated though as well as some kind of focus styles need to be added so that the user can see which button is focused.

PR #817, opened 2022-05-09
Create new SlideTab component

hatched:
https://user-images.githubusercontent.com/532033/167508540-9b41f809-cb64-46cc-9928-ba9299bb57d7.mov

This component is required to complete https://github.com/gravitational/teleport/issues/10342


packages/design/src/SlideTabs/SlideTabs.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+import React, { useEffect, useState } from 'react';
+import styled from 'styled-components';
+import { props } from 'teleport/Apps/Apps.story';
```
hatched, line 18: This shouldn't be here and will be removed.

packages/design/src/SlideTabs/SlideTabs.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  it('calls the onChange handler when the tab is changed', () => {
+    const cb = jest.fn();
+    const { container } = render(
+      <SlideTabs onChange={cb} tabs={['aws', 'automatically', 'manually']} />
+    );
+    fireEvent.click(container.querySelector('label[for=slide-tab-manually]'));
+    expect(cb.mock.calls).toHaveLength(2);
+    expect(cb.mock.calls).toEqual([[0], [2]]);
```
kimlisa, line 52: wasnt sure i was reading it right, does this mean `onChange` is getting called twice for one click?

kimlisa (approved):
we put license headers on every file, but other than that LGTM

packages/design/src/SlideTabs/SlideTabs.test.tsx
hatched: Good question, I'll add a comment to this effect.

But no, what's happening is that when the component initializes it's setting the tab to index `0` then when you click it's setting it to index `2`. This was intentional because this tab control will be the source of truth for the Slider so it needs to know what to display first as you could, in theory, have it start at a different index.

packages/design/src/SlideTabs/SlideTabs.tsx
```diff
@@ -0,0 +1,96 @@
...
+}
+
+type props = {
+  // A list of tab names that you'd like displayed in the list of tabs.
+  tabs: string[];
+  // The name you'd like to use for the form if using multiple tabs on the page.
+  // Default: "slide-tab"
+  name?: string;
+  // To be notified when the selected tab changes supply it with this fn.
+  onChange?: (selectedTab: number) => void;
```
rudream, line 54: Why is this prop optional? Is there a scenario where this component would be used without this prop?

hatched: I have no idea any longer why I made this optional :) I honestly can't think of a time when you'd want to use this component and not be notified of changes. I'll update this to make it required.

rudream (approved):

PR #818, opened 2022-05-11
Refactor Teleport reset invite flow

kimlisa:
#### Description
Part of the passwordless work: users can now register with passwordless device or standard user/pass/mfa on `reset` or `invite`

- display primary and secondary register flows depending on primary authn type
- use sliding transitions
- add a `success` comp after successfully registering (prior you'd auto redirect to dashboard, now you have to click a button)
- sneaked in this link to button request change: https://github.com/gravitational/webapps/pull/816#issuecomment-1122491457

note: i tried to keep each commit specific to something

#### Testing
- [x] Register pwdless device through reset/invite + login with it
- [x] Register password only through reset/invite
- [x] Register password + mfa through reset/invite
- [x] Cloud register, recovery codes, recover (set) password through recovery with passwordless device works

#### Required to merge
- [ ] required backend changes: https://github.com/gravitational/teleport/pull/11748
- [x] merge counter part PR: https://github.com/gravitational/webapps.e/pull/276
- [x] update e-ref

#### Demo with invite password registration
All other variations can be found on storybook under `Teleport/Welcome/Form`

invite-flow


packages/teleport/src/Welcome/NewCredentials/NewMfaDevice.tsx
```diff
@@ -0,0 +1,265 @@
...
+              </Text>
+            </Box>
+          </Flex>
+          {submitAttempt.status === 'failed' && (
+            <Danger children={submitAttempt.statusText} />
+          )}
+          <Text typography="subtitle1" color="light" caps mb={1}>
+            Two-Factor Method
+          </Text>
+          <Box>
```
rudream, line 126: Spacing is a little off below the radio buttons: 
image


```suggestion
          <Box mb={1}>
```

kimlisa: shifted things around a bit, but the change is here: https://github.com/gravitational/webapps/pull/818/files#diff-7a98121ef6a3ad155b6fb45127d9ba29b19bf15393be9bf92c555c0cf7712e33R126

packages/teleport/src/Welcome/NewCredentials/NewMfaDevice.tsx
```diff
@@ -0,0 +1,265 @@
...
+                  .
+                </Text>
+              </>
+            )}
+            {mfaType?.value === 'webauthn' && (
+              <>
+                <Image src={imgSrc} width="220px" height="154px" />
+                <Text fontSize={1} color="text.secondary" textAlign="center">
+                  We support a wide range of hardware devices including yubi
+                  keys, touch ID, watches, and more.
```
ravicious, line 175:
```suggestion
                  keys, Touch ID, watches, and more.
```

packages/teleport/src/Welcome/NewCredentials/NewPassword.tsx
```diff
@@ -0,0 +1,146 @@
...
+      next();
+      return;
+    }
+
+    onSubmit(password);
+  }
+
+  function switchToPasswordlessFlow(e, isBack = false) {
+    e.preventDefault();
+    switchFlow('passwordless', isBack);
```
gzdunek, line 67: Hmm I have problems with figuring out what `isBack` means, can we find more meaningful name?

packages/teleport/src/Welcome/NewCredentials/NewPasswordlessDevice.tsx
```diff
@@ -0,0 +1,115 @@
...
+import React, { useState } from 'react';
+import { Text, Box, ButtonPrimary, ButtonText } from 'design';
+import { Danger } from 'design/Alert';
+import FieldInput from 'shared/components/FieldInput';
+import Validation, { Validator } from 'shared/components/Validation';
+import { requiredField } from 'shared/components/Validation/rules';
+import { SliderProps } from 'teleport/components/StepSlider';
+import { Props, LoginFlow } from './NewCredentials';
+
+export default function NewPwdlessDevice(
```
gzdunek, line 26: `Pwdless` -> `Passwordless`?

packages/teleport/src/Welcome/NewCredentials/NewMfaDevice.tsx
```diff
@@ -0,0 +1,265 @@
...
+    clearSubmitAttempt,
+    auth2faType,
+    onSubmitWithWebauthn,
+    onSubmit,
+    password,
+    prev,
+    refCallback,
+  } = props;
+  const [otp, setOtp] = useState('');
+  const mfaOptions = useMemo(
```
gzdunek, line 47: `useMemo` doesn't seem to be necessary

packages/teleport/src/components/FormLogin/FormLogin.tsx
```diff
@@ -413,19 +418,23 @@ const Divider = () => (
...
     mb={2}
   >
     <StyledOr>Or</StyledOr>
   </Flex>
 );
 
-const StyledPaswordlessBtn = styled(Box)`
-  cursor: pointer;
-  transition: all 0.3s;
+const StyledPaswordlessBtn = styled(ButtonText)`
```
ravicious, line 428: Great, this makes keyboard navigation possible.

One remaining issue is that after the sliding transition, for example going from step 1 to step 2, the focus for the page doesn't get reset. I press enter to submit the first step, but then:

* pressing enter does nothing (I expect it to submit the form)
* pressing tab focuses on the location bar, signaling that I tabbed past the last focusable element on the page

https://user-images.githubusercontent.com/27113/168096363-abbae751-67a8-465d-bc9f-dd9e356ea614.mov

Here I press tab after submitting the first step, expecting the tab to tab me through the form elements so that I can get to the device name field, but instead the browser focuses on the location bar.


https://user-images.githubusercontent.com/27113/168096593-a52d4e9b-86b6-490e-9b77-18705fed0d42.mov

Here I submit the first step and then press enter a couple of times, expecting the second step to be submitted, but nothing happens because there's no focus on the form. I then press tab expecting to be focused on the form, but instead the browser focuses on the location bar.

Off the top of my head I don't remember how to fix that though. 🤔 It's either through the `autofocus` attribute or some JS call that rewinds/resets the focus.

ravicious: @gzdunek See the last paragraph of my comment above, do you remember how to do this? IIRC you might have worked on something similar to that a few weeks ago.

ravicious (approved):
Left a comment about tabbing behavior, but once that's fixed I think this PR is good to go.

I didn't have time to look closely at the component changes in the "Refactor/Create authn step components" commit, I gave it only a cursory look. But I did reset the password and used the invite flow like 10 times using various auth methods and I didn't run into any problems.

packages/teleport/src/Welcome/NewCredentials/NewMfaDevice.tsx
```diff
@@ -0,0 +1,265 @@
...
+
+  useEffect(() => {
+    setTransitionPropertyName('');
+  }, []);
+
+  function onBtnClick(
+    e: React.MouseEvent<HTMLButtonElement>,
+    validator: Validator
+  ) {
+    e.preventDefault(); // prevent form submit default
```
gzdunek, line 69: Is it required? From what I know it is needed for form elements (`<form onSubmit={}` />), but in this case it's just `onClick()`.

gzdunek (approved):
Can we use named for exports as discussed here? https://gravitational.slack.com/archives/C02DQ1C2BMW/p1642178529014800

packages/teleport/src/components/FormLogin/FormLogin.tsx
gzdunek: Yes, but in my case I had no input to focus, so I created a hidden one:
```typescript
const focusGrabber = (
  <input
    style={{
      opacity: 0,
      position: 'absolute',
      height: 0,
      zIndex: -1,
    }}
    autoFocus={true}
  />
);
```
but here we have an input, so `autoFocus` should work 🤔

packages/teleport/src/Welcome/NewCredentials/NewPassword.tsx
```diff
@@ -0,0 +1,146 @@
...
+    if (mfaEnabled) {
+      next();
+      return;
+    }
+
+    onSubmit(password);
+  }
+
+  function switchToPasswordlessFlow(e, isBack = false) {
+    e.preventDefault();
```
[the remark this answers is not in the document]
kimlisa: it's because the parent `NewCredentials.tsx` that renders these childrens uses `form`

packages/teleport/src/components/FormLogin/FormLogin.tsx
kimlisa: Thanks, I didn't notice this. I guess sometimes in Chrome, the `transitionend` event for `height` gets canceled (i wasn't sure why) but the fix was to also look for a `transitioncancel` for `height`. This will always autofocus on transition. Strangely this doesn't happen in firefox.

ravicious: I did a quick check and it looks like the reset form works now, but the problem is still present in the login form.

When I open the login form, then tab to "Other sign-in options" and press enter, the slide transition happens, I see the "Passwordless" button, but pressing enter does nothing and pressing tab focuses on the location bar.

kimlisa: ah, i see what you mean, so on render/transition, focus on the first clickable button (i just assumed to just focus on input box regardless of position). i'll try to sneak this in my next PR

ravicious: > on render/transition, focus on the first clickable button

I'd say more like "on render/transition, focus on the first focusable [form control element](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/elements).

In other words, despite having no "real" page transition here I'd like us to make the form behave as if we did transition into a page with a form that has a form control element with the `autofocus` attribute. I assume that this is what users would expect instead of the focus being somehow set at the last focusable element instead having tab change focus to the location bar.

PR #819, opened 2022-05-11
Update Electron & add shared process

gzdunek:
This PR:
- adds shared gRPC process
- updates Electron to v18

Best to review commit-by-commit.

Previously, `node-pty` which creates PTY process was used in the `preload` script. That blocked as from upgrading the Electron to version above 13 (`allowRendererProcessResuse` flag has been removed in v14). 
In the new implementation we create a new `shared` process that takes care of interaction with `node-pty`. It runs a gRPC server and communicates with renderer via unix sockets for Inter-Proces Communication (same as `tshd`).
The gRPC `PtyHost` service consist of three endpoints:
- `createPtyProces` - unary call
- `exchangeEvents` - bideractional stream (we send `start`(maybe it should be `open`?)/`data`/`resize`/`onOpen`/`onExit` events here)
- `getCwd` - unary call

On the first, client invokes `createPtyProcess`, that creates a `PtyProcess` in the `shared` process and returns `ptyId`. Then `exchangeEvents` channel is established. ~~Unfortunately, there is no way to add any data when the channel is created (so we could set `ptyId` for it). To overcome it, we set `ptyId` in the first event which is sent (`start`)~~, so _sever_ knows which `process` this channel belongs to. (We use stream metadata to set `ptyId`).
I thought about just having an `exchangeEvents` endpoint and converting `createPtyProcess` and `getCwd` to some kind of events (thanks to this we could get rid of `ptyId`s), but I think those are different things than regular pty events. For example, when we send `data` event we don't care about any response. On the contrary, when we ask for the `cwd` we care about the response for that particular request. Additionally, thanks to this, the ui part works without any changes. What do you think about this reasoning?
When client runs `dipose()` we notify the _sever_ by closing the `exchangeEvents` channel (I think it is a nice usage of the gRPC features).

TODO:
- add gRPC files generation from `.proto` files (for now I used `teleport` generator).
- starting a new process can improved 
- add logger

BTW, when you open the devtools, you can see strings like `[76721:0511/133521.851672:ERROR:CONSOLE(6)] "console.assert", source: devtools://devtools/bundled/panels/console/console.js (6)`
[It looks like a bug in the Electron v18](https://stackoverflow.com/questions/71766903/electron-app-is-getting-343520406-075851-692errorconsole6-console-assert) 🤔

Closes https://github.com/gravitational/webapps.e/issues/203


ravicious:
I read just the PR description so far, but I figured I'll answer and ask some questions before jumping into the code.

> I thought about just having an exchangeEvents endpoint and converting createPtyProcess and getCwd to some kind of events (thanks to this we could get rid of ptyIds), but I think those are different things than regular pty events. (…) What do you think about this reasoning?

I think that's a good approach! The reasoning made sense and after reading this I went to check how Tabby does this and it does exactly the same thing: creating a process and getting the PID is request/response, while things like resize/read/write are events.

> When client runs `dispose()` we notify the sever by closing the `exchangeEvents` channel (I think it is a nice usage of the gRPC features).

I haven't looked at how this runs yet, but I remember that we have to be careful there. From what I remember, on Windows when the main process gets closed, the processes it spawn are killed as well. But on macOS & Linux it's easy to leave orphaned processes. Did you perhaps investigate how this setup behaves in different scenarios after the main/renderer process crashes?

> In the new implementation we create a new `shared` process that takes care of interaction with node-pty. It runs a gRPC server (…)

I'm just trying to clearly understand this: the reason we run a gRPC server in that new process as opposed to running a client there is that this just sort of follows the natural way to do it? If the shared process was running a client, then implementing calls like `createPtyProces`/`getCwd` would be a lot harder because then they'd need to be implemented within a bidirectional stream, right?

> This PR (…) updates Electron to v18

Wow, and other than dealing with node-pty it really was just updating package.json and that's all? 😄 Were there any notable breaking changes between v13 and v18 that might affect us right now or in the future?

gzdunek:
> (...) on macOS & Linux it's easy to leave orphaned processes. Did you perhaps investigate how this setup behaves in different scenarios after the main/renderer process crashes?

Good point, I haven't investigated it yet, but I will take a look.

 > I'm just trying to clearly understand this: the reason we run a gRPC server in that new process as opposed to running a client there is that this just sort of follows the natural way to do it? If the shared process was running a client, then implementing calls like createPtyProces/getCwd would be a lot harder because then they'd need to be implemented within a bidirectional stream, right?

So, the main process would run a server, correct?  
I think it would be very hard to implement even a bidirectional stream as it would mean establishing a connection between two clients (`shared` to `main` and `main` to `renderer`). 
And IMO this `shared` can (and should) communicate with the `main` process in the future. Let's imagine a situation where we move `fileStorage` to that process. Then we would like to have clients on both sides (at `renderer` and `main`) and `shared` would become truly shared 😄

> Wow, and other than dealing with node-pty it really was just updating package.json and that's all?

Yes!

> Were there any notable breaking changes between v13 and v18 that might affect us right now or in the future?

I took a look at the breaking changes, it doesn't seem that anything affects us.

packages/teleterm/src/sharedProcess/ptyHost/server/ptyEventsStreamHandler.ts
```diff
@@ -0,0 +1,91 @@
...
+  PtyServerEvent,
+} from '../v1/ptyHostService_pb';
+import { PtyProcess } from '../server/ptyProcess';
+
+export class PtyEventsStreamHandler {
+  private ptyId?: string;
+
+  constructor(
+    private readonly stream: ServerDuplexStream<PtyClientEvent, PtyServerEvent>,
+    private readonly ptyProcesses: Map<string, PtyProcess>
```
ravicious, line 18: > Unfortunately, there is no way to add any data when the channel is created (so we could set `ptyId` for it). To overcome it, we set `ptyId` in the first event which is sent (`start`), so _sever_ knows which `process` this channel belongs to.

It'd be a good idea to document this somewhere, perhaps in this constructor? It'd help explain why `PtyEventsStreamHandler` cannot take `ptyId` as a constructor argument.

packages/teleterm/src/main.ts
```diff
@@ -31,13 +31,20 @@ const mainProcess = MainProcess.create({
...
-// node-pty is not yet context aware
-app.allowRendererProcessReuse = false;
+const sharedProcess = fork(
+  path.join(__dirname, 'sharedProcess.js'),
+  [`--addr=${settings.shared.networkAddr}`],
+  {
+    stdio: 'inherit',
+  }
+);
+
```
ravicious, line 41: Should we move that inside `packages/teleterm/src/mainProcess/mainProcess.ts`? This is where the tshd process is started as well.

packages/teleterm/src/main.ts
```diff
@@ -31,13 +31,20 @@ const mainProcess = MainProcess.create({
   fileStorage,
 });
 
-// node-pty is not yet context aware
-app.allowRendererProcessReuse = false;
+const sharedProcess = fork(
```
ravicious, line 34: Did you consider documenting the whole communication flow somewhere?

PTY process communication is already quite complex, now we're adding another layer (layers?) of indirection and on top of that there's bidirectional streaming. My head hurt after reviewing this for over an hour!

I was thinking of a sequence diagram, created using something like https://www.websequencediagrams.com/ so that it can be extended if needed. But it wouldn't necessarily have to be a diagram.

My main points when trying to fit the code in my head is that there are various events that we register listeners to. Some of those events are sent from the client to the server, some from the server to the client and then there are also events created my the PTY process itself.

It'd be good to have a piece of documentation which could help you grok this faster.

ravicious (commented):
Wow, that's a lot of code. :D Good job figuring this all out. I think the general approach makes sense, in some places we could make some tweaks to make it easier to understand.

I haven't run the code yet, just spend some time reading it. But before I dive deeper I'll wait for your replies, especially since I know you planned to add more stuff.

ravicious:
> So, the main process would run a server, correct?  
I think it would be very hard to implement even a bidirectional stream as it would mean establishing a connection between two clients (`shared` to `main` and `main` to `renderer`). 

I don't think we understood each other here. 😏 I was just wondering if the main process could run only the server and the shared process would be its client. That is, I was trying to understand "by inversion" why it is that the shared process has to run the server and the main process is only its client.

Running two servers on two separate processes seems like bit much? But I haven't thought about that a lot.

ravicious:
Re: what I said about killing child processes:

> If forked from the main process, any child processes you create will be killed when application is terminated. Children forked from a renderer process, on the other hand, will be killed as soon as their containing window is closed. This is an important distinction. OSX terminates applications when the user explicitly quits them, while Windows and Linux terminates applications when all of their windows are closed. For consistency between the two, make sure to call fork from the main process if the child process must exist for the lifetime of the application.

https://www.matthewslipper.com/2019/09/22/everything-you-wanted-electron-child-process.html#child-process-lifetimes

packages/teleterm/src/sharedProcess/ptyHost/client/buildPtyOptions.ts
```diff
@@ -124,11 +70,11 @@ async function buildOptions(
 function prependBinDirToPath(
   env: typeof process.env,
   settings: RuntimeSettings
-) {
+): void {
   let path: string = env['PATH'] || '';
 
   if (!path.trim()) {
-    path = settings.binDir;
+    path = settings.binDir || '';
```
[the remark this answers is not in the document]
gzdunek: Ah no, removed.

packages/teleterm/src/mainProcess/types.ts
```diff
@@ -16,6 +16,9 @@ export type RuntimeSettings = {
     homeDir: string;
     flags: string[];
   };
+  shared: {
```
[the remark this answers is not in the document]
gzdunek: I agree, changed to `sharedProcess`.

packages/teleterm/src/preload.ts
```diff
@@ -16,7 +16,7 @@ const loggerService = createLoggerService({
 PreloadLogger.init(loggerService);
 
 const tshClient = createTshClient(runtimeSettings.tshd.networkAddr);
-const ptyServiceClient = createPtyService(runtimeSettings);
+const ptyServiceClient = createPtyHostServiceClient(runtimeSettings);
```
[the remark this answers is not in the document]
gzdunek: Yeah, that name was a bit confusing, I'm fine with `ptyHost`.

packages/teleterm/src/sharedProcess/ptyHost/client/index.ts
```diff
@@ -0,0 +1 @@
+export * from './ptyHostServiceClient';
```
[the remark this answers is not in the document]
gzdunek: IMO it is easier to think about it when related things are close to each other. Additionally, we don't have to re-export definitions from the generated files.

packages/teleterm/src/sharedProcess/ptyHost/server/ptyEventsStreamHandler.ts
```diff
@@ -0,0 +1,91 @@
...
+        return this.handleDataEvent(event.getData());
+      case PtyClientEvent.EventCase.RESIZE:
+        return this.handleResizeEvent(event.getResize());
+    }
+  }
+
+  private handleStartEvent(event: PtyEventStart): void {
+    this.ptyId = event.getId();
+    const ptyProcess = this.getPtyProcess();
+    if (ptyProcess) {
```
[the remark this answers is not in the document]
gzdunek: Hmm, we don't need this check.

packages/teleterm/src/sharedProcess/ptyHost/server/ptyEventsStreamHandler.ts
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  private handleResizeEvent(event: PtyEventResize): void {
+    this.getPtyProcess()?.resize(event.getColumns(), event.getRows());
+  }
+
+  private handleStreamError(error: Error): void {
+    console.error(error); //TODO(gzdunek) use proper logger
+    this.cleanResources();
+  }
```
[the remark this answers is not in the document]
gzdunek: Generally, error can be emitted for example on network failure or on authentication error, but in our case they won't happen. For now, I can't imagine what else can go wrong, so I would leave this general purpose handler.

packages/teleterm/src/sharedProcess/ptyHost/server/ptyEventsStreamHandler.ts
gzdunek: Good idea, done

packages/teleterm/src/main.ts
gzdunek: Yes, done

packages/teleterm/src/main.ts
gzdunek: Please take a look at the [diagram](https://www.websequencediagrams.com/cgi-bin/cdraw?lz=dGl0bGUgVW4AAwVkCgpEb2N1bWVudCBUZXJtaW5hbC0-UFRZIFByb2Nlc3MgQ2xpZW50OiB3YW50cyBuZXcgUFRZCgAQEgAwBkhvc3QgU2VydmVyOiBjcmVhdGVQdHkARQcob3B0aW9ucykAOgUAIAstAGYNOgBgBgAtCgAeEgCBDRVwdHlJZCBvZiB0aGUgcACBOAdpcyByZXR1cm5lZACBDyZlc3RhYmxpc2ggZXhjaGFuZ2VFdmVudHMoKSBjaGFubmVsCm5vdGUgbGVmdCBvZiAAghkTCmMAgi8FIGhhcyBiZWVuAIICB2QsIApzbwAeEyBjYW4gYXR0YWNoIGV2ADQGbmRsZXJzIAp0bwCBSgUAdAcgKG9uRGF0YS9vbk9wZW4vb25FeGl0KQplbmQgbm90ZQCDAxUAg1ARAIImBQCCGQlvYmplY3QAg18oc3RhcnQAgwsHAIIqIgCCNQ0uADYGAIMiBSkgAIJABnJpZ2gAgj4JAIQwDQoAgnEOAIIZB2VzAIITDwCCGQgAhSkKAIF_JQCENRgAgVcKAIVXDACFSRMAgnsGAIRzKACBeg8AMA0AgwwjAGQJAIcFKGRpc3Bvc2UAgn8pbmQAhV8PAIVjCQCCDx4AYQcAhmUJYW5kIHJlbW92ZSBpdAo&s=default).

Do you think it helps with understanding the flow?

gzdunek:
>I was just wondering if the main process could run only the server and the shared process would be its client. That is, I was trying to understand "by inversion" why it is that the shared process has to run the server and the main process is only its client.
Running two servers on two separate processes seems like bit much?

So how the communication between the pty in the shared process and the pty client would look like?
Wouldn't it be `shared<=>main<=>renderer`?

Two servers sounds like overkill for me 😅

And regarding killing child process, I checked that and it wasn't killed when main process crashed. 
I fixed that by changing `throw error` to `app.quit()` (which run `dispose()`). The app was closed anyway and we have a GH issue to improve this behavior.
I don't fully understand that part: `If forked from the main process, any child processes you create will be killed when application is terminated.`. It sounds like `sharedProcess.kill()` is not necessary, but process is not killed after the app quit without the explicit kill.

gzdunek:
I was wrong about passing data when the stream is created. We can pass `ptyId` to the stream metadata, so it will be sent when the stream is established.

packages/teleterm/src/sharedProcess/ptyHost/server/ptyHostServiceServer.ts
```diff
@@ -0,0 +1,44 @@
+import { unique } from 'teleterm/ui/utils';
+import { PtyProcess } from './ptyProcess';
+import { IPtyHostServer } from '../v1/ptyHostService_grpc_pb';
+import { PtyCwd, PtyId } from '../v1/ptyHostService_pb';
+import { PtyEventsStreamHandler } from './ptyEventsStreamHandler';
+
+export function getPtyHostServiceServerImpl(): IPtyHostServer {
```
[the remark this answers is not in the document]
ravicious: I haven't checked how that looks like in the grand scheme of things, but on its own it sounds good. 👌

packages/teleterm/src/sharedProcess/ptyHost/server/ptyEventsStreamHandler.ts
```diff
@@ -0,0 +1,86 @@
...
+    if (this.ptyId) {
+      this.ptyProcesses.delete(this.ptyId);
+    }
+    this.stream.destroy();
+    this.stream.removeAllListeners();
+  }
+
+  private getPtyProcess(): PtyProcess | undefined {
+    return this.ptyProcesses.get(this.ptyId);
+  }
```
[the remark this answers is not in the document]
gzdunek: Removing the map from the constructor would be nice, but as you said it would make cleaning up the resources complicated.

Removed `getPtyProcess`

packages/teleterm/src/sharedProcess/ptyHost/types.ts
```diff
@@ -0,0 +1,18 @@
+export type PtyProcessOptions = {
+  env: Record<string, string>;
+  path: string;
+  args: string[];
+  cwd?: string;
+  initCommand?: string;
+};
+
+export type PtyProcessType = {
```
ravicious, line 9:
```suggestion
export type IPtyProcess = {
```

I see that we actually use this convention for `IAppContext`, perhaps we could use it here? I generally don't think we should have types called `*Type`, unless it's stuff like `*EventType`.

packages/teleterm/src/services/pty/ptyHost/ptyProcess.ts
```diff
@@ -0,0 +1,57 @@
+import { PtyProcessType } from 'teleterm/sharedProcess/ptyHost';
+import { PtyHostClient } from '../types';
+
+export function createPtyProcess(
```
ravicious, line 4: It's a bummer that we're not able to return a class here. Each place which uses the return value of this function has to import the interface type `PtyProcessType`, rather than the class that actually implements the given interface.

I guess we can live with it. But it's so much easier is to inspect all the places where `PtyProcess` from `sharedProcess/ptyHost/ptyProcess.ts` is used vs the return value of this function. ;(

packages/teleterm/README.md
```diff
@@ -119,21 +119,21 @@ $ rm -rf ./../webapps/packages/teleterm/src/services/tshd/v1/ && cp -R lib/telet
   +---------------+   | tls/tcp on localhost     | |
   |    local      |   |                          | |
   | user profile  |   |                          v v
-  |   (files)     |   |                   +------+-+-------------------+
```
ravicious, line 122: What did you use to modify this diagram? I've used Monodraw in the past but idk if you can just import those diagrams like that there.

ravicious (approved):
Great work Grzegorz, I know it wasn't easy nor fun, but it looks so much better overall now.

I left one minor nit.

packages/teleterm/src/services/pty/ptyHost/ptyProcess.ts
gzdunek: True :(

packages/teleterm/README.md
ravicious: FWIW, when I have to draw a non-sequential diagram, I typically use Graphviz and then put it through https://dreampuf.github.io/GraphvizOnline/

Monodraw is nice but I hated that I had to spend so much time moving boxes around when Graphviz can do that for me (though it's not always ideal and requires some additional work).

ravicious:
Great, now that we're on the current version of Electron, we should probably try to follow its releases more closely. I see that there's 18.2.1 already hehe.

ravicious:
Also, hopefully #221 got easier to implement.

PR #821, opened 2022-05-16
Add gRPC generation and logging to shared process

gzdunek:
This PR addresses:
- gRPC files generation
- missing logging 

I added new script `generate-grpc-shared` in `teleterm/package.json`. Unfortunately, we still can't use `grpc-tools` on M1 Macs, so I use `npx` to install a version for x64 arch. The script command is really long, but I didn't find any way to extract a npx script to something like a file :/ I hope we will be able to get rid of it when `grpc-tools` will have a version for M1.
The proto and generated files are structured similarly to what we have in the `teleport` repo.
I removed `ts-protoc-gen` dependency (it wasn't used) and added `google-protobuf` (it was a dependency of `ts-protoc-gen` and is required for the gRPC to work). 

I also made some changes to the logger. From now, the name for the log file must be set explicitly as we have more than two processes. 
The new log files have names: `main.log`, `shared.log`, `renderer.log`.


packages/teleterm/src/sharedProcess/ptyHost/server/ptyEventsStreamHandler.ts
```diff
@@ -9,9 +9,11 @@ import {
   PtyServerEvent,
 } from '../../api/protogen/ptyHostService_pb';
 import { PtyProcess } from '../server/ptyProcess';
+import Logger from 'teleterm/logger';
 
 export class PtyEventsStreamHandler {
   private readonly ptyId: string;
+  private logger = new Logger('PtyEventsStreamHandler');
```
ravicious, line 16: Since we know the ptyId upfront, could we instead do something like 

```typescript
new Logger(`PtyEventsStreamHandler (id: ${this.ptyId})`)
```

in the constructor? This way we won't have to repeat "stream for pty" in each logger invocation.

packages/teleterm/src/services/logger/loggerService.ts
```diff
@@ -20,7 +20,7 @@ export default function createLoggerService(opts: Options): LoggerService {
         maxsize: 4194304, // 4 MB - max size of a single file
```
ravicious, line 20: TIL we actually limit the logs to some size, that's good to know. I was worried that we're going to leave hundreds of megabytes of logs on the drives of Connect users. ;)

packages/teleterm/package.json
```diff
@@ -17,7 +17,8 @@
     "build-main": "webpack build --config webpack.main.config.js --progress --mode=production",
     "build-renderer": "webpack build --config webpack.renderer.prod.config.js --progress",
     "build-natives": "electron-builder install-app-deps",
-    "package": "electron-builder build --publish never -c.extraMetadata.name=teleconnect"
+    "package": "electron-builder build --publish never -c.extraMetadata.name=teleconnect",
+    "generate-grpc-shared": "npx -y --target_arch=x64 --package=grpc_tools_node_protoc_ts --package=grpc-tools -- grpc_tools_node_protoc -I=src/sharedProcess/api/proto --ts_out=service=grpc-node,mode=grpc-js:src/sharedProcess/api/protogen --grpc_out=grpc_js:src/sharedProcess/api/protogen --js_out=import_style=commonjs,binary:src/sharedProcess/api/protogen src/sharedProcess/api/proto/*.proto"
```
ravicious, line 21: > Unfortunately, we still can't use `grpc-tools` on M1 Macs, so I use `npx` to install a version for x64 arch.

I think there might be an even easier way of dealing with this where we can just tell yarn/npm that we always want the x64. I'll check it out later, I vaguely remember something like that existing.

ravicious, line 21: > The script command is really long, but I didn't find any way to extract a npx script to something like a file :/ I hope we will be able to get rid of it when `grpc-tools` will have a version for M1.

You could put it in a bash file, but I think what we got is good enough for now.

ravicious, line 21: Is the name `generate-grpc-shared` used to make it more clear that we don't mean the gRPC files behind daemon <=> Connect communication?

I think we should add information about this command to the readme next to that diagram.

ravicious (approved):
It looks and works pretty nice I have to say.

We don't log each RPC that comes in through, unlike between tsh daemon and Connect, but that should be fine? I think we didn't use to have extensive logging around PTY processes before. Logging each call would obviously add a lot of noise to the logs, it seems that right now we log the important bits. 

We should add info about `generate-grpc-shared` to the readme though.

Leaving an approve because I think it'll be easier for you to merge this into `gzdunek/electron-update` and then do stuff like renaming the classes and whatnot, instead of trying to deal with merge/rebase conflicts later.

packages/teleterm/package.json
ravicious: I tried to search for that, but I think I meant the `target_arch` flag which you're already using here and which cannot be specified in package.json.

We could probably use one of the package-specific node-pre-gyp flags or whatever, but idk how to do this and it's probably not worth the time anyway since this solution here is good enough.

Could we lock the package versions though? I think it should be a matter of appending `@x.y.z` to package names in the `npx` command.

packages/teleterm/src/sharedProcess/ptyHost/server/ptyEventsStreamHandler.ts
gzdunek: Good idea, done

packages/teleterm/package.json
gzdunek: Yes, I think it is not possible to install all packages for the target arch.  
Added exact versions to the `npx` script.

gzdunek: Done

gzdunek:
>Leaving an approve because I think it'll be easier for you to merge this into gzdunek/electron-update and then do stuff like renaming the classes and whatnot, instead of trying to deal with merge/rebase conflicts later.

Ah I didn't see it earlier, but fortunately these conflicts were easy to solve.

ravicious (approved):

PR #822, opened 2022-05-17
Add tooltips with keyboard shortcuts

gzdunek:
Adds tooltips with format `Label (shortcut)` to:
- Connections 
- Cluster selector
- Identity selector
- Close tab icon
- New tab icon

Also, adds inline shortcut to Quick input:
image

Closes: https://github.com/gravitational/webapps.e/issues/141

### DIFF
--- a/packages/teleterm/src/ui/QuickInput/QuickInput.tsx
+++ b/packages/teleterm/src/ui/QuickInput/QuickInput.tsx
@@ -225,8 +225,8 @@ const Shortcut = styled(Box)`
   right: 12px;
   top: 12px;
   padding: 2px 3px;
-  color: rgba(255, 255, 255, 0.6);
-  background-color: rgba(255, 255, 255, 0.05);
+  color: ${({ theme }) => theme.colors.text.secondary};
+  background-color: ${({ theme }) => theme.colors.primary.light};
   line-height: 12px;
   font-size: 12px;
   border-radius: 2px;
--- a/packages/teleterm/src/ui/QuickInput/QuickInput.tsx
+++ b/packages/teleterm/src/ui/QuickInput/QuickInput.tsx
@@ -17,7 +17,7 @@
 import React, { useEffect, useMemo, useRef, useState } from 'react';
 import styled from 'styled-components';
 import { debounce } from 'lodash';
-import { Flex } from 'design';
+import { Box, Flex } from 'design';
 import { color, height, space, width } from 'styled-system';
 import useQuickInput from './useQuickInput';
 import QuickInputList from './QuickInputList';
@@ -160,6 +160,7 @@ function QuickInput() {
         onKeyDown={handleKeyDown}
         isOpened={visible}
       />
+      {!visible && <Shortcut>{props.keyboardShortcut}</Shortcut>}
       {visible && hasSuggestions && (
         <QuickInputList
           ref={refList}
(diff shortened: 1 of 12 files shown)


packages/teleterm/src/ui/QuickInput/QuickInput.tsx
```diff
@@ -219,6 +220,18 @@ const Input = styled.input(props => {
   };
 });
 
+const Shortcut = styled(Box)`
+  position: absolute;
+  right: 12px;
+  top: 12px;
+  padding: 2px 3px;
+  color: rgba(255, 255, 255, 0.6);
+  background-color: rgba(255, 255, 255, 0.05);
```
ravicious, line 229: What about this?

```suggestion
  color: ${({ theme }) => theme.colors.text.secondary};
  background-color: ${({ theme }) => theme.colors.action.hover};

```

packages/teleterm/src/ui/QuickInput/useQuickInput.ts
```diff
@@ -16,16 +16,20 @@ limitations under the License.
...
+import { KeyboardShortcutType } from 'teleterm/services/config';
 
 export default function useQuickInput() {
-  const { quickInputService, workspacesService, commandLauncher } =
-    useAppContext();
+  const {
+    quickInputService,
+    workspacesService,
+    commandLauncher,
+  } = useAppContext();
```
ravicious, line 32: Could you check if you have Prettier configured properly? 😏

In the PTY PR I noticed that my editor formatted some of your changes, so when I saw this I decided to double check.

`yarn prettier -c packages/teleterm/src/ui/QuickInput/useQuickInput.ts` emits warnings, if I save the buffer in my editor the only remaining warning is about `jsxBracketSameLine` being deprecated.

packages/teleterm/src/ui/TopBar/Identity/IdentitySelector/IdentitySelector.tsx
```diff
@@ -17,12 +18,18 @@ export const IdentitySelector = forwardRef<
...
   const Icon = props.isOpened ? SortAsc : SortDesc;
 
   return (
-    <Container isOpened={props.isOpened} ref={ref} onClick={props.onClick}>
+    <Container
+      isOpened={props.isOpened}
+      ref={ref}
+      onClick={props.onClick}
+      title={getLabelWithShortcut('Open Profile', 'toggle-identity')}
+    >
```
ravicious, line 32: Hmm, this makes it hard to discover the actual shortcut, because the tooltip is different when you hover on the avatar or the arrow and the profile name itself.

Could we make it so that there's only one tooltip for this Container? Something along the lines of `Open Profile <profile name> <shortcut>` I suppose?

ravicious (approved):
Added some minor comments, but looks good overall!

packages/teleterm/src/ui/QuickInput/QuickInput.tsx
gzdunek: Done, I used `theme.colors.primary.light` for background, is a bit darker than `action.hover`

packages/teleterm/src/ui/TopBar/Identity/IdentitySelector/IdentitySelector.tsx
gzdunek: You are right here, having two tooltips for the Container makes one of them hard to access. I discovered that I can use line break in the `title` attribute, so the final tooltip will look like this:
```
<profile name>
Open Profiles <shortcut>
```

PR #823, opened 2022-05-17
Create a LabelPicker component

hatched:
Screen Shot 2022-05-17 at 11 39 39 AM

This component is required to complete https://github.com/gravitational/teleport/issues/10342


packages/teleport/src/components/LabelSelector/LabelSelector.tsx
```diff
@@ -0,0 +1,269 @@
...
+              + Create new label "{newLabel}"
+            </CreateLabel>
+          ) : (
+            <CreateLabelError data-testid="create-label-error">
+              <WarningIconWrapper>
+                <Warning />
+              </WarningIconWrapper>
+              <WarningText>
+                <Text style={{ color: '#D83C31', fontWeight: 700 }}>
+                  Invalid label format
```
kimlisa, line 173: this error appears before i even write anything, on first click, shouldn't it happen after entering a wrong label?

kimlisa, line 173: this error appears before i even write anything on first click, shouldn't it show after entering a wrong label?

packages/teleport/src/components/LabelSelector/LabelSelector.tsx
```diff
@@ -0,0 +1,269 @@
...
+            View Documentation
+          </a>
+        </Text>
+      </Heading>
+      <LabelContainer
+        onClick={() => setShowAdd(!showAdd)}
+        data-testid="label-container"
+      >
+        {labels.length === 0 && (
+          <Text style={{ color: 'rgba(255, 255, 255, 0.1)' }}>
```
kimlisa, line 134: `Text` comp allows you to pass in `color` using styled system

packages/teleport/src/components/LabelSelector/LabelSelector.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+  it('prevents invalid labels to be submitted', () => {
+    const onChange = jest.fn();
+    const { container } = render(<LabelSelector onChange={onChange} />);
+    fireEvent.click(screen.getByTestId('label-container'));
+    const labelInput: HTMLInputElement = container.querySelector(
+      'input[name=addLabel]'
+    );
+    fireEvent.change(labelInput, { target: { value: 'foo bar' } });
+    fireEvent.keyPress(labelInput, { key: 'Enter', charCode: 13 });
+    expect(onChange.mock.calls).toEqual([[[]]]);
```
kimlisa, line 61: i think you explained this before but why does this look so weird `[[[]]]`, i'm not sure what it means, can we acheive something similar with `toHaveBeenCalledWith`?

kimlisa, line 61: i think you explained this before but why does this look so weird `[[[]]]`, i'm not sure what it means, can we acheive something similar with `toHaveBeenCalledWith`?

packages/teleport/src/components/LabelSelector/LabelSelector.tsx
```diff
@@ -0,0 +1,269 @@
+import React, { useEffect, useRef, useState } from 'react';
+import styled from 'styled-components';
+
+import Text from 'design/Text';
+import { Info, Warning } from '../../../../design/src/Icon';
+import Pill from '../../../../design/src/Pill';
```
kimlisa, line 6:
```suggestion
import { Info, Warning } from 'design/Icon';
import Pill from 'design/Pill';
```

kimlisa: you can also export Pill and Popover in `design/src/index.js` and you can do combine imports

`import {Text, Box, Pill, Popover} from 'design'`

packages/teleport/src/components/LabelSelector/LabelSelector.tsx
```diff
@@ -0,0 +1,269 @@
...
+
+function labelList({
+  labels,
+  onDismiss,
+}: {
+  labels: string[];
+  onDismiss: (string) => void;
+}) {
+  return labels.map(label => (
+    <Pill key={label} label={label} dismissable onDismiss={onDismiss} />
```
kimlisa, line 265: i'm curious, couldn't we make `onDismiss` optional, and create the `dismissable` flag inside the `Pill` component `const dismissable = !!onDismiss`? b/c it looks a little redundant to apply both dismissable and a onDismiss func

packages/teleport/src/components/LabelSelector/LabelSelector.tsx
```diff
@@ -0,0 +1,269 @@
...
+            Teleport provides users the ability to add labels (in the form of
+            key:value pairs) to resources. Some valid example labels are “env:
+            prod” and “arch: x86_64”. Labels, used in conjunction with roles,
+            define access in Teleport. For example, you can specify that users
+            with the “on-call” role can access resources labeled “env: prod”.
+            For more information, check out our documentation on{' '}
+            <a
+              href="https://goteleport.com/docs/setup/admin/trustedclusters/"
+              target="_blank"
+            >
```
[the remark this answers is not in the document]
kimlisa: if you're curious where it started: https://github.com/gravitational/teleport/issues/5493

packages/design/src/Pill/index.ts
```diff
@@ -0,0 +1,3 @@
+import Pill from "./Pill";
+
+export default Pill
```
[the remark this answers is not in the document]
hatched: Updated all in this PR, looks like the app does a mix of both, maybe we'll have to go through all the packages and update them.

packages/teleport/src/components/LabelSelector/LabelSelector.test.tsx
hatched: When the mock is called it creates a multi-dimensional array with the arguments from each call. What I'm doing here is checking that it was only called once and with no data, just doing so in a single assertion.

packages/teleport/src/components/LabelSelector/LabelSelector.tsx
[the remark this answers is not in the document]
hatched: Updated these to use the `Link` component.

packages/teleport/src/components/LabelSelector/LabelSelector.tsx
hatched: updated

packages/teleport/src/components/LabelSelector/LabelSelector.tsx
hatched: Yep great idea, I'll do that

packages/teleport/src/components/LabelSelector/LabelSelector.test.tsx
kimlisa: hm, i'm not sure i like this, or will remember the multi-dimensional array part when time passes, i think it's better to do something like this:

```
toHaveBeenCalledWith
toHaveBeenCalledTimes
```
it's one more line of code, but it's explicit and easier to read and understand :woman_shrugging: we could get another opinion on this

packages/design/src/Pill/index.ts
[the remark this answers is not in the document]
kimlisa: yeah, default was standard then, but going forward with new files we decided to avoid using it

packages/teleport/src/components/LabelSelector/LabelSelector.tsx
```diff
@@ -0,0 +1,273 @@
+import React, { useEffect, useRef, useState } from 'react';
+import styled from 'styled-components';
+
+import { Box, Flex, Pill, Popover, Link, Text } from 'design';
+import { Info, Warning } from '../../../../design/src/Icon';
```
kimlisa, line 5:
```suggestion
import { Info, Warning } from 'design/Icon';
```

packages/design/src/Pill/__snapshots__/Pill.test.tsx.snap
```diff
@@ -0,0 +1,53 @@
...
+      color="light"
+    />
+  </button>,
+]
+`;
+
+exports[`design/Pill renders the label without dismissable 1`] = `
+HTMLCollection [
+  .c1 {
+  display: inline-block;
```
kimlisa, line 32: hm these snapshots look identical now?

packages/teleport/src/components/LabelSelector/LabelSelector.tsx
```diff
@@ -0,0 +1,273 @@
...
+
+  return (
+    <div>
+      <Heading>
+        <Flex justifyContent="space-between">
+          <Flex>
+            <Text>Assign Labels (optional)</Text>
+            <div
+              ref={infoIconRef}
+              style={{ marginLeft: '12px', float: 'left' }}
```
kimlisa, line 64: float police, could you check to see other float usage in this file, i see others below

packages/teleport/src/components/LabelSelector/LabelSelector.tsx
```diff
@@ -0,0 +1,273 @@
...
+        >
+          <AddLabelInput
+            name="addLabel"
+            value={newLabel}
+            onChange={e => {
+              setNewLabel(e.target.value);
+            }}
+            onKeyPress={e => {
+              // Add a new label on `Enter` if it's valid.
+              if (e.charCode === 13 && validLabel) {
```
kimlisa, line 157: looks like KeyboardEvent.charCode is also deprecated now? https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/charCode

packages/teleport/src/components/LabelSelector/LabelSelector.test.tsx
hatched: No problem, I've updated it to use those methods.

packages/shared/hooks/useEscape.ts
```diff
@@ -0,0 +1,35 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { useCallback, useEffect } from 'react';
+
+function useEscape(handler) {
+  const handlerMemo = useCallback(handler, [handler]);
```
kimlisa, line 20: i'm curious why we `usecallback` here? wouldn't `handler` reference change every time the caller re-renders? 

in LabelSelector it's used like this:
```

  useClickOutside(addLabelContainerRef, () => {
    setShowAdd(false);
  });

  useEscape(() => {
    setShowAdd(false);
  });
```

kimlisa (commented):
missing license headers on some files

packages/teleport/src/components/LabelSelector/LabelSelector.tsx
hatched: Done

packages/design/src/Pill/__snapshots__/Pill.test.tsx.snap
hatched: Good catch, there was a bug there that I missed 👍

packages/teleport/src/components/LabelSelector/LabelSelector.tsx
hatched: 🚨 😁 
All updated

packages/teleport/src/components/LabelSelector/LabelSelector.tsx
hatched: good catch, updated

packages/shared/hooks/useEscape.ts
hatched: Good catch, I removed that handling from here and put the `useCallback` in `LabelSelector`.

packages/shared/hooks/useClickOutside.ts
```diff
@@ -0,0 +1,42 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import {useCallback, useEffect} from "react";
+
+function useClickOutside(ref, handler) {
+
+  const handlerMemo = useCallback(handler, [handler]);
```
kimlisa, line 21: same issue here as `useEscape`?

kimlisa (approved):

packages/shared/hooks/useClickOutside.ts
hatched: Oops sorry I forgot to commit that file :) updated

PR #824, opened 2022-05-18
Remove use of deprecated POST `/webapi/sessions` login endpoint

strideynet:
As per https://github.com/gravitational/teleport/blob/master/lib/web/apiserver.go#L304 , POST `/webapi/sessions` has been migrated to POST `/webapi/sessions/web`.

### DIFF
--- a/packages/teleport/src/config.ts
+++ b/packages/teleport/src/config.ts
@@ -113,6 +113,7 @@ const cfg = {
     renewTokenPath: '/v1/webapi/sessions/renew',
     resetPasswordTokenPath: '/v1/webapi/users/password/token',
     sessionPath: '/v1/webapi/sessions',
+    webSessionPath: '/v1/webapi/sessions/web',
     userContextPath: '/v1/webapi/sites/:clusterId/context',
     userStatusPath: '/v1/webapi/user/status',
     passwordTokenPath: '/v1/webapi/users/password/token/:tokenId?',
--- a/packages/teleport/src/services/auth/auth.test.ts
+++ b/packages/teleport/src/services/auth/auth.test.ts
@@ -33,7 +33,7 @@ describe('services/auth', () => {
     jest.spyOn(api, 'post').mockResolvedValue({});
 
     await auth.login(email, password, '');
-    expect(api.post).toHaveBeenCalledWith(cfg.api.sessionPath, {
+    expect(api.post).toHaveBeenCalledWith(cfg.api.webSessionPath, {
       user: email,
       pass: password,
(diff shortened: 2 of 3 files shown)


packages/teleport/src/config.ts
```diff
@@ -113,6 +113,7 @@ const cfg = {
     renewTokenPath: '/v1/webapi/sessions/renew',
     resetPasswordTokenPath: '/v1/webapi/users/password/token',
     sessionPath: '/v1/webapi/sessions',
```
Joerger, line 115: Is this still used?

strideynet: Yes, it's still used for deleting a session :)

kimlisa: i think we should also change the delete endpoint too, b/c it looks like we changed the endpoint in the first place to be more specific and differentiate between `sessions/app` and `sessions/web`

kimlisa: or maybe we could keep things as is and assume `/sessions` is for web b/c we also have this endpoint to mean `renew web session`: 	

```
h.POST("/webapi/sessions/renew", h.WithAuth(h.renewSession))
```

if we want to keep things consistent, then that endpoint should probably also be changed to `sessions/renew/web`?

strideynet: > i think we should also change the delete endpoint too, b/c it looks like we changed the endpoint in the first place to be more specific and differentiate between `sessions/app` and `sessions/web`

Coming back to this one thing occurred to me.  Calling `DELETE /webapi/sessions` does not only delete your web session, but also any app sessions associated with your user. So an argument could be made that it could made that it's suitable for it to remain at `/webapi/sessions` rather than `/webapi/sessions/web`.

I think perhaps, if consistency of the endpoints is a goal here, it makes sense to merge this PR so that the long since deprecated endpoint is dealt with, and then raise a future PR/issue in order to move the `renew` and `logout` endpoints, since these will need to include a deprecation of their own first.

Either way, it'd be nice to have some more feedback from @Joerger and @kimlisa , so we can progress with this in one direction or another.

kimlisa: let's just leave it alone, and just delete the endpoint in the back that we were supposed to use but never did

Joerger (approved):

zmb3:
@kimlisa I can't tell from the thread above whether more work is necessary here or if everything has been addressed. Are you looking for more changes here?

strideynet:
Sounds good to me, thank you.

PR #825, opened 2022-05-19
Prevent `restorePersistedState()` crash when there is no workspace for a cluster

gzdunek:
Closes https://github.com/gravitational/webapps.e/issues/275

Main reason of the problem:
`restorePersistedState()` starts iterating through the clusters and tries to restore a workspace for each of them, if it starts with the cluster which doesn't have a workspace it will crash on reading `persistedWorkspace.documents` and it won't restore anything.

Additionally, we couldn't spot the problem easily as we didn't show errors from the `init()` function anywhere.

### DIFF
--- a/packages/teleterm/src/ui/services/workspacesService/workspacesService.test.ts
+++ b/packages/teleterm/src/ui/services/workspacesService/workspacesService.test.ts
@@ -62,7 +62,7 @@ describe('restoring workspace', () => {
     return { workspacesService, clusterDocument };
   }
 
-  it('should restore the workspace if it there is a persisted state for given clusterUri', () => {
+  it('restores the workspace if it there is a persisted state for given clusterUri', () => {
     const testClusterUri = '/clusters/test-uri';
     const testWorkspace: Workspace = {
       localClusterUri: testClusterUri,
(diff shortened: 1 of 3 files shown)


ravicious (commented):
I read the PR description and your comment under the issue, I don't have the time to review it today unfortunately, but:

`WorkspacesService`, like any class that manages global, long running resources , is very error-prone. I know that adding tests for it is a huge undertaking, but could we add a very simple test just for this one scenario that this PR is fixing?

This would already help a lot in the future when there's another bug we have to fix.

gzdunek:
Agree, we should have tests for such critical parts ot the app

packages/teleterm/src/ui/services/workspacesService/workspacesService.test.ts
```diff
@@ -0,0 +1,115 @@
...
+      undefined,
+      statePersistenceService
+    );
+
+    workspacesService.getWorkspaceDocumentService = () => ({
+      // @ts-expect-error using mocks
+      createClusterDocument() {
+        return clusterDocument;
+      },
+    });
```
ravicious, line 60: That's fine for now, but I think we might be able to get rid of `@ts-expect-error` by mocking stuff through Jest. I encourage you to give it a try next time, see `packages/teleterm/src/ui/services/quickInput/quickInputService.test.ts` for an example.

ravicious (approved):
Thanks for adding the tests!

It looks like the branch could use some rebasing though to clean up the unrelated commits from other branches.

packages/teleterm/src/ui/services/workspacesService/workspacesService.test.ts
gzdunek: Thanks, I'll do it next time!

gzdunek:
>It looks like the branch could use some rebasing though to clean up the unrelated commits from other branches.

Oops, I didn't notice it, fixed

PR #826, opened 2022-05-20
[v9] Backport "Update Electron & add shared process"

gzdunek:
Backports https://github.com/gravitational/webapps/pull/819


ravicious (approved):
We should backport the one with logging too, right?

gzdunek:
> We should backport the one with logging too, right?

Yes, I wanted to merge that one first

PR #828, opened 2022-05-21
Create agent type for resources

kimlisa:
part of access request work

#### Description
This refactor should also aid in the refactoring of pagination work

- Create a agent file, containing agent types
- Remove design packing importing from teleport package

** easier to review by commit

#### Required to merge
- [x] merge counter part PR https://github.com/gravitational/webapps.e/pull/286
- [x] update e-ref

#### Testing
Checked each resource was able to render list, next and prev: 
 - nodes, db, apps, desktops, kubes, and console nodes

### DIFF
--- a/packages/teleport/src/services/agents/types.ts
+++ b/packages/teleport/src/services/agents/types.ts
@@ -20,9 +20,9 @@ import { Node } from 'teleport/services/nodes';
 import { Kube } from 'teleport/services/kube';
 import { Desktop } from 'teleport/services/desktops';
 
-export type AgentKinds = App | Database | Node | Kube | Desktop;
+export type AgentKind = App | Database | Node | Kube | Desktop;
 
-export type AgentResponse<T extends AgentKinds> = {
+export type AgentResponse<T extends AgentKind> = {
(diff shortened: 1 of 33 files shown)


packages/teleport/src/services/agents/types.ts
```diff
@@ -0,0 +1,51 @@
...
+import { App } from 'teleport/services/apps';
+import { Database } from 'teleport/services/databases';
+import { Node } from 'teleport/services/nodes';
+import { Kube } from 'teleport/services/kube';
+import { Desktop } from 'teleport/services/desktops';
+
+export type AgentKinds = App | Database | Node | Kube | Desktop;
+
+export type AgentResponse = {
+  agents: AgentKinds[];
```
rudream, line 26: Why use `AgentKinds` instead of generics? 

```
export type AgentsResponse<T> = {
    agents: T[];
    ....
}```

kimlisa: good catch, i think this will be better, revised

rudream (approved):

PR #829, opened 2022-05-23
Do not include secrets in gRPC logs

gzdunek:
Closes https://github.com/gravitational/webapps.e/issues/285

Object values which keys includes `passw` will be changed to `~FILTERED~`.

### DIFF
--- a/packages/teleterm/src/services/tshd/middleware.ts
+++ b/packages/teleterm/src/services/tshd/middleware.ts
@@ -1,10 +1,10 @@
-/* eslint-disable @typescript-eslint/ban-types */
 import * as grpc from '@grpc/grpc-js';
 import Logger from 'teleterm/logger';
+import { isObject, transform } from 'lodash';
 
 export type UnaryInterceptor = (
-  options: any,
-  nextCall: any
+  options: grpc.InterceptorOptions,
+  nextCall: (options: grpc.InterceptorOptions) => grpc.InterceptingCall
 ) => grpc.InterceptingCall;
 
 // This is custom grpc middleware implementation that uses JS Proxy to intercept method calls
@@ -65,18 +65,20 @@ export default function middleware<T extends Record<string, any>>(
   });
 }
 
-export const withLogging = (logger: Logger) => {
-  const interceptor: UnaryInterceptor = (options: any, nextCall: Function) => {
+export const withLogging = (logger: Logger): UnaryInterceptor => {
+  return (options, nextCall) => {
     const method = options.method_definition.path;
-    const params = {
-      start(metadata: any, listener: any, next: Function) {
+    const filterPasswords = object =>
+      filterSensitiveProperties(object, ['passw']);
+    const params: grpc.Requester = {
+      start(metadata, listener, next) {
         next(metadata, {
           onReceiveMetadata(metadata, next) {
             next(metadata);
           },
 
           onReceiveMessage(message, next) {
-            const json = message ? message.toObject() : null;
+            const json = message ? filterPasswords(message.toObject()) : null;
             logger.info(`receive: ${method} -> (${JSON.stringify(json)})`);
             next(message);
           },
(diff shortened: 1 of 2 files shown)


packages/teleterm/src/services/tshd/middleware.ts
```diff
@@ -65,18 +65,20 @@ export default function middleware<T extends Record<string, any>>(
...
 
-export const withLogging = (logger: Logger) => {
-  const interceptor: UnaryInterceptor = (options: any, nextCall: Function) => {
+export const withLogging = (logger: Logger): UnaryInterceptor => {
+  return (options, nextCall) => {
     const method = options.method_definition.path;
-    const params = {
-      start(metadata: any, listener: any, next: Function) {
+    const filterPasswords = object =>
+      filterSensitiveProperties(object, ['passw']);
```
ravicious, line 72: How do you feel about defining `filterPasswords` outside of `withLogging` or not parameterizing `filterSensitiveProperties` at all (by defining `filteredProperties` once in the scope of the file) and using `filterSensitiveProperties` instead of `filterPasswords`?

I know that this style is a popular convention inside React components. But in general I think that from the perspective of code maintainability, it's best to define a function in the outermost possible scope.

When I see `filterPasswords` defined within the scope of the function returned from `withLogging`, I assume that it's because `filterPasswords` needs something from that scope, when actually this is not the case.

Maybe it's just that I have wrong heuristics here, but I believe that it's best to clearly communicate the boundaries of functions, what they do need and what they don't need.

packages/teleterm/src/services/tshd/middleware.ts
```diff
@@ -65,18 +65,20 @@ export default function middleware<T extends Record<string, any>>(
   });
 }
 
-export const withLogging = (logger: Logger) => {
-  const interceptor: UnaryInterceptor = (options: any, nextCall: Function) => {
+export const withLogging = (logger: Logger): UnaryInterceptor => {
+  return (options, nextCall) => {
```
ravicious, line 69: 👍, the types are much easier to read this way.

packages/teleterm/src/services/tshd/middleware.ts
```diff
@@ -91,14 +93,41 @@ export const withLogging = (logger: Logger) => {
...
+    toFilter,
+    (result: object, value: any, key: any) => {
+      if (
+        filteredProperties.some(
+          filteredProp => typeof key === 'string' && key.includes(filteredProp)
+        )
+      ) {
+        result[key] = '~FILTERED~';
+      } else {
+        if (isObject(value)) {
```
ravicious, line 124: Small nit, but I think this could made into `else if` instead which would save one nesting level, thus also making the "default"/"else" case easier to notice.

packages/teleterm/src/services/tshd/middleware.ts
```diff
@@ -91,14 +93,41 @@ export const withLogging = (logger: Logger) => {
...
   };
-
-  return interceptor;
 };
+
+function filterSensitiveProperties(
+  toFilter: object,
+  filteredProperties: string[]
+): object {
+  return transform(
```
ravicious, line 114: We don't send that many messages so it doesn't matter that much, but if the throughput was higher, it'd probably be better to mutate an existing object instead of creating a new one each time.

ravicious (approved):

packages/teleterm/src/services/tshd/middleware.ts
gzdunek: Ah, good intuition, we don't need that parameter.
I defined `filteredProperties` inside the function (I like that style more than using vars in the file scope), I guess it is fine?

packages/teleterm/src/services/tshd/middleware.ts
gzdunek: Done by using early returns, IMO it is even easier to read than `else if`.

packages/teleterm/src/services/tshd/middleware.ts
gzdunek: Yes, I took that into account, but I agree that for now it doesn't matter that much

packages/teleterm/src/services/tshd/middleware.ts
ravicious: Why not define it in the file scope though? 👀 This way you clearly communicate that it won't change during the lifetime of the program. Also, this avoids creating a new array each time the function is called (although idk, maybe modern JS engines are able to optimize that away).

I think it's perfectly okay to define vars in the file scope, it's almost like a constant since there's never a need to change it during the lifetime of the program. You could even uppercase it, from what I've seen we use this convention in our codebase:

```
$ rg "^const [A-Z]{3,}" --pretty --color=never | head
packages/teleterm/src/mainProcess/runtimeSettings.ts
10:const RESOURCES_PATH = app.isPackaged

packages/build/devserver/index.js
39:const PROXY_TARGET = urlObj.host;
40:const ROOT = '/web';
41:const PORT = 8080;

packages/teleterm/src/ui/DocumentTerminal/Terminal/ctrl.ts
25:const WINDOW_RESIZE_DEBOUNCE_DELAY = 200;

ravicious (approved):

PR #830, opened 2022-05-24
Upgrade Electron to 19.0.0

ravicious:
I checked PTY & db access as well as restoring window state, anything other we should do a quick check on during upgrades like that?


gzdunek (approved):
I think that is enough for a quick check, it is unlikely that any other parts are affected

PR #837, opened 2022-05-25
Remove connections when logging out

gzdunek:
Closes https://github.com/gravitational/webapps.e/issues/270

### DIFF
--- a/packages/teleterm/src/ui/services/connectionTracker/connectionTrackerService.ts
+++ b/packages/teleterm/src/ui/services/connectionTracker/connectionTrackerService.ts
@@ -32,6 +32,7 @@ import {
   TrackedGatewayConnection,
 } from './types';
 import { getClusterName } from 'teleterm/ui/utils';
+import { routing } from 'teleterm/ui/uri';
 
 export class ConnectionTrackerService extends ImmutableStore<ConnectionTrackerState> {
   private _trackedConnectionOperationsFactory: TrackedConnectionOperationsFactory;
@@ -64,10 +65,11 @@ export class ConnectionTrackerService extends ImmutableStore<ConnectionTrackerSt
     return this.state.connections.map(connection => {
       const { rootClusterUri, leafClusterUri } =
         this._trackedConnectionOperationsFactory.create(connection);
-      const cluster = this._clusterService.findCluster(
-        leafClusterUri || rootClusterUri
-      );
-      return { ...connection, clusterName: getClusterName(cluster) };
+      const clusterUri = leafClusterUri || rootClusterUri;
+      const clusterName =
+        getClusterName(this._clusterService.findCluster(clusterUri)) ||
+        routing.parseClusterName(clusterUri);
+      return { ...connection, clusterName };
     });
   }
 
@@ -80,6 +82,15 @@ export class ConnectionTrackerService extends ImmutableStore<ConnectionTrackerSt
       await this._workspacesService.setActiveWorkspace(rootClusterUri);
     }
     activate();
+  }
+
+  setState(
+    nextState: (
+      draftState: ConnectionTrackerState
+    ) => ConnectionTrackerState | void
+  ): void {
+    super.setState(nextState);
+    this._statePersistenceService.saveConnectionTrackerState(this.state);
   }
 
   async disconnectItem(id: string): Promise<void> {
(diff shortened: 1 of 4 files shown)


packages/teleterm/src/ui/services/connectionTracker/connectionTrackerService.ts
```diff
@@ -64,10 +65,11 @@ export class ConnectionTrackerService extends ImmutableStore<ConnectionTrackerSt
...
         this._trackedConnectionOperationsFactory.create(connection);
-      const cluster = this._clusterService.findCluster(
-        leafClusterUri || rootClusterUri
-      );
-      return { ...connection, clusterName: getClusterName(cluster) };
+      const clusterUri = leafClusterUri || rootClusterUri;
+      const clusterName =
+        getClusterName(this._clusterService.findCluster(clusterUri)) ||
+        routing.parseClusterName(clusterUri);
+      return { ...connection, clusterName };
```
ravicious, line 72: This whole thing can be removed once we implement gravitational/webapps.e#261 and the cluster name will be known upfront, right?

Then we won't need to do this whole map but instead set the cluster name once when the connection is created and then just return `this.state.connections`. The cluster name never changes.

packages/teleterm/src/ui/services/connectionTracker/connectionTrackerService.ts
```diff
@@ -97,18 +108,23 @@ export class ConnectionTrackerService extends ImmutableStore<ConnectionTrackerSt
     this._statePersistenceService.saveConnectionTrackerState(this.state);
   }
 
+  removeItemsBelongingToRootCluster(clusterUri: string): void {
+    this.setState(draft => {
+      draft.connections = draft.connections.filter(i => {
+        const { rootClusterUri } =
+          this._trackedConnectionOperationsFactory.create(i);
```
ravicious, line 115: It feels like in the future each connection could have the `clusterUri` field instead so that we don't have to do the whole dance with `this._trackedConnectionOperationsFactory.create`.

packages/teleterm/src/ui/services/workspacesService/workspacesService.ts
```diff
@@ -210,7 +210,18 @@ export class WorkspacesService extends ImmutableStore<WorkspacesState> {
...
+                    d.kind === 'doc.terminal_tsh_node' ||
+                    d.kind === 'doc.gateway'
+                  ) {
+                    return {
+                      ...d,
+                      status: 'disconnected',
+                    };
+                  }
+                  return d;
+                }),
```
ravicious, line 224: What's the purpose behind this change?

No matter the purpose though, adding code like this is tricky because then each item we add a new document that should act kinda like `doc.terminal_tsh_node` or `doc.gateway`, we'll have checks like this one scattered across the codebase.

It'd be better to add a function to a file related to documents, which at the moment we don't have I believe.

Still, a lone function that can be imported defined in `documentsService.ts` would be better then this, or maybe we could even have `documentsService/document.ts` with functions that operate on `Document`.

ravicious (commented):
The changes to the connection tracker look fine, but I have a question about changes in the workspaces service.

packages/teleterm/src/ui/services/connectionTracker/connectionTrackerService.ts
gzdunek: Generally yes, however I wanted to avoid saving that information if we can calculate it basing on `clusterUri`.

> The cluster name never changes.

It reminded mi about [that list](https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/).
> People’s names do not change. 

😄

packages/teleterm/src/ui/services/connectionTracker/connectionTrackerService.ts
gzdunek: We already have some `resourceUri` in each connection (that contains `clusterUri`), so I wanted to avoid duplicating that information, but I can agree, it makes working with the connections a bit more difficult.

packages/teleterm/src/ui/services/workspacesService/workspacesService.ts
gzdunek: I decided to remove it from the PR as it doesn't really solve the issue. I wanted to show `Reconnect` screen when the cluster is offline, but we don't keep the information about that state. 
Basically, `useReconnect` will make this check:
```
useEffect(() => {
    if (cluster.connected) {
      markDocumentAsConnected();
    }
  }, []);
```
and change the document state to `connected`.

Let's leave it as it is for now.

ravicious (approved):

packages/teleterm/src/ui/services/connectionTracker/connectionTrackerService.ts
ravicious: But I'm fine with leaving it like this for now, let's just make sure to create such function when another opportunity to use it comes up.

PR #840, opened 2022-05-26
Do not show the login dialog when a user adds a cluster that is connected 

gzdunek:
Related issue: https://github.com/gravitational/webapps.e/issues/242

Thanks to this PR user won't see login dialog if it is already logged to the cluster.

Teleport counterpart: https://github.com/gravitational/teleport/pull/12940

### DIFF
--- a/packages/teleterm/src/ui/ClusterConnect/ClusterConnect.tsx
+++ b/packages/teleterm/src/ui/ClusterConnect/ClusterConnect.tsx
@@ -2,12 +2,23 @@ import React, { useState } from 'react';
 import { ClusterAdd } from './ClusterAdd';
 import { ClusterLogin } from './ClusterLogin';
 import Dialog from 'design/Dialog';
+import { useAppContext } from 'teleterm/ui/appContextProvider';
 
 export function ClusterConnect(props: ClusterConnectProps) {
   const [createdClusterUri, setCreatedClusterUri] = useState<
     string | undefined
   >();
+  const { clustersService } = useAppContext();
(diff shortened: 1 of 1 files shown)


packages/teleterm/src/ui/ClusterConnect/ClusterConnect.tsx
```diff
@@ -2,13 +2,24 @@ import React, { useState } from 'react';
...
+  const { clustersService } = useAppContext();
   const clusterUri = props.clusterUri || createdClusterUri;
 
+  function handleClusterAdd(clusterUri: string): void {
+    const cluster = clustersService.findCluster(clusterUri);
+    if (cluster?.connected) {
+      props.onSuccess(clusterUri);
+    } else {
+      setCreatedClusterUri(clusterUri);
+    }
```
ravicious, line 20: At first I was worried that it's going to interfere with my changes to handling expired certs, but I think we should be fine. I call `openClusterConnectDialog` with the cluster URI already provided, so `ClusterConnect` will render `ClusterLogin` instead.

There might be a situation where someone calls this after the certs have expired and we weren't updated about it, but I think it's unlikely. Besides, if we plan to periodically sync root clusters somehow as I mentioned on Slack today, that should make it even less likely. Plus once that implemented, your code here will elegantly handle a situation in which someone tried to add a cluster which we know that it has expired certs.

ravicious (approved):

ravicious:
Let's backport this to v9 too please, along with gravitational/teleport#12940. I already added a label to that PR, so creating a backport should be straightforward.

Issue #836, opened 2022-05-24
Allow "Activity" tab to be hidden from web UI when you can't use anything under it


webvictim:
Similar to https://github.com/gravitational/webapps/pull/728, we've had a request to allow the "Activity" section of the web UI to be hidden when users don't have access to do anything under it. We actually mention in that issue that the Activity section can't be completely hidden at present.

At the moment, you can deny visibility to "Active Sessions" using a role with a `deny` rule like this:

```yaml
kind: role
metadata:
  name: block-active-sessions
spec:
  deny:
    rules:
    - resources:
      - ssh_session
      verbs:
      - list
      - read
      - create
      - update
      - delete
version: v5
```

...but it generates an error in the UI, and the "Active Sessions" tab is still visible:
![image](https://user-images.githubusercontent.com/897821/170099977-aa2db774-ba94-4a65-b357-e0c9c3444367.png)

It'd be good to make a change to the UI code to hide "Active Sessions" when the user is explicitly denied permission to do anything with `ssh_session`.

Ideally, we'd also be able to hide "Access Requests" with a similar deny rule and ultimately hide the entire "Activity" section too when you can't view any of its submenus (that was the actual customer ask here) but I think for now just hiding "Active Sessions" is good.

PR #844, opened 2022-05-26
Allow "Activity" tab to be hidden

ibeckermayer:
Adds to the `ACL` to hide active sessions when they're not available to the user to list.

closes https://github.com/gravitational/webapps/issues/836

corresponds with https://github.com/gravitational/teleport/pull/12962


kimlisa (approved):

rudream (approved):
lgtm

PR #846, opened 2022-05-27
Retry with relogin on errors related to expired certs

ravicious:
This PR implements the first TODO item from gravitational/webapps.e#282.

This is a draft of the full PR. It adds the retry mechanism only to creating database gateway. I'm writing tests right now and then I have to add this to two more places (`ClusterContext.sync` & `getDatabaseUsers`). Still, I want to get early feedback in case I missed some scenarios.

## How to test it

The easiest way to test this is to edit one of the roles in the cluster and set `spec.options.max_session_ttl` to something low. I usually set it to `30s` and then revert it back to `30h`, but the actual session will have a TTL of at least 1 minute anyway.

Another way to test it is to make this change in the tsh daemon:

<details>
<summary>tsh daemon diff</summary>

```diff
diff --git a/lib/teleterm/apiserver/handler/handler.go b/lib/teleterm/apiserver/handler/handler.go
index 722008bf8..6d0081d0d 100644
--- a/lib/teleterm/apiserver/handler/handler.go
+++ b/lib/teleterm/apiserver/handler/handler.go
@@ -21,8 +21,11 @@ import (
 	"github.com/gravitational/trace"
 )
 
+var counter int
+
 // New creates an instance of Handler
 func New(cfg Config) (*Handler, error) {
+	counter = 0
 	if err := cfg.CheckAndSetDefaults(); err != nil {
 		return nil, trace.Wrap(err)
 	}
diff --git a/lib/teleterm/apiserver/handler/handler_gateways.go b/lib/teleterm/apiserver/handler/handler_gateways.go
index b9b9820cf..bdf5ba29d 100644
--- a/lib/teleterm/apiserver/handler/handler_gateways.go
+++ b/lib/teleterm/apiserver/handler/handler_gateways.go
@@ -26,6 +26,11 @@ import (
 
 // CreateGateway creates a gateway
 func (s *Handler) CreateGateway(ctx context.Context, req *api.CreateGatewayRequest) (*api.Gateway, error) {
+	counter++
+	if counter%2 != 0 {
+		return nil, trace.NotImplemented("ssh: handshake failed")
+	}
+
 	params := clusters.CreateGatewayParams{
 		TargetURI:             req.TargetUri,
 		TargetUser:            req.TargetUser,
```
</details>


packages/teleterm/src/ui/utils/retryWithRelogin.ts
```diff
@@ -0,0 +1,71 @@
...
+): Promise<T> {
+  let retryableErrorFromActionToRepeat: Error;
+  try {
+    return await actionToRepeat();
+  } catch (error) {
+    // TODO(ravicious): Replace this with actual check on metadata.
+    const isRetryable =
+      error instanceof Error &&
+      (error.message.includes('ssh: handshake failed') ||
+        error.message.includes('ssh: cert has expired'));
```
ravicious, line 19: I'll replace this with a check on metadata in another series of PRs. For now this is enough.

packages/teleterm/src/ui/DocumentGateway/useDocumentGateway.ts
```diff
@@ -32,12 +33,14 @@ export default function useGateway(doc: types.DocumentGateway) {
   const cluster = ctx.clustersService.findClusterByResource(doc.targetUri);
 
   const [connectAttempt, createGateway] = useAsync(async () => {
-    const gw = await ctx.clustersService.createGateway({
-      targetUri: doc.targetUri,
-      port: doc.port,
-      user: doc.targetUser,
-      subresource_name: doc.targetSubresourceName,
-    });
+    const gw = await retryWithRelogin(ctx, doc.uri, doc.targetUri, () =>
```
ravicious, line 36: We might be dealing with a race condition here, but now that I wrote it down I think we're fine.

Once you click "Connect" next to a db and pick the db user and db name, this code is executed:

https://github.com/gravitational/webapps/blob/627dea4ad701be3c5fae36e06ee73da4837d855c/packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/useDatabases.ts#L41-L42

This adds the document, which in turn fires this `useAsync` hook and we fire a request to the tsh daemon. If we get an error related to expired certs, there's a check in `retryWithRelogin` to see if the user is still seeing the relevant UI element.

So we could think of a situation where `documentsService.add` is fired and before `documentsService.open` is executed, we already got a cert expired error from the gateway, but since we haven't open the doc yet, the check in `retryWithRelogin` will think that the user is not looking at the relevant doc.

However, as far as I understand JavaScript's runtime, both `add` and `open` are not async functions. This means that any async code executed inside them will happen on the next tick, after the whole `useDatabases.connect` function finishes its execution.

So I think we should be fine here, even in theory? Because in practice it's obviously hard for the daemon to respond before we open the doc.

packages/teleterm/src/ui/utils/retryWithRelogin.ts
```diff
@@ -0,0 +1,71 @@
+import {} from 'teleterm/ui/services/workspacesService/documentsService';
+import { routing } from 'teleterm/ui/uri';
+import AppContext from 'teleterm/ui/appContext';
+
+export async function retryWithRelogin<T>(
+  appContext: AppContext,
+  originatingDocumentUri: string,
+  resourceUri: string,
```
ravicious, line 8: I ended up not needing `document.getResourceUri`, because I just pass this as an argument, so each place calling `retryWithRelogin` will be responsible for fetching its own resource URI.

Actually, `getResourceUri` would be hard to implement for `DocumentGateway`. What is the `resourceUri` for `DocumentGateway`? I'd say that it's `gatewayUri`, but this field is set only after the actual gateway is created. The gateway itself is kind of an ephemeral resource, it's deleted after you close the gateway or the app.

So in the case of `useDocumentGateway`, I actually pass `doc.targetUri` here.

I need `resourceUri` only to extract `rootClusterUri` from it. As I was writing this, I realized I could just accept `rootClusterUri` instead of `resourceUri`. However, it'd increase the complexity of using `retryWithRelogin` at the callsite, requiring each place to use something like `routing.ensureRootClusterUri`.

Fetching a relevant resource URI is much easier, usually we'll just grab it from a field in a doc.

packages/teleterm/src/ui/utils/retryWithRelogin.ts
```diff
@@ -0,0 +1,71 @@
...
+    // The error is retryable, but the user is no longer looking at the relevant UI, for example
+    // they switched to a different document or a different workspace completely.
+    //
+    // Since it might take a few seconds before the execution gets to this point, the user might no
+    // longer remember what their intent was, thus showing them a login modal can be disorienting.
+    //
+    // In that situation, let's just not attempt to relogin and instead let's surface the original
+    // error.
+    throw retryableErrorFromActionToRepeat;
+  }
```
ravicious, line 42: In case we ever need to use `retryWithRelogin`, this part can be extracted to a separate function that `retryWithRelogin` is going to accept as an arg.

The purpose of the function would be the same as this code here – to check if the user is still seeing the relevant UI.

packages/teleterm/src/ui/utils/retryWithRelogin.ts
```diff
@@ -0,0 +1,71 @@
...
+    await login(appContext, rootClusterUri);
+  } catch (error) {
+    // In case the user canceled the login attempt, retry the original call anyway in case the cert
+    // was refreshed externally, for example through tsh.
+    const isRetryable = error instanceof LoginAttemptCanceledError;
+
+    if (!isRetryable) {
+      throw error;
+    }
+  }
```
ravicious, line 56: While writing tests I realized that the custom error class is completely unnecessary. The login attempt either succeeds or the user closes it, so the promise can be just resolved/rejected.

We don't see any errors related to the process of logging in, because those are contained within the login modal.

I fully control when and why the promise is rejected.

packages/teleterm/src/ui/DocumentGateway/useDocumentGateway.ts
gzdunek: Correct. The JS engine will not execute the asynchronous tasks before the synchronous ones finish.

gzdunek:
Generally I like the solution, it's easy to follow.

I am wondering if we can make the function easier to use, because currently we have to pass 4 params to it. What comes to my mind is converting it to a hook, like `useAsyncWithRelogin` ;) 
Thanks to this we could get rid of `appContext`, `originatingDocumentUri` and `resourceUri`. The hook could take these params from the context. 
`originatingDocumentUri` -> `ctx.workspacesService.getWorkspaceDocumentsService().getActive()`
`resourceUri` -> `ctx.workspacesService.getRootClusterUri()` (as you said only root cluster uri is needed).

Under the hood the hook could use `useAsync` and return its values.

OTOH, that change would make it more complex and harder to test.

packages/teleterm/src/ui/utils/retryWithRelogin.ts
```diff
@@ -0,0 +1,71 @@
...
+        error.message.includes('ssh: cert has expired'));
+
+    if (isRetryable) {
+      retryableErrorFromActionToRepeat = error;
+    } else {
+      throw error;
+    }
+  }
+
+  const isDocumentStillActive = appContext.workspacesService.isDocumentActive(
```
gzdunek, line 28: Good you thought about it!

ravicious:
~I'm not opposed to adding a hook, but I think we could just let it stay like this for now (it's only three places) and as we develop the app we might add a hook and keep a regular function too. Testing the hook would be actually trivial then because all the logic tests would be contained in the regular function.~

> The hook could take these params from the context. 
`originatingDocumentUri` -> `ctx.workspacesService.getWorkspaceDocumentsService().getActive()`
`resourceUri` -> `ctx.workspacesService.getRootClusterUri()` (as you said only root cluster uri is needed).

`originatingDocumentUri` doesn't map to the active document but rather the document which started the request. Same thing with `resourceUri`. The way I think about it is that I need URIs related to this specific document I'm calling this function from, while those two functions from `workspacesService` deal with some kind of a global state, returning "active" versions of those URIs.

While writing this I realized that this kinda invalidates the first paragraph then, because you'd still need to pass those arguments to the hook then. I thought about creating a special context which stores doc URI and root cluster URI, but to be honest:

1) It increases complexity just to avoid passing arguments.
2) I don't think that a function accepting four arguments is a problem in itself. If there was a huge ceremony related to obtaining them, making the callsites very noisy and polluted, then I can see how abstracting this away could help. In the case of this function, the first three arguments are easily obtainable from the context of the callsite.

packages/teleterm/src/ui/utils/retryWithRelogin.ts
ravicious: Well, technically `appContext.modalsService.openClusterConnectDialog` could throw an error and thus reject the promise with something else than `LoginAttemptCanceledError`. But we're going to assume that such behavior is a bug in the program rather than something we need to handle.

ravicious: To avoid using exceptions for control flow, we could just resolve the promise to `true` when the user logs in and `false` if they cancel the attempt. Then if the promise is rejected, that means there was an unexpected error somewhere else.

gzdunek:
> `originatingDocumentUri` doesn't map to the active document but rather the document which started the request.

I was thinking about taking a "snapshot" of the current active document URI when hook starts (the same for `resourceUri`), but it would work correctly only for the simplest case (opening only one tab at a time)  :/ 

Let's discard this idea as it is very error prone.

packages/teleterm/src/ui/utils/retryWithRelogin.ts
ravicious: Matter of fact, I don't actually care in this case if they canceled the attempt or not, because I'm going to repeat the call anyway. So I'll just resolve to void in both cases and any unexpected errors will bubble up thanks to `await`.

gzdunek: Returning `true`/`false` sounds good to me.

Why do we want to repeat do call when user cancels the attempt?

ravicious: https://github.com/gravitational/webapps/blob/6ef6affc0d23065ac23a2f22f29733fdf11dc347/packages/teleterm/src/ui/utils/retryWithRelogin.ts#L46-L50

I actually removed that comment by mistake, let me bring it back. ;)

gzdunek:
> How is it usually done, like with a useState or something?

Both `useState/useRef` would work I guess.

>But yeah, caveats when operating on a global state like that usually push me into just explicitly passing arguments around.

👍

ravicious (commented):
Probably the most artisanal & gourmet tests I've written at Teleport so far.

packages/teleterm/src/ui/utils/retryWithRelogin.test.ts
```diff
@@ -0,0 +1,131 @@
...
+    'doc-uri',
+    '/clusters/foo/servers/bar',
+    actionToRetry
+  );
+
+  const openClusterConnectDialogSpy =
+    appContext.modalsService.openClusterConnectDialog;
+  expect(openClusterConnectDialogSpy).toHaveBeenCalledTimes(1);
+  expect(openClusterConnectDialogSpy).toHaveBeenCalledWith(
+    expect.objectContaining({ clusterUri: '/clusters/foo' })
```
gzdunek, line 66: Nice, I didn't know about it

gzdunek:
The tests are really cool!

ravicious:
@gzdunek I just pushed commits that add retry logic when using the Sync button, for both root clusters and leaf clusters. This is a bigger change than I expected, this is because I had to restructure a bit how `ClustersService` throws errors. I tried to explain that in the commit message.

ravicious:
I'll try to add `retryWithRelogic` to fetching db users and I think after that the PR will be finished. Then I'll start working on another one which uses gRPC metadata and another one that restarts gateways on relogin.

Perhaps I should start working on the latter first, as I still haven't decided if this should be done on the tshd's side or just in Connect.

packages/teleterm/src/ui/utils/retryWithRelogin.ts
```diff
@@ -1,6 +1,8 @@
-import {} from 'teleterm/ui/services/workspacesService/documentsService';
 import { routing } from 'teleterm/ui/uri';
 import AppContext from 'teleterm/ui/appContext';
+import Logger from 'teleterm/ui/logger';
```
gzdunek, line 3: Please use `Logger` from `teleport/logger` as I'm going to remove the one you used https://github.com/gravitational/webapps/pull/853/commits/fa37504c33bd338aee213fe2b1403ec6156569dd

packages/teleterm/src/ui/appContext.ts
```diff
@@ -25,8 +25,9 @@ import { KeyboardShortcutsService } from 'teleterm/ui/services/keyboardShortcuts
 import { CommandLauncher } from './commandLauncher';
 import { WorkspacesService } from 'teleterm/ui/services/workspacesService/workspacesService';
 import { NotificationsService } from 'teleterm/ui/services/notifications';
+import { IAppContext } from './types';
 
-export default class AppContext {
+export default class AppContext implements IAppContext {
```
gzdunek, line 30: 👏

packages/teleterm/src/ui/utils/retryWithRelogin.ts
```diff
@@ -2,6 +2,21 @@ import {} from 'teleterm/ui/services/workspacesService/documentsService';
...
+//
+// Each place using retryWithRelogin must be able to show the error to the user in case the relogin
+// attempt fails. Each place should also offer the user a way to manually retry the action which
+// results in a call to the tshd client.
+//
+// retryWithRelogin should wrap calls to the tshd client as tightly as possible. At the moment, it
+// means actionToRetry will usually involve calls to ClustersService, which so far is the only place
+// that has access to the tshd client.
+//
+// resourceUri is needed only in order to extract the root cluster URI from it.
```
gzdunek, line 21: I'd prefer that comment to be in JSDoc format, but it's up to you.

ravicious: Ah, I keep forgetting about this. The main advantage is that editor integrations can display the comment when hovering above the function, right?

gzdunek: The editor will show the panel regardless of the type of comment, [but JSDoc is more readable](https://github.com/gravitational/webapps/pull/754#discussion_r855142548).

packages/teleterm/src/ui/utils/retryWithRelogin.ts
ravicious: You mean I should use `teleterm/logger` I assume?

gzdunek: Ah, yes

packages/teleterm/src/ui/utils/retryWithRelogin.ts
```diff
@@ -4,21 +4,24 @@ import Logger from 'teleterm/ui/logger';
...
+ * attempt fails. Each place should also offer the user a way to manually retry the action which
+ * results in a call to the tshd client.
+ *
+ * retryWithRelogin should wrap calls to the tshd client as tightly as possible. At the moment, it
+ * means actionToRetry will usually involve calls to ClustersService, which so far is the only place
+ * that has access to the tshd client.
+ *
+ * @param {string} resourceUri - The URI used to extract the root cluster URI. That's how we
+ * determine the cluster the login modal should use.
+ */
```
gzdunek, line 24:
```suggestion
/**
 * `retryWithRelogin` executes `actionToRetry`. If `actionToRetry` throws an error, it checks if the error
 * can be resolved by the user logging in, according to metadata returned from the tshd client.
 *
 * If that's the case, it checks if the user is still looking at the relevant UI and if so, it shows
 * a login modal. After the user successfully logs in, it calls `actionToRetry` again.
 *
 * Each place using `retryWithRelogin` must be able to show the error to the user in case the relogin
 * attempt fails. Each place should also offer the user a way to manually retry the action which
 * results in a call to the tshd client.
 *
 * `retryWithRelogin` should wrap calls to the tshd client as tightly as possible. At the moment, it
 * means `actionToRetry` will usually involve calls to `ClustersService`, which so far is the only place
 * that has access to the tshd client.
 *
 * @param resourceUri - The URI used to extract the root cluster URI. That's how we
 * determine the cluster the login modal should use.
 */
```

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/Databases.tsx
```diff
@@ -24,8 +24,8 @@ import { Danger } from 'design/Alert';
 import { MenuLogin, MenuLoginHandle } from 'shared/components/MenuLogin';
 import { MenuLoginTheme } from '../MenuLoginTheme';
 import { useAppContext } from 'teleterm/ui/appContextProvider';
-import { ClustersService } from 'teleterm/ui/services/clusters';
-import { NotificationsService } from 'teleterm/ui/services/notifications';
+import { retryWithRelogin } from 'teleterm/ui/utils/retryWithRelogin';
```
gzdunek, line 27: Let's export that in `utils/index.ts`

gzdunek (approved):

gzdunek (approved):

PR #859, opened 2022-06-01
Fix logout icon

gzdunek:
Closes: https://github.com/gravitational/webapps.e/issues/289
image


packages/design/src/Icon/index.js
```diff
@@ -14,333 +14,7 @@ See the License for the specific language governing permissions and
...
-  Visa,
-  VolumeUp,
-  VpnKey,
-  Wand,
-  Warning,
-  Wifi,
-  Windows,
-  Youtube,
-};
+export * from './Icon';
```
ravicious, line 20: Do you know why we used to export each icon individually? If not, could you ask one more person from the frontend team to review this change specifically?

The change makes sense to me, but the process of exporting those icons one by one seems elaborate enough that I suspect there must've been a reason behind it. Unfortunately, I don't see any documentation about it.

gzdunek: @kimlisa maybe you know the reason for this? For me both versions work identically

kimlisa: i don't know the history of why it was like this. if it doesn't break anything currently, i think it's okay?

kimlisa (approved):

PR #866, opened 2022-06-03
Refactor Serverside Filtering

rudream:
## Purpose

Refactor serverside filtering and remove duplicate code.

#### Test initial sorting, sorting with filters, and SSO and local login redirects for:
- [ ] Nodes
- [ ] Nodes (Console view)
- [ ] Applications
- [ ] Databases
- [ ] Kubernetes
- [ ] Desktops


packages/teleport/src/Kubes/Kubes.tsx
```diff
@@ -65,7 +65,9 @@ export function Kubes(props: State) {
   const [showAddKube, setShowAddKube] = useState(false);
 
   const hasNoKubes =
-    attempt.status === 'success' && results.kubes.length === 0 && isSearchEmpty;
+    attempt.status === 'success' &&
+    results.agents.length === 0 &&
```
hatched, line 69: See previous comment about `?.`

packages/teleport/src/Desktops/Desktops.tsx
```diff
@@ -63,7 +63,7 @@ export function Desktops(props: State) {
 
   const hasNoDesktops =
     attempt.status === 'success' &&
-    results.desktops.length === 0 &&
+    results.agents.length === 0 &&
```
hatched, line 66: See comment about `?.`

packages/teleport/src/Nodes/Nodes.tsx
```diff
@@ -76,7 +76,9 @@ export function Nodes(props: State) {
   }
 
   const hasNoNodes =
-    attempt.status === 'success' && results.nodes.length === 0 && isSearchEmpty;
+    attempt.status === 'success' &&
+    results.agents.length === 0 &&
```
hatched, line 80: See above comment about `?.`

packages/teleport/src/components/hooks/useServersidePagination.ts
```diff
@@ -0,0 +1,99 @@
...
+        setResults({
+          ...results,
+          agents: res.agents,
+          startKey: res.startKey,
+        });
+        setFetchStatus(res.startKey ? '' : 'disabled');
+        setStartKeys([...startKeys, res.startKey]);
+      })
+      .catch((err: Error) => {
+        setAttempt({ status: 'failed', statusText: err.message });
```
hatched, line 57: In the event of an error the `fetchStatus` should probably be set to the default value.

packages/teleport/src/generateResourcePath.ts
```diff
@@ -14,6 +14,17 @@ See the License for the specific language governing permissions and
...
+/* The generatePath function built into react router doesn't use encodeURIComponent. This causes the resulting encoded URL to be
+encoded improperly for this use-case, and the query fails. This custom generateResourcePath function resolves that issue while also
+being more specialized to this use-case and supporting a SortType param.
+
+Example:
+
+Output from generatePath: /v1/webapi/sites/im-a-cluster-name/apps?limit=&startKey=&query=labels.app%20==%20%22banana%22&search=&sort=
+Output from generateResourcePath: /v1/webapi/sites/im-a-cluster-name/apps?limit=5&startKey=&query=labels.app%20%3D%3D%20%22banana%22&search=&sort=name:asc
+
+*/
```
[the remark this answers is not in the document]
kimlisa: couple more things to refactor for this function:
- `params` needs to be typed (not set to accept any params), the URL expects explicit params so we need to be explicit
- the `for loop` may have been nice for many arbitrary params, but we only have 3 explicit params that need to be processed. I would remove the for loop and simplify:

```ts
    const sort = params.sort ? <construct sort param> : ''
    .replace(':query?', params.query? encodeURIComponent(params.query) || '')
    .replace(':search?', params.search? encodeURIComponent(params.search) || '')
    .replace(':sort?', sort');
```

packages/teleport/src/components/hooks/useUrlFiltering/labelClick.ts
```diff
@@ -1,5 +1,21 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 import { AgentLabel } from 'teleport/services/agents';
-import { ResourceUrlQueryParams } from './getUrlQueryParams';
+import { ResourceUrlQueryParams } from './getQueryParams';
 import encodeUrlQueryParams from './encodeUrlQueryParams';
 
 export default function labelClick(
```
kimlisa, line 21: it seems `labelClick.ts` is only used in `useUrlFiltering.ts`, i would move it there and delete this file

packages/teleport/src/components/hooks/useUrlFiltering/useUrlFiltering.ts
```diff
@@ -0,0 +1,61 @@
...
+  function replaceHistory(path: string) {
+    history.replace(path);
+  }
+
+  function setSort(sort: SortType) {
+    setParams({ ...params, sort });
+  }
+
+  const onLabelClick = (label: AgentLabel) =>
+    labelClick(label, params, setParams, pathname, replaceHistory);
```
kimlisa, line 41: we can simplify this, for `labelClick` the main logic is to build the final query string, have `labelClick` return that query string, and we don't need to pass in `setParams, pathname, replaceHistory`

```ts
function onLabelClick(label: AgentLabel) {
   const query = addAgentLabelToQuery(filter: AgentFilter, label: AgentLabel) 

   setParams({ ...params, search: '', query });
   replaceHistory(encodeUrlQueryParams(pathname, query, params.sort, true));
}
```

packages/teleport/src/components/hooks/useUrlFiltering/useUrlFiltering.ts
```diff
@@ -0,0 +1,61 @@
...
+import { AgentLabel } from 'teleport/services/agents';
+import getQueryParams from './getQueryParams';
+import labelClick from './labelClick';
+
+export default function useUrlFiltering(initialSort: SortType) {
+  const { search, pathname } = useLocation();
+  const [params, setParams] = useState<ResourceUrlQueryParams>({
+    sort: initialSort,
+    ...getQueryParams(search),
+  });
```
kimlisa, line 30: i defined a type called `AgentFilter`, we should probably use that instead

packages/teleport/src/components/hooks/useUrlFiltering/getQueryParams.ts
```diff
@@ -24,11 +24,13 @@ export default function getResourceUrlQueryParams(
   const search = searchParams.get('search');
   const sort = searchParams.get('sort');
 
+  const sortParam = sort ? sort.split(':') : null;
+
```
kimlisa, line 28: i think if we move `labelClick` to the file where it is used, this file can also be moved to where it is only used and delete this file

i would always try to define the func closest to where it is used, and only separate it if it needs to be used else where (we have a utils file `teleport/lib/util`)

kimlisa (commented):
first pass and couple of other things:

1) I noticed that when a user clicks `next/prev` button, we disable this button while loading, but we don't disable the search bar. A user could, while loading, enter a search which puts out another fetch request. This could be a race condition? I would just disable the whole search panel easily with `pointer-events: none; opacity: 0.5;`

2) could we revisit the Table `serverSideProps`:
```
       serversideProps={{
          sort: params.sort,
          setSort,
          startKeys,
          serversideSearchPanel: (
            <ServersideSearchPanel
              from={from}
              to={to}
              count={totalCount}
              params={params}
              setParams={setParams}
              pathname={pathname}
              replaceHistory={replaceHistory}
            />
          ),
        }}
```
with the few modifications I committed, i think this could be simpler:

1) we don't need startKeys (as explained in one of the comments)
2) sort related can be pulled out and provided as `customSort`
3) use the `disableFilter` flag so the table doesn't try to apply its default filters 
4) all that's left is to provide a `SearchPanel` component. 

I think it's okay for a little code duplication and split our `Pager` into two types to make it easier to understand: `ClientPager` and `ServerPager` (or something like that), what do you think?

Also, do we need to pass in `pathname and replaceHistory`? Can't we just pass in `updateParams` and then the provider of the func can update the url?

packages/teleport/src/Apps/Apps.tsx
```diff
@@ -62,7 +62,9 @@ export function Apps(props: State) {
   } = props;
 
   const hasNoApps =
-    attempt.status === 'success' && results.apps.length === 0 && isSearchEmpty;
+    attempt.status === 'success' &&
+    results.agents.length === 0 &&
```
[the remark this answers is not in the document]
kimlisa: hm, i don't think it will be, we set a zero value, and our services always (or should be) returning null arrays as empty arrays

PR #867, opened 2022-06-03
Search based request related changes and Table addons

kimlisa:
#### Description
- Adds table addons:
  - Allow `customSort` field to use our own set of sort rules
  - Allow `disableFilter` which is a flag to indicate
    the caller of Table will handle their own filtering (eg. server side)
  - Add non render flag to allow adding a key to be searcheable but not renderable
- Adds `searchAsRoles` query param
- Adds `Access Request` to top level navigation for enterprise only

#### Note
There will be a enterprise PR that will add features to the top level navigation. If this PR gets merged without the enterprise, you'd likely see a Access Request in the nav but nothing in it.

### DIFF
--- a/packages/teleport/src/services/user/makeUser.ts
+++ b/packages/teleport/src/services/user/makeUser.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright 2020 Gravitational, Inc.
+ * Copyright 2020-2022 Gravitational, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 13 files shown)


packages/design/src/DataTable/types.ts
```diff
@@ -17,13 +17,25 @@ export type TableProps<T> = {
   showFirst?: (data: T[]) => T;
   className?: string;
   style?: React.CSSProperties;
+  // customSort contains fields that describe the current sort direction,
+  // the field to sort by, and a custom sort function.
+  customSort?: CustomSort;
+  // disableFilter when true means to skip running
+  // any client table filtering supplied by default.
+  // Use case: filtering is done on the caller side e.g. server side.
+  disableFilter?: boolean;
```
rudream, line 26: Is this flag necessary? We already have an `isSearchable` prop on the Table and an `isSortable` for each column

kimlisa: it's similar to how the existence of `serverSideProps` was working, i just needed a simple flag to do the same thing

rudream (approved):

PR #868, opened 2022-06-03
Add automation to update webassets on push

zmb3:
This will run using a service account which is the only account
permitted to push to webassets and webassets.e.

The action will run when any commits are pushed to the protected
branches in webapps. This means that commits to webapps.e do not
directly trigger updates - this automation will run when the
enterprise submodule ref is updated in webapps.

The "nop" yarn script allows us to separate the building of the docker
image from the build of Teleport, so that GitHub actions can run the
yarn build after buildling an image from our dockerfile. This is
necessary because container steps do not support build args.

### DIFF
--- a/.github/actions/docker-build-webassets/action.yaml
+++ b/.github/actions/docker-build-webassets/action.yaml
@@ -1,0 +1,11 @@
+name: docker-build-webassets
+author: Zac Bergquist <zac@goteleport.com>
+description: Build Teleport webassets in Docker
+
+runs:
+  using: docker
+  image: ../../../Dockerfile
+  entrypoint: sh
+  args:
+    - -c
+    - yarn install && yarn run build-teleport
--- a/.github/workflows/update-webassets.yaml
+++ b/.github/workflows/update-webassets.yaml
@@ -1,0 +1,55 @@
+name: Update Webassets
+
+on:
+  push:
+    branches:
+      - master
+      - teleport-v**
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      # Get webapps, and the webapps.e submodule
+      - name: Checkout webapps
+        uses: actions/checkout@v3
+        with:
+          ssh-key: ${{ secrets.SSH_PRIVATE_KEY }}
+          submodules: recursive
+          fetch-depth: 0
+
+      # Get webassets, and the webassets.e submodule
+      - name: Checkout webassets
+        uses: actions/checkout@v3
+        with:
+          ssh-key: ${{ secrets.SSH_PRIVATE_KEY }}
+          repository: gravitational/webassets
+          submodules: recursive
+          path: dist
+
+      # This workflow only runs on push to a protected branch, so the code
+      # has already been reviewed and is trusted. This means it is safe to
+      # the version of the action that we checked out and we don't need to
+      # always use master.
+      - name: Build webassets
+        uses: ./.github/actions/docker-build-webassets
+
+      - name: Push updates
+        run: |
+          git config --global user.email github-goteleport-core-svc@goteleport.com
+          git config --global user.name "$GITHUB_ACTOR"
+
+          export WEBAPPS_BRANCH=$(git branch --show-current)
+
+          cd dist/e
+          git switch $WEBAPPS_BRANCH
+          git add -A
+          git commit -m "Update webassets.e from gravitational/webapps@$GITHUB_SHA"
+          git push
+
+          cd ..
+          git switch $WEBAPPS_BRANCH
+          git add -A
+          git commit -m "Update webassets from gravitational/webapps@$GITHUB_SHA"
+          git push
--- a/.github/workflows/update-webassets.yaml
+++ b/.github/workflows/update-webassets.yaml
@@ -24,7 +24,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           ssh-key: ${{ secrets.SSH_PRIVATE_KEY }}
-          repository: gravitational/webassets
+          repository: gravitational/webassets-fork # TODO(zmb3): switch off the fork
           submodules: recursive
           path: dist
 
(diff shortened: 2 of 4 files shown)


zmb3:
Opened this up for review, but 🚨do not merge🚨 yet. I'm going to fork the repo and run some tests in a fork before we enable this.

Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM node:16.3-slim
-RUN apt-get update && apt-get install git g++ make python -y
+RUN apt-get update && apt-get install git g++ make python tree -y
```
ryanclark, line 2: What's `tree` needed for?

.github/workflows/update-webassets.yaml
```diff
@@ -0,0 +1,55 @@
...
+          ssh-key: ${{ secrets.SSH_PRIVATE_KEY }}
+          repository: gravitational/webassets
+          submodules: recursive
+          path: dist
+
+      # This workflow only runs on push to a protected branch, so the code
+      # has already been reviewed and is trusted. This means it is safe to
+      # the version of the action that we checked out and we don't need to
+      # always use master.
+      - name: Build webassets
```
ryanclark, line 35: Could we use https://github.com/docker/build-push-action instead? It supports build args, and allow us to use GHA caching for speedier CI runs

ryanclark (approved):
LGTM!

(I deleted my comment about the Docker build action as I had mistook the Docker step for a Docker image build rather than a Docker container action)

.github/actions/docker-build-webassets/action.yaml
```diff
@@ -0,0 +1,11 @@
...
+author: Zac Bergquist <zac@goteleport.com>
+description: Build Teleport webassets in Docker
+
+runs:
+  using: docker
+  image: ../../../Dockerfile
+  entrypoint: sh
+  args:
+    - -c
+    - yarn install && yarn run build-teleport
```
ryanclark, line 11: Does this need `yarn install`? The Dockerfile runs it already

Dockerfile
zmb3: I was testing this and wanted to be able to dump the directory structure to the logs to ensure the build got put in the right place. Not needed anymore, so I can remove, but it also doesn't hurt anything.

.github/actions/docker-build-webassets/action.yaml
zmb3: Maybe not? I was just concerned about some sort of caching if GHA were to reuse a previously-built image that had old dependencies in it.

package.json
```diff
@@ -23,7 +23,8 @@
     "eslint": "eslint --ext .js,.jsx,.ts,.tsx packages/",
     "type-check": "tsc --noEmit",
     "prettier-check": "yarn prettier --check 'packages/**/*.{ts,tsx,js,jsx,md}'",
-    "prettier-write": "yarn prettier --write 'packages/**/*.{ts,tsx,js,jsx,md}'"
+    "prettier-write": "yarn prettier --write 'packages/**/*.{ts,tsx,js,jsx,md}'",
+    "nop": "exit 0"
```
hatched, line 27: Can you add a comment to this - someone coming by this in the future will be very confused :)

.github/actions/docker-build-webassets/action.yaml
hatched: If it did cache then the install would mostly be a noop and it should pick up any changes....should being the key word there :) 

If you want to be sure install is run again then you might want something like...

```suggestion
    - yarn install --force && yarn run build-teleport
```

ryanclark: I don't think it's an issue to keep in, as it'll just be a no-op and take less than a couple of seconds.

But I'm not sure on the caching mechanism for Docker container actions in GHA - I took a look at the docs and couldn't see a mention of it

.github/workflows/update-webassets.yaml
```diff
@@ -0,0 +1,55 @@
...
+        uses: actions/checkout@v3
+        with:
+          ssh-key: ${{ secrets.SSH_PRIVATE_KEY }}
+          repository: gravitational/webassets
+          submodules: recursive
+          path: dist
+
+      # This workflow only runs on push to a protected branch, so the code
+      # has already been reviewed and is trusted. This means it is safe to
+      # the version of the action that we checked out and we don't need to
```
hatched, line 33: I think you missed a word or two here?

hatched (approved):

.github/actions/docker-build-webassets/action.yaml
zmb3: I also didn't see a mention in the docs.

Sounds like we're all on board with keeping this as-is?

package.json
zmb3: Sure.

It is in the commit message, so a `git blame` would tell the story at least.

.github/workflows/update-webassets.yaml
zmb3: I don't think so?

ravicious:
The `nop` workaround is interesting, I guess in the end it might be the simplest approach.

In the past, I've dealt with a similar problem in two ways:

1. Simply [do not pass `NPM_SCRIPT` at all](https://github.com/gravitational/webapps/blob/589e9f643bf439c5595e5fe63f184c47bbeab27b/.cloudbuild/ci/build.yaml#L20-L24). Docker will execute `yarn` which is the equivalent of `yarn install` which is already done in the Dockerfile so it'll be a slower `nop`.
2. We could avoid executing the last RUN altogether by using [different stages](https://stackoverflow.com/a/60820156/742872) or [inspecting the value of the arg](https://stackoverflow.com/a/43656644/742872). I used the first approach [in `Dockerfile-teleterm`](https://github.com/gravitational/teleport/blob/24fcadc375d8359e80790b3ebeaa36bd8dd2822f/build.assets/Dockerfile-teleterm) – check how the `GRPC_NODE_PLUGIN_BINARY_TYPE` arg is used.

As I said, `nop` might be the simplest approach here, I just wanted to list out some alternatives.

PR #870, opened 2022-06-07
Add `TextArea` and `FieldTextArea` components

gzdunek:
`FieldTextArea` will be used in `ShareFeedback` dialog in Connect.
It is based on the `Input` component.


packages/design/src/TextArea/TextArea.story.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export default {
+  title: 'Design/TextArea',
+};
+
+export const TextAreas = () => (
+  <>
+    <TextArea mb={4} placeholder="Enter Some long text" />
+    <TextArea mb={4} hasError={true} defaultValue="This field has an error" />
+    <TextArea mb={4} autoFocus defaultValue="This field is focused" />
```
ravicious, line 28: It seems that the `autoFocus` is not passed to the actual `<textarea>` tag, at least in the storybook the `≤textarea>` doesn't seem to have `autofocus` attribute.

gzdunek: It's weird, in a regular app it works, so probably it is something related to `Storybook`.

I will remove these examples (also in `Input.story`) as they are not working.

packages/shared/components/FieldTextArea/FieldTextArea.tsx
```diff
@@ -0,0 +1,82 @@
...
+        name={name}
+        css={textAreaCss}
+        hasError={hasError}
+        placeholder={placeholder}
+        value={value}
+        autoComplete={autoComplete}
+        autoFocus={autoFocus}
+        onChange={onChange}
+        readOnly={readOnly}
+      />
```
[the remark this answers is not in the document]
gzdunek: Done (I needed it also for tests of `ShareFeedback` - to find element by label).

I adjusted margins between elements so there shouldn't be any difference in their look.

Uh, 41 snapshots updated.

PR #873, opened 2022-06-07
Show node specific ssh logins options

mcbattirola:
PR in draft while I do some more testings and check the boxes below

TODO:

- [ ] remove `sshLogins` from user context if it isn't necessary anymore
___

This PR uses a new field `sshLogins` added in [this teleport PR](https://github.com/gravitational/teleport/pull/13266) instead of using the `sshLogins` from the user context.

The `sshLogins` from the context that are being used currently include all logins. This can be confusing for users, since the UI will show some users when they click `Connect` that they can't use for that specific node.

This PR changes this so the logins showed to the user is specific to the node the user is connecting.

Closes https://github.com/gravitational/teleport/issues/6822

### DIFF
--- a/packages/teleport/src/Nodes/useNodes.ts
+++ b/packages/teleport/src/Nodes/useNodes.ts
@@ -22,7 +22,7 @@ import history from 'teleport/services/history';
 import Ctx from 'teleport/teleportContext';
 import { StickyCluster } from 'teleport/types';
 import cfg from 'teleport/config';
-import { NodesResponse } from 'teleport/services/nodes';
+import { Node, NodesResponse } from 'teleport/services/nodes';
 import { openNewTab } from 'teleport/lib/util';
 import getResourceUrlQueryParams, {
   ResourceUrlQueryParams,
@@ -37,7 +37,6 @@ export default function useNodes(ctx: Ctx, stickyCluster: StickyCluster) {
   const { attempt, setAttempt } = useAttempt('processing');
   const [isAddNodeVisible, setIsAddNodeVisible] = useState(false);
   const canCreate = ctx.storeUser.getTokenAccess().create;
-  const logins = ctx.storeUser.getSshLogins();
   const [fetchStatus, setFetchStatus] = useState<FetchStatus>('');
   const [params, setParams] = useState<ResourceUrlQueryParams>({
     sort: { fieldName: 'hostname', dir: 'ASC' },
@@ -67,7 +66,8 @@ export default function useNodes(ctx: Ctx, stickyCluster: StickyCluster) {
   }
 
   function getNodeLoginOptions(serverId: string) {
-    return makeOptions(clusterId, serverId, logins);
+    const node = results.nodes.find(node => node.id == serverId);
+    return makeOptions(clusterId, node);
   }
 
   const startSshSession = (login: string, serverId: string) => {
@@ -187,15 +187,14 @@ export default function useNodes(ctx: Ctx, stickyCluster: StickyCluster) {
   };
 }
 
-function makeOptions(
-  clusterId: string,
-  serverId = '',
-  logins = [] as string[]
-) {
-  return logins.map(login => {
+function makeOptions(clusterId: string, node: Node | undefined) {
+  const nodeLogins = node?.sshLogins || [];
+  nodeLogins.unshift('root');
+
+  return nodeLogins.map(login => {
     const url = cfg.getSshConnectRoute({
       clusterId,
-      serverId,
+      serverId: node?.id || '',
       login,
     });
 
--- a/packages/teleport/src/Console/DocumentNodes/DocumentNodes.story.tsx
+++ b/packages/teleport/src/Console/DocumentNodes/DocumentNodes.story.tsx
@@ -97,6 +97,7 @@ const clusters = [
 const nodes: Node[] = [
   {
     tunnel: false,
+    sshLogins: ['dev', 'root'],
     id: '104',
     clusterId: 'cluseter-1',
     hostname: 'fujedu',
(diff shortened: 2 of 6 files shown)


packages/teleport/src/Console/DocumentNodes/useNodes.ts
```diff
@@ -19,7 +19,7 @@ import { useLocation } from 'react-router';
 import { FetchStatus, SortType } from 'design/DataTable/types';
 import useAttempt from 'shared/hooks/useAttemptNext';
 import history from 'teleport/services/history';
-import { NodesResponse } from 'teleport/services/nodes';
+import { Node, NodesResponse } from 'teleport/services/nodes';
```
hatched, line 22: I figure that we could  start using `import type` to be more explicit.

```suggestion
import type { Node, NodesResponse } from 'teleport/services/nodes';
```

packages/teleport/src/Console/DocumentNodes/useNodes.ts
```diff
@@ -156,10 +152,26 @@ export default function useNodes({ clusterId, id }: stores.DocumentNodes) {
...
-      login,
-      url: consoleCtx.getSshDocumentUrl({ serverId, login, clusterId }),
-    }));
+    const node = results.nodes.find(node => node.id == serverId);
+    return makeOptions(clusterId, node);
+  }
+
+  function makeOptions(clusterId: string, node: Node | undefined) {
+    const nodeLogins = node?.sshLogins || [];
+    nodeLogins.unshift('root');
```
hatched, line 161: I'm not sure that we can always assume that there will always be a `root` login?

mcbattirola: You're right, thanks for pointing that out. I missed that `root` is added to the top of the list if it is present in the logins in the first place. 

Fixed in https://github.com/gravitational/webapps/pull/873/commits/076772774718c896a6aa3704c6132885cfadaf6a

packages/teleport/src/Console/DocumentNodes/useNodes.ts
```diff
@@ -186,3 +198,14 @@ export default function useNodes({ clusterId, id }: stores.DocumentNodes) {
...
+// sort logins by making 'root' as the first in the list
+const sortLogins = (logins: string[]) => {
+  const noRoot = logins.sort().filter(l => l !== 'root');
+  if (noRoot.length === logins.length) {
+    return logins;
+  }
+
+  noRoot.unshift('root');
+  return noRoot;
+};
```
kimlisa, line 211:
```suggestion
const sortLogins = (logins: string[]) => {
  const noRoot = logins.filter(l => l !== 'root').sort();
  if (noRoot.length === logins.length) {
    return logins;
  }

  return ['root', ...noRoot];
};
```

packages/teleport/src/Nodes/useNodes.ts
```diff
@@ -210,4 +209,15 @@ function makeOptions(
   });
 }
 
+// sort logins by making 'root' as the first in the list
+const sortLogins = (logins: string[]) => {
```
kimlisa, line 213: i thought i was going crazy b/c i swear i just reviewed this :laughing:, we should probably put this and `makeOptions` in a shared utils file, maybe export them from `useNodes`? since they both share them :woman_shrugging:

kimlisa (commented):
pinging @ravicious or @gzdunek, i saw `getSshLogins` used in `teleterm` as well (but they aren't shared) just letting you know of this change

packages/teleport/src/Console/DocumentNodes/useNodes.ts
mcbattirola: Thanks, looks much better. I used this on [packages/teleport/src/Nodes/useNodes.ts](https://github.com/gravitational/webapps/pull/873/files/c211f70a376b4f31adb555d0e3647e2f73ca9919#diff-7d4b0ce921483e33f08ef1df96185b0e604089b55a02e7eb3b2455d24a392eb4) and exported it to use on this file, addressing both this and the other comment (https://github.com/gravitational/webapps/pull/873/commits/6f67c0adec0efa1ca36acfccdac56e3a338c42b4).

Thanks!

mcbattirola:
I had to commit an update to webapps.e (https://github.com/gravitational/webapps/pull/873/commits/ecd5f51e9d16d11e61135ba26f71d5baacf762e9) so the build works correctly.

I will change this to point to webapps.e's master once https://github.com/gravitational/webapps.e/pull/310 is merged.

packages/teleport/src/Nodes/useNodes.ts
```diff
@@ -191,15 +191,14 @@ export default function useNodes(ctx: Ctx, stickyCluster: StickyCluster) {
...
-) {
+function makeOptions(clusterId: string, node: Node | undefined) {
+  const nodeLogins = node?.sshLogins || [];
+  const logins = sortLogins(nodeLogins);
+
   return logins.map(login => {
     const url = cfg.getSshConnectRoute({
       clusterId,
-      serverId,
+      serverId: node?.id || '',
```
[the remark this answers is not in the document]
mcbattirola: I don't think so, but I added it since the node ID was being checked before.

I think we can remove it since this code only runs once the user clicked on Connect, and they can't do it unless there is a node to click in the first place.

ravicious:
> pinging @ravicious or @gzdunek, i saw `getSshLogins` used in `teleterm` as well (but they aren't shared) just letting you know of this change

Thanks, Connect so far just reads the logins from the cert (as the Web UI used to do before this PR).

Connect doesn't even support pagination at the moment, so we'll have to implement that first before we add fetching logins for each server.

PR #874, opened 2022-06-08
Build Connect in Webapps-Build pipeline 

ravicious:
This PR adds Connect to the Webapps-Build pipeline that's executed for each commit to the repo.

* Building the Web UI and Connect within Webapps-Build is done in parallel.
* Lint and type check are ran just once before both artifacts are built.
* GCB runs on Linux, so it'll unfortunately create a Linux build of Connect which we don't really support at the moment. It's better than nothing though.
* Creating a full-fledged macOS build that's signed and packaged just as the real releases will be is already being done on each commit to master in the teleport repo (gravitational/teleport#12257).

I also removed the `build` script from the main package.json, because it seems that after my changes (and one-line change to Makefile) it's not being used anywhere.

---

Notes:

* `make update-webassets` in the teleport repo runs `make build-teleport` in the webapps repo, which doesn't depend on the `build` npm script.

TODO:

* [x] Check if `yarn build` is used anywhere else. This PR changes what it does (it no longer runs lint and type check).
   * `make build` in webapps uses the `build` script.
   * If it's not used anywhere, remove it.
* [x] Instead of modifying `build`, use `build-teleport` in the pipeline instead.


gzdunek (approved):
LGTM

PR #875, opened 2022-06-08
Add `access_request.search` event to audit log

kimlisa:
image

### DIFF
--- a/packages/teleport/src/services/audit/makeEvent.ts
+++ b/packages/teleport/src/services/audit/makeEvent.ts
@@ -43,7 +43,7 @@ export const formatters: Formatters = {
   },
   [eventCodes.ACCESS_REQUEST_RESOURCE_SEARCH]: {
     type: 'access_request.search',
-    desc: 'Access Request Resource Searched',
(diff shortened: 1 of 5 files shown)


packages/teleport/src/services/audit/makeEvent.ts
```diff
@@ -41,6 +41,12 @@ export const formatters: Formatters = {
     desc: 'Access Request Deleted',
     format: ({ id }) => `Access request [${id}] has been deleted`,
   },
+  [eventCodes.ACCESS_REQUEST_RESOURCE_SEARCH]: {
+    type: 'access_request.search',
+    desc: 'Access Request Resource Searched',
```
nklaassen, line 46:
```suggestion
    desc: 'Resource Access Request Search',
```
Hard to come up with a good wording for this but here's my suggestion. I think we are going to be calling these "Resource Access Requests" so this might be a bit more consistent

nklaassen (approved):
Approved with a wording suggestion, not sure if mine is much better but it's what I've got 🤷‍♂️

PR #876, opened 2022-06-08
Various small fixes and touch ups

kimlisa:
#### Description
- v10 teleterm: handles older auth cfg (v9) returning `u2f` as second factor. Instead of allowing users to continue with this deprecated option, we push users to change cluster settings to `webauthn`
- check if passwordless was enabled before rendering the option to add a passwordless device
- fix the brief rendering of buttons before final attempt state (success or failed), which may not include the buttons
- adds a soft warning to firefox users that passwordless may not be supported. I decided on a soft warning because firefox may provide support in the very near future. Currently in firefox:
   - upon register, if users try to add a passwordless device, they are in `soft lock`, firefox prompts but does nothing
   - upon login, it will just return an ambiguous error

firefox support watch:
https://bugzilla.mozilla.org/show_bug.cgi?id=1530370
https://bugzilla.mozilla.org/show_bug.cgi?id=1752089

### DIFF
--- a/packages/teleport/src/Welcome/NewCredentials/NewPasswordlessDevice.tsx
+++ b/packages/teleport/src/Welcome/NewCredentials/NewPasswordlessDevice.tsx
@@ -16,7 +16,7 @@
 
 import React, { useState } from 'react';
 import { Text, Box, ButtonPrimary, ButtonText } from 'design';
-import { Danger } from 'design/Alert';
+import { Danger, Info } from 'design/Alert';
 import FieldInput from 'shared/components/FieldInput';
 import Validation, { Validator } from 'shared/components/Validation';
 import { requiredField } from 'shared/components/Validation/rules';
@@ -55,6 +55,8 @@ export function NewPasswordlessDevice(props: Props & SliderProps<LoginFlow>) {
     switchFlow('local', applyNextAnimation);
   }
 
+  const isFirefox = navigator?.userAgent?.toLowerCase().includes('firefox');
+
   return (
     <Validation>
       {({ validator }) => (
@@ -64,6 +66,12 @@ export function NewPasswordlessDevice(props: Props & SliderProps<LoginFlow>) {
           </Text>
           {submitAttempt.status === 'failed' && (
             <Danger children={submitAttempt.statusText} />
+          )}
+          {isFirefox && (
+            <Info mt={3}>
+              Firefox may not support passwordless register. Please try Chrome
+              or Safari.
+            </Info>
           )}
           <FieldInput
             rule={requiredField('Device name is required')}
--- a/packages/teleport/src/components/FormLogin/FormLogin.tsx
+++ b/packages/teleport/src/components/FormLogin/FormLogin.tsx
@@ -121,31 +121,40 @@ const SsoList = ({ attempt, authProviders, onLoginWithSso }: Props) => {
   );
 };
 
-const Passwordless = ({ onLoginWithWebauthn, attempt }: Props) => (
-  <Box px={5} pt={2} data-testid="passwordless" pb={1}>
-    <StyledPaswordlessBtn
-      mt={3}
-      py={2}
-      px={3}
-      width="100%"
-      onClick={() => onLoginWithWebauthn()}
-      disabled={attempt.isProcessing}
-    >
-      <Flex alignItems="center" justifyContent="space-between">
-        <Flex alignItems="center">
-          <Key mr={3} fontSize={16} />
-          <Box>
-            <Text typography="h6">Passwordless</Text>
-            <Text fontSize={1} color="text.secondary">
-              Follow the prompt from your browser
-            </Text>
-          </Box>
+const Passwordless = ({ onLoginWithWebauthn, attempt }: Props) => {
+  const isFirefox = navigator?.userAgent?.toLowerCase().includes('firefox');
+  return (
+    <Box px={5} pt={2} data-testid="passwordless" pb={1}>
+      {isFirefox && (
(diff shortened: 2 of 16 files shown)


packages/teleterm/src/ui/ClusterConnect/ClusterLogin/useClusterLogin.ts
```diff
@@ -75,7 +75,15 @@ export default function useClusterLogin(props: Props) {
...
-    init();
+    init().then(resp => {
+      const [cfg, err] = resp;
+      if (!err && cfg.preferredMfa === 'u2f') {
+        const statusText = `the U2F API for hardware keys is deprecated, \
+        please notify your system administrator to update cluster \
+        settings to use WebAuthn as the second factor protocol.`;
+        setInitAttempt({ status: 'error', statusText });
+      }
+    });
```
ravicious, line 86: Tested this myself on a cluster in our Cloud. u2f handling has to be moved a bit deeper in the call stack, otherwise `init` finishes with success and `ClusterLogin` tries to render `LoginForm` which throws an error because it tries to read `.value` from undefined somewhere.

I made a quick patch to see if changing it to this would work and it does.

<details>
<summary>Patch to useClusterLogin.ts</summary>

```diff
diff --git a/packages/teleterm/src/ui/ClusterConnect/ClusterLogin/useClusterLogin.ts b/packages/teleterm/src/ui/ClusterConnect/ClusterLogin/useClusterLogin.ts
index 03c989be..3ec35641 100644
--- a/packages/teleterm/src/ui/ClusterConnect/ClusterLogin/useClusterLogin.ts
+++ b/packages/teleterm/src/ui/ClusterConnect/ClusterLogin/useClusterLogin.ts
@@ -28,8 +28,16 @@ export default function useClusterLogin(props: Props) {
   const [shouldPromptSsoStatus, promptSsoStatus] = useState(false);
   const [shouldPromptHardwareKey, promptHardwareKey] = useState(false);
 
-  const [initAttempt, init, setInitAttempt] = useAsync(() => {
-    return clustersService.getAuthSettings(clusterUri);
+  const [initAttempt, init] = useAsync(async () => {
+    const authSettings = await clustersService.getAuthSettings(clusterUri);
+
+    if (authSettings.preferredMfa === 'u2f') {
+      throw new Error(`the U2F API for hardware keys is deprecated, \
+        please notify your system administrator to update cluster \
+        settings to use WebAuthn as the second factor protocol.`);
+    }
+
+    return authSettings;
   });
 
   const [loginAttempt, login] = useAsync((opts: types.LoginParams) => {
@@ -75,15 +83,7 @@ export default function useClusterLogin(props: Props) {
   };
 
   useEffect(() => {
-    init().then(resp => {
-      const [cfg, err] = resp;
-      if (!err && cfg.preferredMfa === 'u2f') {
-        const statusText = `the U2F API for hardware keys is deprecated, \
-        please notify your system administrator to update cluster \
-        settings to use WebAuthn as the second factor protocol.`;
-        setInitAttempt({ status: 'error', statusText });
-      }
-    });
+    init();
   }, []);
 
   useEffect(() => {
```
</details>

packages/teleport/src/Welcome/NewCredentials/NewPasswordlessDevice.tsx
```diff
@@ -55,6 +55,8 @@ export function NewPasswordlessDevice(props: Props & SliderProps<LoginFlow>) {
     switchFlow('local', applyNextAnimation);
   }
 
+  const isFirefox = navigator?.userAgent?.toLowerCase().includes('firefox');
```
ravicious, line 58: TIL how the safe navigation operator actually works in JS. I've always thought that it's similar to `&.` in Ruby, so I thought the code would fail if `navigator.userAgent` resolves to undefined for some reason, this code is going to attempt to call `toLowerCase` on undefined, but it'll actually short-circuit just before that.

Though does `navigator?.` make sense actually? In case it's not defined in the browser for some reason, isn't the browser going to throw a ReferenceError anyway?

Perhaps we could do something like this:

```suggestion
  const isFirefox = window.navigator?.userAgent
    ?.toLowerCase()
    .includes('firefox');
```

packages/teleport/src/components/FormLogin/FormLogin.tsx
```diff
@@ -121,31 +121,40 @@ const SsoList = ({ attempt, authProviders, onLoginWithSso }: Props) => {
...
-        <Flex alignItems="center">
-          <Key mr={3} fontSize={16} />
-          <Box>
-            <Text typography="h6">Passwordless</Text>
-            <Text fontSize={1} color="text.secondary">
-              Follow the prompt from your browser
-            </Text>
-          </Box>
+const Passwordless = ({ onLoginWithWebauthn, attempt }: Props) => {
+  const isFirefox = navigator?.userAgent?.toLowerCase().includes('firefox');
```
ravicious, line 125: Same as above.

ravicious (approved):
The code for Connect needs to be changed a bit, but once you deal with that I think the rest looks good.

packages/teleport/src/Welcome/NewCredentials/NewPasswordlessDevice.tsx
hatched: If you'd like to look up more about it, it's called the "optional chaining operator"

kimlisa:
> Looks good but could you add the firefox check rationale into the code? 

done

> Maybe it could even be abstracted out so there is only one place where we store that logic.

i decided not to do this, b/c i'm expecting to delete all this in the near future

PR #877, opened 2022-06-09
Add error callback to `handleRequest()` method on devserver

mcbattirola:
This PR adds an error callback to the `proxyServer.web` call on the devserver. This way, the server won't crash when the target is not running.

The goal is to improve DX and not have to restart the dev server if you restart teleport (or if you run the web UI before running teleport), which can take several seconds.

### DIFF
--- a/packages/build/devserver/index.js
+++ b/packages/build/devserver/index.js
@@ -114,7 +114,7 @@ function serveIndexHtml(req, res) {
 
   function handleRequest() {
     proxyServer.web(req, res, getTargetOptions(), (err, req, res) => {
-      const msg = `error handling request: ${err.message}. Is the target running an accessible at ${target}?`;
+      const msg = `error handling request: ${err.message}. Is the target running and accessible at ${target}?`;
       console.log(msg);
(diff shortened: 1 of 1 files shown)


packages/build/devserver/index.js
```diff
@@ -112,7 +113,12 @@ function serveIndexHtml(req, res) {
   }
 
   function handleRequest() {
-    proxyServer.web(req, res, getTargetOptions());
+    proxyServer.web(req, res, getTargetOptions(), (err, req, res) => {
+      const msg = `error handling request: ${err.message}. Is the target running and accessible at ${target}?`;
+      console.log(msg);
```
gzdunek, line 118: For other errors in `devserver` we use `console.error`, so we can use that function to be consistent (but the output looks the same).

gzdunek (approved):
Thanks for this improvement!

packages/build/devserver/index.js
mcbattirola: Awesome, thanks

PR #878, opened 2022-06-10
Add share feedback

gzdunek:
image

The white inputs will be fixed in another PR.
Also, I decreased height of the status bar (it was 32px on a design) to 28px, I think it looks better. Its color probably should be changed, it is too similar to color of the table.


packages/teleterm/src/ui/StatusBar/ShareFeedback/ShareFeedback.tsx
```diff
@@ -0,0 +1,67 @@
...
+    <Flex bg="primary.main" p={3} borderRadius={3}>
+      <Validation>
+        {({ validator }) => (
+          <Flex
+            flexDirection="column"
+            as="form"
+            onSubmit={e => {
+              e.preventDefault();
+              if (validator.validate()) {
+                console.log('submit', formValues); //TODO (gzdunek): connect to a real service
```
ravicious, line 30: I was thinking how we could make it so that we're able to still execute this code in development.

Perhaps if the app is not packaged, we could just change the endpoint to which it sends the request?

I did a bit of googling and found that [Postman has its own echo service](https://documenter.getpostman.com/view/5025623/SWTG5aqV#0cc70253-566a-4fc8-b311-2bda4d955b84).

This would mean that in development we could just point our code to `https://postman-echo.com/post` and it should still work.

packages/teleterm/src/ui/StatusBar/ShareFeedback/ShareFeedback.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { ShareFeedbackFormValues } from './types';
+
+interface ShareFeedbackProps {
+  onClose(): void;
+}
+
+export function ShareFeedback(props: ShareFeedbackProps) {
+  const [formValues, setFormValues] = useState<ShareFeedbackFormValues>({
+    feedback: '',
+    email: '',
```
ravicious, line 15: Could we pre-fill the email field with the username if the username looks like an email (`/^\S+@\S+$/`)?

packages/teleterm/src/ui/StatusBar/ShareFeedback/ShareFeedbackForm.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  setFormValues(values: ShareFeedbackFormValues): void;
+}
+
+export function ShareFeedbackForm({
+  formValues,
+  setFormValues,
+}: ShareFeedbackFormProps) {
+  const isFieldRequiringEmailEnabled =
+    formValues.newsletterEnabled || formValues.salesContactEnabled;
```
ravicious, line 20: Should we just make it to be always required? IIRC that's what we decided on during our recent conversation in DMs and the designs also seem to indicate that the email field is required.

ravicious (commented):
I took a very cursory look, I'll come back to this on Monday and we should also make a decision on those inputs then, yeah?

packages/teleterm/src/ui/StatusBar/ShareFeedback/ShareFeedbackForm.tsx
```diff
@@ -0,0 +1,86 @@
...
+      <Flex>
+        <Toggle
+          isToggled={formValues.newsletterEnabled}
+          onToggle={() => {
+            updateFormField('newsletterEnabled', !formValues.newsletterEnabled);
+          }}
+        />
+        <Text ml={2} color="text.primary">
+          Sign me up for the newsletter
+        </Text>
```
ravicious, line 68: Let's make this a label for the checkbox so that it's possible to click on the label to toggle the checkbox please.

I experimented with it a bit and it looks like we could modify `<Toggle>` to accept a child and then render the child within the label.

This works and doesn't change the visual look of the UI:

```html
<label class="Toggle__StyledWrapper-kQsuFA ekHuMv">
  <input type="checkbox" class="Toggle__StyledInput-jBswXp eetIWj">
  <div class="Toggle__StyledSlider-iAHUMy gedMTV"></div>
  <div color="text.primary" class="Text-pXEYb OoHRQ">Sign me up for the newsletter</div>
</label>
```

packages/teleterm/src/ui/StatusBar/ShareFeedback/ShareFeedback.tsx
ravicious, line 30: Any ideas on what it should say after submitting? Something like "Feedback submitted!" should be fine for now I guess?

gzdunek: This is a great idea, I use that service in the next PR when I will connect to the real backend.

packages/teleterm/src/ui/StatusBar/ShareFeedback/ShareFeedback.tsx
gzdunek: Ah yes, done

packages/teleterm/src/ui/StatusBar/ShareFeedback/ShareFeedbackForm.tsx
gzdunek: I think we didn't decide whether it should be required or not, but let's keep things simple. I got rid of this condition.

packages/teleterm/src/ui/StatusBar/ShareFeedback/ShareFeedbackForm.tsx
gzdunek: Done

packages/teleterm/src/ui/StatusBar/ShareFeedback/ShareFeedback.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+          loggedInUser: { name: 'alice' },
+        } as Cluster.AsObject;
+
+      if (clusterUri === '/clusters/production') {
+        return {
+          loggedInUser: {
+            name: 'bob@prod.com',
+          },
+        } as Cluster.AsObject;
+      }
```
ravicious, line 36: Could you do this within the scope of the test itself? From my experience, a test setup like this (a single mock which behavior depends on a value from tests) gets progressively harder to understand and maintain as the size of the test suite grows.

Instead of having to look just at the test body itself to determine what the behavior will be, you now also have to inspect functions defined in other parts of the test suite.

packages/teleterm/src/ui/StatusBar/ShareFeedback/ShareFeedback.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+  jest
+    .spyOn(WorkspacesService.prototype, 'getRootClusterUri')
+    .mockReturnValue('/clusters/localhost');
+
+  // @ts-expect-error - using mocks
+  const appContext: AppContext = {
+    clustersService,
+    workspacesService,
+  };
```
ravicious, line 47: Did you consider using `MockAppContext()` to get rid of `@ts-expect-error`?

I used it in tests for `retryWithRelogin`: I call `new MockAppContext()` first and then add spies for individual functions on services.

Not the ideal solution, but it keeps typechecking intact by not using `@ts-expect-error`. TDD purists would be rather unhappy there because we mock objects that we don't own: services are instantiated within AppContext (are "owned" by AppContext) and not passed from above ("owned" by us), but well… 🙊

I think ideally we'd do something like

```typescript
const AppContextMock = AppContext as jest.MockedClass<
  typeof AppContext
>;
```

but then we'd have to mock those services anyway due to how AppContext is structured, so for now I opted to just do it as described above.

---

In general, I think [`@ts-expect-error` should be used for tests](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#ts-ignore-or-ts-expect-error) where you write type signatures so that certain code is impossible to compile (because it results in type errors) and then you can use `@ts-expect-error` to write tests for that code. But I don't think it should be used as an escape hatch for situations where the way you mock stuff doesn't play well with TypeScript. ;)

gzdunek: Thanks for the advices, looks better now :)

ravicious: Nice, yeah I agree!

The conditionals you left in the mocked implementations, like

```typescript
if (clusterUri === '/clusters/localhost')
```

I don't think they're bad, it's just not something I've seen being used that often. What I'd usually see in other projects is that you'd mock a function and then verify that that function was called with correct arguments.

packages/teleterm/src/ui/StatusBar/ShareFeedback/ShareFeedbackForm.tsx
```diff
@@ -0,0 +1,77 @@
...
+        isToggled={formValues.salesContactEnabled}
+        onToggle={() => {
+          updateFormField(
+            'salesContactEnabled',
+            !formValues.salesContactEnabled
+          );
+        }}
+      >
+        <Text ml={2} color="text.primary">
+          I want to be contacted by the sales team
```
ravicious, line 72: Sasha posted this copy on #dev-connect:

> I want your sales team to demo me Teleport Enterprise features

It makes the form quite wide, perhaps we should put some max-width on it? The label would need to be fixed though because the spacing is broken when the label text is split into two lines.

ravicious (approved):
So is this the final form of this PR and then the upcoming ones will connect the form to the endpoint and add a visual effect to the icon if the app is launched for the first time?

packages/teleterm/src/ui/StatusBar/ShareFeedback/ShareFeedback.test.tsx
gzdunek: Good point, I replaced these conditionals with a function call check.

packages/teleterm/src/ui/StatusBar/ShareFeedback/ShareFeedbackForm.tsx
gzdunek: Done.
image

ravicious: Cool, did you do anything special to make those two checkbox labels line up or did I just get used to them looking like that? I couldn't find anything in the code. 🤔

ravicious: Actually……………………… could you put nbsp between Teleport and Enterprise?

ravicious: tbh I cannot decide which version looks worse; I'd say the one without nbsp looks better? At least it's not cut in an awkward way.

feedback form with nbsp

gzdunek: I'd vote for the one with `&nbsp;` - if we can avoid breaking the product name, let's do it.

>did you do anything special to make those two checkbox labels line up?
Hmm the only thing which I did was decreasing the `line-height` for the second label.

ravicious: > I'd vote for the one with `&nbsp;` - if we can avoid breaking the product name, let's do it.

👍

PR #882, opened 2022-06-14
[V9] Show node specific ssh logins options (#873)

mcbattirola:
Backports https://github.com/gravitational/webapps/pull/873 to v9

`yarn tsc` and `yarn test` are passing

Apparently there is no need to update `webapps.e` like in the original PR since both files updates on it don't exist on v9 (see https://github.com/gravitational/webapps.e/commit/4004ee8d94074f5197949fba6ba46ec6e6f4813a for the PR)

___

* Show node specific ssh logins options

* Use ssh logins from node response instead of user context on console

* Remove logins from user context


hatched (approved):

kimlisa (approved):
yeah refactor to access request is for v10 only, thanks for the check!

PR #883, opened 2022-06-14
Add ability to change db name for db proxy

gzdunek:
Closes: https://github.com/gravitational/webapps.e/issues/308
tshd counterpart: https://github.com/gravitational/teleport/pull/13476

Best to review commit by commit.

image

**OUTDATED**
~The `SET DATABASE NAME` button looks a bit different than in the issue, we can create a specialised version of `MenuLoginThemeProvider` if it's needed.~
image
image


ravicious:
I haven't looked at the code yet, but I noticed that the db name is not persisted between app restarts.

packages/shared/components/MenuLogin/types.ts
```diff
@@ -26,6 +26,9 @@ export type MenuLoginProps = {
   transformOrigin?: any;
   placeholder?: string;
   required?: boolean;
+  disabled?: boolean;
+  buttonText?: string;
+  info?: string;
```
ravicious, line 31: > The SET DATABASE NAME button looks a bit different than in the issue, we can create a specialised version of MenuLoginThemeProvider if it's needed.

Yeah, I totally forgot about that side effect of reusing MenuLogin. I have two ideas which will let us customize MenuLogin without having to use `MenuLoginThemeProvider`:

1. Add `caret` boolean prop to `MenuLogin` that will not render the caret if false, defaults to true.
   * I think that caret makes sense only when submitting the input will transition you somewhere + when we provide some suggestions (like on our DocumentCluster or in the server list in the Web UI). IMHO it feels weird for the db name and port buttons, especially since we won't even provide suggestions at the beginning.
   * idk why the icon is called `CarrotDown`. 😅 Even ["caret"](https://en.wikipedia.org/wiki/Caret_(computing)) is not the right name for it but I think "caret down" is a common name for it in icon sets.
3. Instead of using `ButtonBorder` in `MenuLogin`, use `Button` and make `buttonKind` a prop with a default of `"border"`.

I think those will elegantly solve this problem, won't they?

packages/shared/components/MenuLogin/MenuLogin.tsx
```diff
@@ -20,20 +20,28 @@ import { NavLink } from 'react-router-dom';
...
 
 export const MenuLogin = React.forwardRef<MenuLoginHandle, MenuLoginProps>(
   (props, ref) => {
-    const { onSelect, anchorOrigin, transformOrigin, required = true } = props;
+    const {
+      onSelect,
+      anchorOrigin,
+      transformOrigin,
+      required = true,
+      buttonText = 'CONNECT',
```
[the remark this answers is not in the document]
gzdunek: I reverted all of the changes to `MenuLogin`.

packages/shared/components/MenuLogin/MenuLogin.tsx
```diff
@@ -20,20 +20,28 @@ import { NavLink } from 'react-router-dom';
 import Menu, { MenuItem } from 'design/Menu';
 import { space } from 'design/system';
 import { MenuLoginProps, LoginItem, MenuLoginHandle } from './types';
-import { ButtonBorder, Flex, Indicator } from 'design';
-import { CarrotDown } from 'design/Icon';
+import { ButtonBorder, Flex, Indicator, Text } from 'design';
+import { CarrotDown, Info } from 'design/Icon';
 import { useAsync, Attempt } from 'shared/hooks/useAsync';
 
 export const MenuLogin = React.forwardRef<MenuLoginHandle, MenuLoginProps>(
```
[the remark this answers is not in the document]
gzdunek: I reverted all of the changes to `MenuLogin`.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/Databases.tsx
```diff
@@ -78,40 +76,22 @@ function ConnectButton({
...
-          />
-        </OverlayGrid>
+        <MenuLogin
+          placeholder="Enter username"
+          getLoginItems={() => getDatabaseUsers(appContext, documentUri, dbUri)}
+          onSelect={(_, user) => {
+            onConnect(user);
+          }}
+          transformOrigin={transformOrigin}
+          anchorOrigin={anchorOrigin}
```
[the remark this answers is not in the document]
gzdunek: Done

packages/teleterm/src/ui/DocumentGateway/DocumentGateway.tsx
```diff
@@ -87,8 +101,34 @@ export function DocumentGateway(props: State) {
...
+        isLoading={isLoading}
+        onRun={runCliCommand}
+      />
+      <Flex justifyContent="end">
+        <MenuLoginTheme>
+          <MenuLogin
+            disabled={isLoading}
+            placeholder="Enter database name"
+            required={false}
+            buttonText="SET DATABASE NAME"
```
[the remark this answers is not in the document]
gzdunek: No longer needed.

packages/teleterm/src/ui/DocumentGateway/DocumentGateway.tsx
```diff
@@ -87,8 +101,34 @@ export function DocumentGateway(props: State) {
         </ButtonSecondary>
       </Flex>
       <Text bold>Connect with CLI</Text>
-      <CliCommand cliCommand={gateway.cliCommand} onClick={runCliCommand} />
-      <Text bold>Connect with GUI</Text>
+      <CliCommand
+        cliCommand={gateway.cliCommand}
+        isLoading={isLoading}
```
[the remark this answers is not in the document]
gzdunek: Yes, I added also a kind of _debounce_ to `isLoading` so the command won't blink when user changes the db name.

packages/shared/components/MenuLogin/types.ts
gzdunek: No longer needed.

packages/teleterm/src/ui/DocumentGateway/DocumentGateway.tsx
```diff
@@ -147,3 +221,26 @@ function CliCommand({
     </Flex>
   );
 }
+
+const ConfigInput: typeof FieldInput = styled(FieldInput)`
```
gzdunek, line 225: I'm not a fan of this, but the white `input` looked really awful.

packages/teleterm/src/ui/DocumentGateway/DocumentGateway.tsx
```diff
@@ -87,8 +110,29 @@ export function DocumentGateway(props: State) {
         </ButtonSecondary>
       </Flex>
       <Text bold>Connect with CLI</Text>
```
ravicious, line 112: Those uppercased labels look a bit weird next to a section header that's almost the same size as them.

```suggestion
      <Text typography="h4">Connect with CLI</Text>
```

packages/teleterm/src/ui/DocumentGateway/DocumentGateway.tsx
```diff
@@ -87,8 +110,29 @@ export function DocumentGateway(props: State) {
...
+        onRun={runCliCommand}
+      />
+      {changeDbNameAttempt.status === 'error' && (
+        <Alerts.Danger>
+          Could not change the database name: {changeDbNameAttempt.statusText}
+        </Alerts.Danger>
+      )}
+      <Text bold mt={3}>
+        Connect with GUI
+      </Text>
```
ravicious, line 135: Same as above.

packages/teleterm/src/ui/DocumentGateway/DocumentGateway.tsx
```diff
@@ -45,8 +57,19 @@ export function DocumentGateway(props: State) {
...
+    changeDbName,
+    changeDbNameAttempt,
+    doc,
   } = props;
 
+  const handleChangeDbName = useMemo(() => {
+    return debounce((value: string) => {
+      changeDbName(value);
+    }, 150);
+  }, []);
```
ravicious, line 69: Changing the db name fails for me ("Could not change the database name: Could not find gateway undefined") until I change some code and cause React to reload the components.

My bet is that `useMemo` memoizes `changeDbName` when `doc.gatewayUri` is still undefined (it gets updated after the gateway is actually created).

To fix this, this `useMemo` would need to depend on `changeDbName` and `changeDbName` would need to have a stable identity itself.

packages/teleterm/src/ui/DocumentGateway/DocumentGateway.tsx
gzdunek: Good idea, looks better

packages/teleterm/src/ui/DocumentGateway/DocumentGateway.tsx
gzdunek: It didn't happen to me, but right, it can be caused by the `useMemo`.

I updated `useAsync` too.

ravicious:
I haven't looked at the code yet, but I tried to use it to test my refactor of the backend part.

It still fails for me with the same error message after creating a gateway. After app restart and reopening tabs, I also can't change the db name but this time it fails with another message: "Could not change the database name: Could not find gateway /gateways/445685d4-447d-4ce7-af8d-d8b212b28c8f".

I suspect that's because the gateway URI is persisted in the doc, but after restart it's no longer valid (the gateway no longer exists). `DocumentGateway` creates a new gateway and updates its gateway URI but the old one still gets used due to caching, presumably.

packages/teleterm/src/ui/DocumentGateway/DocumentGateway.tsx
```diff
@@ -86,9 +109,30 @@ export function DocumentGateway(props: State) {
...
-      <Text bold>Connect with GUI</Text>
+      <Text typography="h4">Connect with CLI</Text>
+      <Flex>
+        <Validation>
+          <ConfigInput
+            label="Database name"
+            defaultValue={doc.targetSubresourceName}
+            onChange={e => handleChangeDbName(e.target.value)}
+            spellCheck={false}
+          />
```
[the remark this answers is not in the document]
gzdunek: Agree, seems like all we have to do is just change the max-width to 580px.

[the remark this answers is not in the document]
ravicious: That doesn't seem to be true. But let's leave this for now and come back to it when we're solving the same problem is some other tab.

https://user-images.githubusercontent.com/27113/174812828-cd480472-5e69-4768-8a0b-ac89fcf70bbb.mov

ravicious (approved):
Let's wait with merging this until the backend part is done.

PR #884, opened 2022-06-14
Refactor StepSlider Component

kimlisa:
#### Description
Moves switching to a new flow logic to parent's responsibility instead of children with the `swithFlow` method (which is now removed)

Note: easier to review by commit, I ended up having to move this component to `design` package b/c I also need it for `teleterm` package


kimlisa:
@rudream @hatched friendly ping

PR #886, opened 2022-06-14
additional set up for prettier

michellescripts:
A few things to make prettier easier to use: 

 - add prettier-check & prettier-write scripts to package.json
  - add prettierignore so that the script can be run from root (& abc'd scripts for easier reading)
  - replaced deprecated jsxBracketSameLine with bracketSameLine
  - commiting lock file after yarn install


package.json
```diff
@@ -4,24 +4,25 @@
   "main": "index.js",
   "license": "Apache-2.0",
   "scripts": {
```
michellescripts, line 6: For this section:

- added "prettier-check": "yarn prettier --check ./packages"
- abc'd the scripts to make them more readable (^ line changes, no changed functionality)

hatched (approved):
Thanks!

.prettierignore
```diff
@@ -0,0 +1,10 @@
+# Ignore directories:
+packages/build
+packages/teleterm/src/services/tshd/v1/**
+packages/teleterm/src/sharedProcess/**
```
ravicious, line 4: I'm pretty sure this should be `packages/teleterm/src/sharedProcess/api/protogen/**` instead, as `protogen` is the actual directory that holds automatically generated files.

yarn.lock
```diff
@@ -2631,11 +2631,6 @@
   resolved "https://registry.yarnpkg.com/@types/google-protobuf/-/google-protobuf-3.15.6.tgz#674a69493ef2c849b95eafe69167ea59079eb504"
   integrity sha512-pYVNNJ+winC4aek+lZp93sIKxnXt5qMkuKmaqS3WGuTq0Bw1ZDYNBgzG5kkdtwcv+GmYJGo3yEg6z2cKKAiEdw==
 
-"@types/google-protobuf@^3.7.2":
-  version "3.15.5"
-  resolved "https://registry.yarnpkg.com/@types/google-protobuf/-/google-protobuf-3.15.5.tgz#644b2be0f5613b1f822c70c73c6b0e0b5b5fa2ad"
-  integrity sha512-6bgv24B+A2bo9AfzReeg5StdiijKzwwnRflA8RLd1V4Yv995LeTmo0z69/MPbBDFSiZWdZHQygLo/ccXhMEDgw==
-
```
ravicious, line 2638: I'm not 100% sure, but I think this was here because some stuff from [webapps.e](https://github.com/gravitational/webapps.e) depends on it, but if you haven't initialized that submodule, then yarn wouldn't see the package.json from that package.

package.json
ravicious: > abc'd the scripts to make them more readable

Do we plan to enforce that or will that depend on people remembering to keep the order?

I don't know if I'd call that more readable myself. 😅 Before this change, the scripts were more or less in some logical groups (stuff for tests, then stuff related to Teleport Connect, then the Web UI, then some linting), now they're just in alphabetical order. To be honest I'd be for keeping them that way.

package.json
```diff
@@ -4,24 +4,26 @@
...
-    "build-oss": "yarn build-teleport-oss",
-    "build-e": "yarn build-teleport-e",
-    "type-check-and-lint": "yarn type-check && yarn lint",
+    "build-teleport-oss": "yarn workspace @gravitational/teleport build --output-path=../../dist/teleport/app",
+    "build-term": "yarn workspace @gravitational/teleterm build",
     "lint": "eslint --ext .js,.jsx,.ts,.tsx packages/",
-    "type-check": "tsc --noEmit"
+    "package-term": "yarn workspace @gravitational/teleterm package",
+    "prettier-check": "yarn prettier --check ./packages",
+    "prettier-write": "yarn prettier --write ./packages",
```
kimlisa, line 17: maybe i'm overthinking it, but does the `write` command apply prettier on every file in `./packages`? b/c if it does, i don't think this repo is ready for this command yet. maybe after we go through a review process of prettifying every file (ensuring no intended changes went in), then we can introduce this and a checker for the CI (or could we have something like prettifying `modified` files only?)

PR #890, opened 2022-06-15
Prevent recreating a gateway when it is removed

gzdunek:


ravicious (approved):
I'm not sure if the title of this PR is correct, perhaps something like "Ensure the gateway is created only once when opening the doc" better illustrates the intent?

ravicious:
The logic here is tricky and we modified it a couple of times, so I think it's important for the git history of this file to be as clear as possible.

PR #899, opened 2022-06-21
Connect `ShareFeedback` with API

gzdunek:
I didn't expect to do so many changes, sorry for that :( 
I decided to to move the `Popover` and `Icon` components from the `StatusBar` to `ShareFeedbackContainer`, because we I had to add more logic related to opening the popover and displaying things like indicator on the icon. I don't think `StatusBar` should be aware of that.   

Please review commit-by-commit, or at least without the last, renaming commit.

Indicator at the bottom right corner:
image

I tested `Popover` changes in the web ui, it looks like I don't break anything :)  

The form API is not working properly due to CORS issues, but it should work without any changes.


gzdunek:
I forgot to add, in the previous version state of the form was lost when it was closed. Now it is kept during the lifetime of the app, which is desirable IMO.

packages/design/src/Popover/Popover.jsx
```diff
@@ -203,6 +219,8 @@ export default class Popover extends React.Component {
     return {
       top: `${top}px`,
       left: `${left}px`,
+      bottom: `${containerWindow.innerHeight - bottom}px`,
+      right: `${containerWindow.innerWidth - right}px`,
```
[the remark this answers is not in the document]
ravicious: Before the initial definition of `bottom` and `right`, I'd also be for adding this comment, otherwise the variable names might be a bit misleading.

```typescript
    // bottom and right correspond to the calculated position of the element from the top left, not
    // from the bottom right, meaning they must be inverted before using them as `bottom` and
    // `right` CSS properties.
    let bottom = top + elemRect.height;
    let right = left + elemRect.width;
```

packages/design/src/Popover/Popover.jsx
```diff
@@ -129,11 +129,27 @@ export default class Popover extends React.Component {
...
-      element.style.left = positioning.left;
+    if (this.props.transformOrigin.horizontal === 'right') {
+      if (positioning.right) {
+        element.style.right = positioning.right;
+      }
+    } else {
+      if (positioning.left) {
+        element.style.left = positioning.left;
+      }
     }
```
ravicious, line 153: What's the reason behind tying top-left/bottom-right to `props.transformOrigin`?

`transformOrigin` dictates the point of the popover that will attach to the anchor's origin. Using top-left vs bottom-right dictates how the component will grow in height and width. I can imagine situations where someone would want those two to not be coupled to each other.

Tying this to `props.transformOrigin` also means that we're changing how the Popover behaves not only in the case of the feedback form but also every other instance of popover that used right or bottom for its `transformOrigin`.

Perhaps a safer and more extensible approach would be to introduce a new prop? For now it wouldn't have to be super-extensible and handle all imaginable cases. Something like `positionalProperties` which accepts either `top-left` (default) or `bottom-right`.

ravicious (commented):
I started to look at the Popover changes yesterday, just submitting the feedback for that right now and I'm diving into the rest of the code right after.

packages/design/src/Popover/Popover.jsx
ravicious: In case of adding new props, please remember to add new items to the story as well.

packages/teleport/src/components/Toggle/Toggle.tsx
```diff
@@ -63,7 +77,7 @@ const StyledSlider = styled.div`
 const StyledInput = styled.input.attrs({ type: 'checkbox' })`
```
ravicious, line 77: Let's add this at the end of styles for `StyledInput` to visually differentiate disabled Toggles from enabled ones:

```typescript
  &:disabled + ${StyledSlider} {
    background: ${props => props.theme.colors.primary.light};

    &:before {
      background: ${props => props.theme.colors.grey[700]};
    }
  }
```

packages/teleport/src/components/Toggle/Toggle.tsx
```diff
@@ -17,10 +17,19 @@ limitations under the License.
...
+export default function Toggle({
+  isToggled,
+  onToggle,
+  children,
+  disabled,
+}: Props) {
   return (
-    <StyledWrapper>
-      <StyledInput checked={isToggled} onChange={() => onToggle()} />
+    <StyledWrapper disabled={disabled}>
```
ravicious, line 27: I don't think labels support the `disabled` attribute, but I'm okay with leaving this as it is for now.

packages/teleterm/src/ui/services/statePersistence/statePersistenceService.ts
```diff
@@ -58,8 +75,11 @@ export class StatePersistenceService {
       workspacesState: {
         workspaces: {},
       },
+      shareFeedback: {
+        hasBeenOpened: false,
+      },
     };
-    return this._fileStorage.get('state') || defaultState;
+    return { ...defaultState, ...this._fileStorage.get('state') };
```
ravicious, line 82: While the change here itself is fine IMHO, I'm commenting on this just because I want us to be super careful here. This encroaches on that "migrations for the persisted state" territory that we've talked about.

Any top-level additions to the default state should be safe but anything other than that is a breaking change, isn't it?

From my experience, once we'll be up against making a breaking change, it'll be much easier to add some hacks and defensive code inside the app code itself rather than build a storage system that can enforce that the saved state has certain fields, for example through migrations.

packages/teleterm/src/ui/StatusBar/ShareFeedback/ShareFeedbackForm.tsx
```diff
@@ -39,9 +39,20 @@ export function ShareFeedbackForm({
         value={formValues.email}
         onChange={e => updateFormField('email', e.target.value)}
       />
+      <FieldInput
+        label="Company name (optional)"
+        autoFocus
```
ravicious: The company name field shouldn't have autofocus, should it?

packages/teleterm/src/ui/StatusBar/ShareFeedback/ShareFeedbackFormFields.tsx
```diff
@@ -0,0 +1,100 @@
...
+        css={`
+          input {
+            font-size: 14px;
+          }
+        `}
+        value={formValues.company}
+        onChange={e => updateFormField('company', e.target.value)}
+      />
+      <FieldTextArea
+        label="Suggestions (Optional)"
```
ravicious, line 58:
```suggestion
        label="Suggestions (optional)"
```

packages/teleterm/src/ui/StatusBar/ShareFeedback/ShareFeedbackForm.tsx
```diff
@@ -39,9 +39,20 @@ export function ShareFeedbackForm({
...
+        readonly={disabled}
```
ravicious: We should be using the `disabled` attribute here but that goes beyond the scope of the PR as I see we use it because of the way `FieldInput` is written.

packages/teleterm/src/ui/StatusBar/ShareFeedback/useShareFeedback.ts
```diff
@@ -0,0 +1,99 @@
...
+    formData.set('company', formValues.company);
+    formData.set('use-case', formValues.feedback);
+    formData.set('newsletter-opt-in', formValues.newsletterEnabled ? 'y' : 'n');
+    formData.set('sales-opt-in', formValues.salesContactEnabled ? 'y' : 'n');
+
+    const headers = new Headers();
+    headers.set('content-type', 'multipart/form-data');
+
+    const response = await fetch(
+      'https://kcwm2is93l.execute-api.us-west-2.amazonaws.com/prod',
```
ravicious, line 40: How do you plan to provide different URLs here?

packages/teleterm/src/ui/StatusBar/ShareFeedback/ShareFeedbackForm.tsx
```diff
@@ -1,83 +1,124 @@
...
-    field: T,
-    value: ShareFeedbackFormValues[T]
-  ) {
-    setFormValues({ ...formValues, [field]: value });
+function SubmissionError(props: { submitFeedbackAttempt: Attempt<Response> }) {
+  function getErrorText() {
+    if (props.submitFeedbackAttempt.statusText === FEEDBACK_TOO_LONG_ERROR) {
+      return (
+        <span>
+          That's a very long reason why. Please let us know more in{' '}
```
ravicious, line 100:
```suggestion
          That's a very long reason suggestion. Please let us know more in{' '}
```

ravicious (commented):
The code around the feedback form itself is solid, I think we need to make a couple of changes to those shared components though.

packages/design/src/Popover/Popover.jsx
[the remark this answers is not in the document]
gzdunek: Thanks, done!

packages/design/src/Popover/Popover.jsx
gzdunek: Ah, I was thinking about it incorrectly. You're absolutely right, `top-left`/`bottom-right` shouldn't be tied to `props.transformOrigin`. 
I added `growDirections` prop to the popover with the values you proposed (`bottom-right` is a default value).
There is also an updated story for this.

packages/teleport/src/components/Toggle/Toggle.tsx
gzdunek: Good idea, thanks!

packages/teleterm/src/ui/services/statePersistence/statePersistenceService.ts
gzdunek: >While the change here itself is fine IMHO, I'm commenting on this just because I want us to be super careful here. This encroaches on that "migrations for the persisted state" territory that we've talked about.

I agree, that we should be really careful here.

>Any top-level additions to the default state should be safe but anything other than that is a breaking change, isn't it?

Yes. I did this change to allow us to add new top-level sections.

packages/teleterm/src/ui/StatusBar/ShareFeedback/ShareFeedbackForm.tsx
gzdunek: No, it shouldn't.

packages/teleterm/src/ui/StatusBar/ShareFeedback/ShareFeedbackFormFields.tsx
gzdunek: Done

packages/teleterm/src/ui/StatusBar/ShareFeedback/useShareFeedback.ts
gzdunek: I added 
```typescript  
const feedbackUrl = ctx.mainProcessClient.getRuntimeSettings().dev
    ? 'https://kcwm2is93l.execute-api.us-west-2.amazonaws.com/prod'
    : 'https://usage.teleport.dev';
```
what do you think?

packages/teleterm/src/ui/StatusBar/ShareFeedback/ShareFeedbackForm.tsx
gzdunek: Done, I think you meant `That's a very long suggestion.` :)

packages/design/src/Popover/Popover.jsx
```diff
@@ -361,6 +380,11 @@ Popover.propTypes = {
     left: PropTypes.number.isRequired,
     top: PropTypes.number.isRequired,
   }),
+  /**
+   * These are the directions in which `Popover` will grow
+   * if its content increase it dimensions after it is opened.
```
ravicious, line 385:
```suggestion
   * if its content increases its dimensions after `Popover` is opened.
```

packages/teleterm/src/ui/StatusBar/ShareFeedback/useShareFeedback.ts
ravicious: Cool for now, but in the future let's see if we can utilize another mechanism for this. Usually I'd do it through some kind of env variables inserted during compilation, in general we need some kind of config that will be different for dev vs prod.

ravicious (approved):
That story you made for Popover changing in size is pretty cool!

PR #900, opened 2022-06-21
Add ability to change port for db proxy

gzdunek:
Closes https://github.com/gravitational/webapps.e/issues/302

---

From Grzegorz:

I had to refactor `DocumentGateway` as it became too long. I created a few smaller components.
Now, we have a `DocumentGateway` which only renders `OfflineDocumentGateway` or `OnlineDocumentGateway` basing on a condition.
There are no changes to `CliCommand`, `ConfigFieldInput` and stuff related to db name.

I wasn't sure what do you mean here:
> In that situation, when the gateway hasn't been created yet, under the hood we should change the port number of the doc and then attempt to create the gateway as usual.

What port you would like to set in the doc?

---

From Rafał:

gravitational/teleport#14690 implements the backend part and is already in master so you just need a fresh tsh build to test this PR.

I added storybooks for `DocumentGateway` which should offer an easy way to see how the component looks like in different states.

The easiest way to test occupied ports:

1. Log in to two clusters.
2. Create a db gateway in one of them and set its port to 12345.
3. Switch to the other cluster.
4. Restart the app.
5. Create a db gateway in the other cluster and set its port to 12345.
6. Switch to the first cluster and restore tabs.

## Changes compared to the original mockups from https://github.com/gravitational/webapps.e/issues/302

The UI is no longer in the form of a dropdown, just a regular field with debounced update.

> What if the RPC to set the port fails… …and the gateway is not started back up?

The original plan included starting a new listener on the port, checking if it's free then closing the listener and starting a new one on that port. During that time, someone else could take the port so the original issue describes a plan to address this.

What I didn't realize at the time is that we can just create that listener on that new port and then pass it around. So if the new gateway on the new port fails to start, we simply never close the old one and return an error. This means that if the call to change the port fails, we don't need to sync the gateway list or anything like that, unlike what the original plan says.

---

Todo:

* [x] https://github.com/gravitational/webapps/pull/900#discussion_r902773230
* [x] Add more stories, particularly around different error states of `useAsync` attempts.


packages/teleterm/src/ui/DocumentGateway/common.tsx
```diff
@@ -25,8 +25,21 @@ export const ConfigFieldInput: typeof FieldInput = styled(FieldInput)`
...
+  ::-webkit-outer-spin-button {
+    -webkit-appearance: none;
+    margin: 0;
+  }
+`;
+
 export const PortFieldInput: typeof ConfigFieldInput = props => (
-  <ConfigFieldInput
+  <ConfigFieldInputWithoutStepper
     type="number"
```
ravicious, line 43: Many answers on the internet recommend using `type="text" inputmode="numeric"` if you don't need the stepper. But if we use that, we lose all the nice validations that come with `type="number"` + it's possible to type any character on a regular keyboard.

I decided to hide the stepper then and categorize this as lack of browser's support for hiding the stepper through another attribute. 😏

packages/teleterm/src/ui/DocumentGateway/common.tsx
```diff
@@ -25,8 +25,21 @@ export const ConfigFieldInput: typeof FieldInput = styled(FieldInput)`
...
+)`
+  ::-webkit-inner-spin-button {
+    -webkit-appearance: none;
+    margin: 0;
+  }
+  ::-webkit-outer-spin-button {
+    -webkit-appearance: none;
+    margin: 0;
+  }
+`;
```
ravicious, line 39: I have no idea why this works but nesting those styles under `input {}` didn't.

If I simply add `color: red` here, the style seems to be applied to the `input` itself rather than the whole component. @gzdunek maybe you have an idea what's going on here?

packages/teleterm/src/ui/DocumentGateway/common.tsx
gzdunek: Totally agree!

packages/teleterm/src/ui/DocumentGateway/common.tsx
gzdunek: Yes, it took me a while to notice that there was no closing bracket for the `input` selector in `ConfigFieldInput`. It was breaking the styles for the `ConfigFieldInputWithoutStepper`. Pushed a fix for that.

gzdunek:
I also fixed `Alerts` imports. React was throwing an exception because it was expecting a component, but got undefined.

ravicious:
> I wasn't sure what do you mean here:
> 
> > In that situation, when the gateway hasn't been created yet, under the hood we should change the port number of the doc and then attempt to create the gateway as usual.
> 
> What port you would like to set in the doc?

I meant that on the next attempt to reconnect, the app should use the port from the input. Updating the `port` field on `DocumentGateway` seemed like one of the ways to accomplish that. But I see that you already did that in a slightly different and perhaps better way.

ravicious:
@gzdunek I can't tag you for a review so I'm doing this manually. ;)

Everything below "Realign form elements in OfflineDocumentGateway" was added by me, as the commit list suggests.

ravicious:
FYI @avatus, I took ownership of this PR as Grzegorz started working on the Windows version.

packages/teleterm/src/ui/services/connectionTracker/connectionTrackerService.test.ts
```diff
@@ -78,3 +85,75 @@ it('removeItemsBelongingToRootCluster removes connections', () => {
     { clusterName: 'remote_leaf', ...connections[3] },
   ]);
 });
+
+it('updates the port of a gateway connection when the underlying doc gets updated', () => {
```
ravicious, line 89: @avatus I'm curious what you think about the way we write tests. I've never really TypeScript tests in the past so the whole `X as jest.MockedClass<typeof X>` is new to me and I'm not sure if there's a better way.

Arguably `ConnectionTrackerService` is quite complex, as shown by how complex the test setup is.

packages/teleterm/src/ui/DocumentGateway/OfflineDocumentGateway.tsx
```diff
@@ -0,0 +1,72 @@
...
+        alignItems="flex-end"
+        flexWrap="wrap"
+        justifyContent="space-between"
+        mt={3}
+        gap={2}
+      >
+        {shouldShowPortInput && (
+          <Validation>
+            <PortFieldInput
+              label="Port (optional)"
```
gzdunek, line 58: Let's add `autoFocus` to this input.

gzdunek:
### APPROVED

packages/teleterm/src/ui/DocumentGateway/OfflineDocumentGateway.tsx
ravicious: I'm not sure if we should do it here. I think autofocus makes sense in places where typing something is the only natural way to proceed, like the login form. Here creating a gateway might fail due to a variety of reasons (no network connection, lost access to the database server, expired certs). A port being occupied is only one of them so I don't think we should guide the user towards changing the port.

Is it okay if I leave it without autofocus then?

packages/teleterm/src/services/tshd/createClient.ts
```diff
@@ -287,6 +287,21 @@ export default function createClient(
         });
       });
     },
+
+    async setGatewayLocalPort(gatewayUri = '', localPort = '') {
```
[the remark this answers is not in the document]
ravicious: I was cargo culting what was already there. 😶‍🌫️

One explanation that I came up with: this might make sense if we were trying to mimic Go API where an empty string is the default value you'd use for something of type `string` since you cannot pass nil where a string is expected.

But for me the tshd client level is something that should be grounded in the JS world. Here I'd rather just force those args to be present.

I'm going to remove this function call but leave them for others as I'm not sure if we're free to do that.

packages/teleterm/src/ui/DocumentGateway/OfflineDocumentGateway.tsx
gzdunek: Yeah, I'm convinced.

packages/teleterm/src/services/tshd/createClient.ts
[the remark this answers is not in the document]
gzdunek: OK, let's do this.

packages/teleterm/src/ui/services/connectionTracker/connectionTrackerService.test.ts
avatus: I see this normally when using typescript to make writing the tests easier (the mocks now have the right class signatures). This is a complex test but I don't really know of a better way (also not the strongest with deep deep testing)

JanKaczmarkiewicz: Casting here is not needed. StatePersistenceService and ClustersService can be used directly

ravicious: But then the tests would use the real implementation of `ClustersService` and `StatePersistenceService`, wouldn't they?

I do want to mock those objects. They are not the subjects of the test. For the most part I don't care how they behave as long as they fulfill the contract they have with `ConnectionTrackerService`, that is `StatePersistenceService.prototype.getConnectionTrackerState` returns an object of a particular shape.

JanKaczmarkiewicz: > But then the tests would use the real implementation of ClustersService and StatePersistenceService, wouldn't they?

No. `as` keyword in typescript changes the only type of expression.

JanKaczmarkiewicz: To mock them you would need 
```
jest.mock('../statePersistence', () => ({
    StatePersistenceService: ... // you mock implementation
}))
```

ravicious: Ah, sorry, I misunderstood you completely.

So I just call `jest.mock(path)` and then use automocked classes from that path just as if `jest.mock()` wasn't called at all?

I checked the docs and it looks like `jest.MockedClass` was removed from the docs starting with Jest v27.

ravicious (approved):
LGTM 😎

packages/teleterm/src/ui/services/connectionTracker/connectionTrackerService.test.ts
JanKaczmarkiewicz: I only mean that in
```ts
const StatePersistenceServiceMock =
    StatePersistenceService as jest.MockedClass<typeof StatePersistenceService>;
  const ClustersServiceMock = ClustersService as jest.MockedClass<
    typeof ClustersService
  >;
```
we don't need `ClustersService` and `StatePersistenceService` to have `jest.MockedClass<...>` type because we don't use special `jest.MockedClass` methods/properties of that type

PR #901, opened 2022-06-22
Show cluster breadcrumbs

gzdunek:
Closes https://github.com/gravitational/webapps.e/issues/318
image


packages/teleterm/src/ui/services/workspacesService/documentsService/documentsUtils.ts
```diff
@@ -0,0 +1,25 @@
...
+    case 'doc.cluster':
+      return document.clusterUri;
+    case 'doc.gateway':
+      return document.targetUri;
+    case 'doc.terminal_tsh_node':
+      return document.serverUri;
+    case 'doc.terminal_tsh_kube':
+      return document.kubeUri;
+    case 'doc.terminal_shell':
+      return routing.getClusterUri(document);
```
gzdunek, line 15: I'm considering adding a `clusterUri` field to `DocumentPtySession` (that field would be constructed in the same way as here, basing on `rootClusterId` and `leafClusterId`).

packages/teleterm/src/ui/StatusBar/useActiveDocumentClusterBreadcrumbs.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  const rootCluster = ctx.clustersService.findCluster(rootClusterUri);
+  const leafCluster =
+    clusterUri === rootClusterUri
+      ? undefined
+      : ctx.clustersService.findCluster(clusterUri);
+
+  return [rootCluster, leafCluster]
+    .filter(Boolean)
+    .map(c => c.actualName)
```
ravicious, line 34: Are we sure that `actualName` will always be available here?

packages/teleterm/src/ui/services/workspacesService/documentsService/documentsUtils.ts
ravicious: Good idea, although I'd probably leave this for later. I looked into that a couple of weeks ago and that whole document could use some refactoring if we were to start changing those fields.

ravicious (approved):

packages/teleterm/src/ui/StatusBar/useActiveDocumentClusterBreadcrumbs.ts
gzdunek: Yes, for leaf clusters we just return cluster name as `actualName`.

PR #906, opened 2022-06-23
Show connection type in connection tracker

gzdunek:
image

### DIFF
--- a/packages/teleterm/src/ui/TopBar/Connections/ConnectionsFilterableList/ConnectionItem.tsx
+++ b/packages/teleterm/src/ui/TopBar/Connections/ConnectionsFilterableList/ConnectionItem.tsx
@@ -120,10 +120,6 @@ export function ConnectionItem(props: ConnectionItemProps) {
 }
 
 function getKindName(kind: ExtendedTrackedConnection['kind']): string {
-  function assertUnreachable(x: never): never {
-    throw new Error(`Unhandled case: ${x}`);
-  }
-
   switch (kind) {
     case 'connection.gateway':
       return 'DB';
@@ -132,4 +128,8 @@ function getKindName(kind: ExtendedTrackedConnection['kind']): string {
     default:
       assertUnreachable(kind);
   }
+}
+
+function assertUnreachable(x: never): never {
+  throw new Error(`Unhandled case: ${x}`);
 }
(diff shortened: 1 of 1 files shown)


packages/teleterm/src/ui/TopBar/Connections/ConnectionsFilterableList/ConnectionItem.tsx
```diff
@@ -100,3 +112,18 @@ export function ConnectionItem(props: ConnectionItemProps) {
     </ListItem>
   );
 }
+
+function getKindName(kind: ExtendedTrackedConnection['kind']): string {
+  function assertUnreachable(x: never): never {
+    throw new Error(`Unhandled case: ${x}`);
+  }
```
ravicious, line 119: Could you move this out of `getKindName`? :D

packages/teleterm/src/ui/TopBar/Connections/ConnectionsFilterableList/ConnectionItem.tsx
```diff
@@ -75,7 +75,19 @@ export function ConnectionItem(props: ConnectionItemProps) {
...
+                background: rgba(255, 255, 255, 0.25);
+                opacity: 0.85;
+                padding: 1px 2px;
+                margin-right: 4px;
+                border-radius: 4px;
+              `}
+            >
+              {getKindName(props.item.kind)}
+            </span>
+            <span>{props.item.title}</span>
```
ravicious, line 90: Could you add `vertical-align: middle` to the span with connection title? It should align those two spans that have different font sizes.

ravicious (approved):
Tested with long resource name and it appears to work just fine.

packages/teleterm/src/ui/TopBar/Connections/ConnectionsFilterableList/ConnectionItem.tsx
gzdunek: Done

packages/teleterm/src/ui/TopBar/Connections/ConnectionsFilterableList/ConnectionItem.tsx
gzdunek: Done 👍

PR #908, opened 2022-06-23
Small fixes and tweaks while going through test plan

kimlisa:
#### Description

- Updating outdated doc links
- Adding white border around QR code (the scanner wasn't picking it up otherwise)
- Fix sorting bug for Users `role` column
- Change language `reset password` to `reset autthentication` since users can register passwordlessly


packages/teleport/src/Users/UserList/UserList.tsx
```diff
@@ -40,6 +40,12 @@ export default function UserList({
           key: 'roles',
           headerText: 'Roles',
           isSortable: true,
+          onSort: (a: string[], b: string[]) => {
```
hatched, line 43: Interesting approach to sort based on a list of roles :)

packages/teleport/src/services/user/makeUser.ts
```diff
@@ -21,7 +21,7 @@ export default function makeUser(json): User {
   const [name, roles, authType] = at(json, ['name', 'roles', 'authType']);
   return {
     name,
-    roles: roles || [],
+    roles: roles ? roles.sort() : [],
```
hatched, line 24: Should this use the same sort method as above? Looking at the snapshot it looks like the sort is now out of order (unless I'm reading it wrong).

kimlisa: You're right the sort is now out of order b/c in storybook we mock the users and it doesn't go through the `makeUser` that sorts it, which i think it's fine. I moved the sorting here so we are not sorting the list of roles every time the table re-renders (and I needed sorted list to sort the column).

hatched (approved):

packages/teleport/src/Users/UserList/UserList.tsx
kimlisa: i reverted back to basic sorting, so we use basic sorting when we sort the list of roles

Issue #863, opened 2022-06-03
Update Design System


rudream:
## Purpose

Our design system is outdated and still not typed. We need to update the `styled-system` package to the latest version and update our implementation and stop using deprecated methods. We also need to add typing to our design components to make them easier to use. 

### Steps

- [ ] Update `styled-system` package to `v5`
- [ ] Update `styled-system` implementation (stop using `.propTypes` and update `theme`)
- [ ] Add typing to all `design` components

PR #918, opened 2022-06-24
Update styled-system package to v5.1.5

rudream:
## Purpose

Part of #863 

Updates the `styled-system` package to the latest version 

## Implementation

This implementation with the `@styled-system/prop-types` package and `space.px` is temporary and just there to allow us to update to the latest version of `styled-system` without breaking the app. This added package will be removed in the next PR.

PS: Review the first commit on its own to avoid the clutter caused by the trivial snapshot updates

### DIFF
--- a/packages/design/package.json
+++ b/packages/design/package.json
@@ -16,7 +16,8 @@
   },
   "dependencies": {
     "cross-env": "5.0.5",
-    "styled-system": "^3.1.11",
+    "styled-system": "^5.1.5",
+    "@styled-system/prop-types": "^5.1.5",
     "@gravitational/build": "^1.0.0"
   }
 }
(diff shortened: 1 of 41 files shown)


gzdunek (commented):
Good job, but I found a few regressions :( 

1. Table no longer has rounded corners.
2. Margins are incorrect on the `Text` elements, example: 
image
I guess it's because in the new version, applying margin on the element doesn't play well with the default props in `Text` component. Previously, they were merged into a single value: 

```css
margin: 0px 0px 16px;
``` 
and now we have two: 
```css
margin-bottom: 16px; 
margin: 0;
```

I saw this problem in `Empty` component, in the text `SHOWING 1 - 1 of 1`.

I can recommend opening `localhost:8080` and your target ( in my case `localhost:3080`) in separate tabs and switching between them with Ctrl-Tab to see the differences.

alex-kovoy:
@rudream [what are you waiting for? Xmass? :-)](https://www.youtube.com/watch?v=vifwar7jxcQ&ab_channel=HappYpupValento)

PR #919, opened 2022-06-24
Add default username for Redis

gzdunek:
Closes https://github.com/gravitational/webapps.e/issues/320 

I was unsure whether I should pass `default` value from the `MenuLogin` picker when it is empty (for redis protocol) or it is better to do have a default value in `connect` in `useDatabases`. IMO the second way is more universal 🤔

### DIFF
--- a/packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/Databases.tsx
+++ b/packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/Databases.tsx
@@ -25,6 +25,7 @@ import { MenuLoginTheme } from '../MenuLoginTheme';
 import { useAppContext } from 'teleterm/ui/appContextProvider';
 import { retryWithRelogin } from 'teleterm/ui/utils';
 import { IAppContext } from 'teleterm/ui/types';
+import { GatewayProtocol } from 'teleterm/ui/services/clusters';
 
 export default function Container() {
   const state = useDatabases();
@@ -56,7 +57,7 @@ function DatabaseList(props: State) {
               <ConnectButton
                 documentUri={props.documentUri}
                 dbUri={db.uri}
-                protocol={db.protocol}
+                protocol={db.protocol as GatewayProtocol}
                 onConnect={dbUser => props.connect(db.uri, dbUser)}
               />
             ),
@@ -77,7 +78,7 @@ function ConnectButton({
 }: {
   documentUri: string;
   dbUri: string;
-  protocol: string;
+  protocol: GatewayProtocol;
   onConnect: (dbUser: string) => void;
 }) {
   const appContext = useAppContext();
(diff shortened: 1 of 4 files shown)


packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/Databases.tsx
```diff
@@ -83,7 +86,8 @@ function ConnectButton({
     <Cell align="right">
       <MenuLoginTheme>
         <MenuLogin
-          placeholder="Enter username"
+          {...getMenuLoginOptions(protocol)}
+          width="195px"
```
ravicious, line 90: As much as I dislike setting a specific width to elements, I can't really think of a better solution here. ;f

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/useDatabases.ts
```diff
@@ -36,12 +36,20 @@ export function useDatabases() {
...
-      targetUser: dbUser,
+      targetUser: getTargetUser(db.protocol, dbUser),
     });
     documentsService.add(doc);
     documentsService.open(doc.uri);
   }
 
+  function getTargetUser(protocol: string, providedDbUser: string): string {
+    if (protocol === 'redis') {
+      return providedDbUser || 'default';
```
ravicious, line 47: Could you mention in a comment that we're replicating [tsh logic](https://github.com/gravitational/teleport/blob/796e37bdbc1cb6e0a93b07115ffefa0e6922c529/tool/tsh/db.go#L240-L244) here but unlike tsh, in Connect each gateway document assumes that the username is present?

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/Databases.tsx
```diff
@@ -71,10 +72,12 @@ function DatabaseList(props: State) {
 function ConnectButton({
   documentUri,
   dbUri,
+  protocol,
   onConnect,
 }: {
   documentUri: string;
   dbUri: string;
+  protocol: string;
```
ravicious, line 80: Could you use `GatewayProtocol` as the type here? https://github.com/gravitational/webapps/blob/79547b2d7399bfa611e3fd1e9dd04452a718ad2b/packages/teleterm/src/services/tshd/types.ts#L13-L23

ravicious (approved):
I'd just add a comment explaining why Redis is getting a special treatment here but other than that it looks fine.

Defining the default username in `useDatabases` seems like a good move to me. 👍

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/useDatabases.ts
gzdunek: Done

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/Databases.tsx
gzdunek: Done

PR #920, opened 2022-06-24
Fall back to leaf cluster id when restoring leaf cluster terminal document

gzdunek:
Closes https://github.com/gravitational/webapps.e/issues/327

### DIFF
--- a/packages/teleterm/src/ui/DocumentTerminal/useDocumentTerminal.ts
+++ b/packages/teleterm/src/ui/DocumentTerminal/useDocumentTerminal.ts
@@ -70,7 +70,9 @@ async function initState(
     const parsed = routing.parseClusterUri(clusterUri);
 
     if (!parsed?.params?.leafClusterId) {
-      throw new Error('Incorrect cluster URI');
+      throw new Error(
+        'The leaf cluster URI was expected, but the URI does not contain the leaf cluster ID'
+      );
     }
     return parsed.params.leafClusterId;
   };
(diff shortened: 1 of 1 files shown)


packages/teleterm/src/ui/DocumentTerminal/useDocumentTerminal.ts
```diff
@@ -48,10 +56,28 @@ async function initState(
...
+    if (cluster) {
+      return cluster.actualName;
+    }
+
+    /*
+     When restoring the documents, we do not always have the leaf clusters already fetched.
+     In that case we can fall back to `clusterId` from a leaf cluster URI
+     (for a leaf cluster `clusterId` === `actualName`)
+    */
+    const parsed = routing.parseClusterUri(clusterUri);
```
ravicious, line 70: Good comment, thanks.

packages/teleterm/src/ui/DocumentTerminal/useDocumentTerminal.ts
```diff
@@ -48,10 +56,28 @@ async function initState(
...
+
+    if (!parsed?.params?.leafClusterId) {
+      throw new Error('Incorrect cluster URI');
```
ravicious, line 73: This error message could provide more information though. :D Maybe something that at this point a leaf cluster URI was expected but the URI doesn't seem to include a leaf cluster ID?

ravicious (approved):
Nice, I'm glad we managed to fix this before the GA release!

PR #925, opened 2022-06-27
Sort connections only when the list opens

gzdunek:
I wanted to add some tests, but I have troubles with mocking the app context :/
I didn't want to spend more time on it.

I updated the icons for disconnect/remove actions, the previous ones were confusing and had too low contrast.  

image


gzdunek:
I realized there is a bug when a new connection is added when the list is open. It just won't show up. I'm wondering what we should do with it, maybe it should be added on the top of the list?

gzdunek:
I think adding new items at the beginning of the list make the most sense. They will just stay in the same place after sorting (reopening the menu).

ravicious (approved):
We might want to add later a comment explaining why we need the connection tracker to be updated only on open.

PR #926, opened 2022-06-27
Update "Connect with GUI" section, add universal context menu

ravicious:
Closes gravitational/webapps.e#304.

---

Since the "Connect with GUI" section now includes the port number, I figured it'd be nice to be able to copy it with a context menu. Before this it was still possible to just copy it with Cmd + C, but I'm not sure how clear this was for the users.

The universal context menu can be overridden by overriding the `contextmenu` event handler.

Since the menu is "universal" and doesn't need any extra dependencies (it's just an array of objects), I decided to define it in `windowsManager.ts` rather than in `packages/teleterm/src/mainProcess/contextMenus`. The latter holds more complex menus that require IPC and some dependencies being passed in to the function.

---

Also, I found out that we should be able to add "Select All" to the terminal context menu so I created an issue for it (gravitational/webapps.e#334).

### DIFF
--- a/packages/teleterm/src/mainProcess/windowsManager.ts
+++ b/packages/teleterm/src/mainProcess/windowsManager.ts
@@ -1,4 +1,4 @@
-import { BrowserWindow, Rectangle, screen } from 'electron';
+import { BrowserWindow, Menu, Rectangle, screen } from 'electron';
 import { getAssetPath } from 'teleterm/mainProcess/runtimeSettings';
 import path from 'path';
 import { FileStorage } from 'teleterm/services/fileStorage';
@@ -43,6 +43,17 @@ export class WindowsManager {
     } else {
       window.loadFile(path.join(__dirname, '../renderer/index.html'));
     }
+
+    // Taken from https://github.com/electron/electron/issues/4068#issuecomment-274159726
+    // selectall was removed from menus because it doesn't make sense in our context.
+    window.webContents.on('context-menu', (_, props) => {
+      const { selectionText, isEditable } = props;
+      if (isEditable) {
+        inputContextMenu.popup({ window });
+      } else if (selectionText && selectionText.trim() !== '') {
+        selectionContextMenu.popup({ window });
+      }
+    });
   }
 
   private saveWindowState(window: BrowserWindow): void {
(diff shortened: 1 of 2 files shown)


packages/teleterm/src/mainProcess/windowsManager.ts
```diff
@@ -43,6 +43,17 @@ export class WindowsManager {
     } else {
       window.loadFile(path.join(__dirname, '../renderer/index.html'));
     }
+
+    // Taken from https://github.com/electron/electron/issues/4068#issuecomment-274159726
+    // selectall was removed from menus because it doesn't make sense in our context.
+    window.webContents.on('context-menu', (_, props) => {
```
gzdunek, line 49: I wasn't aware that is possible to listen to events from the `window` in the `main` process, nice!

From what I see, we usually create methods as event handlers, like this:
```typescript
window.once('close', () => {
  this.saveWindowState(window);
});
```

I'd do the same here and maybe keep `selectionContextMenu` and `inputContextMenu` inside that method?

packages/teleterm/src/ui/DocumentGateway/DocumentGateway.tsx
```diff
@@ -135,14 +135,20 @@ export function DocumentGateway(props: State) {
         Connect with GUI
       </Text>
       <Text>
-        To connect with a GUI database client, see our{' '}
+        To connect with a GUI database client, configure the client to connect
```
gzdunek, line 138: I'm not a fan o this repetition:
image

What would you say for removing the repeated part completely? So it would be just `Configure the client to connect to host localhost on port 64307.` 

I'd also change `See our documentation` to `See the documentation`. Idk, it seems more professional to me :)

ravicious: Yeah, I noticed that repetition too, I think initially the copy was taken from the Web UI. I'll change it to "Configure the GUI database client to connect…".

packages/teleterm/src/mainProcess/windowsManager.ts
ravicious: Ah, that approach with a method makes sense, I didn't notice it.

`selectionContextMenu` and `inputContextMenu` are completely standalone vars that are only needed in the context of this file so I wanted to define them just once. 😏

That being said, it did occur to me that maybe it's not best to do this in the main scope since it calls some Electron code when the file is loaded. I'm going to move them to the constructor so that they're called only when the `WindowsManager` is initialized.

PR #932, opened 2022-06-28
Filter out logins starting with dash

gzdunek:
Closes https://github.com/gravitational/webapps.e/issues/294


ravicious (approved):
I've just realized that we should've probably done that on the level of tshd. But at this point we should probably wait to see what happens with this in tsh.

PR #938, opened 2022-06-29
Add `NotificationContext` to `FeatureHeader`

mcbattirola:
Changed to draft while I address the changes discussed on the weekly update and await to possible copy/design changes

Ignore the text below since I'll update it soon
___

This PR is part of adding license warnings for the web UI (https://github.com/gravitational/cloud/issues/1673).

The objective of this PR is to create a warning message and enable it to be triggered from other parts of the app.

To follow the design we can see [here](https://github.com/gravitational/webapps.e/issues/314), I had to add a new way to send data from the license enforcer (in the webapps.e project) to the `FeatureHeader`. To achieve this, I created a new context that the enforcer (or any component) can write on and the header will read and display.

Here's the PR on `webapps.e` that issues the license warnings: https://github.com/gravitational/webapps.e/pull/341

The `FeatureHeader` component now has a wrapping div, since we need it to wrap both the header and the warning messages, while still applying the margin to both together. This caused many snapshot updates, but since it is just a `Box` (which is just a div), no visual changes are expected.

Here's what the warnings look like:

Screenshot from 2022-06-29 11-08-06

___

Note that this warning is unrelated to the current error messages that we display, and they can both be visible:

Screenshot from 2022-06-29 11-12-40

This doesn't look very good and I think we should change it so all error messages are shown through the `FeatureHeader`. The component already accepts multiple messages, we just have to make the pages pass their errors to it instead of displaying the message themselves.

I didn't do it in this PR tough because it would be too much to review since I'd have to change all pages. Let me know what you think.


packages/teleport/src/NotificationContext.tsx
```diff
@@ -0,0 +1,30 @@
...
+const NotificationContext = createContext<NotificationContextType | null>(null);
+
+// Context Value
+const makeNotificationContextValue = () => {
+  const [notifications, setNotifications] = useState([]);
+
+  const removeNotification = (index: number) => {
+    setNotifications([
+      ...notifications.slice(0, index),
+      ...notifications.slice(index + 1),
```
hatched, line 15: You can reduce these two `slice` calls into a single [splice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice) call. 

(disregard this if you make the architecture change below)

packages/teleport/src/NotificationContext.tsx
```diff
@@ -0,0 +1,30 @@
+import { createContext, useState } from 'react';
+
+// Notification Context
+type NotificationContextType = ReturnType<typeof makeNotificationContextValue>;
+
+const NotificationContext = createContext<NotificationContextType | null>(null);
+
+// Context Value
+const makeNotificationContextValue = () => {
+  const [notifications, setNotifications] = useState([]);
```
hatched, line 10:
```suggestion
  const [notifications, setNotifications] = useState<string[]>([]);
```

(disregard this if you make the architecture change below)

packages/teleport/src/NotificationContext.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+  const removeNotification = (index: number) => {
+    setNotifications([
+      ...notifications.slice(0, index),
+      ...notifications.slice(index + 1),
+    ]);
+  };
+
+  const addNotification = (message: string) => {
+    setNotifications([...notifications, message]);
```
hatched, line 20: I think that `notifications` needs to be something like `Notification[]` instead of `string` type where...

```ts
type Level = "message" | "warning" | "alert";

type Notification = {
  message: string;
  id: number;
  level: Level;
};
```

or even

```ts
type Level = "message" | "warning" | "alert";

type NotificationList = {
  [id: string]: Notification;
};

type Notification = {
  message: string;
  level: Level;
}
```

This is to support having different colours of notification and to be able to remove each notification by ID instead of the index in the list as that index will change over time.

hatched (changes requested):

PR #941, opened 2022-06-29
wrap switchTab in a conditional

avatus:
Closes https://github.com/gravitational/webapps.e/issues/344

Solution:
Check if `allDocuments` has a length before doing any index searching. This will reduce any extra compute if no tabs exist, as well as ensure that `[index].uri` exists when it gets there

### DIFF
--- a/packages/teleterm/src/ui/TabHost/useTabShortcuts.ts
+++ b/packages/teleterm/src/ui/TabHost/useTabShortcuts.ts
@@ -61,16 +61,19 @@ function buildTabsShortcuts(
   const handleTabSwitch = (direction: 'previous' | 'next') => () => {
     const activeDoc = documentService.getActive();
     const allDocuments = documentService.getDocuments();
-    if (allDocuments.length > 0) {
-      const activeDocIndex = allDocuments.indexOf(activeDoc);
-      const getPreviousIndex = () =>
-        (activeDocIndex - 1 + allDocuments.length) % allDocuments.length;
-      const getNextIndex = () => (activeDocIndex + 1) % allDocuments.length;
-      const indexToOpen =
-        direction === 'previous' ? getPreviousIndex() : getNextIndex();
 
-      documentService.open(allDocuments[indexToOpen].uri);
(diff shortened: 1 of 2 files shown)


gzdunek (approved):

packages/teleterm/src/ui/TabHost/useTabShortcuts.ts
```diff
@@ -61,14 +61,16 @@ function buildTabsShortcuts(
...
+      const activeDocIndex = allDocuments.indexOf(activeDoc);
+      const getPreviousIndex = () =>
+        (activeDocIndex - 1 + allDocuments.length) % allDocuments.length;
+      const getNextIndex = () => (activeDocIndex + 1) % allDocuments.length;
+      const indexToOpen =
+        direction === 'previous' ? getPreviousIndex() : getNextIndex();
 
-    documentService.open(allDocuments[indexToOpen].uri);
+      documentService.open(allDocuments[indexToOpen].uri);
+    }
```
ravicious, line 73: For me personally having an early return here would be preferable so that we don't have to nest all the "happy path" code.

```typescript
  const handleTabSwitch = (direction: 'previous' | 'next') => () => {
    const activeDoc = documentService.getActive();
    const allDocuments = documentService.getDocuments();

    if (allDocuments.length === 0) {
      return;
    }

    const activeDocIndex = allDocuments.indexOf(activeDoc);
    const getPreviousIndex = () =>
      (activeDocIndex - 1 + allDocuments.length) % allDocuments.length;
    const getNextIndex = () => (activeDocIndex + 1) % allDocuments.length;
    const indexToOpen =
      direction === 'previous' ? getPreviousIndex() : getNextIndex();

    documentService.open(allDocuments[indexToOpen].uri);
  };
```

ravicious, line 73: Also, could you add tests for this behavior? `useTabShortcuts` has a test suite so it should be pretty straightforward.

ravicious (approved):

PR #942, opened 2022-06-29
Discover wizard screen POC + boilerplating

kimlisa:
#### Description
- Created a plugin like flow to hopefully make it easy to work on individual steps and customizing steps for different agents. Open to suggestions.
- The feature is turned off by default, so in order to turn it on go to `teleport/src/config.ts` and set the flag `cfg.discoverWizard` to true.
- The feature can be accessed by route specifically or by going to `dashboard > click on your name (top right corner) > click Discovery Wizard`

Nothing is set in stone and needs more polish. Most components are just placeholders and demo.

#### Recap on terminology (`agents` vs `resources`)
In webapps, we use the term `resources` and `agents`. Before `agents` came along, we already had a type [Resource](https://github.com/gravitational/webapps/blob/master/packages/teleport/src/services/resources/types.ts#L17) that referred to the types defined in the link (roles, auth connectors, trusted clusters). 

Then came along [Agents](https://github.com/gravitational/webapps/blob/master/packages/teleport/src/services/agents/types.ts#L23) which was a way to separate the types defined with `Resource` with other resources: apps, dbs, nodes, desktops, and kubes.

So we want to refer to resources like `apps, dbs, nodes, desktops, and kubes` as `agents` where we can. When we use the term `resources` it can mean all inclusive, or the other three depending on context. 

#### Demo
discover-poc


packages/teleport/src/Discover/discoverContextProvider.tsx
```diff
@@ -0,0 +1,39 @@
...
+  value: DiscoverContext;
+}> = props => {
+  return <DiscoverReactContext.Provider {...props} />;
+};
+
+export default DiscoverContextProvider;
+
+export function useDiscoverContext() {
+  const ctx = React.useContext(DiscoverReactContext);
+  if (!ctx) {
```
ryanclark, line 32: I _think_ this won't be reached - if `useDiscoverContext` is used without the context provider being in the parent tree, `React.useContext(DiscoverReactContext)` would throw the error for us

packages/teleport/src/Discover/AgentConnect.tsx
```diff
@@ -0,0 +1,221 @@
...
+import styled from 'styled-components';
+import { Flex, ButtonPrimary, ButtonSecondary, Text, Box } from 'design';
+import * as Icons from 'design/Icon';
+import { State, AgentKind } from './useDiscover';
+
+// agentStepTitles defines the titles per steps defined by enum `AgentStep`.
+//
+// We use the enum `AgentStep` numerical values to access the list's value,
+// so this list's order and length must be equal to the enum.
+export const agentStepTitles: string[] = [
```
ryanclark, line 27: nit pick: no need to specify the type here as it's inferred

hatched (approved):
After our discussion offline we can land this as-is and then make the updates as needed as we move forward 👍

PR #951, opened 2022-07-05
Add directory sharing to the ACL

ibeckermayer:
The setting is also modulated by the `enableDirectorySharing` config variable while the feature is still in development.

Corresponds with https://github.com/gravitational/teleport/pull/14103

v9 and v10 backport required

### DIFF
--- a/packages/teleport/src/config.ts
+++ b/packages/teleport/src/config.ts
@@ -28,6 +28,8 @@ import { RecordingType } from 'teleport/services/recordings';
 import generateResourcePath from './generateResourcePath';
 
 const cfg = {
+  // TODO(isaiah): remove after feature is finished.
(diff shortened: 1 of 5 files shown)


kimlisa (approved):

packages/teleport/src/config.ts
```diff
@@ -28,6 +28,8 @@ import { RecordingType } from 'teleport/services/recordings';
 import generateResourcePath from './generateResourcePath';
 
 const cfg = {
+  // TODO(isaiah): remove after feature is finished.
+  enableDirectorySharing: false, // note to reviewers: should be false in any PRs.
```
hatched, line 32: Nice note! 👍

hatched (approved):

PR #952, opened 2022-07-05
Directory sharing menu item

ibeckermayer:
Adds a new "Share Directory" item to the dropdown menu of a Desktop Session's top bar (when RBAC enabled)

image

Also refactors the top bar to push the (new) "directory sharing in progress", clipboard sharing enabled, and recording enabled icons to the right, next to the menu button ("...")

image

This is just the UI scaffolding, it's not meant to do anything yet except cause the UI to properly change according to internal state changes.

Requires backport to v9/v10


packages/teleport/src/DesktopSession/ActionMenu.tsx
```diff
@@ -41,6 +47,9 @@ export default function ActionMenu(props: Props) {
 }
 
 type Props = {
+  canShareDirectory: boolean;
+  isSharingDirectory: boolean;
+  onShareDirectory: VoidFunction;
```
hatched, line 52: As these three props aren't ever used in isolation I feel like you can simplify this logic by combining the two boolean values into one `shouldShowDirectoryShare` and then pass in the callback as well.

zmb3 (approved):

packages/design/src/Icon/Icon.jsx
```diff
@@ -151,6 +151,8 @@ export const Expand = makeFontIcon('Expand', 'icon-frame-expand');
 export const Facebook = makeFontIcon('Facebook', 'icon-facebook');
 export const FacebookSquare = makeFontIcon('FacebookSquare', 'icon-facebook2');
 export const FileCode = makeFontIcon('Youtube', 'icon-file-code');
+export const FolderPlus = makeFontIcon('FolderPlus', 'icon-folder-plus');
```
kimlisa, line 154: could we also add to story, i always look at the story `Design/Icon` to help me see what we already have

packages/teleport/src/lib/tdp/client.ts
```diff
@@ -39,11 +39,10 @@ export enum TdpClientEvent {
 // calling Client.nuke() (typically after Client emits a TdpClientEvent.DISCONNECT or TdpClientEvent.ERROR event) in order to clean
 // up its websocket listeners.
 export default class Client extends EventEmitterWebAuthnSender {
-  codec: Codec;
-  socket: WebSocket;
-  socketAddr: string;
-  username: string;
-  logger = Logger.create('TDPClient');
+  protected codec: Codec;
+  protected socket?: WebSocket;
```
kimlisa, line 43: what does the question mark do here?

packages/teleport/src/lib/tdp/client.ts
```diff
@@ -184,33 +183,33 @@ export default class Client extends EventEmitterWebAuthnSender {
   }
 
   sendMouseMove(x: number, y: number) {
-    this.socket.send(this.codec.encodeMouseMove(x, y));
+    this.socket?.send(this.codec.encodeMouseMove(x, y));
```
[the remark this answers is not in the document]
kimlisa: i also like not failling silently

packages/teleport/src/DesktopSession/TopBar.tsx
```diff
@@ -79,6 +103,12 @@ const StyledClipboard = styled(Clipboard)`
   align-self: 'center';
 `;
 
+const StyledFolderShared = styled(FolderShared)`
+  font-weight: ${({ theme }) => theme.fontWeights.bold};
+  font-size: ${({ theme }) => theme.fontSizes[6] + 'px'};
```
kimlisa, line 108: i like this, we should be doing this more to keep in line with our theme @rudream

packages/teleport/src/lib/tdp/client.ts
ibeckermayer: I had assumed it meant optional (like in interfaces or types), equivalent to:

```typescript
protected socket: WebSocket | undefined;
```

However it turns out that in order for it to be interpreted that way, we need to add `strictNullChecks` to our `tsconfig.json`, which then causes a ton of other errors elsewhere in the codebase. I think its a good option to turn on, because it prevents us from using any property on a `null` or `undefined`, so I've added an issue for it: https://github.com/gravitational/webapps/issues/957.

I've made it explicit for now, like in the example above.

packages/teleport/src/lib/tdp/client.ts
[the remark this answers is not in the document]
ibeckermayer: Is `#_` a shorthand for private?

packages/teleport/src/lib/tdp/client.ts
```diff
@@ -178,47 +178,58 @@ export default class Client extends EventEmitterWebAuthnSender {
...
 
+  protected send(
+    data: string | ArrayBufferLike | Blob | ArrayBufferView
+  ): void {
+    if (this.socket && this.socket.readyState === 1) {
+      this.socket.send(data);
+    } else {
+      this.handleError(new Error('websocket unavailable'));
+      return;
+    }
```
kimlisa, line 189: won't allow me to edit code 

```ts
if (this.socket && this.socket.readyState === 1) {
      this.socket.send(data);
      return;
}

this.handleError(new Error('websocket unavailable'));
```

kimlisa (approved):

packages/teleport/src/lib/tdp/client.ts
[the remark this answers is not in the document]
hatched: Sorry that's just me mixing old and new conventions.

`#` is the new syntax for [private class methods](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields#private_methods). Before you could make real private methods any method/property starting with an underscore would indicate that it's a private method and that you shouldn't use it directly. So I kind of adopted that abomination of a prefix 😁

Issue #887, opened 2022-06-14
Prettier fixes


michellescripts:
**What**
- add a pre-commit hook to check changed files for prettier formatting (i.e., prevent commits with improper formatting)
- see if there are any files with existing errors that can be fixed without introducing errors
- re-run snapshots for tests that break (likely to see this with design files that are missing leading zeros)

**Why**
- clean code, enforced!

ravicious:
> add a pre-commit hook to check changed files for prettier formatting (i.e., prevent commits with improper formatting)

What do you think about adding a linter rule for that instead of a pre-commit hook? The linter must be satisfied before merging a PR to the main branch. Most of the time we do squash and merge, so this would ensure that the tip of the main branch is formatted correctly.

IMHO, pre-commit hooks are a rather invasive way to take care of this. I wouldn't want to be prevented from making a commit by a problem that can be fixed after the fact (before the PR is merged).

PR #955, opened 2022-07-06
[2/3] Prettify shared, build, and teleport packages

kimlisa:
part of https://github.com/gravitational/webapps/issues/887

Suggest looking by commit

Ran `yarn test; yarn tsc`, no snapshot updates needed


packages/teleport/src/Player/ProgressBar/Slider/Slider.jsx
```diff
@@ -625,7 +650,9 @@ THE SOFTWARE.
     },
 
     _getValueFromPosition: function (position) {
-      var diffValue = position / (this.state.sliderLength - this.state.handleSize) * (this.props.max - this.props.min);
+      var diffValue =
+        (position / (this.state.sliderLength - this.state.handleSize)) *
+        (this.props.max - this.props.min);
```
kimlisa, line 655: i checked this was the same (also i learned that multiplication and division is whichever comes first not necessarily PEMDAS)

.prettierignore
```diff
@@ -1,2 +1,7 @@
 # assets contains pre-generated files like iconmoon, fonts, svgs.
 packages/design/src/assets/*
+
+# Ignore file types:
+packages/**/*.css
+packages/**/*.json
+packages/**/*.yaml
```
ibeckermayer, line 7: Why do we want to ignore these?

ibeckermayer (approved):
See my inline question

.prettierignore
kimlisa: we have huge json files used for fixtures that doesn't benefit from formatting, and it also slowed down vscode when viewing the file formatted vs when it wasn't

we also have like yaml template files (like auth connectors, trusted cluster etc), it was adding extra spaces so figured we leave those alone

we probably don't need css here since i ignored the entire directory that contained the pre generated css files from like icomoon, but we can remove it later if we want

PR #956, opened 2022-07-06
Prettify teleterm package

kimlisa:
part of https://github.com/gravitational/webapps/issues/887

#### Question
I thought teleterm package to have the least amount of prettying to do because it's the newest, so i'm curious if our prettier setup is just different?


package.json
```diff
@@ -19,9 +19,11 @@
     "build-teleport-e": "yarn workspace @gravitational/teleport.e build --output-path=../../../dist/e/teleport/app",
     "build-oss": "yarn build-teleport-oss",
     "build-e": "yarn build-teleport-e",
-    "type-check-and-lint": "yarn type-check && yarn lint",
+    "type-check-and-lint": "yarn type-check && yarn prettier-check && yarn lint",
```
kimlisa, line 22: i figured prettier fell under the `lint` category so I put it here and i also saw that our CI runs the [type-check-and-lint](https://github.com/gravitational/webapps/blob/master/.cloudbuild/ci/build.yaml#L21), i'm not sure if this means it will run prettier now? (i need to merge the PRs that this PR bases from to see if it can be applied to this PR, like add a file to deliberately fail the check to test it)

gzdunek (approved):

ravicious (approved):

package.json
```diff
@@ -19,9 +19,11 @@
...
     "lint": "eslint --ext .js,.jsx,.ts,.tsx packages/",
-    "type-check": "tsc --noEmit"
+    "type-check": "tsc --noEmit",
+    "prettier-check": "yarn prettier --check packages/",
+    "prettier-write": "yarn prettier --write packages/"
```
ibeckermayer, line 26: We can be explicit about which types of files we want to prettify like

```suggestion
    "prettier-check": "yarn prettier --check 'packages/**/*.{ts,tsx,js,jsx,md}'",
    "prettier-write": "yarn prettier --write 'packages/**/*.{ts,tsx,js,jsx,md}'"
```

ibeckermayer (changes requested):

.prettierrc
```diff
@@ -2,9 +2,9 @@
   "arrowParens": "avoid",
   "printWidth": 80,
   "bracketSpacing": true,
-  "jsxBracketSameLine": false,
+  "bracketSameLine": false,
```
michellescripts, line 5: 🎉

Issue #959, opened 2022-07-06
Handle directory sharing in non-chrome browsers


ibeckermayer:

zmb3:
How will this work?

ibeckermayer:
I just mean to show an error message to the user if they attempt to start file sharing in an unsupported browser.

Thinking about it in more detail now -- we may want to make this error a dismissible modal, so that we're not abruptly ending their session.

Issue #958, opened 2022-07-06
Handle directory sharing initialization before login completes


ibeckermayer:
Currently, if a user attempts to share a directory before login is completed, the system fails silently (and they are unable to retry, because the internal frontend state is set as if the process succeeded, and prevents them from re-sharing).

Figure out what is going on under the hood and come up with a better solution.

Relates to https://github.com/gravitational/teleport/issues/8583

PR #960, opened 2022-07-06
`SharedDirectoryAnnounce`

ibeckermayer:
## UX
Allows the user to select "Share Directory" from the dropdown
image

And choose a directory to share
image

If necessary, the user will be prompted to allow access to the directory

image

At this point, the session will terminate because we will receive an unimplemented TDP message.

### Couldn't find a permission name

Unfortunately, unlike the clipboard feature, I wasn't able to find a permission name to query for this permission preemptively
https://github.com/gravitational/webapps/blob/bc6586b32cc8eb7ec6ae25e1d125d6b5d907bfb4/packages/teleport/src/DesktopSession/useClipboard.ts#L61

The list of permission names for Chrome is linked by MDN's documentation to [here](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/third_party/blink/renderer/modules/permissions/permission_descriptor.idl), but none of these are what we are looking for.

### Mysterious DOMException
We will also occasionally get a `Uncaught (in promise) DOMException: Document is not focused.` error when this permission modal pops up, or maybe more accurately, once `sharedDirHandle` is initialized:
https://github.com/gravitational/webapps/blob/bc6586b32cc8eb7ec6ae25e1d125d6b5d907bfb4/packages/teleport/src/DesktopSession/DesktopSession.tsx#L162-L171

It's not entirely clear to me why this happens -- I've struggled to find a way to reproduce it consistently, though denying the permission in the modal and then asking for it again tends to work _most_ of the time. [There is limited information on Google about it](https://www.google.com/search?q=Uncaught+(in+promise)+DOMException%3A+Document+is+not+focused.+file+access+api&oq=Uncaught+(in+promise)+DOMException%3A+Document+is+not+focused.+file+access+api&aqs=chrome..69i64j69i57.795j0j9&sourceid=chrome&ie=UTF-8). Most of the information relates to the clipboard API, which we dealt with here by checking for focus:
https://github.com/gravitational/webapps/blob/bc6586b32cc8eb7ec6ae25e1d125d6b5d907bfb4/packages/teleport/src/DesktopSession/useTdpClientCanvas.tsx#L150-L159

However even if I try adding such a check like so
```typescript
        onShareDirectory={() => {
          window
            .showDirectoryPicker()
            .then(sharedDirHandle => {
              console.log('in then');
              if (document.hasFocus()) {
                setIsSharingDirectory(true);
                tdpClient.sharedDirectory = sharedDirHandle;
                tdpClient.sendSharedDirectoryAnnounce();
              } else {
                throw new Error(
                  'Cannot handle directory while document does not have focus'
                );
              }
            })
            .catch(err => {
              console.log('in catch');
              console.log('err = ' + err);
              setIsSharingDirectory(false);
            });
        }}
```

I still encounter the error. The odd thing is, without the check, everything still seems to run according to plan, even if the error occurs. With the check, the error occurs but then `tdpClient.sendSharedDirectoryAnnounce();` never gets called. 

As far as I can tell, the worst thing that will happen by letting this error slide is that the console will have some uncaught exceptions. It could just be a function of this being a bleeding edge feature.

### Other known issues
- This feature will only be supported in Chromium based browsers. For now I am ignoring other browsers: https://github.com/gravitational/webapps/issues/959
- https://github.com/gravitational/webapps/issues/958

### Opportunistically
Also removes the recording symbol, closes https://github.com/gravitational/teleport/issues/13973

#### `teleport`
Build and run the `windows-desktop-directory-sharing` branch, editing [this line](https://github.com/gravitational/teleport/blob/15439227ea32be084cf5723ceed0aae72e79dc7a/Makefile#L232) in the Makefile to have the `directory_sharing` build flag:
```
$(BUILDDIR)/teleport: ensure-webassets bpf-bytecode rdpclient
	GOOS=$(OS) GOARCH=$(ARCH) $(CGOFLAG) go build -tags "$(PAM_TAG) $(FIPS_TAG) $(BPF_TAG) $(WEBASSETS_TAG) $(RDPCLIENT_TAG) directory_sharing" -o $(BUILDDIR)/teleport $(BUILDFLAGS) ./tool/teleport
```

#### `webapps`
Change the value below to `true` and run the dev server
https://github.com/gravitational/webapps/blob/becfad38776a68e0d8be634476f24cead7890ffc/packages/teleport/src/config.ts#L32

Requires backport to v9/v10


ravicious:
Is there a cluster I could connect to to try this out?

packages/teleport/src/DesktopSession/DesktopSession.tsx
```diff
@@ -164,6 +164,11 @@ function Session(props: PropsWithChildren<State>) {
         isSharingDirectory={isSharingDirectory}
         onShareDirectory={() => {
           setIsSharingDirectory(true);
+          const sharedDirHandle = window
+            .showDirectoryPicker()
```
ravicious, line 163: Do we need to handle browsers which don't support this API?

packages/teleport/src/config.ts
```diff
@@ -29,7 +29,7 @@ import generateResourcePath from './generateResourcePath';
 
 const cfg = {
   // TODO(isaiah): remove after feature is finished.
-  enableDirectorySharing: false, // note to reviewers: should be false in any PRs.
+  enableDirectorySharing: true, // note to reviewers: should be false in any PRs.
```
ravicious: This feels like one of those [logic puzzles](https://en.wikipedia.org/wiki/The_Hardest_Logic_Puzzle_Ever)… 🤔

But in all seriousness, is it supposed to be true or false in the end? 😅

packages/teleport/package.json
```diff
@@ -24,6 +24,7 @@
   },
   "devDependencies": {
     "@gravitational/build": "^1.0.0",
+    "@types/wicg-native-file-system": "^2020.6.0",
```
ravicious, line 27: Much better! 👍

ravicious (approved):
A preemptive approval with two caveats:

1) I didn't have a chance to actually test this.
2) I don't know what our requirements are when it comes to browser support for desktop access. If we run some checks before the execution gets to `window.showDirectoryPicker()` then it's fine. But if not then we should add some feature checks.

packages/teleport/src/DesktopSession/DesktopSession.tsx
```diff
@@ -153,17 +151,24 @@ function Session(props: PropsWithChildren<State>) {
...
           tdpClient.nuke();
         }}
         userHost={`${username}@${hostname}`}
         clipboardSharingEnabled={clipboardSharingActive}
-        isRecording={isRecording}
         canShareDirectory={canShareDirectory}
         isSharingDirectory={isSharingDirectory}
         onShareDirectory={() => {
-          setIsSharingDirectory(true);
+          window
```
hatched, line 162: Should probably add a check for the existence of `window.showDirectoryPicker` so that this doesn't throw on unsupported browsers. 

And as a little nitpick, I don't like defining functions within the render portion of a component as it bloats the markup and makes it harder to follow.

packages/teleport/src/lib/tdp/codec.ts
```diff
@@ -389,6 +397,31 @@ export default class Codec {
     return buffer;
   }
 
+  // | message type (11) | completion_id uint32 | directory_id uint32 | name_length uint32 | name []byte |
+  encodeSharedDirectoryAnnounce(
```
hatched, line 401: As discussed offline, it would be nice if we could add a test for this but as this is still in flux we can just make sure to revisit after the api's have stabilized.

packages/teleport/src/DesktopSession/DesktopSession.tsx
ibeckermayer: Yeah, I left that as a "known issue" for now: https://github.com/gravitational/webapps/issues/959

packages/teleport/src/DesktopSession/DesktopSession.tsx
ibeckermayer: > Should probably add a check for the existence of window.showDirectoryPicker so that this doesn't throw on unsupported browsers.

I left that as a "known issue" for now: https://github.com/gravitational/webapps/issues/959, since we will want to add some error handling and UI message, but I am just focusing on getting the core functionality working.

packages/teleport/src/config.ts
ibeckermayer: Haha

You got it correct, its supposed to be `false` in the end. Fixed.

kimlisa (approved):
> We will also occasionally get a Uncaught (in promise) DOMException: Document is not focused. error when this permission modal pops up, or maybe more accurately, once sharedDirHandle is initialized

not sure what the error is without manually testing it but did you try focusing before calling the func?

ibeckermayer:
> ```ts
> onShareDirectory
> ```

No I hadn't tried that. I just tried

```typescript
const onShareDirectory = () => {
  window.focus();
  ...
};
```
and 
```typescript
const onShareDirectory = () => {
  document.getElementById('app').focus();
  ...
};
```

to no avail

hatched (approved):

ravicious:
Re: "Document is not focused" error, it's interesting that it's reported as

```
Uncaught (in promise) DOMException: Document is not focused.
```

This code doesn't seem to be what's triggering it as the event handler function itself isn't executed within a promise. The promise returned by `showDirectoryPicker` has a `catch` so it can't be the source of it as well.

https://github.com/gravitational/webapps/blob/bc6586b32cc8eb7ec6ae25e1d125d6b5d907bfb4/packages/teleport/src/DesktopSession/DesktopSession.tsx#L161-L172

Matter of fact, when I changed the event handler to this:

```typescript
        onShareDirectory={() =>
          Promise.resolve()
            .then(() => window.showDirectoryPicker())
            .then(sharedDirHandle => {
              setIsSharingDirectory(true);
              tdpClient.sharedDirectory = sharedDirHandle;
              tdpClient.sendSharedDirectoryAnnounce();
            })
            .catch(e => {
              console.log('onShareDirectory error', { e });
              setIsSharingDirectory(false);
            })
        }
```

the error was still reported as "Uncaught (in promise)". 🤔

packages/teleport/src/DesktopSession/DesktopSession.tsx
ibeckermayer: Basically I'm trying to charge forward with the core implementation in browsers that actually support it, and then go back at the end and write the code that makes sure the UX isn't totally confusing in browsers that don't.

ravicious: FWIW, I tried it with Firefox and I wasn't able to get to the point where the UI would allow me to trigger this code. The Web UI blocked me from using remote desktops and showed this message:

> Your user role supports clipboard sharing over desktop access, however this feature is only available on chromium based browsers like Brave or Google Chrome. Please switch to a supported browser.

It didn't start the session so there was no risk of this code crashing.

PR #968, opened 2022-07-08
Add SFTP audit events

capnspacehook:
Add support for SFTP audit events from https://github.com/gravitational/teleport/pull/13491.

Example of new events:

image


kimlisa:
hi @capnspacehook could you also screenshot the added events, b/c designs also need review

capnspacehook:
@kimlisa added screenshot to description

packages/teleport/src/services/audit/types.ts
```diff
@@ -249,7 +286,258 @@ export type RawEvents = {
...
+      path: string;
+      ['addr.local']: string;
+    }
+  >;
+  [eventCodes.SFTP_SYMLINK]: RawEvent<
+    typeof eventCodes.SFTP_SYMLINK,
+    {
+      path: string;
+      ['addr.local']: string;
+    }
```
kimlisa, line 532: since `sftpXXX` has common fields we can create a type just for this event so we don't define these repeatedly
```
type RawEventSftp<T extends EventCode> = RawEvent<
  T,
  {
    path: string;
    ['addr.local']: string;
  }
>;
```

kimlisa (approved):

nklaassen (approved):
I really hope most of these files are generated. Not quite sure why I got added on this but LGTM

capnspacehook:
@kimlisa would you mind taking a quick look at this again, I added logic to display SFTP errors in audit events. I forgot to add an error field to the SFTP audit events when I originally opened this PR, but the change has been merged into Teleport now.

packages/teleport/src/Audit/__snapshots__/Audit.story.test.tsx.snap
```diff
@@ -4121,7 +4121,7 @@ exports[`list of all events 1`] = `
         <td
           style="word-break: break-word;"
         >
-          User [root] failed to query attributes of file [/tmp/file] on node [[::1]:3022]
+          User [root] failed to query attributes of file [/tmp/file] on node [[::1]:3022]: [EOF]
```
kimlisa, line 4124: it looks a bit weird to wrap error messages in a bracket, especially since we have a `:` separator to mean anything coming after it is a error message. i don't think there is a hard rule that every param needs to be in a bracket (it just so happens we do have pretty much everything wrapped in bracket but i think that's to make it like a visual separator from other words)

kimlisa (approved):
minor comment

packages/teleport/src/Audit/__snapshots__/Audit.story.test.tsx.snap
capnspacehook: That makes sense, I originally wrapped the error in brackets because I noticed that's what SCP events did as well. I can change that, but if I do there should probably be a follow up PR that makes formatting consistent between all failure events.

kimlisa: ah ok, i missed that, lets just leave it as is then, it's not that big of a deal

PR #969, opened 2022-07-11
Remove EOT fonts (IE8 support)

gzdunek:
For some reason, EOT fonts can't be loaded by Webpack on Windows and cause `Connect` build failure. 
I could debug this, but it doesn't make sense as these fonts are needed only for IE8 support https://caniuse.com/eot (IE9+ supports WOFF).

After merging this, I will uncheck IE8 support from IcoMoon:
image


ravicious (approved):

hatched (approved):

kimlisa (approved):
this is nice (less files :tada:)

PR #970, opened 2022-07-11
Add an event for when a session recording is accessed

ryanclark:
As part of https://github.com/gravitational/teleport/issues/13880, a new event has been added (`SESSION_RECORDING_ACCESS`, code `T2011I`).

This adds it to the frontend so it displays correctly -
image

image


packages/teleport/src/services/audit/makeEvent.ts
```diff
@@ -680,6 +680,13 @@ export const formatters: Formatters = {
       return `Certificate of type [${cert_type}] issued for [${user}]`;
     },
   },
+  [eventCodes.SESSION_RECORDING_VIEW]: {
+    type: 'session.recording.view',
+    desc: 'Session Recording Viewed',
+    format: ({ sid, user }) => {
+      return `User [${user}] viewed a session recording [${sid}]`;
```
zmb3, line 687: I feel like "accessed" a session recording is a better term than "viewed".

(You can download sessions via tsh, for example)

ryanclark: Ah yes I was going to rename the event as per your other PR review comment - I'll change the wording too

packages/teleport/src/Audit/__snapshots__/Audit.story.test.tsx.snap
```diff
@@ -2636,13 +2636,13 @@ exports[`list of all events 1`] = `
               color="light"
               font-size="3"
             />
-            App Session Started
```
ibeckermayer, line 2639: Confused as to why this would be removed given this change.

ibeckermayer: It looks like `APP_SESSION_START` has the same event code:
https://github.com/gravitational/webapps/blob/723fc13fe0184a9e67bbb6893232f0a55c136254/packages/teleport/src/services/audit/types.ts#L49

kimlisa (commented):
i think it's worth to add instructions on how to add events on our root README, it's a very common addition (and various devs come in to add them) and the steps are all the same as well as the files touched, see example here: https://github.com/gravitational/webapps/pull/561,

the instructions are also detailed in this file: https://github.com/gravitational/webapps/blob/6f1d453ab4e94b73095c2d2cea732e02b01bfd4c/packages/teleport/src/services/audit/types.ts#L30, but i guess it's hard to see

strideynet (approved):

zmb3:
@ryanclark is there any work left here? We can merge this at any time (backend doesn't need to be ready)

ryanclark:
> @ryanclark is there any work left here? We can merge this at any time (backend doesn't need to be ready)

@zmb3 Yeah it's ready to go, I'll sort it out in a couple of hours once I'm at my MacBook

PR #971, opened 2022-07-11
Add Windows support for Connect

gzdunek:
Closes https://github.com/gravitational/webapps.e/issues/196 
Tshd counterpart: https://github.com/gravitational/teleport/pull/14301

Best to review commit-by-commit (although the first one is quite big :/)

The main change is the use of TCP with mTLS instead of UDS. It requires from us to generate the certs when the app launches. I'm not sure if the high-level `mkcert` library is the best option, it's not really popular, however its source code look good. Another option would be to use low-level library `node-forge` (`mkcert` uses it) and just have all this logic in our codebase). 

~~The certs are generated regardless of the platform. We can change it (Mac/Linux use UDS, so it doesn't need encryption), but that will make the code more complicated, so probably I would leave it as it is.~~

Updated certs generation process: https://github.com/gravitational/webapps/pull/971#discussion_r926630971

The icon I added looks the same as for Mac, it only doesn't have a margin around it.


ravicious (commented):
I didn't make it past the first commit and some other stuff that caught my eye while skimming through the commit list, I'll continue the review tomorrow.

packages/teleterm/src/mainProcess/runtimeSettings.ts
```diff
@@ -101,18 +104,73 @@ function getAssetPath(...paths: string[]): string {
...
+async function getAvailablePort(): Promise<number> {
+  const server = net.createServer();
+  return new Promise(resolve => {
+    // OS will find a free port automatically
+    server.listen({ port: 0 }, () => {
+      const { port } = server.address() as net.AddressInfo;
+      server.close(() => {
+        resolve(port);
+      });
+    });
```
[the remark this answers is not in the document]
ravicious: I saw a couple of answers on SO that suggested using a command for that on Windows (`netstat` I think?). The downside is that the output might be different depending on the system language of the user. A Powershell command has also been mentioned which might be more reliable.

packages/teleterm/package.json
```diff
@@ -46,6 +46,7 @@
     "file-loader": "6.0.0",
     "google-protobuf": "^3.20.1",
     "immer": "^9.0.7",
+    "mkcert": "^1.5.0",
```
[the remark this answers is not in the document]
gzdunek: Yes, I mentioned in the PR description that we may want to keep the code for generating certs in our code base. I will do it.

packages/teleterm/src/main.ts
```diff
@@ -9,103 +9,123 @@ import { ConfigServiceImpl } from 'teleterm/services/config';
...
-      }
+  app.on(
+    'certificate-error',
+    (event, webContents, url, error, certificate, callback) => {
+      // allow certs errors for localhost:8080
       if (
-        url.host === 'github.com' &&
-        url.pathname.startsWith('/gravitational/')
+        settings.dev &&
+        url.includes('://localhost:8080/') &&
```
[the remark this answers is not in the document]
gzdunek: Thanks, I don't why I chose such weird way to check it ;)

packages/teleterm/src/mainProcess/runtimeSettings.ts
```diff
@@ -16,19 +17,28 @@ const dev = env.NODE_ENV === 'development' || env.DEBUG_PROD === 'true';
...
     homeDir: getTshHomeDir(),
-    networkAddr: tshNetworkAddr,
-    flags: ['daemon', 'start', `--addr=${tshNetworkAddr}`],
+    networkAddr: tshAddress,
+    flags: [
+      'daemon',
+      'start',
+      `--addr=${
+        // grpc-js requires us to pass localhost:port for TCP connections,
+        // for tshd we have to specify the protocol as well protocol
```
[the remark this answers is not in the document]
gzdunek: Done

packages/teleterm/src/mainProcess/runtimeSettings.ts
```diff
@@ -16,19 +17,28 @@ const dev = env.NODE_ENV === 'development' || env.DEBUG_PROD === 'true';
...
+        tshAddress.startsWith('unix') ? tshAddress : 'tcp://' + tshAddress
+      }`,
```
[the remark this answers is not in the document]
gzdunek: Looks better, done

packages/teleterm/src/mainProcess/runtimeSettings.ts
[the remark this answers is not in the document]
gzdunek: I considered doing this like you wrote, but unfortunately it will increase the complexity of establishing the connections significantly. 
1. Currently, both the UDS and TCP connections are created in the same way, so both clients just read addresses from `runtimeSettings` and connect to them. After changing this, we would have to use different logic for both types of connections.
2. Checking the port by PID requires spawning a new process (and probably it is not a solution, I guess that `netstat` returns all ports that a process use?)
3. The option that would work is to return ports from processes via stdout, but it requires looking for some string in process output and blocking the creation of clients until we get the correct addresses. 

I was curious how the OS assigns port numbers, so I performed quick test. I invoked `getAvailablePort()` and then wait 300 ms, 10 times. The ports I got were not repeated, OS assigned a new port for each connection, event the previous one was already closed. 
This doesn't give 100% confidence, for sure some apps do not relay on OS, instead they just randomly create the port number.

It would be really good to have this done right, but it requires quite a few changes and I feel like this problem is unlikely to happen... I don't know, @zmb3 do you have any opinion?

packages/teleterm/package.json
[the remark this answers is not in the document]
ravicious: Ah sorry, I completely missed it. 🤦

packages/teleterm/src/ui/DocumentGateway/DocumentGateway.tsx
```diff
@@ -161,6 +164,23 @@ export function DocumentGateway(props: State) {
...
+    gateway.localAddress === 'localhost'
+  ) {
+    return {
+      cliCommand: gateway.cliCommand.replace('localhost', '127.0.0.1'),
+      localAddress: '127.0.0.1',
+    };
+  }
+  return { cliCommand: gateway.cliCommand, localAddress: gateway.localAddress };
+}
+
```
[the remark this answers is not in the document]
ravicious: > (…) but maybe I will remove that commit from the PR? It's not necessary for the Windows support.

Sounds fine.

packages/teleterm/src/mainProcess/runtimeSettings.ts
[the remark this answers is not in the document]
ravicious: I said this in DMs between us but I'd rather do it properly rather than keep our fingers crossed each time someone launches the app that everything goes okay. ;)

I'm not sure though what's the best way for a program to advertise what port it occupies. Using a system like Bonjour is probably an overkill in this situation. Reading from stdout of the process seems alright but idk if there's a better way to do it. Those processes will always be launched by our app only and never by the user, so we can even emit the port as a specific string like `CONNECT_GRPC_PORT: 1337`.

---

When it comes to changes necessary in our startup logic: the network addresses are supplied in the runtime settings right now and the runtime settings is the first thing we load when we start the app. Changing this logic would be in fact quite cumbersome.

However, what if instead we just changed the logic that sets up the clients for tshd and shared process? `MainProcess.create` would need to become an async function that returns the ports used by gRPC servers. Then we'd need to change only the code which starts the clients rather than changing all code which reads the runtime settings.

packages/teleterm/src/sharedProcess/ptyHost/ptyProcess.ts
```diff
@@ -178,5 +178,7 @@ async function getWorkingDirectory(pid: number): Promise<string> {
     case 'linux':
       const asyncReadlink = promisify(readlink);
       return await asyncReadlink(`/proc/${pid}/cwd`);
+    case 'win32':
+      return undefined;
```
ravicious, line 182: Did you try to implement it for Windows? What problems did you run into?

packages/teleterm/src/mainProcess/windowsManager.ts
```diff
@@ -38,6 +38,7 @@ export class WindowsManager {
       backgroundColor: theme.colors.primary.darker,
       minWidth: 400,
       minHeight: 300,
+      autoHideMenuBar: true,
```
ravicious, line 41: This one is interesting though, I wonder if Windows users usually expect that bar to be present?

packages/teleterm/src/mainProcess/runtimeSettings.ts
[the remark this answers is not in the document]
gzdunek: Ok, I'm thinking about this again now, and it seems that we can do it in a way where there is no difference between UDS and TCP. Processes can return whole addresses like `tcp://host:port` or `unix://path` which we will provide to clients. The idea with changing `MainProcess.create` to be async makes sense, I'll work on that. Thanks for your comments!

packages/teleterm/src/sharedProcess/ptyHost/ptyProcess.ts
zmb3: This whole function looks rather complicated. Does `process.cwd()` not work?

packages/teleterm/src/mainProcess/runtimeSettings.ts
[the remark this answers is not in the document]
gzdunek: In the end, I decided to return only the port from the child processes. Returning the full address was a bit problematic, for example, tshd retuned `127.0.0.1` instead of `localhost` (`grpc-js` uses DNS to resolve the address) and I had to remove `tcp://` from returned address manually. 

I modified @ravicious idea a bit. Because we have to provide the resolved addresses to the `preload` script, we need to add a new `ipcMain` handler. That handler can be the only place that interacts with the logic that resolves the addresses. By doing this, we don't have to change the code in `main.ts`.

packages/teleterm/src/main.ts
```diff
@@ -9,103 +9,123 @@ import { ConfigServiceImpl } from 'teleterm/services/config';
...
       ) {
-        return true;
+        event.preventDefault();
+        callback(true);
+      } else {
+        callback(false);
+        console.error(error);
       }
     }
+  );
```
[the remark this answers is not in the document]
gzdunek: Yes, I wanted to make sure that we only suppress errors for Webpack that servers our app over HTTPS using self-signed cert. Previously, we suppressed all certs-related errors.

If there are any errors related to do the generated certificates we will see them.

packages/teleterm/src/sharedProcess/ptyHost/ptyProcess.ts
gzdunek: >This whole function looks rather complicated. Does process.cwd() not work?

No, `process.cwd()` returns only the directory where the process was spawned. 
The method from above is used also by `vscode` or `tabby`.

>Did you try to implement it for Windows? What problems did you run into?

Hmm is there any way to do it? The apps I mentioned don't support CWD for Windows. From what I see, there is a [Get-Process](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/get-process?view=powershell-7.2) command, but it doesn't return a working directory.

packages/teleterm/src/mainProcess/windowsManager.ts
gzdunek: I don't think so. If I remember correctly, starting from Windows 7, the menu bar was hidden by default and you had to press `Alt` to show it.

Another reason is that we don't show anything interesting in that menu ;)

packages/teleterm/src/ui/ThemeProvider/globals.ts
```diff
@@ -37,10 +38,28 @@ const GlobalStyle = createGlobalStyle`
...
+    height: 8px;
+  }
+
+  ::-webkit-scrollbar-thumb {
+    background: #757575;
+  }
+
+  ::-webkit-scrollbar-corner {
+    background: rgba(0, 0, 0, 0.5);
+  }
```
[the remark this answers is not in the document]
gzdunek: By default Chrome renders that ugly, wide scrollbar :/ 

image

packages/teleterm/src/mainProcess/runtimeSettings.ts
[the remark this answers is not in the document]
zmb3: This sounds like a good approach. I agree with Rafal that we should do it right rather than hope that the ports are still free by the time we launch tshd.

packages/teleterm/package.json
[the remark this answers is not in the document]
gzdunek: Done, I improved serial number generation basing on issues in `node-forge` (previously it was a random number from 50 000 - 99 999 range, now it is a 16 bytes length random number).

packages/teleterm/src/services/grpcCerts/generateGrpcCerts.ts
```diff
@@ -0,0 +1,39 @@
...
+    caKey: ca.key,
+    caCert: ca.cert,
+  });
+
+  await Promise.all([
+    fs.writeFile(getPath('ca.crt'), ca.cert),
+    fs.writeFile(getPath('server.crt'), server.cert),
+    fs.writeFile(getPath('server.key'), server.key),
+    fs.writeFile(getPath('client.crt'), client.cert),
+    fs.writeFile(getPath('client.key'), client.key),
```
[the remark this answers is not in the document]
espadolini: My idea was to avoid ever having private key material on disk and to essentially tie the authorization to the ability to write in what is (hopefully) a private directory (to overwrite the certificate) rather than just the ability to read in that directory.

This warrants some testing, but I'm pretty sure that if you have a certificate in the trust store and the end-entity cert presented by the other end of a TLS connection is _that_ certificate then any authentication is assumed to pass, so you can just put a self-signed certificate in `rootCerts`/`ClientCAs` without having to generate a certificate tree.

If you don't want tsh to generate its own cert/key then you should be generating a single self-signed cert/key, store it on disk, and use it for both ends of the connection.

packages/teleterm/src/sharedProcess/ptyHost/ptyProcess.ts
ravicious: > Hmm is there any way to do it?

I don't know, that's why I asked. 😏 If there's no easy way to do that then it's fine I guess. I haven't installed Parallels yet so I cannot check if cmd.exe or Powershell show something (if anything) instead of cwd.

packages/teleterm/src/services/grpcCerts/makeCert/makeCert.ts
```diff
@@ -0,0 +1,198 @@
...
+      digitalSignature: true,
+      keyEncipherment: true,
+      critical: true,
+    },
+    { name: 'extKeyUsage', serverAuth: true, clientAuth: true },
+    {
+      name: 'subjectAltName',
+      altNames: domains.map(domain => {
+        const types = { domain: 2, ip: 7 }; // available Types: https://git.io/fptng
+        const ipRegex = /^(?!0)(?!.*\.$)((1?\d?\d|25[0-5]|2[0-4]\d)(\.|$)){4}$/;
```
ravicious, line 123: Nit: I'd recommend moving it to `const IP_REGEX` somewhere at the top of the file.

packages/teleterm/src/mainProcess/mainProcess.ts
```diff
@@ -104,11 +116,52 @@ export default class MainProcess {
     });
   }
 
+  private resolveNetworkAddress(
```
ravicious, line 119: Looks good, though the only missing thing is some kind of a timeout mechanism. Otherwise the initialization process can hang forever waiting for the process to output the matching string.

packages/teleterm/src/mainProcess/mainProcess.ts
```diff
@@ -59,7 +71,7 @@ export default class MainProcess {
   private _initTshd() {
     const { binaryPath, flags, homeDir } = this.settings.tshd;
     this.tshdProcess = spawn(binaryPath, flags, {
-      stdio: [null, 'pipe', 'pipe'],
```
ravicious, line 62: I wonder why it was set to `null` in the first place. [The docs](https://nodejs.org/api/child_process.html#optionsstdio) say that providing `null` or `undefined` has this effect:

> Use default value. For stdio fds 0, 1, and 2 (in other words, stdin, stdout, and stderr) a pipe is created. For fd 3 and up, the default is `'ignore'`.

packages/teleterm/src/mainProcess/mainProcess.ts
```diff
@@ -91,7 +103,7 @@ export default class MainProcess {
       path.join(__dirname, 'sharedProcess.js'),
       [`--runtimeSettingsJson=${JSON.stringify(this.settings)}`],
       {
-        stdio: 'inherit',
+        stdio: 'pipe',
```
ravicious, line 106: I was curious if this is going to have any bearing on logs but it looks like they're still collected properly.

packages/teleterm/src/preload.ts
```diff
@@ -7,28 +7,38 @@ import { ElectronGlobals } from './types';
...
-  const grpcCerts = await readGrpcCerts(runtimeSettings.certsDir);
+contextBridge.exposeInMainWorld('electron', getElectronGlobals());
 
-  const tshClient = createTshClient(runtimeSettings, grpcCerts);
-  const ptyServiceClient = createPtyService(runtimeSettings, grpcCerts);
+async function getElectronGlobals(): Promise<ElectronGlobals> {
+  try {
+    const [grpcCerts, resolvedAddresses] = await Promise.all([
+      readGrpcCerts(runtimeSettings.certsDir),
+      mainProcessClient.getResolvedChildProcessAddresses(),
```
ravicious, line 26: How do we guarantee that these address are set before the execution gets to this point?

ravicious (commented):
I looked through the new commits, I'll install Parallels on Monday and take the app for a spin.

packages/teleterm/src/services/grpcCerts/generateGrpcCerts.ts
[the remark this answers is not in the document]
gzdunek: Cool, I wasn't aware that I do not have to create that whole certificate tree!
Thanks to this, I can simplify the `makeCerts.ts` module.

>If you don't want tsh to generate its own cert/key then you should be generating a single self-signed cert/key, store it on disk, and use it for both ends of the connection.

Still, it seems to me that the simplest solution is to have one place in the app that generates certificates.

packages/teleterm/src/services/grpcCerts/makeCert/makeCert.ts
gzdunek: Done

packages/teleterm/src/mainProcess/mainProcess.ts
gzdunek: Done, I set 15s timeout, should be fine?

packages/teleterm/src/mainProcess/mainProcess.ts
gzdunek: I don't know tbh, this doesn't seem to have any effect on the collected logs (as you noticed below).

packages/teleterm/src/preload.ts
gzdunek: Please take a look at `_init()` method in `mainProcess.ts`. The promise that resolves that addresses is set in that method. This means that by the time the execution reaches that point, the promise is already created and our async handler will return it.

To improve readability, I extracted the code that sets ` this.resolvedChildProcessAddresses...` to a new method and changed the order of execution. Now `this._initIpc()` is set last.

packages/teleterm/src/mainProcess/mainProcess.ts
ravicious: > Done, I set 15s timeout, should be fine?

Yeah, I think even 10s would be enough. The server should get spun up pretty quickly, like within the first second or two so at 15 I feel like we're just making the user wait unnecessarily. ;)

packages/teleterm/src/preload.ts
ravicious: Great, that makes sense, thanks.

ravicious (commented):
I managed to run the app but couldn't get the PTY to work, I'll try again tomorrow!

packages/teleterm/src/mainProcess/mainProcess.ts
gzdunek: Done

packages/teleterm/src/mainProcess/runtimeSettings.ts
```diff
@@ -101,18 +102,53 @@ function getAssetPath(...paths: string[]): string {
...
-    logger.error(
-      `Failed to read ${process.platform} platform default shell, using fallback: ${fallbackShell}.\n`
-    );
+function requestChildProcessesAddresses(): ChildProcessAddresses {
+  switch (process.platform) {
+    case 'win32': {
+      return {
+        tsh: 'localhost:0',
+        shared: 'localhost:0',
+      };
```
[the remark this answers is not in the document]
gzdunek: These errors are related to this change: https://github.com/gravitational/webapps/blob/649aaf35a5f4baf6c910890ce7aee9d26dcc5209/packages/teleterm/src/main.ts#L46-L47

I couldn't find more info on the internet.
My guess is that when the certificate error occurs in C ++ code, it is reported to the JS layer, then we let the cert be trusted, but the original error is still printed.

packages/teleterm/src/services/grpcCerts/generateGrpcCerts.ts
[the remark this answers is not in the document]
espadolini: How does `tshd` get told about this "CA" to verify client connections? Is there any chance that it will also trust it for outbound connections?

If that's not a possibility then LGTM, although the cert generation could be slimmed down a fair bit - we shouldn't need any SANs, or even _any_ field in the subject (although a CN of `teleport-connect-internal` or something along those lines could be useful if it ends up being printed anywhere).

Basically the equivalent of
```
% openssl req -new -x509 -newkey ec:<( openssl ecparam -name prime256v1 ) -keyout key.pem -nodes -out cert.pem -subj /CN=teleport-connect -set_serial 0 -days 3650
Generating a 256 bit EC private key
writing new private key to 'key.pem'
-----
espadolini@cantor:~% openssl x509 -noout -text -in cert.pem
Certificate:
    Data:
        Version: 1 (0x0)
        Serial Number: 0 (0x0)
    Signature Algorithm: ecdsa-with-SHA256
        Issuer: CN=teleport-connect
        Validity
            Not Before: Jul 19 09:44:09 2022 GMT
            Not After : Jul 16 09:44:09 2032 GMT
        Subject: CN=teleport-connect
        Subject Public Key Info:
            Public Key Algorithm: id-ecPublicKey
                Public-Key: (256 bit)
                pub:
                    04:10:8e:b7:f4:42:d6:c9:c5:df:dc:94:42:ea:12:
                    8b:cd:20:e1:0f:1c:91:c0:95:87:41:84:55:42:32:
                    b2:36:dc:78:0d:9c:21:75:82:0b:fe:7a:85:cc:c5:
                    e6:91:9a:29:ef:14:68:99:88:07:91:49:87:07:eb:
                    9d:bc:70:78:66
                ASN1 OID: prime256v1
                NIST CURVE: P-256
    Signature Algorithm: ecdsa-with-SHA256
         30:45:02:20:1a:13:1d:ea:50:ab:dd:02:04:98:9e:86:0c:7a:
         ff:86:db:ed:19:65:3f:e9:3f:f7:76:37:a2:26:d7:3b:bf:06:
         02:21:00:81:be:0a:eb:f2:c7:50:09:6f:a1:c1:af:8e:98:93:
         df:1f:e4:1d:08:b9:04:e3:a4:16:d7:ff:9b:2b:93:72:71
```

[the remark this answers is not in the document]
gzdunek: Thx, I removed unneeded fields in the cert.

>How does tshd get told about this "CA" to verify client connections?
[`tshd` reads `certs.crt` from the `certs` directory](https://github.com/gravitational/teleport/blob/d959e315cd2e865d68ab3a27659e9b0b01028535/lib/teleterm/apiserver/apiserver.go#L121) which is located in user space. Only the user should be able to create a connection. 

>Is there any chance that it will also trust it for outbound connections?
No, it should only trust `localhost` connections that come from a user who has access to the certs.

I tested using `teleport-connect-internal` as CN, but it ended with the error:
`Hostname/IP does not match certificate's altnames: Host: localhost. is not cert's CN: teleport-connect-internal`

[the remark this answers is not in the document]
gzdunek: I thought more about it and decided to not save the private key on a disk to increase the security and enable TLS only for TCP connections to save the resources.
Thanks @espadolini for pointing this out!

Now we store these three files:
- `shared_server.crt` generated by shared process 
- `tsh_server.crt` generated by tsh process  
- `client.crt` generated by window preload script

Private keys are stored in the memory of each process.

Then the servers wait for the client's cert, and the client waits for the servers' cert. Creating a `tsh` server is a bit simpler because we can use the [`GetConfigForClient` function, which reads the client's cert from disk when it tries to establish a connection](https://github.com/gravitational/teleport/blob/a15c606bf0ed568fdf6a9e6e59d6e5ce93f21d16/lib/teleterm/apiserver/apiserver.go#L150). In the case of a shared server, this is no that option, we have to wait with the creation of the server until the client's cert is read.

I encapsulated the logic involved in generating & reading certs, and then creating `ServerCredentials` & `ClientCredentials` in the `grpcCredentials` module. This makes the shared server and preload unaware of all the machinery needed to create the credentials. In addition, I could easily disable the use of TLS for UDS connections where it is not needed. 

Waiting for files is done using node.js `watch`, this should be the most efficient option. We can't watch a file that doesn't exist, so instead the entire directory is watched. Then we wait for the first event with a non-empty file with the specified name. After that, the watcher is closed.

packages/teleterm/src/services/config/providers/keyboardShortcutsConfigProvider.ts
```diff
@@ -50,7 +50,7 @@ export const keyboardShortcutsConfigProvider: ConfigServiceProvider<KeyboardShor
         'toggle-identity': 'Command-I',
       };
 
-      const linuxShortcuts: KeyboardShortcutsConfig = {
+      const windowsAndLinuxShortcuts: KeyboardShortcutsConfig = {
```
ravicious, line 53: I just noticed that the shortcut for `tab-next` is `Ctrl-PageDown`. Shouldn't it be `Ctrl-Tab`? AFAIK that's what browsers use on those systems and that's what I found in [help articles for Edge](https://support.microsoft.com/en-us/microsoft-edge/keyboard-shortcuts-in-microsoft-edge-50d3edab-30d9-c7e4-21ce-37fe2713cfad) too.

packages/teleterm/src/services/pty/ptyHost/buildPtyOptions.ts
```diff
@@ -5,6 +5,7 @@ import {
   ResolveShellEnvTimeoutError,
 } from './resolveShellEnv';
 import { PtyCommand, PtyProcessCreationStatus } from '../types';
+import { delimiter } from 'path';
 
 export async function buildPtyOptions(
```
ravicious, line 10: Connect sets the default working directory for new shell sessions on Windows to the directory where the app is installed. It seems that we should default it to the user home directory instead, just as cmd and powershell do.

packages/teleterm/src/services/grpcCredentials/helpers.ts
```diff
@@ -0,0 +1,91 @@
...
+  } finally {
+    abortController.abort();
+  }
+}
+
+/**
+ * Checks if the gRPC connection should be encrypted.
+ * The only source of truth is the type of tshd protocol.
+ * Any other protocol than `unix` should be encrypted.
+ * The same check is performed on the tshd site.
```
ravicious, line 83:
```suggestion
 * The same check is performed on the tshd side.
```

packages/teleterm/src/services/grpcCredentials/grpcClientCredentials.ts
```diff
@@ -0,0 +1,59 @@
+import { ChannelCredentials, credentials } from '@grpc/grpc-js';
+import { RuntimeSettings } from 'teleterm/mainProcess/types';
+import {
+  readGrpcCert,
+  generateAndSaveGrpcCert,
+  shouldEncryptConnection,
+} from './helpers';
+import { GrpcCertName } from './types';
+
+export async function getGrpcClientCredentials(
```
ravicious, line 10: What do you think about dropping `grpc` from function and file names? They're already living in the `grpcCredentials` directory anyway.

packages/teleterm/src/preload.ts
```diff
@@ -16,12 +17,28 @@ const loggerService = createLoggerService({
...
+    );
+
+    return {
+      mainProcessClient,
+      tshClient,
+      ptyServiceClient,
+      loggerService,
+    };
+  } catch (e) {
+    return e;
```
ravicious, line 42: Couldn't we remove this `catch` branch altogether and get rid of the error handling here?

https://github.com/gravitational/webapps/blob/344e62cf6f3143b6efebd2745dbbcc2bcea14dc1/packages/teleterm/src/ui/boot.tsx#L31-L37

AFAIR context bridge handles errors just fine, minus any custom properties which we don't need to worry about in this case anyway.

ravicious (commented):
Getting close!

packages/teleterm/src/services/grpcCredentials/helpers.ts
gzdunek: Done

packages/teleterm/src/services/grpcCredentials/grpcClientCredentials.ts
gzdunek: Ah yes, look much better now

packages/teleterm/src/preload.ts
gzdunek: Right, these error handlers are unnecessary here. I've just tested it, the app works the same without them.

I left the handler in the `boot` function to show the actual error instead of a blank screen.

packages/teleterm/src/services/config/providers/keyboardShortcutsConfigProvider.ts
gzdunek: I'll address it in the next PR as discussed.

packages/teleterm/src/services/pty/ptyHost/buildPtyOptions.ts
gzdunek: I'll address it in the next PR as discussed.

ravicious (approved):
LGTM. @espadolini, could you take one final look at the cert situation? https://github.com/gravitational/webapps/pull/971#discussion_r926630971 I'd really appreciate it!

espadolini:
Left some comments on gravitational/teleport#14301 but they also apply here.

alex-kovoy:
Curious what did not work on Windows as Microsoft added a support for unix-sockets [back in 2018.](https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/) ?

gzdunek:
@alex-kovoy node.js doesn't support UDS on Windows :/

espadolini:
The TLS stuff LGTM 👍

PR #972, opened 2022-07-11
Add k8s to recordings and active sessions list

kimlisa:
#### Description
resolves https://github.com/gravitational/teleport/issues/14009

- Allow table column sorting to take in an alternate sorting key (we had a `durationText` but could not accurately sort by the text, so we need to sort by the field that was used to calculate `durationText`)
- rename conflicting `session.ts` to more accurate `websession` and changed `ssh.ts` to more accurate `session.ts`
- Adds k8s sessions to recordings list (tested it was playable in Player)
- Adds k8s sessions to active sessions list (not joinable at this time). This required a refactor in how we display information because it had node only language. I removed some data that I thought wasn't necessary (and wasn't available for k8 sessions) as depicted in screenshots below

#### Dependent PR
- [x] merge first, backend change: https://github.com/gravitational/teleport/pull/14329
- [x] merge enterprise PR: https://github.com/gravitational/webapps.e/pull/351
- [x] requires update e-ref to pass CI
- [ ] backport to v10

#### Screenshot Before and After
before (node only information)

image

after (node + kube)

image

recordings

image


packages/teleport/src/services/recordings/makeRecording.ts
```diff
@@ -80,7 +80,7 @@ function makeSshRecording({
     users: participants.join(', '),
     hostname,
     description,
-    recordingType: 'ssh',
+    recordingType: kubernetes_cluster ? 'k8s' : 'ssh',
```
ibeckermayer, line 83: We should rename this function to something more accurate like `makeSshOrKubeRecording`, or abstract out the common logic and call it from `makeSshRecording` and a new `makeKubeRecording`

packages/teleport/src/services/ssh/types.ts
```diff
@@ -16,10 +16,10 @@ limitations under the License.
 
 export interface Participant {
   user: string;
-  remoteAddr: string;
 }
 
 export interface Session {
+  kind: SessionKind;
```
ibeckermayer, line 22: It's imprecise to export this `Session` from `services/ssh` and then also use it for k8s sessions. What about adding something like this to `services/kube/types.ts`

```typescript
import { Session } from 'teleport/services/ssh';

export interface KubeSession extends Session {
  kubeClusterId: string;
}
```

ibeckermayer (changes requested):
Would the same argument in https://github.com/gravitational/teleport/issues/14009 not also apply to Desktop Sessions and Database Sessions?

packages/teleport/src/services/recordings/makeRecording.ts
kimlisa: renamed to `makeSshOrKubeRecording`

packages/teleport/src/services/ssh/types.ts
kimlisa: i meant to rename this file to `session`, and rename the existing `session` to `websession`, see here: https://github.com/gravitational/webapps/pull/972/commits/de7bddd306bd74b05d60a8ba978c0d937910acd3

kimlisa:
> Would the same argument in https://github.com/gravitational/teleport/issues/14009 not also apply to Desktop Sessions and Database Sessions?

@ibeckermayer If it does, then it will have to be another issue (setup takes a while to get). This kube issue turned out to be a top priority for a customer so I kinda dropped my previous work to do this one quickly.

packages/teleport/src/services/ssh/types.ts
kimlisa: sigh... renaming things kinda screwed up the file diff... just fyi

ibeckermayer (approved):

ibeckermayer (approved):

PR #979, opened 2022-07-13
disable DocumentCluster sync button when syncing

avatus:
Fixes https://github.com/gravitational/webapps.e/issues/207

Problem:
The current Sync button shows no indication that anything is syncing.

Solution:
Listen to state changes for when syncing starts/stops and disable the button accordingly. 

https://user-images.githubusercontent.com/5201977/178826490-c26bb14a-1fcb-4fdf-ad59-70cd4005ec3a.mp4

Thoughts:
There is a method currently named `getSyncStatus()` which I originally wanted to use although I'm not quite sure its the best use for it. I think that function has some race-y conditions as well due to the "sync" status being returned if any of the services are in a 'syncing' state. But imagine a scenario where one service starts syncing (button disabled), stops (button enabled), then the next service starts syncing for some reason (button disabled again? flickering button?).

The `getSyncStatus()` method is great for grabbing the individual status of each service which is the only use of it in the codebase I've seen. For example `getSyncStatus().dbs` would return the status of that individual service. I like this in that context.

Regardless, the Sync button (and any button that performs an action) should have its state abstracted to the highest order call of that 'action'. To use another example, if I add a "todo" on a todo website and click Save, I don't want my save button determined by the state of the web server, the authentication, the database action, etc etc all individually, I just want the state of my button to be "i made my request, and now the request is done". 

Long story longer, I added `syncing` state to the clusterContext which is changed when the `sync` function is called.

Ok rambling over, please let me know if I'm over thinking this.

### DIFF
--- a/packages/teleterm/src/ui/DocumentCluster/DocumentCluster.tsx
+++ b/packages/teleterm/src/ui/DocumentCluster/DocumentCluster.tsx
@@ -90,7 +90,12 @@ export function Cluster() {
         </Text>
         <Flex>
           <ClusterSearch onChange={clusterCtx.changeSearchValue} />
-          <ButtonPrimary ml={2} size="small" onClick={clusterCtx.sync}>
+          <ButtonPrimary
+            disabled={state.syncing}
+            ml={2}
+            size="small"
+            onClick={clusterCtx.sync}
+          >
             Sync
           </ButtonPrimary>
         </Flex>
--- a/packages/teleterm/src/ui/DocumentCluster/clusterContext.tsx
+++ b/packages/teleterm/src/ui/DocumentCluster/clusterContext.tsx
@@ -28,6 +28,7 @@ type State = {
   searchValue: string;
   leaf: boolean;
   leafConnected: boolean;
+  syncing: boolean;
   status: 'requires_login' | 'not_found' | '';
   statusText: string;
 };
@@ -47,6 +48,7 @@ class ClusterContext extends Store<State> {
     searchValue: '',
     leaf: false,
     leafConnected: false,
+    syncing: false,
     status: '',
     statusText: '',
   };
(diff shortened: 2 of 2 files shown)


ravicious:
The rationale about abstracting the state to the highest order of an action makes sense if clicking that button would be the only way to trigger sync. But sync is also triggered when you start the app or after you log in to a cluster – that's why `ClustersService` acts as a single source of truth for that information.

I think the sync button actually should be disabled when any of the resource types is getting synced. In general, it's not possible to sync just a single resource type, so the "Sync" button or any other action which causes sync syncs all resources.

The individual syncing states could be used to show progress bars on each individual resource tab. Take a look at the `DocumentGateway` component – it uses another component, `LinearProgress`, to show the progress bar. I think it'd be great to use that in the servers tab and the databases tab.

---

While you're at it, could you also remove some unnecessary code? The app was supposed to support app access and k8s too so we already sync that info but we don't actually show it anywhere. On top of that, k8s is synced twice. 🤔

We can leave `syncApps` defined in the code `syncKubes`, but we should probably remove calls to them in the sync functions.

https://github.com/gravitational/webapps/blob/6f1d453ab4e94b73095c2d2cea732e02b01bfd4c/packages/teleterm/src/ui/services/clusters/clustersService.ts#L127-L132

https://github.com/gravitational/webapps/blob/6f1d453ab4e94b73095c2d2cea732e02b01bfd4c/packages/teleterm/src/ui/services/clusters/clustersService.ts#L147-L155

ravicious:
The issue also says this:

> We have similar issue when fetching leaf clusters - when a leaf cluster is the active cluster, you will see ... cluster not found message on startup.

This problem stems from the fact that we don't track the sync status of leaf cluster information (the call to `ClustersService.syncLeafClustersList` specifically). So we can't even show progress for that. I believe this deserves to have its own issue created as piling it on top of this one might be too much work.

ravicious:
At this point it'd probably be best to address this when adding advanced search, wouldn't it? Since the logic for "things are being loaded" will be probably much different.

avatus:
Closing this as this functionality will change and be handled [in this PR](https://github.com/gravitational/webapps/pull/1307)

PR #981, opened 2022-07-14
Add gap property to Flex

ravicious:
This will help us stop adding margins on individual items.

styled-system doesn't seem to support gap yet [1] but we can just add it
ourselves.

This code works for styled-system v3.1.11 [2]. In the future we'll need
to simply replace `style` with `system` [3].

[1] https://github.com/styled-system/styled-system/issues/1159
[2] https://github.com/styled-system/styled-system/blob/v3.1.11/docs/custom-props.md
[3] https://styled-system.com/custom-props/


kimlisa (approved):
cool

PR #984, opened 2022-07-15
Remove unnecessary tranformValue for gap

ravicious:
It makes it impossible to specify gap as a specific value, for example "8px"
as it'll get transformed to "8pxpx".


hatched (approved):
Thanks I noticed this and made a similar change locally but you beat me to it 👍

PR #985, opened 2022-07-15
Add source maps and type checking to Webpack, config improvements

ryanclark:
### General improvements

Change `url-loader` and `svg-url-loader` to [asset modules](https://webpack.js.org/guides/asset-modules/) instead, as `url-loader` has been deprecated since webpack 5.

Change the `tsconfig.json` at the root to support the type checking additions below.

Remove `raw-loader` and instead use the `resourceQuery` `?raw` to provide a static asset module instead. https://webpack.js.org/guides/asset-modules/#replacing-inline-loader-syntax

Use `[contenthash]` instead of `[hash]` for file names as `[hash]` has been deprecated.

Remove a few unused dependencies.

### Source Maps

This changes the webpack config to enable source maps during development.

Making this change - 
image

Shows up in dev tools correctly from `Audit.tsx` (no idea why it renders 4 times, probably state changes on initial effects)
image

Clicking through shows the correct source -
image

### Type Checking

I noticed that we don't have type checking with webpack. I make this deliberate incorrect type change -

image

Webpack is fine with it -

image

To fix this, as `babel-loader` + TypeScript presets only work to transpile TypeScript and not type check it, I've added `ts-loader` and `fork-ts-checker-webpack-plugin`.

`ts-loader` will type check in production builds (and fail CI if there are any issues), and just transpile to es6 for babel to take over in development. 

`fork-ts-checker-webpack-plugin` runs during development. This runs on a separate thread so it doesn't block the compilation of assets, so your development build is still as fast as possible. An example of this type checking the issue above -

image


packages/build/jest/config.js
```diff
@@ -21,7 +21,7 @@ module.exports = {
   moduleNameMapper: {
     // mock all imports to asset files
     '\\.(css|scss|stylesheet)$': path.join(__dirname, 'mockStyles.js'),
-    '\\.(png|svg|yaml)$': path.join(__dirname, 'mockFiles.js'),
+    '\\.(png|svg|yaml|yaml\\?raw)$': path.join(__dirname, 'mockFiles.js'),
```
zmb3, line 24: I'm not familiar with the `yaml?raw` - what's going on here?

packages/build/webpack/webpack.base.js
```diff
@@ -37,20 +47,30 @@ const configFactory = {
...
     fonts() {
       return {
         test: /fonts\/(.)+\.(woff|woff2|ttf|eot|svg)/,
-        loader: 'url-loader',
-        options: {
-          limit: 102400, // 100kb
-          name: '/assets/fonts/[name].[ext]',
+        type: 'asset',
+        generator: {
+          filename: 'assets/fonts/[name][ext]',
```
zmb3, line 61: I assume the `.` is now part of the extension?

ryanclark: Yeah - https://webpack.js.org/guides/asset-modules/#custom-output-filename

packages/build/jest/config.js
ryanclark: You can use `resourceQuery` to match with `?whatever` on imports, and the webpack docs recommend changing `raw-loader` to use a `resourceQuery` match instead, and specify it as a `asset/source` type - https://webpack.js.org/guides/asset-modules/#replacing-inline-loader-syntax

kimlisa:
i'm running into some errors that maybe I'm missing a step? the below commands are how i restarted things:

```ssh
$ make clean
$ yarn cache clean
$ yarn install
```

Problems:

- running `yarn storybook` is not working anymore, i think the common issue was `unexpected token`
- when i run `yarn tsc` (or i think `yarn type-check` is equiv), for every existing file, it creates a new equivalent `.js` and a `js.map` file. I use this command a lot: right before i push, and to help me find files that may be affected by a type change (i rarely look at my dev server outputs)
- i was clicking around the app and for some reason, this branch removes styling on these components (i haven't looked into why):

**this branch (removes styling):**

image
image


**master branch:**

image
image

ryanclark:
@kimlisa I've fixed the styling issues - thanks for finding that!

`yarn tsc` will run the TypeScript compiler, and will emit `.js` files - `yarn type-check` will run the TypeScript compiler with `--noEmit`, which won't - I recommend using the dev server for faster feedback than `yarn type-check`, but you can continue on just using `yarn type-check` if you wish, which won't emit files.

I've also fixed storybook - please can you test this again? Thank you 😀

gzdunek (commented):
Thanks for running TS with webpack :)

Any idea why building a storybook is much slower?
Previously: 14s
Now: 52s

packages/build/webpack/webpack.base.js
```diff
@@ -37,20 +47,30 @@ const configFactory = {
...
   rules: {
+    raw() {
+      return {
+        resourceQuery: /raw/,
+        type: 'asset/source',
+      };
+    },
     fonts() {
       return {
         test: /fonts\/(.)+\.(woff|woff2|ttf|eot|svg)/,
```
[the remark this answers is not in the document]
ryanclark: Done!

packages/build/webpack/webpack.dev.config.js
```diff
@@ -29,16 +29,19 @@ module.exports = {
     filename: '[name].js',
     chunkFilename: '[name].js',
   },
-  devtool: false,
+  devtool: 'source-map',
```
ravicious, line 32: Should we use source-map in development? [The docs](https://webpack.js.org/configuration/devtool/) say it's a setting recommended for production builds. Did you consider using `eval-source-map` or any other setting?

ravicious, line 32: Btw, here's our [previous discussion about source maps](https://gravitational.slack.com/archives/C02DQ1C2BMW/p1651593622899069). From what I can gather, Alexey was against enabling source maps by default. Nowadays I think most of us would rather have them enabled by default.

Perhaps we could add an env var here so that if someone wants to keep it set to `off` they could just set the env var?

packages/build/.babelrc.js
```diff
@@ -22,12 +22,11 @@ const plugins = [
...
 
 function makePresents(test = false) {
-  const presents = ['@babel/preset-react', '@babel/preset-typescript'];
-
   if (test) {
     return [
       ['@babel/preset-env', { targets: { node: 'current' } }],
-      ...presents,
+      '@babel/preset-react',
+      '@babel/preset-typescript',
```
ravicious, line 29: So now only tests need those babel presets because in dev/production builds React and TypeScript stuff is going to be handled by ts-loader, do I understand this right?

packages/teleterm/tsconfig.json
```diff
@@ -0,0 +1,48 @@
...
+    "jsx": "react",
+    "module": "esnext",
+    "moduleResolution": "node",
+    "noEmitHelpers": true,
+    "pretty": true,
+    "sourceMap": true,
+    "stripInternal": true,
+    "target": "esnext",
+    "types": [
+      "@emotion/core"
```
ravicious: What does this add?

.cloudbuild/ci/build.yaml
```diff
@@ -18,17 +18,17 @@ steps:
       - "go run ./cmd/checkout -w=/workspace"
 
   - name: gcr.io/cloud-builders/docker
-    id: type-check-and-lint
+    id: prettier-check-and-lint
```
ravicious, line 21: I think this step and the underlying npm script can be renamed to just `lint` – it can run both eslint and prettier check. `type-check-and-lint` is used only in CI, I added it in #874.

ravicious, line 21: Edit: nvm, I wrote this comment and then saw you addressed this in the next comment. ;f

---

With the new config instead of typechecking the code once we're going to check it twice, first when compiling the Web UI and then again when compiling Connect, right?

I don't think that's necessarily bad, one could even say it provides some additional granularity so that you can see your change broke the Web UI but not Connect.

---

tl;dr should this step run in parallel to other build steps now?

For me type checking was the most important part of `type-check-and-lint` – if the code doesn't type check then it's probably broken.

Then there's building which is another important step because if it doesn't build then it means that your PR broke something.

So one piece that we're left with is `prettier-check-and-lint` which now does just linting. IMHO most of the time that's a largely inconsequential check. I reeeally dislike when a minor problem with linting prevents the build from getting to the real deal which is type checking and actually attempting to build the code.

With that in mind, should we make it so that this step, renamed to lint, runs in parallel with `build-teleport` and `build-term`?

packages/teleport/package.json
```diff
@@ -25,6 +25,7 @@
   "devDependencies": {
     "@gravitational/build": "^1.0.0",
     "@types/wicg-native-file-system": "^2020.6.0",
-    "jest-canvas-mock": "^2.3.1"
+    "jest-canvas-mock": "^2.3.1",
+    "ts-loader": "^9.3.1"
```
ravicious, line 29: Hmm, that's suspicious, the addition of ts-loader to `packages/build/package.json` should've solved the issue. 🤔

packages/teleterm/webpack.renderer.dev.config.js
```diff
@@ -26,6 +26,6 @@ devCfg.devServer = {
 };
 
 devCfg.output.publicPath = '';
-devCfg.plugins = [configFactory.plugins.reactRefresh(), createHtmlPlugin()];
+devCfg.plugins.push(createHtmlPlugin());
```
ravicious, line 29: I didn't have time to investigate what causes this problem but this PR seems to break refreshing React components in Connect.

This line might not be related to that problem, I just had nowhere else to place this comment. ;)

If I open a tab in Connect and then edit its source file and save it, the dev version of Connect still shows the old version.

.storybook/main.js
```diff
@@ -60,9 +62,25 @@ module.exports = {
...
+
     storybookConfig.module.rules.push({
       test: /\.(ts|tsx)$/,
-      loader: require.resolve('babel-loader'),
+      use: [
+        {
+          loader: require.resolve('babel-loader'),
+        },
+        {
+          loader: require.resolve('ts-loader'),
```
ravicious, line 77: > I couldn't get fork-ts-checker-webpack-plugin to work with storybook - that would type check in another thread and speed up storybook. I think the initial slowdown isn't too much, but if it is I can set ts-loader to transpile only which should speed it up, but not type check.

Yeah the startup of the storybook is way to long IMHO. Making sure that my changes don't break stories is already something that felt like a chore so if doing that will mean spending a minute just waiting it'll make people even less likely to do it.

At the moment I couldn't even get the story book to run because of type errors.

I guess we could turn off type checking there until we figure out how to make it faster?

packages/build/webpack/webpack.dev.config.js
```diff
@@ -29,12 +29,16 @@ module.exports = {
...
   mode: 'development',
-  plugins: [configFactory.plugins.reactRefresh()],
+  plugins: [
+    configFactory.plugins.tsChecker(),
```
ravicious, line 35: The problems with the storybook got me thinking: what's the advantage of type checking code within webpack rather than as a separate step before building?

The only thing that comes to my mind is that the dev server will show you a nice error but is there anything else that this gives us?

packages/build/webpack/webpack.prod.config.js
```diff
@@ -32,6 +32,7 @@ module.exports = {
   module: {
```
ravicious, line 32: How do the changes in this PR impact compilation times of prod builds?

Connect is built after every commit to the teleport repo so I'm mostly worried about that right now + time spent in CI on building the Web UI and Connect.

ravicious (commented):
Whoa that's a lot of changes but well we had to try to modernize our setup at some point!

In terms of testing this I only tried to run Connect in dev mode and the storybook but stopped soon after running into some problems, I'll try to do more testing once those problems are fixed!

packages/build/webpack/webpack.prod.config.js
zmb3: Why does this matter though? I'm not aware of any part of the dev workflow that depends on the results of a build after something is committed to main. What's the difference to you if it takes 1 minute or 10?

ravicious: I remember [Trent pointing out that the macOS builders take a lot of time](https://gravitational.slack.com/archives/C0DF0TPMY/p1655950985331039?thread_ts=1655940010.768399&cid=C0DF0TPMY) so I didn't want to make this even worse if we don't have to.

And when it comes to CI it's best to just keep it as short as possible, I think.

packages/teleterm/tsconfig.json
ryanclark: Adds typings for the `css` prop

packages/teleterm/webpack.renderer.dev.config.js
ryanclark: Hm, I haven't tested Connect, but I will take a look. I'm not sure if this is the cause, because the react refresh plugin will already be in the array from the base config

.storybook/main.js
ryanclark: Sure thing, I'll turn it off!

packages/build/webpack/webpack.dev.config.js
ryanclark: Type checking is a lot faster when you can run a TypeScript compiler continuously instead of from fresh every time. I much prefer the development flow of having Webpack type check instead of having to run another command to do so, or rely on my IDE picking up changes. I left the `yarn type-check` script in for those who still want to use the command.

packages/build/.babelrc.js
ryanclark: Yeah, you're spot on

packages/build/webpack/webpack.dev.config.js
ryanclark: Sure thing, I'll add an env var to override it

ryanclark: I'm happy to use `eval-source-map`

packages/build/.babelrc.js
```diff
@@ -22,12 +22,11 @@ const plugins = [
 ];
 
 function makePresents(test = false) {
```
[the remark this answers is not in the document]
ryanclark: Sure!

.cloudbuild/ci/build.yaml
ryanclark: I'll rename

ryanclark: Yeah, but the `ts-loader` is set only compile bundled files, so it shouldn't type check anything that doesn't end up in the main bundle. This means the web UI will only type check the web UI code, and Connect will only type check Connect code

ryanclark: Also, I tried to change the `waitFor` from linting to checkout so they run in parallel, but it then builds the docker container three times - I'm not familiar with cloud build's config, so I didn't look into it much further and reverted the change

packages/build/.babelrc.js
ryanclark: Well, JSX is still handled by babel for the styled components babel plugin to work

.cloudbuild/ci/build.yaml
ravicious: I'll see if adding an empty intermediate step helps with that.

ravicious: Something like this should help with caching the docker image between steps:

https://github.com/gravitational/webapps/commit/c278ceca8b9f2c7ad1637b7433b204c3d435b2d9

https://console.cloud.google.com/cloud-build/builds;region=global/3911680d-bd1a-46c9-85d0-280faf69e9ef;step=2?project=ci-account

packages/build/webpack/webpack.prod.config.js
ravicious: I know this is based on just three runs but it looks like the CI times are not affected overall. Linting step takes one minute less but both build steps take two minutes more. When we run all three in parallel this results in no difference in time.

* Current master CI https://console.cloud.google.com/cloud-build/builds/d67c329e-1d5f-41a2-866a-06228714acfc?project=ci-account
* Current master CI https://console.cloud.google.com/cloud-build/builds/174cbd8b-9028-4891-bd1a-c27989149622?project=ci-account
* This PR but all three steps in parallel https://console.cloud.google.com/cloud-build/builds;region=global/3911680d-bd1a-46c9-85d0-280faf69e9ef;step=1;tab=detail?project=ci-account

packages/build/webpack/webpack.dev.config.js
alex-kovoy: Here is some recollection from my memory :) 

Source maps were disabled because they added extra overhead to build and debugging time. They also were not very useful (chrome struggled at that time to provide quick variable lookups and on the fly editing) so we disabled the source maps completely. Also, once you disable ES2015 compilation target (in babel) you should be able to see transpiled code pretty much unchanged. 

As for webpack ts complication in dev mode, the reason why it was disabled because it was slow. And since vscode has built-in ts compiler (which runs by default) it seemed to work fine. The feedback loop was < 1 sec. These days when everyone has M1 it might not be an issue anymore :)

packages/build/.babelrc.js
```diff
@@ -21,13 +21,12 @@ const plugins = [
   '@babel/plugin-syntax-dynamic-import',
 ];
 
-function makePresents(test = false) {
-  const presents = ['@babel/preset-react', '@babel/preset-typescript'];
-
+function makePresets(test = false) {
```
hatched, line 24: Aww I wanted presents!

hatched (approved):
Thanks for making these improvements!

packages/build/.babelrc.js
ryanclark: Apologies 😔

ravicious (approved):

gzdunek (approved):

PR #986, opened 2022-07-15
deny any permission requests until needed

avatus:
Closes https://github.com/gravitational/teleport-private/issues/146

For now, due to us not needing any explicit permissions that request this handler, we are just going to deny all until needed.

In the future, we will need a custom confirmation dialog to handle these requests, assuming we pass the request to the user.


avatus:
note: If we do need to ask for permission in the future, we can use our [DialogConfirmation](https://github.com/gravitational/webapps/tree/master/packages/design/src/DialogConfirmation)

gzdunek (approved):

ravicious (approved):

zmb3 (approved):

PR #987, opened 2022-07-15
added a content-security-policy

avatus:
Closes: https://github.com/gravitational/teleport-private/issues/147

Due to electron not being able to set HTTP request headers in production (because files are sent using the `file://` protocol), we can define a Content-Security-Policy in our "index.html" page.

This csp mostly matches the webui csp, with the exception of `font-src` where I've added `data:` to keep our icons working.

Note about `connect-src`: Adding in the urls `https://kcwm2is93l.execute-api.us-west-2.amazonaws.com/prod` and `https://usage.teleport.dev` are required for the feedback form to work on dev (insecure) and prod respectively

### DIFF
--- a/packages/teleterm/webpack.renderer.extend.js
+++ b/packages/teleterm/webpack.renderer.extend.js
@@ -28,7 +28,7 @@ function createHtmlPlugin() {
         <meta http-equiv="X-UA-Compatible" content="IE=edge" />
         <meta name="referrer" content="no-referrer" />
         <meta name="viewport" content="width=device-width, initial-scale=1" />
-        <meta http-equiv="Content-Security-Policy" content="default-src 'self'; connect-src 'self' https://kcwm2is93l.execute-api.us-west-2.amazonaws.com/prod; style-src 'self' 'unsafe-inline'; img-src 'self' data: blob:; object-src 'none'; font-src 'self' data:">
+        <meta http-equiv="Content-Security-Policy" content="default-src 'self'; connect-src 'self' https://kcwm2is93l.execute-api.us-west-2.amazonaws.com/prod https://usage.teleport.dev; style-src 'self' 'unsafe-inline'; img-src 'self' data: blob:; object-src 'none'; font-src 'self' data:">
       </head>
       <body>
         <div id="app"></div>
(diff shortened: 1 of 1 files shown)


packages/teleterm/webpack.renderer.extend.js
```diff
@@ -28,6 +28,7 @@ function createHtmlPlugin() {
         <meta http-equiv="X-UA-Compatible" content="IE=edge" />
         <meta name="referrer" content="no-referrer" />
         <meta name="viewport" content="width=device-width, initial-scale=1" />
+        <meta http-equiv="Content-Security-Policy" content="default-src 'self'; script-src 'self'; style-src 'self' 'unsafe-inline'; img-src 'self' data: blob:; object-src 'none'; font-src 'self' data:">
```
gzdunek, line 31: We have to add two domains to `connect-src` directive for the share feedback form (it uses two URLs, please take a look at `useShareFeedback`).

Additionally, we can remove redundant `script-src` directive (it has the same value as `default-src`).

avatus: thanks! done.

packages/teleterm/webpack.renderer.extend.js
```diff
@@ -28,6 +28,7 @@ function createHtmlPlugin() {
         <meta http-equiv="X-UA-Compatible" content="IE=edge" />
         <meta name="referrer" content="no-referrer" />
         <meta name="viewport" content="width=device-width, initial-scale=1" />
+        <meta http-equiv="Content-Security-Policy" content="default-src 'self'; connect-src 'self' https://kcwm2is93l.execute-api.us-west-2.amazonaws.com/prod; style-src 'self' 'unsafe-inline'; img-src 'self' data: blob:; object-src 'none'; font-src 'self' data:">
```
ravicious, line 31: I think we're still missing the `https://usage.teleport.dev` URL here.

avatus: whoops, removed during testing and forgot to throw back in. thank you!

gzdunek (approved):

ravicious (approved):

PR #988, opened 2022-07-15
updated webPreferences config

avatus:
Closes: https://github.com/gravitational/teleport-private/issues/148

Notes:
- `enableRemoteModule` defaults to `false` from electron 10+, so we do not need to explicitly set it here. [source](https://www.electronjs.org/docs/latest/breaking-changes#default-changed-enableremotemodule-defaults-to-false)
- `sandbox: true` has much bigger consequences and would require quite a bit of work to enable. We should discuss the implications of this in a separate issue. [docs](https://www.electronjs.org/docs/latest/tutorial/sandbox). As usual, if I'm overthinking this let me know, but the app does not currently work if sandbox is true and the changes appear to be substantial.

### DIFF
--- a/packages/teleterm/src/mainProcess/windowsManager.ts
+++ b/packages/teleterm/src/mainProcess/windowsManager.ts
@@ -41,6 +41,10 @@ export class WindowsManager {
       title: 'Teleport Connect Preview',
       icon: getAssetPath('icon.png'),
       webPreferences: {
+        devTools: this.settings.dev,
+        webgl: false,
+        enableWebSQL: false,
+        safeDialogs: true,
         contextIsolation: true,
         nodeIntegration: false,
         preload: path.join(__dirname, 'preload.js'),


packages/teleterm/src/mainProcess/windowsManager.ts
```diff
@@ -41,6 +41,10 @@ export class WindowsManager {
       title: 'Teleport Connect Preview',
       icon: getAssetPath('icon.png'),
       webPreferences: {
+        devTools: this.settings.dev,
+        webgl: false,
+        enableWebSQL: false,
+        safeDialogs: true,
```
ravicious, line 47: What were the errors you were running into when `sandbox` was set to `true`?

From reading the docs alone it seems that sandbox should just work for us but obviously I'm missing something here.

ravicious (approved):
Thanks for creating separate issues for those, it makes it easier to look up what vulnerabilities the PRs are dealing with.

I see that [`nativeWindowOpen` has been removed](https://www.electronjs.org/docs/latest/breaking-changes#removed-nativewindowopen) and now Electron acts as if it was set to `true`.

Since TEL-Q122-17 is only of informational severity, I think we can consider it done even without enabling sandbox.

packages/teleterm/src/mainProcess/windowsManager.ts
avatus: The `loggerService` would start to become undefined causing the app the break. Removing it would cascade into other problems. Didn't follow it to the end but that was the thread I started pulling before thinking it could just be a new issue

gzdunek: It can be caused by my changes here: https://github.com/gravitational/webapps/pull/853#discussion_r887943357 :/ 
Maybe we should think about it one more time?

ravicious: Oh I see, this might be a problem with Winston being a fundamentally a node package which requires node APIs. Inside renderers, any interaction with Winston would have to be done through IPC/gRPC, I guess any Winston objects could live only in the main and the shared process.

That's definitely a bigger change, I'll create a task for it just to have a dedicated place in case we want to come back to this in the future.

PR #991, opened 2022-07-18
Display UTC time in audit log and session recording log

avatus:
Fixes https://github.com/gravitational/teleport/issues/14562

I've updated the column header with UTC on audit events and display only the date string in raw format.

### DIFF
--- a/packages/teleport/src/services/audit/makeEvent.ts
+++ b/packages/teleport/src/services/audit/makeEvent.ts
@@ -702,7 +702,7 @@ export default function makeEvent(json: any): Event {
     id: getId(json),
     code: json.code,
     user: json.user,
-    time: new Date(json.time),
+    time: json.time,
     raw: json,
   };
 
--- a/packages/teleport/src/services/audit/makeEvent.ts
+++ b/packages/teleport/src/services/audit/makeEvent.ts
@@ -703,7 +703,7 @@ export default function makeEvent(json: any): Event {
     id: getId(json),
     code: json.code,
     user: json.user,
-    time: dateToUtc(new Date(json.time)),
+    time: new Date(json.time),
     raw: json,
   };
 
--- a/packages/teleport/src/services/audit/makeEvent.ts
+++ b/packages/teleport/src/services/audit/makeEvent.ts
@@ -15,6 +15,7 @@ limitations under the License.
 */
 
 import { formatDistanceStrict } from 'date-fns';
+import { dateToUtc } from 'shared/services/loc/loc';
 import { Event, RawEvent, Formatters, eventCodes } from './types';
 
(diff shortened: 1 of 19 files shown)


packages/teleport/src/services/audit/makeEvent.ts
```diff
@@ -702,7 +703,7 @@ export default function makeEvent(json: any): Event {
     id: getId(json),
     code: json.code,
     user: json.user,
-    time: new Date(json.time),
+    time: dateToUtc(new Date(json.time)),
```
zmb3, line 706: Shouldn't the JSON already be in UTC?

avatus: the date in the JSON is in UTC but `new Date` automatically creates/converts to the browser's timezone

zmb3: Can we use `Date.parse(json.time)` instead? This is fine, but we're allocating two `Date` objects for every event..

hatched: What is the data type of `json.time` it's untyped (`any`) in the app. `Data.parse()` parses a string representation of a datetime. It doesn't convert an epoch into a datetime whereas `Date(epoch)` will parse it into a new `Date` instance that can then be output into any format.

avatus: `json.time` is a date string format like
```
2022-07-19T19:32:26.755Z
```

We can just display the string if we don't want to have _any_ date objects created. 

If we just did something like `new Date(json.time).toUTCString()` we could eliminate the second `Date` object but then we'd get some gnarly string such as `Tue, 19 Jul 2022 19:32:26 GMT` which I guess isn't the worst thing. Maybe some string manipulation after? 

`date-fns/format` requires a number or `Date` object as it's first argument. If we did `Date.parse`, it returns in milliseconds so we could use in the `format` function but it still will output timezoned unless we add the offset prior.

---------

on second thought, we can just move some of the date manipulation into the `displayDateTime` function like this
```js
    return format(
      date.getTime() + date.getTimezoneOffset() * 60 * 1000,
      cfg.dateTimeFormat
    );
```

Only one `Date` would be allocated, which is used to parse from the raw date string

zmb3: Let's just display the string from the audit event. We don't need a date object at all.

avatus: The audit event will now show the same raw string event stored in the event json
```
2021-05-25T07:34:22.204Z
```

hatched (approved):

zmb3 (approved):

PR #992, opened 2022-07-19
Create LoginTrait Component

kimlisa:
#### Description

I made this component specific to `logins` (ssh) for the first phase. There were too many unknowns to make it generic for all traits (if that's even reasonable to do).

- Create a LoginTrait component that allows user's to add or remove new login principles
- Added trait fields to User response
- Created a shared file that exports shared/common components predicted to be used in each step

#### Dependent PR
- [ ] requires this small modification: https://github.com/gravitational/teleport/pull/14611

#### Demo
It starts with existing traits that I added earlier.

login-trait-demo

### DIFF
--- a/packages/teleport/src/Discover/LoginTrait/LoginTrait.test.tsx
+++ b/packages/teleport/src/Discover/LoginTrait/LoginTrait.test.tsx
@@ -1,0 +1,116 @@
+/**
+ * Copyright 2022 Gravitational, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+// eslint-disable-next-line import/named
+import { RenderResult } from '@testing-library/react';
+import { render, screen, act, fireEvent } from 'design/utils/testing';
+import type { User } from 'teleport/services/user';
+import { DiscoverContext } from '../discoverContext';
+import ContextProvider from '../discoverContextProvider';
+import LoginTrait from './LoginTrait';
+
+describe('login trait comp behavior', () => {
+  const ctx = new DiscoverContext();
+  const userSvc = ctx.userService;
+
+  let Component;
+
+  beforeEach(() => {
+    Component = (
+      <ContextProvider value={ctx}>
+        <LoginTrait
+          // TODO we don't need all of this
+          attempt={null}
+          joinToken={null}
+          createJoinToken={null}
+          agentMeta={null}
+          updateAgentMeta={null}
+          nextStep={null}
+          prevStep={null}
+        />
+      </ContextProvider>
+    );
+  });
+
+  test('add a new login', async () => {
+    jest.spyOn(userSvc, 'fetchUser').mockResolvedValue(mockUser);
+
+    let r: RenderResult;
+    await act(async () => {
+      r = render(Component);
+    });
+
+    // Expect no checkboxes to be rendered.
+    const checkboxes = r.container.querySelectorAll('input[type=checkbox]');
+    expect(checkboxes).toHaveLength(0);
+
+    // Test adding a new login name.
+    fireEvent.click(screen.getByText(/add new/i));
+    const inputEl = screen.getByPlaceholderText('name');
+    fireEvent.change(inputEl, { target: { value: 'banana' } });
+    fireEvent.click(screen.getByText('Add'));
+    expect(screen.getByText('banana')).toBeInTheDocument();
+  });
+
+  test('rendering of init logins', async () => {
+    jest.spyOn(userSvc, 'fetchUser').mockResolvedValue({
+      ...mockUser,
+      traits: {
+        ...mockUser.traits,
+        logins: ['apple', 'banana'],
+      },
+    });
+
+    let r: RenderResult;
+    await act(async () => {
+      r = render(Component);
+    });
+
+    // Test init logins to be rendered.
+    let checkboxes: NodeListOf<HTMLInputElement> =
+      r.container.querySelectorAll('input:checked');
+    expect(checkboxes).toHaveLength(2);
+    expect(checkboxes[0].name).toBe('apple');
+    expect(checkboxes[1].name).toBe('banana');
+
+    // Test init logins to be rendered with a new login name.
+    fireEvent.click(screen.getByText(/add new/i));
+    const inputEl = screen.getByPlaceholderText('name');
+    fireEvent.change(inputEl, { target: { value: 'carrot' } });
+    fireEvent.click(screen.getByText('Add'));
+
+    checkboxes = r.container.querySelectorAll('input:checked');
+    expect(checkboxes).toHaveLength(3);
+    expect(checkboxes[0].name).toBe('apple');
+    expect(checkboxes[1].name).toBe('banana');
+    expect(checkboxes[2].name).toBe('carrot');
+  });
+});
+
+const mockUser: User = {
+  name: 'foo',
+  roles: [],
+  traits: {
+    logins: [],
+    dbUsers: [],
+    dbNames: [],
+    kubeUsers: [],
+    kubeGroups: [],
+    windowsLogins: [],
+    awsRoleArns: [],
+  },
+};
--- a/packages/teleport/src/Discover/LoginTrait/LoginTrait.test.tsx
+++ b/packages/teleport/src/Discover/LoginTrait/LoginTrait.test.tsx
@@ -15,8 +15,7 @@
  */
 
 import React from 'react';
-// eslint-disable-next-line import/named
-import { RenderResult } from '@testing-library/react';
+import type { RenderResult } from '@testing-library/react';
 import { render, screen, act, fireEvent } from 'design/utils/testing';
 import type { User } from 'teleport/services/user';
 import { DiscoverContext } from '../discoverContext';
@@ -106,8 +105,8 @@ const mockUser: User = {
   roles: [],
   traits: {
     logins: [],
-    dbUsers: [],
-    dbNames: [],
+    databaseUsers: [],
+    databaseNames: [],
     kubeUsers: [],
     kubeGroups: [],
     windowsLogins: [],
--- a/packages/teleport/src/Discover/LoginTrait/LoginTrait.tsx
+++ b/packages/teleport/src/Discover/LoginTrait/LoginTrait.tsx
@@ -14,14 +14,20 @@
  * limitations under the License.
  */
 
-import React, { useState } from 'react';
-import { Flex, ButtonPrimary, ButtonText, Text, Box, Indicator } from 'design';
+import React, { useState, useRef } from 'react';
+import {
+  Flex,
+  ButtonPrimary,
+  ButtonText,
+  Text,
+  Box,
+  Indicator,
+  Input,
+} from 'design';
 import { Danger } from 'design/Alert';
 import * as Icons from 'design/Icon';
-import FieldInput from 'shared/components/FieldInput';
-import Validation from 'shared/components/Validation';
 import { Header, CancelButton } from '../Shared';
-import type { AgentStepProps } from '../Shared';
+import type { AgentStepProps } from '../types';
 import { useDiscoverContext } from '../discoverContextProvider';
 import { useLoginTrait, State } from './useLoginTrait';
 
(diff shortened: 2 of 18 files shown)


packages/teleport/src/Discover/LoginTrait/LoginTrait.tsx
```diff
@@ -0,0 +1,169 @@
...
+                      setNewLogin('');
+                      setShowInputBox(false);
+                    }}
+                  >
+                    Add
+                  </ButtonPrimary>
+                </Validation>
+              </Flex>
+            )}
+            {!showInputBox && (
```
ryanclark, line 130: This could be `||` instead as we're checking `showInputBox` above

packages/teleport/src/Discover/LoginTrait/LoginTrait.tsx
```diff
@@ -0,0 +1,169 @@
...
+            )}
+          </Box>
+          <ButtonPrimary
+            width="165px"
+            onClick={() => {
+              let checkedLogins = Object.keys(loginMap).filter(login =>
+                loginMap[login] ? login : false
+              );
+              nextStep(checkedLogins);
+            }}
```
ryanclark, line 159: I'd move this into a callback using `React.useCallback` -

```typescript
const handleClick = React.useCallback(() => {
  // etc...
}, [loginMap, nextStep]);
```

packages/teleport/src/Discover/LoginTrait/useLoginTrait.ts
```diff
@@ -0,0 +1,77 @@
...
+import { useState, useEffect } from 'react';
+import useAttempt from 'shared/hooks/useAttemptNext';
+import type { User } from 'teleport/services/user';
+import { DiscoverContext } from '../discoverContext';
+import type { AgentStepProps } from '../Shared';
+
+export function useLoginTrait({ ctx, props }: Props) {
+  const [user, setUser] = useState<User>();
+  const { attempt, run, setAttempt, handleError } = useAttempt('processing');
+  const [loginMap, setLoginMap] = useState<Record<string, boolean>>({});
```
ryanclark, line 26: Can we use a map here? `Map<string, boolean>`

packages/teleport/src/Discover/LoginTrait/useLoginTrait.ts
```diff
@@ -0,0 +1,77 @@
...
+  function nextStep(logins: string[]) {
+    setAttempt({ status: 'processing' });
+    ctx.userService
+      .updateUser({
+        ...user,
+        traits: { ...user.traits, logins },
+      })
+      .then(props.nextStep)
+      .catch(handleError);
+  }
```
ryanclark, line 48: This could use async/await for a nicer flow

```typescript
  async function nextStep(logins: string[]) {
    setAttempt({ status: 'processing' });
    
    try {
      const user = await ctx.userService.updateUser({ ...user, etc });
      
      props.nextStep(user);
    } catch (err) {
      handleError(err);
    }
  }
```

packages/teleport/src/services/user/makeUser.ts
```diff
@@ -14,16 +14,26 @@
  * limitations under the License.
  */
 
-import { at } from 'lodash';
 import { User } from './types';
 
 export default function makeUser(json): User {
-  const [name, roles, authType] = at(json, ['name', 'roles', 'authType']);
+  json = json || {};
```
ryanclark, line 20:
```typescript
export default function makeUser(json = {}): User {
  const { ... etc } = etc;
}
```

Also can we get a type for `json`?

packages/teleport/src/Discover/LoginTrait/LoginTrait.tsx
ryanclark: Also, if you use a map instead of a record here, you can do something like

```
const checkedLogins = [...loginMap.entries()]
  .filter(([, value]) => value)
  .map(([key]) => key)
```

It's a bit nicer to iterate on a map than the keys of an object

packages/teleport/src/services/user/makeUser.ts
kimlisa: the `json` param does not require typing. in our `services/*` any `makeXXX` funcs are typically used to process a untyped api response and return it typed (we just call it `json` maybe not the best naming)

packages/teleport/src/Discover/LoginTrait/useLoginTrait.ts
kimlisa: tbh, i think the chaining is cleaner and easier to read :sweat_smile:  (maybe it's b/c i'm so used to it). but i am open to discussion with the team to vote for writing things a certain way for consistency sake, otherwise i'm keeping it as is atm

packages/teleport/src/Discover/LoginTrait/LoginTrait.tsx
kimlisa: hm, do you mean `{showInputBox || ...`? If yes, it looks a bit confusing to me, and will keep `&&`

packages/teleport/src/Discover/LoginTrait/LoginTrait.tsx
kimlisa: > I'd move this into a callback using React.useCallback -

hm.. i don't agree in using `useCallback` here, the logic is not intensive enough to warrant a optimizing hook and i don't expect the list of user names to be gigantic. 

i'd be curious in a third opinion though b/c when to use these optimizing hooks is confusing and i typically reach for them (rarely) when "is it computationally intensive? or can the performance problem be measured or is visibly problematic?". Opinions? @hatched @ravicious @gzdunek

packages/teleport/src/Discover/LoginTrait/LoginTrait.tsx
ryanclark: You can chain onto the `{showInputBox && (` above by either `{showInputBox && (<element>) || (<other>)}` or `{showInputBox ? (<element>) : (<other>)}`

packages/teleport/src/services/user/makeUser.ts
ryanclark: Can you make it the `any` type then? Not typing it adds more technical debt for if we enable strict mode

packages/teleport/src/Discover/LoginTrait/LoginTrait.tsx
ryanclark: I agree that the performance of it is barely noticeable, I just prefer to keep the business logic for a handler out of the JSX code.

Note that not using inline functions is quite a popular approach, there's an [ESLint rule](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md) for it too.

Not gonna push on this as it comes down to personal preference really

packages/teleport/src/Discover/LoginTrait/useLoginTrait.ts
ryanclark: Ah see I find async/await easier to read, and it's more flexible when doing more with promises - https://dev.to/gafi/7-reasons-to-always-use-async-await-over-plain-promises-tutorial-4ej9

Happy for the team to chime in on this

packages/teleport/src/services/user/makeUser.ts
kimlisa: the `any` sounds reasonable to me

packages/teleport/src/Discover/LoginTrait/useLoginTrait.ts
ryanclark: After thinking about this, we might be able to use `Set<string>` instead - do we need to keep the entries where the value is `false`? If not, we can use a set and remove the value if it's set to false

packages/teleport/src/Discover/LoginTrait/LoginTrait.tsx
hatched: I typically try and keep function definitions out of the `render` section of a react component for legibility but in this specific case it's small enough I'm not super concerned if you felt strongly about leaving it here.

Now as far as a map is concerned....generally you don't want to use user defined input as keys as you could end up with some unintended consequences but using a `Map` would give you some protections against this.

For this specific case however....outside of maintaining UI state I don't see the `checked` logic being used for anything before the 'form' is to be submitted. So I would let the browser handle the state. By this I mean wrap the inputs in a form element with a `ref`. Then use "uncontrolled" components with a defined `name` for the checkboxes and only on "next" grab the ref and parse the values into an array of values `['root', 'ubuntu', ...]` and store that.

packages/teleport/src/Discover/LoginTrait/useLoginTrait.ts
hatched: If you take my suggestion above then this becomes
```suggestion
  const [loginMap, setLoginMap] = useState<string[]>([]);
```

packages/teleport/src/Discover/LoginTrait/useLoginTrait.ts
hatched: Let's leave this for a discussion out of the scope of this PR as convention in the project is to use chaining.

packages/teleport/src/Discover/LoginTrait/useLoginTrait.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  function toggleLoginSelect(login: string) {
```
hatched, line 50: These two functions can disappear if you use my suggestion above.

packages/teleport/src/Discover/Shared.tsx
```diff
@@ -0,0 +1,64 @@
...
+    mr={3}
+    mt={3}
+    width="165px"
+    onClick={() => history.push(cfg.routes.root, true)}
+  >
+    Go To Dashboard
+  </ButtonSecondary>
+);
+
+export type AgentStepProps = {
```
hatched, line 44: Can you move this out into a `types.ts` file plz. 

It's a bit odd to be mixed in with unrelated components.

packages/teleport/src/services/user/makeUser.ts
hatched: Yeah I've been asking for us to type these requests and responses but that'll be a larger longer running undertaking. I've also previously brought up the concern around `json` naming as it's not json 😄 but it's one of those conventions we should discuss and then change all at once.

packages/teleport/src/Discover/LoginTrait/LoginTrait.tsx
kimlisa: > I would let the browser handle the state. By this I mean wrap the inputs in a form element with a ref. Then use "uncontrolled" components with a defined name for the checkboxes and only on "next" grab the ref and parse the values into an array of values ['root', 'ubuntu', ...] and store that.

@hatched that was a brilliant suggestion, simplified a lot of things

packages/teleport/src/Discover/LoginTrait/LoginTrait.tsx
```diff
@@ -0,0 +1,169 @@
...
+                    `}
+                  >
+                    {login}
+                  </label>
+                </Flex>
+              );
+            })}
+            {showInputBox && (
+              <Flex alignItems="end" mt={3}>
+                <Validation>
```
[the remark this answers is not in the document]
kimlisa: leftover, removed

packages/teleport/src/Discover/LoginTrait/LoginTrait.test.tsx
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+// eslint-disable-next-line import/named
```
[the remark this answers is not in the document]
kimlisa: tbh, i have no idea, trivial at the moment since it's in a test file, so i'll leave it be

packages/teleport/src/Discover/LoginTrait/LoginTrait.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+        />
+      </ContextProvider>
+    );
+  });
+
+  test('add a new login', async () => {
+    jest.spyOn(userSvc, 'fetchUser').mockResolvedValue(mockUser);
+
+    let r: RenderResult;
+    await act(async () => {
```
JanKaczmarkiewicz, line 52: render is internally wrapped in act so we don't have to do this manually
https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#wrapping-things-in-act-unnecessarily

packages/teleport/src/Discover/LoginTrait/LoginTrait.tsx
```diff
@@ -0,0 +1,169 @@
...
+  return (
+    <Box>
+      <Header>Set Up Access</Header>
+      {attempt.status === 'processing' && (
+        <Box textAlign="center" m={10}>
+          <Indicator />
+        </Box>
+      )}
+      {attempt.status === 'failed' && <Danger>{attempt.statusText}</Danger>}
+      {attempt.status === 'success' && (
```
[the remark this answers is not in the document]
JanKaczmarkiewicz: also there is also another way to express that using object literals
```
const content = {
   processing: <Box textAlign="center">...etc</Box>,
   failed: ... ,
   ...
}[attempt.status]
```

This compact syntax comes with slight perf hit since all "branches" needs to allocate.

packages/teleport/src/Discover/LoginTrait/LoginTrait.tsx
hatched: Haha thanks! Happy to help 😁

hatched (approved):
Thanks for the updates @kimlisa I think this is good enough to land so we can continue iterating.

ryanclark (approved):

packages/teleport/src/Discover/LoginTrait/LoginTrait.test.tsx
kimlisa: i found that you need them with async calls

packages/teleport/src/Discover/LoginTrait/LoginTrait.tsx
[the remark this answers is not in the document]
alex-kovoy: Using flags for hidding/showing components is a way to keep JSX scoped in
one place in regards to layout. This also makes it more readable as it reduces 
the number of "jumps" to see what supposed to be rendered (bigger picture). 
It allows people to read JSX from up to bottom in one flow as it does not break the scope (more declarative). 

Essentially this is a shortcut for this:

```
<Box visible={true}>
  <Content/>
</Box>
``` 

which becomes

```
{ flag &&  (  <Content/> )}
```

and which is better than just: 

```
 {$content} // provides no information about what and how supposed to be rendered
```

One thing to keep in mind is to avoid complex expressions and nested conditions. Usually it means refactoring the code to a separate component by declaratively passing props to it.

PR #995, opened 2022-07-19
Fix bug caused by having no participants in a Kube session recording

rudream:
## Purpose

Fixes a bug that caused the "Session Recordings" tab to break with a `Cannot read properties of null (reading 'join')` error when there was a Kube session with en empty participants list. 

## Implementation

We defaulted the `participants` to `[]` when destructuring the object, however this only works for `undefined` values, and not for `null` values which is what is returned in this case.

### DIFF
--- a/packages/teleport/src/services/recordings/makeRecording.ts
+++ b/packages/teleport/src/services/recordings/makeRecording.ts
@@ -41,7 +41,7 @@ function makeDesktopRecording({
 }
 
 function makeSshOrKubeRecording({
-  participants = [],
+  participants,
   time,
   session_start,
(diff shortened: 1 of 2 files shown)


kimlisa (commented):
thanks for fix! can you add a test for this file, i forgot to add one which would've caught it, shame on me :sweat_smile:

zmb3 (approved):

zmb3:
Please backport to v10 and v9.

kimlisa (approved):

kimlisa (approved):

ryanclark (approved):

PR #996, opened 2022-07-19
`SharedDirectoryInfoResponse`

ibeckermayer:
Besides adding the plumbing for [`SharedDirectoryInfoResponse`](https://github.com/gravitational/teleport/blob/6aa7c239412b3677efce972941e85755b483ab57/rfd/0067-desktop-access-file-system-sharing.md?plain=1#L222-L234), this PR also creates a new class `SharedDirectoryManager` for managing access into the `FileSystemDirectoryHandle` that represents the directory being shared. It is written in such a way that it will be easy to extend to share multiple directories at once.

Unfortunately I wasn't able to test it with `Jest`, see the Slack thread [here](https://gravitational.slack.com/archives/C02DQ1C2BMW/p1658201986124609) for details. I did however manually hardcode various paths to confirm that `SharedDirectoryManager.walkPath` is working as expected.

Requires backport to v9/v10

#### `webapps`

## What to expect

image

Navigating to the shared drive should give you

image


packages/teleport/src/lib/tdp/sharedDirectoryManager.ts
```diff
@@ -0,0 +1,119 @@
...
+    if (this.dir) {
+      throw new Error(
+        'SharedDirectoryManager currently only supports sharing a single directory'
+      );
+    }
+    this.dir = sharedDirectory;
+  }
+
+  getName(): string {
+    this.checkReady();
```
hatched, line 31: As discussed offline these functions will be updated to use [jsdoc throws](https://jsdoc.app/tags-throws.html) and this call will add a catch.

hatched (approved):

packages/teleport/src/lib/tdp/sharedDirectoryManager.ts
ibeckermayer: https://github.com/gravitational/webapps/pull/1010

kimlisa (approved):

PR #999, opened 2022-07-19
`SharedDirectoryListRequest`

ibeckermayer:
Requires backport to v9/v10

#### `webapps`

## What to expect

image

Navigating to the shared drive should give you

image

in the console.


hatched (approved):

packages/teleport/src/lib/tdp/codec.ts
```diff
@@ -41,6 +41,7 @@ export enum MessageType {
   SHARED_DIRECTORY_ACKNOWLEDGE = 12,
   SHARED_DIRECTORY_INFO_REQUEST = 13,
   SHARED_DIRECTORY_INFO_RESPONSE = 14,
+  SHARED_DIRECTORY_LIST_REQUEST = 25,
```
kimlisa, line 44: is the jump in number expected? (just curious)

kimlisa (approved):

packages/teleport/src/lib/tdp/codec.ts
ibeckermayer: Yeah, these codes were doled out sequentially as I was writing the [spec](https://github.com/gravitational/teleport/blob/master/rfd/0067-desktop-access-file-system-sharing.md). Eventually the other numbers will get filled in.

PR #1000, opened 2022-07-20
`SharedDirectoryListResponse`

ibeckermayer:
Adds `listContents` to get the information about each item in a directory and send it back in the `fsoList` of the [`SharedDirectoryListResponse`](https://github.com/gravitational/teleport/blob/a0fd84793b52dc58d326bdfcb54d0f7e3a8acae0/rfd/0067-desktop-access-file-system-sharing.md?plain=1#L427-L442).

I initially tried to create a version of `listContents` which [gathered all of the promises in a list and then returned them wrapped in a `Promise.all`](https://github.com/gravitational/webapps/commit/e7551ced6337e61eb80645ffab7faf6b069384af#diff-6eedaa848a12f318902ec3bcbd03172ccb01660fa22e3419cfb4aa832bad7515R58-R82). However with this structure, sharing a directory with a large number of items (like the `node_modules` directory of this repo after `yarn install`) would cause the system to freeze up for a while. 

I think its a similar problem to the one I encountered way back when working on the display pipeline, namely that piling up a lot of asynchronous calls clogs up the microtask queue and can make performance appear worse, because new regular tasks can't be executed while all the microtasks are processed.

Performance still isn't fantastic when sharing a large directory, but its good enough to be useable.

Requires backport to v9/v10

#### `webapps`

## What to expect

image

image

Trying to open a file should result in a

image


JanKaczmarkiewicz:
> I initially tried to create a version of listContents which [gathered all of the promises in a list and then returned them wrapped in a Promise.all](https://github.com/gravitational/webapps/commit/e7551ced6337e61eb80645ffab7faf6b069384af#diff-6eedaa848a12f318902ec3bcbd03172ccb01660fa22e3419cfb4aa832bad7515R58-R82). However with this structure, sharing a directory with a large number of items (like the node_modules directory of this repo after yarn install) would cause the system to freeze up for a while.
> 
> I think its a similar problem to the one I encountered way back when working on the display pipeline, namely that piling up a lot of asynchronous calls clogs up the microtask queue and can make performance appear worse, because new regular tasks can't be executed while all the microtasks are processed.

Do you think we could use something like promise pool eg: https://stackoverflow.com/questions/63618914/how-to-use-promise-all-to-handle-thousands-of-request?

ibeckermayer:
> We could use a promise pool eg: https://stackoverflow.com/a/63620273/12860195

@JanKaczmarkiewicz very clever idea and implementation. I tried to copy it in another branch [here](https://github.com/gravitational/webapps/blob/83af4b3e9a3748600197f689ea7a318b217ffd66/packages/teleport/src/lib/tdp/sharedDirectoryManager.ts#L152-L182). Our system is complicated by the fact that `dir.values()` gives us an `AsyncIterableIterator`, so we are most naturally inclined to create an [`AsyncGenerator`](https://github.com/gravitational/webapps/blob/83af4b3e9a3748600197f689ea7a318b217ffd66/packages/teleport/src/lib/tdp/sharedDirectoryManager.ts#L72-L84). When I attempted to run the code as written, enumerating the shared directory broke the tab. 

There may yet be a way to use this, but I'm going to kick the can down the road. Added this discussion thread to the relevant issue: https://github.com/gravitational/webapps/issues/1001

hatched (approved):

kimlisa (approved):

PR #1003, opened 2022-07-20
`SharedDirectoryReadRequest`

ibeckermayer:
Adds `SharedDirectoryReadRequest`

Requires backport to v9/v10

#### `webapps`

## What to expect

image

image

Trying to open a file should result in something like the following in the debug console

image


hatched (approved):

kimlisa (approved):

packages/teleport/src/lib/tdp/codec.ts
```diff
@@ -682,6 +693,37 @@ export default class Codec {
     };
   }
 
+  // | message type (19) | completion_id uint32 | directory_id uint32 | path_length uint32 | path []byte | offset uint64 | length uint32 |
```
JanKaczmarkiewicz, line 696: [out of scope, discussion]
If we will add more request types I think we should create some simple utility:
```ts
const sharedDirectoryReadRequestSchema = schema({
   completion_id: uint32,
   directory_id: uint32,
   path_length: uint32,
   path: byteArray,
   offset: uint64,
   length: uint32
});

type SharedDirectoryReadRequest = ReturnType<typeof sharedDirectoryReadRequestSchema.try_parse>;

const data = sharedDirectoryReadRequestSchema.try_parse(buffer);
```

ibeckermayer: I'm not sure if I'm understanding correctly. `createSchema` is returning a class which generates this `decodeSharedDirectoryReadRequest` function (as `tryParse`)?

JanKaczmarkiewicz: something like https://codesandbox.io/s/goofy-hermann-vdrx01?file=/src/index.ts

Issue #1004, opened 2022-07-21
`SharedDirectoryManager` performance improvements


ibeckermayer:
I think it should be possible to cache each `File` by pathname when its first read, and then close it when its written to in the `SharedDirectoryManager`, in order to avoid potentially continually reading it from disk unnecessarily. 

Watch out for the possibility that big writes are sent as multiple `SharedDirectoryWriteRequest`s in a row -- maybe you don't want to remove the `File` from the cache every write, but only if its <= some max write number (look into whether there's a defined write-at-once limit within RDP to start).

ibeckermayer:
There's no maximum number of bytes that can be written to a file at a single time specified by RDP. As far as I've been able to tell, the number of bytes written at a time is determined by the program on the windows side. For example, Notepad appears to be relatively rudimentary, and will just write an entire file on every "save", no matter how big the file (up to a certain maximum size the program will handle). In contrast, the File Explorer program will limit writes (and reads) to 1MB, and run multiple in a row if a file exceeds 1MB.

It would be possible to implement an optimization like the one described in this issue, but would require us to significantly change TDP to be more similar to RDP, where the server tells the client when to create and delete file handles.

PR #1005, opened 2022-07-21
`SharedDirectoryReadResponse`

ibeckermayer:
Adds `SharedDirectoryReadResponse`

Requires backport to v9/v10

Raises https://github.com/gravitational/webapps/issues/1004

#### `webapps`

## What to expect

image

image

Trying to open a file should show you the file contents (if you open it in Notepad for example). You will be able to type into the file, but if you save the session will crash.

image

image


packages/teleport/src/lib/tdp/sharedDirectoryManager.ts
```diff
@@ -81,6 +81,26 @@ export class SharedDirectoryManager {
...
 
+  // Reads length bytes starting at offset from a file at path.
+  async readFile(
+    path: string,
+    offset: bigint,
+    length: number
+  ): Promise<Uint8Array> {
+    this.checkReady();
+
+    const fileHandle = await this.walkPath(path);
```
hatched, line 92: These awaits should probably be wrapped with a try/catch

hatched (approved):

packages/teleport/src/lib/tdp/sharedDirectoryManager.ts
ibeckermayer: Makes sense as a first impression, but actually I want them to just throw that up to the `client.ts::Client`

kimlisa (approved):

PR #1021, opened 2022-07-22
Connect: Wait for tshd gRPC server to start

ravicious:
Closes gravitational/webapps.e#124.

The problem this PR solves is described here: https://github.com/gravitational/webapps.e/issues/124#issuecomment-1192354642. I also improved the function so that if the child process exits, the function rejects immediately instead of waiting for the timeout.

---

This PR is based on the `gzdunek/add-windows-support` branch because Grzegorz made changes on that branch which massively helped with implementing this solution. As such, the code from this PR needs tsh compiled off `gzdunek/add-windows-support`.

To reproduce the problem on `gzdunek/add-windows-support`, add `time.Sleep(time.Second * 3)` just before `net.Listen` in `lib/teleterm/apiserver/apiserver.go`.

### DIFF
--- a/packages/teleterm/src/mainProcess/mainProcess.ts
+++ b/packages/teleterm/src/mainProcess/mainProcess.ts
@@ -1,13 +1,10 @@
 import { ChildProcess, fork, spawn } from 'child_process';
-
 import path from 'path';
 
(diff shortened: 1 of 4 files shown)


gzdunek (approved):
I really like the code! Clean, functional and well tested.

avatus (approved):

PR #1022, opened 2022-07-22
Add section to README about audit events

mcbattirola:
As stated in https://github.com/gravitational/webapps/pull/970#pullrequestreview-1037970565, adding a new event is very common, but involves many parts that are easy to miss.

This PR adds a new section to the root README on how to add an audit event.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,9 +187,11 @@ via `yarn workspace <workspace-name> add <package-name>`, it will create a `pack
 
 When a new event is added to Teleport, the web UI has to be updated to display it correctly:
 
-1. Add a new entry to [`eventCodes`](https://github.com/gravitational/webapps/blob/master/packages/teleport/src/services/audit/types.ts).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -182,3 +182,14 @@ the root of this repository.
 
 Keep in mind that there should only be a single `yarn.lock` in this repository, here at the top level. If you add packages
 via `yarn workspace <workspace-name> add <package-name>`, it will create a `packages/<package-name>/yarn.lock` file, which should not be checked in.
+
+### Adding an Audit Event
+
+When a new event is added to Teleport, the web UI has to be updated to display it correctly:
+
+1. Add a new entry to [`eventCodes`](https://github.com/gravitational/webapps/blob/master/packages/teleport/src/services/audit/types.ts).
```
kimlisa, line 190: we should probably use like the [latest commit](https://github.com/gravitational/webapps/tree/8a0201667f045be7a46606189a6deccdaee2fe1f) links, the master could always change either by file moving/deleting/renaming and so we will get not found error

and should we link this: https://github.com/gravitational/webapps/pull/561, as an example? i always like examples :) up to you

kimlisa (approved):
thank you for doing this :)

README.md
mcbattirola: Good ideas, thanks :)

PR #1024, opened 2022-07-22
Added eslint rule to enforce the order of file imports

hatched:
There have been a few discussions lately about this so this adds an eslint rule to enforce the order of the imports. The order that I picked fit very closely to what we were already doing without this rule so while there are changes to a lot of files the actual changes are minor.

This is an auto-fixable rule so you can enable 'auto fix' in your editor and you never have to think about this order again 😁


ravicious (approved):
Looks good but it seems that webapps.e needs to be formatted as well. Could you also add autogenerated protobuf files (`**/*_pb.*`) to ignored files and revert changes made to them?

---

Personally I never felt too strong about it but I have to say in files where there's a lot of imports it helps to strengthen [the "narrative"](https://practicingruby.com/articles/confident-ruby#losing-the-plot) that the file provides. For example, I like the changes to mainProcess.ts: [before](https://github.com/gravitational/webapps/blob/5d7fa2c473c4d3f65a02c3a1e8e5d3d3b5e7b300/packages/teleterm/src/mainProcess/mainProcess.ts) & [after](https://github.com/gravitational/webapps/blob/c7668e849659de6e9fda80d6c091499f28c5d682/packages/teleterm/src/mainProcess/mainProcess.ts).

That extra space between groups seems like a good choice. Besides, we manually order import groups in the Go code so it makes sense to have this for our JS files as well.

hatched:
Thanks for the comment @ravicious but this PR should be marked as a draft :) Doing so now as you mentioned `webapps.e` needs to be done first and some additional tweaks, like spaces between the groups, still needs to be added.

ryanclark (approved):

PR #1025, opened 2022-07-24
Add protocol interceptor

avatus:
Closes: https://github.com/gravitational/teleport-private/issues/154

-------
#### The `index.html` file is only served over the `file://` protocol in a packaged build. If its in dev mode, the app will be loaded over https from the webserver and this intercept handler will not be called. I believe you can still test the vulnerability in dev mode as I've prescribed below, but all my testing was done on a packaged build
-------

Problem:
In a production build, electron will load files with the `loadFile` api pretty much unrestricted. This could mean that, if a user was able to inject malicious scripts to the loaded content, they could read the files of the local computer. 

Solution:
A few parts to this solution, the main one being the `interceptFileProtocol` handler. This will take any file protocol request and provide a new handler which we can use to sanitize and validate the request. the TL;DR of the function is making sure the requested file is in the install path or user data path of the app and if not, deny.

The second part of the solution is disabling the other protocols since we do not use them but thats pretty straight forward

Notes:
I passed the logger into the `interceptFileProtocol` function itself from main.ts, but I don't know what standard procedure would be here. Happy to update it or create a new logger? Let me know. 

If you wanna reproduce the vuln, the steps in the audit will work but an easier way would be
1. open console
2. create an iframe element `var iframe = document.createElement('iframe')`
3. set src to some system file like `iframe.src = 'file:///etc/hosts'
4. select any element on the screen and append the iframe child to the element, `var thing = document.querySelector('.sc-fzokvW')` and then `thing.appendChild(iframe)`

Before this change, your file will be posted into the document. A more sophisticated attack could allow this information to be sent back to an attackers server or whatever. After this change, the requested file is denied.


packages/teleterm/src/mainProcess/protocolHandler.ts
```diff
@@ -0,0 +1,88 @@
...
+  if (index < 0) {
+    return string;
+  }
+
+  return string.slice(0, index);
+}
+
+function urlToPath(targetUrl, { isWin }) {
+  const decoded = decodeURIComponent(targetUrl);
+  const withoutScheme = decoded.slice(isWin ? 8 : 7);
```
ravicious, line 46: Could you add a comment explaining why `isWin` is needed here?

packages/teleterm/src/mainProcess/protocolHandler.ts
```diff
@@ -0,0 +1,88 @@
...
+  logger,
+}) {
+  protocol.interceptFileProtocol('file', (request, callback) => {
+    const target = path.normalize(urlToPath(request.url, { isWin }));
+    const realPath = fs.existsSync(target) ? fs.realpathSync(target) : target;
+    const properCasing = isWin ? realPath.toLowerCase() : realPath;
+    if (
+      !properCasing.startsWith(
+        isWin ? userDataPath.toLowerCase() : userDataPath
+      ) &&
```
ravicious, line 70: What's the reason behind having to call `toLowerCase` on Windows?

What do you think about moving those transformations out of the conditional, so we'd have something like:

```typescript
const userDataPath = isWin ? userDataPath.toLowerCase() : userDataPath
```

packages/teleterm/src/mainProcess/protocolHandler.ts
```diff
@@ -0,0 +1,88 @@
...
+      logger.error(
+        `Denying request to path '${realPath}' (userDataPath: '${userDataPath}', installPath: '${installPath}'`
+      );
+      return callback();
+    }
+
+    if (!path.isAbsolute(realPath)) {
+      logger.error(`Denying request to non-absolute path '${realPath}'`);
+      return callback();
+    }
```
ravicious, line 82: What would be an example of a request which passes the check for installPath & userPath but doesn't pass this one? It might be good to add that as a comment.

packages/teleterm/src/mainProcess/protocolHandler.ts
```diff
@@ -0,0 +1,88 @@
...
+    if (
+      !properCasing.startsWith(
+        isWin ? userDataPath.toLowerCase() : userDataPath
+      ) &&
+      !properCasing.startsWith(isWin ? installPath.toLowerCase() : installPath)
+    ) {
+      logger.error(
+        `Denying request to path '${realPath}' (userDataPath: '${userDataPath}', installPath: '${installPath}'`
+      );
+      return callback();
```
ravicious, line 76: Do you happen to know what's the difference in behavior when calling `callback` with an error vs without? I tried to check it by requesting `/etc/passwd` in dev mode but I couldn't get any errors to show up.

packages/teleterm/src/mainProcess/protocolHandler.ts
```diff
@@ -0,0 +1,88 @@
...
+  const withoutHash = eliminateAllAfterCharacter(withoutQuerystring, '#');
+
+  return withoutHash;
+}
+
+// intercept, clean, and validate the requested file path.
+export function interceptFileProtocol({
+  protocol,
+  isWin,
+  userDataPath,
```
ravicious, line 59: I think the renderer doesn't need access to `userDataPath`. I removed the check for this dir from this function and everything seemed to work.

packages/teleterm/src/main.ts
```diff
@@ -55,6 +60,16 @@ app.whenReady().then(() => {
     });
   }
 
+  const isWin = process.platform === 'win32';
+
+  const userDataPath = fs.realpathSync(app.getPath('userData'));
+  const installPath = fs.realpathSync(app.getAppPath());
```
ravicious, line 66: Don't the `app.getPath`/`app.getAppPath` already return full paths?

packages/teleterm/src/mainProcess/protocolHandler.ts
avatus: The errors would only happen in production. In dev, files are still requested over https (from the web server) and only in production do files get requested with the `file://` protocol

packages/teleterm/src/mainProcess/protocolHandler.ts
avatus: Will add comment. For conversation, `isWin` is used to slice the files names slightly differently based on their shape.
Windows: `file:///C:/path/to/file`
unix: `file:///path/to/file`

slicing by 8 on windows is needed to get to that root `C:` drive (or whatever drive they need)

packages/teleterm/src/mainProcess/protocolHandler.ts
avatus: unix is sensitive and windows isn't when it comes to file names. So this was just to make sure the check was always happening on the same casing but... I suppose it isn't needed. a few things in this PR were 'super duper safety checks'.

I'm down to keep it but to move it out of the conditional 👍

packages/teleterm/src/mainProcess/protocolHandler.ts
avatus: To answer the first question about callbacks, the error is not needed in the callback. The callback requires an object with a `path` key to be valid, if it doesn't receive one, the request won't work. I will add an error code to them all to make sure it's clear we are NOT returning a successful callback

packages/teleterm/src/mainProcess/protocolHandler.ts
avatus: This was supposed to be above that conditional but I must have missed it in a refactor. Thank you! To answer your question, I have no idea what it would have been because its supposed to be above 😅 , but I think making the chest _before_ `installPath` and `userDataPath` makes much more sense. Thank you

packages/teleterm/src/mainProcess/protocolHandler.ts
avatus: True, was more 'future proofing' just in case. Should I remove it?

packages/teleterm/src/main.ts
avatus: In testing, yes. 
The docs say `Returns string - The current application directory.`.  so I wasn't 100% sure it would always return so a bit cautious here.

packages/teleterm/src/main.ts
```diff
@@ -55,6 +60,16 @@ app.whenReady().then(() => {
     });
   }
 
+  const isWin = process.platform === 'win32';
```
ravicious, line 63: You should be able to read platform from `mainProcess.settings`. Long term I think we should stick to a single way of obtaining platform details.

packages/teleterm/src/main.ts
ravicious: What do you think about getting rid of it and addressing it if it ever becomes a problem?

packages/teleterm/src/mainProcess/protocolHandler.ts
ravicious: FWIW, you could also do `new URL()` and then `url.pathname` so that you only need to get rid of a single character.

packages/teleterm/src/mainProcess/protocolHandler.ts
ravicious: Yeah, I say let's remove it.

packages/teleterm/src/mainProcess/protocolHandler.ts
ravicious: Makes sense, I'd also write that down in a comment.

packages/teleterm/src/mainProcess/protocolHandler.ts
ravicious: Yeah that error -3 was pretty nifty but I noticed this one was blank.

I was testing this on the packaged built, but when I added a disabled handler under `file` the app wasn't showing any errors. Where should they show up?

I'm mostly worried about future devs running into this and not knowing what's happening.

avatus: The error was showing for me in `main.log`
example when trying to reference my `/etc/hosts` from the iframe
```
[24-07-22 14:47:16] [Main] error: Denying request to path '/private/etc/hosts' (userDataPath: '/Users/michael/Library/Application Support/Teleport Connect', installPath: '/Users/michael/code/gravitational/webapps/packages/teleterm/build/release/mac-arm64/Teleport Connect.app/Contents/Resources/app.asar'
```

avatus: I think I understand you're question now. Are you asking "where does the `-3` error show up?"? I'm not sure. I defaulted to returning the 'proper' error code to electron and then using our own error logging in tandem to make sure we see it (in `main.log`). Are there other places we should be seeing logs?

ravicious: Ah, yeah, I just realized that `interceptFileProtocol` will show up in logs. I just realized where my mistake was – I was testing this by adding `protocol.interceptFileProtocol('mailto', _disabledHandler)` but `_disabledHandler` doesn't really have any logging. Could you add it there?

packages/teleterm/src/mainProcess/protocolHandler.ts
```diff
@@ -58,28 +58,30 @@ function urlToPath(targetUrl, { isWin }) {
...
   logger,
 }) {
   protocol.interceptFileProtocol('file', (request, callback) => {
     const target = path.normalize(urlToPath(request.url, { isWin }));
-    const realPath = fs.existsSync(target) ? fs.realpathSync(target) : target;
-    const properCasing = isWin ? realPath.toLowerCase() : realPath;
+    const realPathWithCasing = fs.existsSync(target)
+      ? fs.realpathSync(target)
+      : target;
+    // normalize casing across windows (case-insensitive) and unix (case-sensitive)
```
ravicious, line 69: Isn't macOS case-insensitive by default? Tbh it's the first time I see case normalization being done when looking at the paths. I'm not sure if we should keep it or remove it.

packages/teleterm/src/mainProcess/protocolHandler.ts
```diff
@@ -0,0 +1,93 @@
...
+  protocol as ElectronProtocol,
+  ProtocolRequest,
+  ProtocolResponse,
+} from 'electron';
+import * as path from 'path';
+import fs from 'fs';
+
+type CallbackType = (response: string | ProtocolResponse) => void;
+
+function _disabledHandler(
```
ravicious, line 11: Could you rename it to `disabledHandler`? We don't follow the `_private` convention within. I think it's not even used for unused args here (like `_request` below).

ravicious (approved):
Looks good. I added some nits that I should've added right from the start rather than this late in the review process, sorry for that.

packages/teleterm/src/mainProcess/protocolHandler.ts
avatus: Yeah for sure. This was straight from the audit and not really a conscious decision so I appreciate the catch

packages/teleterm/src/mainProcess/protocolHandler.ts
avatus: I'm ok with removing it and handling later if needed to keep the code cleaner if thats ok with you

packages/teleterm/src/mainProcess/protocolHandler.ts
avatus: And honestly, since we are only pulling in the index.html file that we are casing ourselves, we shouldn't run into this.

avatus: Now that I think about it, rather than passing a logger to a bunch of different handler functions, this `_disabledHanlder` was only calling the callback, so I removed it all together and just disabled the schemes with a `forEach`. Much cleaner now

avatus:
I know you approved @ravicious but I made a slightly significant refactor in the `disabledHandler` (as in, I removed it and put inside a foreach) that i'd like a second review on before we get this in.

packages/teleterm/src/main.ts
```diff
@@ -55,6 +59,16 @@ app.whenReady().then(() => {
     });
   }
 
+  const isWin = mainProcess.settings.platform === 'win32';
+
+  const installPath = app.getAppPath();
+  interceptFileProtocol({ protocol, isWin, installPath, logger });
```
gzdunek, line 65: Hmm would it be possible to export only one function from the module, like `enableWebHandlersProtection()`? Then I would rename `installWebHandler` to `blockUnusedProtocols`. 
From looking at the functions calls it's hard to tell why we need them and what they do 🤔

Moreover, you don't have to pass `protocol` and `logger` to the function. You can just import `protocol`, because you're in the scope of the main process. 
`logger` can be created in the file directly (`const logger = new Logger('WebHandlerProtection')`), but it's up to you.

gzdunek (commented):
I left some comments, sorry for the late review :(

packages/teleterm/src/mainProcess/protocolHandler.ts
```diff
@@ -0,0 +1,84 @@
...
+// intercept, clean, and validate the requested file path.
+export function interceptFileProtocol({
+  protocol,
+  isWin,
+  installPath,
+  logger,
+}) {
+  protocol.interceptFileProtocol('file', (request, callback) => {
+    const target = path.normalize(urlToPath(request.url, { isWin }));
+    const realPath = fs.existsSync(target) ? fs.realpathSync(target) : target;
```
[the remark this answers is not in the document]
ravicious: Wouldn't this allow the attacker to probe if a given path exists on the filesystem by measuring the response time?

ravicious (approved):
The change with `forEach` looks good. I think it's worth using that Node function for converting file to path that Grzegorz suggested.

packages/teleterm/src/mainProcess/protocolHandler.ts
avatus: I love this!

packages/teleterm/src/main.ts
avatus: exporting just one function and moving to the default node url method removed about half the code. THANK YOU!!

packages/teleterm/src/mainProcess/protocolHandler.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    if (!path.isAbsolute(realPath)) {
+      logger.error(`Denying request to non-absolute path '${realPath}'`);
+      return callback({ error: -3 });
+    }
+
+    if (!realPath.startsWith(installPath)) {
+      logger.error(
+        `Denying request to path '${realPath}' (installPath: '${installPath}'`
```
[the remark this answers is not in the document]
avatus: how about `Not in installPath ${installPath}`? make the reason more clear I think

packages/teleterm/src/mainProcess/protocolHandler.ts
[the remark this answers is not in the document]
gzdunek: If so, isn't the first problem a conditional call to `fs.realpathSync`? But we can leave it as it is, that early return doesn't improve the function significantly.

packages/teleterm/src/mainProcess/protocolHandler.ts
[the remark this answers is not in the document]
gzdunek: Fine, but why does it start with a capital letter?  
And please add `)` at the end.

packages/teleterm/src/main.ts
gzdunek: 💪
I was a bit imprecise when it comes to naming (I used single form in `new Logger('WebHandlerProtection')`) , let's name the function enableWebHandler**s**Protection, because we are handling multiple protocols. 
And also, I think, I liked that separation into two functions, it's easier to read, so I would see it this way:

```typescript
export function enableWebHandlersProtection({
  installPath,
}: {
  installPath: string;
}) {
  interceptFileProtocol();
  blockUnusedProtocols();
}
``` 
But this is a matter of preference, so I leave it to you.

packages/teleterm/src/main.ts
```diff
@@ -1,7 +1,8 @@
 import { spawn } from 'child_process';
-import { app, globalShortcut, shell } from 'electron';
+import { app, globalShortcut, shell, protocol } from 'electron';
```
gzdunek, line 2: Unused `protocol`

packages/teleterm/src/mainProcess/protocolHandler.ts
```diff
@@ -0,0 +1,55 @@
...
+  'gopher',
+  'javascript',
+  'mailto',
+];
+
+const logger = new Logger('WebHandlerProtection');
+
+// intercept, clean, and validate the requested file path.
+export function enableWebHandlerProtection({
+  installPath,
```
gzdunek, line 24: I've just realized that we don't need that argument too, we can use `app.getAppPath()` directly.

gzdunek (approved):
LGTM, I left some nits.

gzdunek (approved):

packages/teleterm/src/mainProcess/protocolHandler.ts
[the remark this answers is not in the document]
ravicious: @avatus Do you remember what was the reason for `existsSinc` and `realpathSync` calls here in the first place?

While working on another part of the project that involves paths, I realized that here we could probably depend on just `path.normalize` as well, though I haven't checked that.

PR #1026, opened 2022-07-24
enable secure keyboard entry during password and OTP

avatus:
Closes: https://github.com/gravitational/teleport-private/issues/155

Problem: 
We currently do not leverage `secureKeyboardEntry` for mac devices. A quick rundown is, we can have our app 'take over' the keyboard event process to remove the chance for other processes to intercept our input. However, we must be cautious and [use secure keyboard entry fairly](https://developer.apple.com/library/archive/technotes/tn2150/_index.html)

Solution:
I've added a new ipc listener for `set-secure-keyboard-entry` and added a service for it in the `mainProcessClient` and then hooked into it with the `useClusterLogin` hook. We _could_ abstract this to an entire service on its own but, I don't see another time we'd use this functionality outside of cluster login. This is the first time I'm doing any sort of larger 'data flow' type of deal so, please let me know if you'd like this in different files/folders/exports. I don't mind switching it up!

A way to test this is using `ioreg -l -w 0 | grep kCGSSessionSecureInputPID` which will return back empty if no secure keyboard entry is enabled anywhere on the system. To test this outside of Connect,:
1. open a window of `Terminal` on macos 
2. in the top menu, choose `Terminal > Secure Keyboard Entry`. This will enable it for your window and you can again 
3. run `ioreg -l -w 0 | grep kCGSSessionSecureInputPID` and see a result. 
	-- The important key being `"kCGSSessionSecureInputPID"=YOUR_TERMINAL_PID_HERE`
4. Disable secure keyboard entry in the terminal (i forgot this step on accident and boy howdy did it take me too long to figure out what was going on!!).

Similarly, to test inside Connect, follow the steps above, but for step 3, add a `sleep 5 &&` in front of the command and then tab into the Connect password field. You should see your Connect PID is in charge now (and then unfocus the input and try again).

### DIFF
--- a/packages/teleterm/src/mainProcess/mainProcess.ts
+++ b/packages/teleterm/src/mainProcess/mainProcess.ts
@@ -108,6 +108,14 @@ export default class MainProcess {
       event.returnValue = this.settings;
     });
 
+    ipcMain.on('set-secure-keyboard-entry', (event, enabled) => {
+      const isMac = this.settings.platform === 'darwin';
+      if (isMac) {
+        app.setSecureKeyboardEntryEnabled(enabled);
+        this.logger.info(`Secure keyboard entry: ${enabled}`);
+      }
+    });
+
     subscribeToTerminalContextMenuEvent();
     subscribeToTabContextMenuEvent();
     subscribeToConfigServiceEvents(this.configService);
--- a/packages/teleterm/src/mainProcess/mainProcessClient.ts
+++ b/packages/teleterm/src/mainProcess/mainProcessClient.ts
@@ -4,6 +4,7 @@ import { MainProcessClient } from './types';
 import { createConfigServiceClient } from '../services/config';
 import { openTabContextMenu } from './contextMenus/tabContextMenu';
 import { createFileStorageClient } from 'teleterm/services/fileStorage';
+import { setSecureKeyboardEntry } from './secureKeyboardEntry';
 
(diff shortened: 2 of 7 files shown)


ravicious:
I have some… surprising news. I was curious why is it that my keyboard shortcuts stop working whenever I'm inputting passwords in Connect. Could it be that Chromium already utilizes secure keyboard entry?

Checking with `ioreg -l -w 0 | grep kCGSSessionSecureInputPID` is difficult because Terminal.app seems to acquire SKE only when the app is focused. So if I enable it in Terminal.app and then open Connect to execute the command, it doesn't return anything. But if I execute it within Terminal.app itself, I sure see the Terminal.app PID there.

So, instead of using `ioreg`, I decided to install a keylogger. The one linked to by doyensec didn't work so I used this one: https://github.com/caseyscarborough/keylogger I saw people in the issues complaining that it doesn't work with SKE, as in isn't able to listen to keystrokes when SKE is enabled – everything works as intended.

Imagine my surprise when typing in a password in Connect 10.1.2 didn't make anything show up in the logs.

Could you check if newer versions of Electron or Chromium utilize SKE already? Perhaps we don't need to add any extra code, other than using appropriate input type for passwords.

ravicious:
I just used that method with `sleep` you described and it looks that Connect already uses SKE.

avatus:
Closing: https://github.com/gravitational/teleport-private/issues/155#event-7407019560

PR #1027, opened 2022-07-25
Minor Windows fixes

gzdunek:
This PR addresses three issues:
- fixes incorrect default cwd for terminal sessions
- adds `About` menu
- makes keyboard shortcuts similar to shortcuts in a browser

### DIFF
--- a/packages/teleterm/src/mainProcess/mainProcess.ts
+++ b/packages/teleterm/src/mainProcess/mainProcess.ts
@@ -190,29 +190,46 @@ export default class MainProcess {
   private _setAppMenu() {
     const isMac = this.settings.platform === 'darwin';
 
-    const template: MenuItemConstructorOptions[] = [
-      ...(isMac ? ([{ role: 'appMenu' }] as const) : []),
-      ...(isMac ? [] : ([{ role: 'fileMenu' }] as const)),
+    const macTemplate: MenuItemConstructorOptions[] = [
+      { role: 'appMenu' },
       { role: 'editMenu' },
       { role: 'viewMenu' },
-      isMac
-        ? { role: 'windowMenu' }
-        : {
-            label: 'Window',
-            submenu: [{ role: 'minimize' }, { role: 'zoom' }],
-          },
+      {
+        label: 'Window',
+        submenu: [{ role: 'minimize' }, { role: 'zoom' }],
+      },
       {
         role: 'help',
         submenu: [
           {
             label: 'Learn More',
-            click: () => {},
+            click: () => {}, // TODO: add link to Connect docs
           },
         ],
       },
     ];
 
-    const menu = Menu.buildFromTemplate(template);
+    const otherTemplate: MenuItemConstructorOptions[] = [
+      { role: 'fileMenu' },
+      { role: 'editMenu' },
+      { role: 'viewMenu' },
+      { role: 'windowMenu' },
+      {
+        role: 'help',
+        submenu: [
+          {
+            label: 'Learn More',
+            click: () => {}, // TODO: add link to Connect docs
+          },
+          {
+            label: 'About',
+            role: 'about',
+          },
+        ],
+      },
+    ];
+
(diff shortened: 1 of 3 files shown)


packages/teleterm/src/services/config/providers/keyboardShortcutsConfigProvider.ts
```diff
@@ -50,7 +50,27 @@ export const keyboardShortcutsConfigProvider: ConfigServiceProvider<KeyboardShor
...
+        'tab-3': 'Ctrl-3',
+        'tab-4': 'Ctrl-4',
+        'tab-5': 'Ctrl-5',
+        'tab-6': 'Ctrl-6',
+        'tab-7': 'Ctrl-7',
+        'tab-8': 'Ctrl-8',
+        'tab-9': 'Ctrl-9',
+        'tab-close': 'Ctrl-W',
+        'tab-new': 'Ctrl-T',
+        'tab-previous': 'Ctrl-Shift-Tab',
```
gzdunek, line 65: @avatus I would be thankful if you could test `tab-previous` and `tab-next` on Windows. For some reason these shortcuts don't work on Parallels (I tested them in Edge).

avatus: They work great on windows 👍

packages/teleterm/src/mainProcess/mainProcess.ts
```diff
@@ -183,29 +183,46 @@ export default class MainProcess {
...
+      {
+        label: 'Window',
+        submenu: [{ role: 'minimize' }, { role: 'zoom' }],
+      },
+      {
+        role: 'help',
+        submenu: [
+          {
+            label: 'Learn More',
+            click: () => {}, // TODO: add link to Connect docs
```
ravicious, line 199: Haha okay, I was wondering why this button doesn't do anything when I was testing out the Windows build last week.

packages/teleterm/src/mainProcess/mainProcess.ts
```diff
@@ -183,29 +183,46 @@ export default class MainProcess {
...
         role: 'help',
         submenu: [
           {
             label: 'Learn More',
-            click: () => {},
+            click: () => {}, // TODO: add link to Connect docs
+          },
+          {
+            label: 'About',
+            role: 'about',
```
ravicious, line 219: At some point we should also add something like this:

```typescript
      {
        label: 'Open logs folder',
        click: async () => {
          const { shell } = require('electron')
          await shell.openPath(logsPath);
        }
      }
```

packages/teleterm/src/sharedProcess/ptyHost/ptyProcess.ts
```diff
@@ -47,7 +47,7 @@ export class PtyProcess extends EventEmitter implements IPtyProcess {
       name: 'xterm-color',
       // HOME should be always defined. But just in case it isn't let's use the cwd from process.
       // https://unix.stackexchange.com/questions/123858
-      cwd: this.options.cwd || this.options.env['HOME'] || process.cwd(),
+      cwd: this.options.cwd || getDefaultCwd(),
```
ravicious, line 50: Let's pass `this.options.env` to `getDefaultCwd` and use it there for `USERPROFILE`/`HOME` instead of `process.env`. I assume we pass that env along for a reason.

ravicious (approved):
Looks good, but please address the `this.options.env` comment before merging.

packages/teleterm/src/sharedProcess/ptyHost/ptyProcess.ts
gzdunek: Correct, it should be done this way

packages/teleterm/src/mainProcess/mainProcess.ts
gzdunek: Nice idea

PR #1030, opened 2022-07-25
Added eslint rule to enforce the order of file imports

hatched:
There have been a few discussions lately about this so this adds an eslint rule to enforce the order of the imports. The order that I picked fit very closely to what we were already doing without this rule so while there are changes to a lot of files the actual changes are minor.

This is an auto-fixable rule so you can enable 'auto fix' in your editor and you never have to think about this order again 😁


avatus:
Big fan of the newlines between. LGTM (i mean, the rule... i did a cursory glance of the other 647 files)

avatus (approved):

ryanclark (approved):

packages/teleport/src/Discover/LoginTrait/useLoginTrait.ts
```diff
@@ -16,8 +16,10 @@
 
 import { useState, useEffect } from 'react';
 import useAttempt from 'shared/hooks/useAttemptNext';
-import type { User } from 'teleport/services/user';
+
 import { DiscoverContext } from '../discoverContext';
+
+import type { User } from 'teleport/services/user';
```
kimlisa, line 22: don't think these are correct order?

packages/teleterm/src/ui/ClusterLogout/ClusterLogout.tsx
```diff
@@ -22,9 +22,11 @@ import DialogConfirmation, {
 } from 'design/DialogConfirmation';
 import * as Alerts from 'design/Alert';
 import { ButtonIcon, ButtonPrimary, Text } from 'design';
-import { Props, State, useClusterLogout } from './useClusterLogout';
+
```
kimlisa, line 25: line 26 is kicked out of the group above

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Applications/Applications.tsx
```diff
@@ -29,12 +29,17 @@ import {
   deepOrange,
   blueGrey,
 } from 'design/theme/palette';
-import { Table } from 'teleterm/ui/components/Table';
+
 import { Cell } from 'design/DataTable';
+
+import { Danger } from 'design/Alert';
+
```
kimlisa, line 36: here too

packages/teleport/src/Apps/AddApp/AddApp.tsx
```diff
@@ -17,10 +17,15 @@
 import React from 'react';
 import { Flex } from 'design';
 import Dialog, { DialogTitle } from 'design/Dialog';
-import useTeleport from 'teleport/useTeleport';
-import Manually from './Manually';
+
 import * as Icons from 'design/Icon';
+
+import useTeleport from 'teleport/useTeleport';
```
kimlisa, line 23: isn't the `teleport/*` supposed to be grouped? line 25

also line 27-29 isn't grouped either

packages/teleport/src/Apps/Apps.tsx
```diff
@@ -24,10 +25,12 @@ import {
 } from 'teleport/components/Layout';
 import Empty, { EmptyStateInfo } from 'teleport/components/Empty';
 import ErrorMessage from 'teleport/components/AgentErrorMessage';
+
+import AgentButtonAdd from 'teleport/components/AgentButtonAdd';
```
kimlisa, line 29: this too, it should be grouped

packages/teleport/src/Console/DocumentNodes/DocumentNodes.story.tsx
```diff
@@ -15,10 +15,12 @@ limitations under the License.
 */
 
 import React from 'react';
+
+import { Node } from 'teleport/services/nodes/types';
+
 import DocumentNodes from './DocumentNodes';
 import ConsoleCtx from './../consoleContext';
 import { TestLayout } from './../Console.story';
```
kimlisa, line 23: i've seen the `./../*` come before the `./*` or maybe that was coincidence?

packages/teleport/src/Console/DocumentNodes/useNodes.ts
```diff
@@ -18,17 +18,22 @@ import { useEffect, useState } from 'react';
...
+
 import labelClick from 'teleport/labelClick';
 import { AgentLabel } from 'teleport/services/agents';
 import { sortLogins } from 'teleport/Nodes/useNodes';
 
+import * as stores from './../stores';
+import { useConsoleContext } from './../consoleContextProvider';
+
+import type { Node, NodesResponse } from 'teleport/services/nodes';
+
```
kimlisa, line 36: couple things wrong with `teleport/*` imports

- some are grouped eg line 28-30
- other has new lines
- one is out of order line 35

packages/teleport/src/Console/DocumentSsh/DocumentSsh.story.tsx
```diff
@@ -15,11 +15,13 @@ limitations under the License.
 */
 
 import React from 'react';
-import type { Session } from 'teleport/services/session';
+
 import DocumentSsh from './DocumentSsh';
 import { TestLayout } from './../Console.story';
 import ConsoleCtx from './../consoleContext';
 
+import type { Session } from 'teleport/services/session';
```
kimlisa, line 23: out of order

packages/teleport/src/Console/DocumentSsh/DocumentSsh.tsx
```diff
@@ -15,16 +15,21 @@ limitations under the License.
...
-import Terminal from './Terminal';
-import Document from '../Document';
+
 import AuthnDialog from 'teleport/components/AuthnDialog';
 import useWebAuthn from 'teleport/lib/useWebAuthn';
+
+import Document from '../Document';
+
+import Terminal from './Terminal';
+import FileTransfer, { useFileTransferDialogs } from './../FileTransfer';
```
kimlisa, line 32: goupring error

kimlisa (commented):
i stopped reviewing halfway (and may have not looked at every file while going halfway after seeing inconsistency) but there seems to be ordering and grouping problems

packages/teleport/src/Console/DocumentSsh/DocumentSsh.story.tsx
hatched: types are last in the config so this is correct.

packages/teleport/src/Console/DocumentSsh/DocumentSsh.tsx
hatched: This is correct grouping as per the rules, but it's confusing the linter because of the `./`. So it's grouping it with `./Terminal` instead of `../Document`.

packages/teleport/src/Console/DocumentNodes/useNodes.ts
hatched: These are all correct as per the config.

packages/teleport/src/Account/ManageDevices/AddDevice/AddDevice.tsx
```diff
@@ -38,9 +38,12 @@ import {
   requiredField,
 } from 'shared/components/Validation/rules';
 import FieldSelect from 'shared/components/FieldSelect';
+
+import createMfaOptions, { MfaOption } from 'shared/utils/createMfaOptions';
```
ravicious, line 42: I think Lisa was onto something. 🤔 Why is there a new line between line 42 and 40?

packages/teleterm/src/ui/DocumentGateway/DocumentGateway.tsx
```diff
@@ -24,16 +24,20 @@ import {
   Link,
```
ravicious, line 24: (Adding a comment here just to create a separate GH discussion)

I have a commit on my branch (that I rebased on master) where the types seem to be out of order:

https://github.com/gravitational/webapps/blob/04de0d9edd43bf25c6870a717b00275f10200f4c/packages/teleterm/src/ui/DocumentGateway/OfflineDocumentGateway.tsx#L1-L10 

I suppose `design` is treated like a package but `design/Alert` is treated as another group.

`teleterm/ui/components/LinearProgress` is put into a completely different group than `design/Alert` and `shared/components/Validation` which I guess makes sense as this file is within `packages/teleterm`.

I think top-level imports from `design` and our other packages should be treated the same way as imports from `design/Alert`, no?

packages/build/.eslintrc.js
```diff
@@ -63,6 +63,22 @@ module.exports = {
...
+          'builtin',
+          'external',
+          'internal',
+          'parent',
+          'sibling',
+          'index',
+          'object',
+          'type',
+        ],
+        'newlines-between': 'always-and-inside-groups',
```
ravicious, line 79: Now that I look at it, I wish we went with `always`. 😶

Does `always-and-inside-groups` have any advantages over `always`? It seems to me that you'd always want to keep the groups together.

packages/teleport/src/Account/ManageDevices/AddDevice/AddDevice.tsx
ravicious: I think this newline might be the result of using `always-and-inside-groups` instead of `always`.

packages/teleterm/src/ui/DocumentGateway/DocumentGateway.tsx
ravicious: `design` and `design/Alert` not being together is another result of using `always-and-inside-groups`. It's just what autofix did to the existing code, if I remove all newlines and run it again then the imports look like this:

```typescript
import React, { useState } from 'react';
import { ButtonPrimary, Flex, Text } from 'design';
import * as Alerts from 'design/Alert';
import Validation from 'shared/components/Validation';

import LinearProgress from 'teleterm/ui/components/LinearProgress';

import { PortFieldInput } from './common';
import { DocumentGatewayProps } from './DocumentGateway';
```

I think we could configure the rule to treat our packages as the internal group, so that the final imports would look like this:

```
import React, { useState } from 'react';

import { ButtonPrimary, Flex, Text } from 'design';
import * as Alerts from 'design/Alert';
import Validation from 'shared/components/Validation';

import LinearProgress from 'teleterm/ui/components/LinearProgress';

import { PortFieldInput } from './common';
import { DocumentGatewayProps } from './DocumentGateway';
```

What do you think?

packages/build/.eslintrc.js
ravicious: I just ran autofix on master with `always` and it looks like there's a lot of files where the spaces seem to be arbitrarily injected from when the autofix was first ran.

ravicious:
Could you backport it to v10? I often check if v10 of Connect has everything backported by doing `git diff teleport-v10..master -w -- packages/teleterm` and currently the diff includes a lot of those import order changes.

PR #1033, opened 2022-07-26
Connect: Set up entitlements and provisioning profile

ravicious:
Closes gravitational/webapps.e#325.

This should let us use the same Touch ID credentials that tsh.app uses.

- [x] Try using another method for notarizing locally to avoid sending emails each time you notarize.
- [x] Make it work.
   - [x] Check if hardware keys on a separate laptop still work.
- [x] Grant `keychain-access-groups` only to the tsh process or the main process + tsh if possible.
- [x] Grant `com.apple.security.cs.allow-jit` only to the main process and the Electron child processes.
- [ ] Move tsh.app to correct location within Connect.app (it shouldn't be under Resources).
- [ ] Come up with a solution to making different kinds of builds:
   - packaged, without signing tsh
   - packaged, with tsh signed with dev cert
   - packaged, with tsh signed with prod cert 
- [ ] Describe how the macOS build is made in the readme.
    - Add info on how to check which cert a provisioning profile is bound to.


ravicious:
I'm running [a build on 11.0.0-dev.4](https://drone.platform.teleport.sh/gravitational/teleport/13951/12/1) which is this branch rebased on top of `lisa/tconnect-pwdless-main` to see if this indeed works.

packages/teleterm/package.json
```diff
@@ -55,7 +55,7 @@
   },
   "productName": "Teleport Connect",
   "build": {
-    "appId": "gravitational.teleport.connect",
+    "appId": "com.gravitational.teleport.connect",
```
ravicious, line 58: @gzdunek Most of the changes probably will make more sense to Alan, but off the top of your head do you remember if changing this value is going to have any impact on the app?

packages/teleterm/assets/entitlements.mac.plist
```diff
@@ -6,5 +6,16 @@
     <true/>
     <key>com.apple.security.cs.allow-jit</key>
     <true/>
+
+    <!-- All three keys must be identical to those in tsh.app so that Connect has access to the same
+    Touch ID credentials as tsh.app. -->
```
ravicious, line 11: @codingllama Is "Touch ID credentials" the correct term?

codingllama (commented):
LGTM, just a few minor adjustments.

Checked and double-checked, but it looks like you got it all down. 👍 

The real test is whether the resulting binary works. Give me a nudge if you have any problems, I can help you diagnose/debug.

packages/teleterm/assets/entitlements.mac.plist
```diff
@@ -6,5 +6,16 @@
...
+    <key>com.apple.developer.team-identifier</key>
+    <string>QH8AA5B8UP</string>
+    <key>com.apple.application-identifier</key>
+    <string>QH8AA5B8UP.com.gravitational.teleport.tsh</string>
```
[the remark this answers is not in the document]
ravicious: Ah, I read your notes wrong, thanks.

packages/teleterm/package.json
```diff
@@ -68,6 +68,8 @@
       "hardenedRuntime": true,
       "entitlements": "assets/entitlements.mac.plist",
       "entitlementsInherit": "assets/entitlements.mac.plist",
```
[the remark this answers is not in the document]
ravicious: I have no idea if we need both and unfortunately I don't have time to address this right now.

I did a quick google and found that there's a lot of people having weird issues with entitlements and providing the same file helps ((https://github.com/electron-userland/electron-builder/issues/4040#issuecomment-510507888)). If we don't provide one, I assume electron-builder will use [probably this default file](https://github.com/electron/osx-sign/blob/db611f0eda8931f1ac8caf11a11dfb08187f1d3b/entitlements/default.mas.child.plist).

Also, apparently [that's how VS Code does things](https://github.com/microsoft/vscode/blob/079c49ed1277fc42d8a49823dc6fbab473223503/build/darwin/sign.js#L33-L34).

It seems like one of those cargo culted solutions that no one bothers to look up.

codingllama (approved):

codingllama:
LGTM.

Let us know how it goes after you take the final binary for a spin.

ravicious:
> Let us know how it goes after you take the final binary for a spin.

Sure thing, though idk when that will happen as the darwin builders seem to remain broken.

ravicious:
11.0.0-dev.5 got built but I made a mistake when rebasing my branch on top of @kimlisa's and I forgot to include the change to the bundle identifier. 🤦 I'll give it another go tomorrow.

ravicious:
I built 11.0.0-dev.8 which is again just this branch rebased on top of Lisa's. Here's [the .dmg file](https://drive.google.com/file/d/1uDabEJk9TUHKh0AK6oVrY-s-SGPTcMi_/view?usp=sharing). It fails because it seems that its signature is invalid. (cc @kimlisa)

I need to find a faster way to verify if everything's okay than building a tag on drone which takes ~45 minutes, I'll look at the instructions Alan left in tshdev.

<details>
<summary>codesign -dv --verbose=4</summary>

```
$ codesign -dv --verbose=4 ~/Downloads/Teleport\ Connect.app
Executable=/Users/rav/Downloads/Teleport Connect.app/Contents/MacOS/Teleport Connect
Identifier=com.gravitational.teleport.connect
Format=app bundle with Mach-O thin (x86_64)
CodeDirectory v=20500 size=494 flags=0x10000(runtime) hashes=4+7 location=embedded
VersionPlatform=1
VersionMin=658176
VersionSDK=787200
Hash type=sha256 size=32
CandidateCDHash sha1=c5e101d10cdddfda8a0e8464fe0e16a1f4be4cc9
CandidateCDHashFull sha1=c5e101d10cdddfda8a0e8464fe0e16a1f4be4cc9
CandidateCDHash sha256=26b83b35e4787b334f2277ed2b4375c00ccc8be0
CandidateCDHashFull sha256=26b83b35e4787b334f2277ed2b4375c00ccc8be0b8c48c91c387a1c7bdbd8b37
Hash choices=sha1,sha256
CMSDigest=ae83bf67c6bd94bbda836f4de7f692c74c9791cc6fd7b0d54e069a2d3c892293
CMSDigestType=2
Executable Segment base=0
Executable Segment limit=8192
Executable Segment flags=0x1
Page size=4096
CDHash=26b83b35e4787b334f2277ed2b4375c00ccc8be0
Signature size=9067
Authority=Developer ID Application: Gravitational Inc. (QH8AA5B8UP)
Authority=Developer ID Certification Authority
Authority=Apple Root CA
Timestamp=28 Jul 2022 at 17:24:14
Info.plist entries=30
TeamIdentifier=QH8AA5B8UP
Runtime Version=12.3.0
Sealed Resources version=2 rules=13 files=18
Internal requirements count=1 size=196
```
</details>

<details>
<summary>codesign -v -vvvv</summary>

```
$ codesign -v -vvvv ~/Downloads/Teleport\ Connect.app
--prepared:/Users/rav/Downloads/Teleport Connect.app/Contents/Frameworks/Teleport Connect Helper (Plugin).app
--validated:/Users/rav/Downloads/Teleport Connect.app/Contents/Frameworks/Teleport Connect Helper (Plugin).app
--prepared:/Users/rav/Downloads/Teleport Connect.app/Contents/Frameworks/Teleport Connect Helper (GPU).app
--validated:/Users/rav/Downloads/Teleport Connect.app/Contents/Frameworks/Teleport Connect Helper (GPU).app
--prepared:/Users/rav/Downloads/Teleport Connect.app/Contents/Frameworks/Teleport Connect Helper (Renderer).app
--validated:/Users/rav/Downloads/Teleport Connect.app/Contents/Frameworks/Teleport Connect Helper (Renderer).app
--prepared:/Users/rav/Downloads/Teleport Connect.app/Contents/Frameworks/Electron Framework.framework/Versions/Current/.
--validated:/Users/rav/Downloads/Teleport Connect.app/Contents/Frameworks/Electron Framework.framework/Versions/Current/.
--prepared:/Users/rav/Downloads/Teleport Connect.app/Contents/Frameworks/Teleport Connect Helper.app
--validated:/Users/rav/Downloads/Teleport Connect.app/Contents/Frameworks/Teleport Connect Helper.app
--prepared:/Users/rav/Downloads/Teleport Connect.app/Contents/Frameworks/ReactiveObjC.framework/Versions/Current/.
--validated:/Users/rav/Downloads/Teleport Connect.app/Contents/Frameworks/ReactiveObjC.framework/Versions/Current/.
--prepared:/Users/rav/Downloads/Teleport Connect.app/Contents/Frameworks/Squirrel.framework/Versions/Current/.
--validated:/Users/rav/Downloads/Teleport Connect.app/Contents/Frameworks/Squirrel.framework/Versions/Current/.
--prepared:/Users/rav/Downloads/Teleport Connect.app/Contents/Frameworks/Mantle.framework/Versions/Current/.
--validated:/Users/rav/Downloads/Teleport Connect.app/Contents/Frameworks/Mantle.framework/Versions/Current/.
/Users/rav/Downloads/Teleport Connect.app: valid on disk
/Users/rav/Downloads/Teleport Connect.app: satisfies its Designated Requirement
```
</details>

<details>
<summary>The report compiled by macOS</summary>

```
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Incident Identifier: F7A3F558-CD8C-49F5-AEE9-5E0B221A5B37
CrashReporter Key:   CEED24F1-A75C-89A9-E06E-1B0451613C4D
Hardware Model:      MacBookPro18,3
Process:             Teleport Connect [33400]
Path:                /Users/USER/Downloads/Teleport Connect.app/Contents/MacOS/Teleport Connect
Identifier:          com.gravitational.teleport.connect
Version:             11.0.0-dev.8 (11.0.0-dev.8.14075)
Code Type:           X86-64 (Native)
Role:                Default
Parent Process:      launchd [1]
Coalition:           com.gravitational.teleport.connect [98527]

Date/Time:           2022-07-28 17:44:41.3263 +0200
Launch Time:         2022-07-28 17:44:41.1436 +0200
OS Version:          macOS 12.5 (21G72)
Release Type:        User
Report Version:      104

Exception Type:  EXC_CRASH (SIGKILL (Code Signature Invalid))
Exception Codes: 0x0000000000000000, 0x0000000000000000
Exception Note:  EXC_CORPSE_NOTIFY
Termination Reason: CODESIGNING 1 

Triggered by Thread:  0

Thread 0 Crashed:
0                                 	    0x7ff7fff60a8c 0x7ff7fff5d000 + 14988


Thread 0 crashed with ARM Thread State (64-bit):
    x0: 0x0000000000000000   x1: 0x0000000000000000   x2: 0x0000000000000000   x3: 0x0000000000000000
    x4: 0x0000000000000000   x5: 0x0000000000000000   x6: 0x0000000000000000   x7: 0x0000000000000000
    x8: 0x0000000000000000   x9: 0x0000000000000000  x10: 0x0000000000000000  x11: 0x0000000000000000
   x12: 0x0000000000000000  x13: 0x0000000000000000  x14: 0x0000000000000000  x15: 0x0000000000000000
   x16: 0x0000000000000000  x17: 0x0000000000000000  x18: 0x0000000000000000  x19: 0x0000000000000000
   x20: 0x0000000000000000  x21: 0x0000000000000000  x22: 0x0000000000000000  x23: 0x0000000000000000
   x24: 0x0000000000000000  x25: 0x0000000000000000  x26: 0x0000000000000000  x27: 0x0000000000000000
   x28: 0x0000000000000000   fp: 0x0000000000000000   lr: 0x0000000000000000
    sp: 0x000000030a46bb68   pc: 0x00007ff7fff60a8c cpsr: 0x00001000
   far: 0x0000000000000000  esr: 0x00000000  Address size fault

Binary Images:
    0x7ff7fff5d000 -     0x7ff7fff8cfff  (*) <ef33add1-6b70-3cc9-8bbc-c8544b609d2b> ???

Error Formulating Crash Report:
dyld_process_snapshot_get_shared_cache failed

EOF

-----------
Full Report
-----------

{"app_name":"Teleport Connect","timestamp":"2022-07-28 17:44:41.00 +0200","app_version":"11.0.0-dev.8","slice_uuid":"4c4c44a6-5555-3144-a12e-12d3e30796b2","build_version":"11.0.0-dev.8.14075","platform":0,"bundleID":"com.gravitational.teleport.connect","share_with_app_devs":0,"is_first_party":0,"bug_type":"309","os_version":"macOS 12.5 (21G72)","incident_id":"F7A3F558-CD8C-49F5-AEE9-5E0B221A5B37","name":"Teleport Connect"}
{
  "uptime" : 180000,
  "procLaunch" : "2022-07-28 17:44:41.1436 +0200",
  "procRole" : "Default",
  "version" : 2,
  "userID" : 501,
  "deployVersion" : 210,
  "modelCode" : "MacBookPro18,3",
  "procStartAbsTime" : 4518759004042,
  "coalitionID" : 98527,
  "osVersion" : {
    "train" : "macOS 12.5",
    "build" : "21G72",
    "releaseType" : "User"
  },
  "captureTime" : "2022-07-28 17:44:41.3263 +0200",
  "incident" : "F7A3F558-CD8C-49F5-AEE9-5E0B221A5B37",
  "bug_type" : "309",
  "pid" : 33400,
  "procExitAbsTime" : 4518762888468,
  "translated" : true,
  "cpuType" : "X86-64",
  "procName" : "Teleport Connect",
  "procPath" : "\/Users\/USER\/Downloads\/Teleport Connect.app\/Contents\/MacOS\/Teleport Connect",
  "bundleInfo" : {"CFBundleShortVersionString":"11.0.0-dev.8","CFBundleVersion":"11.0.0-dev.8.14075","CFBundleIdentifier":"com.gravitational.teleport.connect"},
  "storeInfo" : {"deviceIdentifierForVendor":"93F9A63C-EEFB-59C1-B0E5-AF7DB456D7C1","thirdParty":true},
  "parentProc" : "launchd",
  "parentPid" : 1,
  "coalitionName" : "com.gravitational.teleport.connect",
  "crashReporterKey" : "CEED24F1-A75C-89A9-E06E-1B0451613C4D",
  "wakeTime" : 9261,
  "sleepWakeUUID" : "CEE63F45-D468-49AC-9740-925E2B68501F",
  "sip" : "enabled",
  "isCorpse" : 1,
  "exception" : {"codes":"0x0000000000000000, 0x0000000000000000","rawCodes":[0,0],"type":"EXC_CRASH","signal":"SIGKILL (Code Signature Invalid)"},
  "termination" : {"namespace":"CODESIGNING","flags":66,"code":1},
  "extMods" : {"caller":{"thread_create":0,"thread_set_state":0,"task_for_pid":0},"system":{"thread_create":0,"thread_set_state":0,"task_for_pid":0},"targeted":{"thread_create":0,"thread_set_state":0,"task_for_pid":0},"warnings":0},
  "faultingThread" : 0,
  "threads" : [{"triggered":true,"id":1920940,"threadState":{"x":[{"value":0},{"value":0},{"value":0},{"value":0},{"value":0},{"value":0},{"value":0},{"value":0},{"value":0},{"value":0},{"value":0},{"value":0},{"value":0},{"value":0},{"value":0},{"value":0},{"value":0},{"value":0},{"value":0},{"value":0},{"value":0},{"value":0},{"value":0},{"value":0},{"value":0},{"value":0},{"value":0},{"value":0},{"value":0}],"flavor":"ARM_THREAD_STATE64","lr":{"value":0},"cpsr":{"value":4096},"fp":{"value":0},"sp":{"value":13057309544},"esr":{"value":0,"description":" Address size fault"},"pc":{"value":140703127964300,"matchesCrashFrame":1},"far":{"value":0}},"frames":[{"imageOffset":14988,"imageIndex":0}]}],
  "usedImages" : [
  {
    "source" : "P",
    "arch" : "arm64",
    "base" : 140703127949312,
    "size" : 196608,
    "uuid" : "ef33add1-6b70-3cc9-8bbc-c8544b609d2b",
    "name" : ""
  }
],
  "vmSummary" : "ReadOnly portion of Libraries: Total=840K resident=0K(0%) swapped_out_or_unallocated=840K(100%)\nWritable regions: Total=21.4M written=0K(0%) resident=0K(0%) swapped_out=0K(0%) unallocated=21.4M(100%)\n\n                                VIRTUAL   REGION \nREGION TYPE                        SIZE    COUNT (non-coalesced) \n===========                     =======  ======= \nSTACK GUARD                       56.0M        1 \nStack                             8176K        1 \nVM_ALLOCATE                       13.3M        5 \n__DATA                              80K        3 \n__DATA_CONST                        80K        1 \n__LINKEDIT                         240K        4 \n__TEXT                             624K        2 \nmapped file                        6.1G       33 \n===========                     =======  ======= \nTOTAL                              6.2G       50 \n",
  "legacyInfo" : {
  "threadTriggered" : {

  }
},
  "trialInfo" : {
  "rollouts" : [
    {
      "rolloutId" : "5fc94383418129005b4e9ae0",
      "factorPackIds" : {

      },
      "deploymentId" : 240000413
    },
    {
      "rolloutId" : "60186475825c62000ccf5450",
      "factorPackIds" : {
        "SIRI_VALUE_INFERENCE_CONTACT_RESOLUTION" : "62bdc4218fe09c1632d841e9"
      },
      "deploymentId" : 240000023
    }
  ],
  "experiments" : [

  ]
},
  "reportNotes" : [
  "dyld_process_snapshot_get_shared_cache failed"
]
}

Model: MacBookPro18,3, BootROM 7459.141.1, proc 10:8:2 processors, 32 GB, SMC 
Graphics: Apple M1 Pro, Apple M1 Pro, Built-In
Display: DELL U2414H, 1920 x 1080 (1080p FHD - Full High Definition), Main, MirrorOff, Online
Display: Color LCD, 3024 x 1964 Retina, MirrorOff, Online
Memory Module: LPDDR5
AirPort: Wi-Fi, wl0: Apr  6 2022 05:55:54 version 20.90.45.0.8.7.118 FWID 01-e7138ff2
Bluetooth: Version (null), 0 services, 0 devices, 0 incoming serial ports
Network Service: Wi-Fi, AirPort, en0
USB Device: USB31Bus
USB Device: USB31Bus
USB Device: 4-Port USB 3.0 Hub
USB Device: USB 10/100/1000 LAN
USB Device: 4-Port USB 2.0 Hub
USB Device: BillBoard Device
USB Device: Advantage2 Keyboard
USB Device: USB31Bus
USB Device: YubiKey FIDO+CCID
Thunderbolt Bus: MacBook Pro, Apple Inc.
Thunderbolt Bus: MacBook Pro, Apple Inc.
Thunderbolt Bus: MacBook Pro, Apple Inc.
```
</details>

codingllama:
@ravicious 

> I need to find a faster way to verify if everything's okay than building a tag on drone which takes ~45 minutes, I'll look at the instructions Alan left in tshdev.

Yep, that's tough.

> The report compiled by macOS

Looking at Console.app during launch, the following caught my attention:

```
error	14:00:00.549418-0300	taskgated-helper	com.gravitational.teleport.connect: Unsatisfied entitlements: com.apple.developer.team-identifier, keychain-access-groups
```

Followed by (abridged):

```
error	14:00:00.549436-0300	taskgated-helper	Disallowing: com.gravitational.teleport.connect

default	14:00:00.549948-0300	amfid	/Users/alan/Downloads/Teleport Connect.app/Contents/MacOS/Teleport Connect signature not valid: -67671

default	14:00:00.549993-0300	kernel	mac_vnode_check_signature: /Users/alan/Downloads/Teleport Connect.app/Contents/MacOS/Teleport Connect: code signature validation failed fatally: When validating /Users/alan/Downloads/Teleport Connect.app/Contents/MacOS/Teleport Connect:
  Code has restricted entitlements, but the validation of its code signature failed.
Unsatisfied Entitlements:

default	14:00:00.550030-0300	kernel	proc 36104: load code signature error 4 for file "Teleport Connect"
```

Comparing `tsh.app` with `Teleport Connect.app`, I noticed a difference in how the identifier is formed:

```shell
$ codesign -dvv /Applications/tsh-v10.0.2.app
Executable=/Applications/tsh-v10.0.2.app/Contents/MacOS/tsh
Identifier=QH8AA5B8UP.com.gravitational.teleport.tsh <-- STARTS WITH TEAM ID
Format=app bundle with Mach-O thin (x86_64)
CodeDirectory v=20500 size=628469 flags=0x10300(hard,kill,runtime) hashes=19628+7 location=embedded
Signature size=8987
Authority=Developer ID Application: Gravitational Inc. (QH8AA5B8UP)
Authority=Developer ID Certification Authority
Authority=Apple Root CA
Timestamp=19 Jul 2022 12:22:00
Info.plist entries=23
TeamIdentifier=QH8AA5B8UP
Runtime Version=11.3.0
Sealed Resources version=2 rules=13 files=2
Internal requirements count=1 size=204

$ codesign -dvv Teleport\ Connect.app
Executable=/Users/alan/Downloads/Teleport Connect.app/Contents/MacOS/Teleport Connect
Identifier=com.gravitational.teleport.connect <-- NO TEAM ID
Format=app bundle with Mach-O thin (x86_64)
CodeDirectory v=20500 size=494 flags=0x10000(runtime) hashes=4+7 location=embedded
Signature size=9067
Authority=Developer ID Application: Gravitational Inc. (QH8AA5B8UP)
Authority=Developer ID Certification Authority
Authority=Apple Root CA
Timestamp=28 Jul 2022 12:24:14
Info.plist entries=30
TeamIdentifier=QH8AA5B8UP
Runtime Version=12.3.0
Sealed Resources version=2 rules=13 files=18
Internal requirements count=1 size=196
```

Based on the above, my guess would be to prepend the Team ID to Connect's Identifier: `QH8AA5B8UP.com.gravitational.teleport.connect`.

Note that there's also a difference between tsh flags (`flags=0x10300(hard,kill,runtime)`) and Connect (`flags=0x10000(runtime)`). I _think_ `runtime` implies `hard,kill`, but I'd keep an eye on that one too.

ravicious:
[11.0.0-dev.11](https://drive.google.com/file/d/1jISEMU1Ogp_AREPnoR7sbSYsWMzIhcea/view?usp=sharing) has updated identifier but fails with the same error. Looking at codesign output, the remaining differences between Connect and tsh.app is the flags that you mentioned Alan and Runtime Version.

I haven't looked at Console yet, first I want to make a local dev setup, otherwise I'm finding myself just doing the least amount of changes possible in hopes that the ~45 minutes build comes out alright.

I forgot yesterday that I need 2FA to log in to the dev account so I'll get the dev cert today and try again on Monday.

packages/teleterm/package.json
```diff
@@ -57,19 +57,22 @@
   },
   "productName": "Teleport Connect",
   "build": {
-    "appId": "gravitational.teleport.connect",
+    "appId": "com.gravitational.teleport.connect",
```
codingllama, line 60: I suppose we should change this one too?

```suggestion
    "appId": "QH8AA5B8UP.com.gravitational.teleport.connect",
```

ravicious: `build.mac.appId` will override `build.appId` when building for macOS. `bulid.appId` is going to be used on other platforms and I figured out we don't want to use Apple-specific IDs there.

ravicious:
Made some progress today. I set up a local env for signing. Thanks to adding`entitlementsInherit` and `com.apple.security.inherit` set to `true`, the app no longer crashes on launch due to unsatisfied entitlements for child processes.

Now the tshd process launched by the main process seems to get killed soon after launching but at least this doesn't cause any error alerts pop up. I'll look into that on Tuesday once I'm back.

[`ravicious/connect-signing`](https://github.com/gravitational/teleport/compare/ravicious/connect-signing) in the teleport repo fixes the problem with drone choosing the wrong dev ID cert for signing. I'll merge that next week.

packages/teleterm/package.json
```diff
@@ -57,19 +57,22 @@
...
       "build/app/dist"
     ],
     "mac": {
+      "appId": "QH8AA5B8UP.com.gravitational.teleport.connect",
       "target": "dmg",
       "type": "distribution",
       "hardenedRuntime": true,
       "entitlements": "assets/entitlements.mac.plist",
-      "entitlementsInherit": "assets/entitlements.mac.plist",
+      "entitlementsInherit": "assets/entitlements-inherit.mac.plist",
```
codingllama, line 73: Have you tried it without letting the other processes inherit the entitlements? Ideally we don't the permissions to propagate to binaries other than `tsh`.

ravicious: No, I haven't but that's a good point. Matter of fact, we need only the tsh process to have access to those keychain-access-groups as for now it'd be the only process interacting with Touch ID.

I believe the Electron child processes need to have `com.apple.security.cs.allow-jit` set to true and guessing from the errors that I've seen, all child processes must have at least `com.apple.developer.team-identifier` set to the same value as the main process.

I'll see what I can do here.

ravicious:
I understand entitlements a little bit better now so I went a step back.

tl;dr The entitlements for keychain-access-groups are granted for the main process but not the helper processes somehow.

What I did today:

I made a working build with minimal entitlements, that is only the ones required by Electron (`com.apple.security.cs.allow-jit`, `com.apple.security.cs.debugger`) and the app worked without problems.

---

`com.apple.security.inherit` was a red herring. It is meant for Electron apps distributed through App Store.

> To enable sandbox inheritance, a child target must use exactly two App Sandbox entitlement keys: com.apple.security.app-sandbox and com.apple.security.inherit. If you specify any other App Sandbox entitlement, the system aborts the child process.

([source](https://developer.apple.com/library/archive/documentation/Miscellaneous/Reference/EntitlementKeyReference/Chapters/EnablingAppSandbox.html#//apple_ref/doc/uid/TP40011195-CH4-SW15))

---

`entitlements` and `entitlementsInherit` are keys used by electron-builder. `entitlements` is used for the entitlements of the main binary while `entitlementsInherit` are used for any other binary within the bundle – that includes the Electron helper processes under `Teleport Connect.app/Contents/Frameworks/*.app` as well as the tsh binary under `Teleport Connect.app/Contents/Resources/bin/tsh`. I haven't looked into it too deep but it seems that it'll be rather hard to have custom entitlements _just_ for tsh – electron-builder's code simply doesn't provide such options.

---

So, the app with minimal entitlements works just fine. Mind you, I just simply provide the same entitlements file for both the main binary as well as the child binaries for now (`entitlements` and `entitlementsInherit` point to the same file, just as is VS Code).

The moment I add `keychain-access-groups` to the entitlements, the app breaks. In Console I can see taskgated-helper messages like 

> K497G57PDJ.com.goteleport.connectdev.helper.Renderer: Unsatisfied entitlements: keychain-access-groups

What's interesting is that those messages are shown only for the helper processes and not for the main process. To me this suggests that somehow the main process is eligible to those entitlements but child processes are not.

Adding `com.apple.developer.team-identifier` to entitlements doesn't help as this simply adds that entitlement to the list of unsatisfied entitlements seen in Console (again, shown for helper processes but not the main process). Changing the keychain access group from `…tshdev` to an arbitrary string like `…SharedItems` doesn't change much.

When viewing all messages in Console and not just errors I can see this:

> amfid: /Users/rav/Projects/webapps/packages/teleterm/build/release/mac-arm64/TeleportConnect.app/Contents/Frameworks/TeleportConnect Helper.app/Contents/MacOS/TeleportConnect Helper signature not valid: -67671

> mac_vnode_check_signature: /Users/rav/Projects/webapps/packages/teleterm/build/release/mac-arm64/TeleportConnect.app/Contents/Frameworks/TeleportConnect Helper.app/Contents/MacOS/TeleportConnect Helper: code signature validation failed fatally: When validating /Users/rav/Projects/webapps/packages/teleterm/build/release/mac-arm64/TeleportConnect.app/Contents/Frameworks/TeleportConnect Helper.app/Contents/MacOS/TeleportConnect Helper:
  Code has restricted entitlements, but the validation of its code signature failed.
Unsatisfied Entitlements:

---

I also discovered that for notarization my dev build used the wrong app ID.

---

I think the next step would be to check why those errors are thrown for helper processes but not for the main process. I checked the entitlements with `codesign --display --entitlements - path/to/bin` and they appear the same in main vs helpers. `codesign --verify --verbose path/to/bin` reports all signatures as valid.

ravicious:
@codingllama Do you know if Connect will be able to simply provide tsh.app app ID as the keychain access group and call it a day? The example in [Sharing Access to Keychain Items Among a Collection of Apps](https://developer.apple.com/documentation/security/keychain_services/keychain_items/sharing_access_to_keychain_items_among_a_collection_of_apps) shows two apps sharing an arbitrary key rather than the app ID of the first app. OTOH I'm not sure how macOS would differentiate between an arbitrary key and an app ID when reading keychain access group but maybe there's something to it.

> Because app IDs are unique across all apps, and because the app ID is stored in an entitlement protected by code signing, no other app can use it, and so no other app is in this group. Any keychain items stored with this access group are private to App One. Similarly, if you have a second app with a bundle ID of com.example.AppTwo, it automatically belongs to its own private group.

From a section talking about app groups vs keychain access groups:

> The system considers the first item in the list of access groups to be the app’s default access group. This is the access group that keychain services assumes if you don’t otherwise specify one when adding keychain items. An app group can’t ever be the default, because the app ID is always present and appears earlier in the list. However, a keychain access group can be the default, because it appears before the app ID. In particular, the first keychain access group, if any, that you specify in the corresponding capability becomes the app’s default access group. If you don’t specify any keychain access groups, then the app ID is the default.

codingllama:
Good to see you are making progress, @ravicious!


> I haven't looked into it too deep but it seems that it'll be rather hard to have custom entitlements just for tsh – electron-builder's code simply doesn't provide such options.

I think that's _the_ question: ideally we want to entitle tsh and nothing else. I wonder if we can get the Electron builder's output, take it apart and tweak it ourselves, so we get the exact setup we want. Signing and notarizing apps is nothing new to us at this point.

> I also discovered that for notarization my dev build used the wrong app ID.

Yep, that's why I have separate folders for "tsh" and "tshdev".

> @codingllama Do you know if Connect will be able to simply provide tsh.app app ID as the keychain access group and call it a day? (...)

That's a good question. My assumption was that yes, it would work, but we could try something like `QH8AA5B8UP.com.gravitational.teleport.tsh.webauthn` and see if that makes a difference. If it does it means a couple of changes in tsh, but we can manage.

> From a section talking about app groups vs keychain access groups (...)

Yep, I saw that too. Default groups do play a part, but we can tweak the tsh code to use specific groups if necessary. Let's figure out the setup first, then we can consider further changes.

ravicious:
> I think that's _the_ question: ideally we want to entitle tsh and nothing else. I wonder if we can get the Electron builder's output, take it apart and tweak it ourselves, so we get the exact setup we want. Signing and notarizing apps is nothing new to us at this point.

Right, I can see how this would be very important. We certainly can run whatever electron-builder is running manually at the cost of having to maintain those scripts ourselves.

From what I managed to gather, there was at least one instance where [a change in Chromium required changes to how Electron apps are signed](https://www.electronjs.org/blog/electron-6-0#electron-helper-rendererapp-electron-helper-gpuapp-and-electron-helper-pluginapp).

Though tbh, even at this point it might be easier to do all of this manually and just add those entitlements to tsh somehow rather than trying to figure out how to make those entitlements work when electron-builder adds them to all binaries within the bundle.

ravicious:
Actually, we might be able to have custom signing process just for tsh binary…

ravicious:
Okay, I'm getting somewhere. I managed to sign tsh with a custom set of entitlemets but…

> If your product includes a non-bundled executable that uses a restricted entitlement, you must package that executable in an app-like structure. For the details, see [Signing a Daemon with a Restricted Entitlement](https://developer.apple.com/documentation/xcode/signing-a-daemon-with-a-restricted-entitlement).

https://developer.apple.com/forums/thread/128166

I thought I'd be able to avoid this but it seems that it's actually required. I'll create a new provisioning profile for tshdev that's tied to my cert and I'll see how it goes. I might not have time for that today though.

---

BTW, up until today I didn't know how to check which cert exactly a provisioning profile is tied to. Turn out you just right click -> Get Info and it's there in the preview section.

codingllama:
Glad to see you are making progress, Rafal!

> Signing a Daemon with a Restricted Entitlement

That's exactly what we do for tsh.app

> BTW, up until today I didn't know how to check which cert exactly a provisioning profile is tied to. Turn out you just right click -> Get Info and it's there in the preview section.

Good to know, I have a bunch of exoteric commands noted somewhere to get cert from the profile.

ravicious:
Closing this PR in favor of #1116.

This one is a little messy and my final solution is much different than what I initially thought I'd need to do.

PR #1035, opened 2022-07-26
Add resource selection scaffold

hatched:


packages/design/src/SlideTabs/SlideTabs.tsx
```diff
@@ -25,32 +31,51 @@ function SlideTabs({ tabs, name = 'slide-tab', onChange }: props) {
...
-            itemCount={tabs.length}
-            key={`${tabName}-${tabIndex}`}
-          >
-            {tabName}
-            <TabInput type="radio" name={name} id={`${name}-${tabName}`} />
-          </TabLabel>
-        ))}
+      <TabNav appearance={appearance} size={size}>
+        {tabs.map((tabData, tabIndex) => {
+          const tabDataType = typeof tabData === 'string';
```
kimlisa, line 36: `isTabDataTypeStr`?

packages/design/src/SlideTabs/SlideTabs.tsx
```diff
@@ -25,32 +31,51 @@ function SlideTabs({ tabs, name = 'slide-tab', onChange }: props) {
...
+          let tabName = tabDataType ? tabData : tabData.name;
+          let tabContent = tabDataType ? tabData : tabData.component;
```
kimlisa, line 38:
```suggestion
          const tabName = tabDataType ? tabData : tabData.name;
          const tabContent = tabDataType ? tabData : tabData.component;
```

packages/teleport/src/Discover/SelectResource.tsx
```diff
@@ -0,0 +1,245 @@
...
+        <SelectDBDeploymentType
+          selectedType={selectedType}
+          setSelectedType={setSelectedType}
+          resourceTypes={resourceTypes}
+        />
+      )}
+      {selectedResource === 'application' && (
+        <Empty
+          clusterId="abc123"
+          canCreate={true}
```
kimlisa, line 121: i think this flag should come from perms defined in user context?

kimlisa (approved):
i did more a cursory look, as we might need to change things, but lgtm so far

packages/teleport/src/Discover/SelectResource.tsx
hatched: Oh yeah 100% this doesn't work at all, it was just put here as a placeholder to see how it would work. :)

PR #1038, opened 2022-07-27
Allow only one instance of Connect

gzdunek:
To my surprise, you can open the second instance of the app on Windows (on macOS it is not possible when you run the app from Finder). More information [here](https://www.electronjs.org/docs/latest/api/app#apprequestsingleinstancelockadditionaldata).

I was wondering how to implement it and IMO the best option is to only initialize the app, if we know that it's the first instance. Otherwise, some side effects from main process/file storage/logger/etc can occur.  

This change doesn't mean that we can have only one window in the app - when the `app.on('second-instance')` callback is invoked we can simply open a new window. 
By doing it this way, both windows will talk to the same `main` and `shared` process. 

When you try to open the second instance in development you will see the following message in the terminal (I don't know if we can do any better):
```
Starting inspector on 127.0.0.1:9229 failed: address already in use
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

### DIFF
--- a/packages/teleterm/src/main.ts
+++ b/packages/teleterm/src/main.ts
@@ -16,7 +16,7 @@ import { WindowsManager } from 'teleterm/mainProcess/windowsManager';
 if (app.requestSingleInstanceLock()) {
   initializeApp();
 } else {
-  console.log('Attempted to open a second instance of the app.');
+  console.log('Attempted to open a second instance of the app, exiting.');
   // All windows will be closed immediately without asking the user,
   // and the before-quit and will-quit events will not be emitted.
   app.exit(1);
--- a/packages/teleterm/src/main.ts
+++ b/packages/teleterm/src/main.ts
@@ -16,6 +16,7 @@ import { WindowsManager } from 'teleterm/mainProcess/windowsManager';
 if (app.requestSingleInstanceLock()) {
   initializeApp();
 } else {
+  console.log('Attempted to open a second instance of the app.');
   // All windows will be closed immediately without asking the user,
   // and the before-quit and will-quit events will not be emitted.
   app.exit(1);
--- a/packages/teleterm/src/main.ts
+++ b/packages/teleterm/src/main.ts
@@ -13,118 +13,132 @@ import { ConfigServiceImpl } from 'teleterm/services/config';
 import { createFileStorage } from 'teleterm/services/fileStorage';
 import { WindowsManager } from 'teleterm/mainProcess/windowsManager';
 
-const settings = getRuntimeSettings();
-const logger = initMainLogger(settings);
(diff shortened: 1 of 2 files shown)


packages/teleterm/src/main.ts
```diff
@@ -13,118 +13,132 @@ import { ConfigServiceImpl } from 'teleterm/services/config';
...
-app.on('web-contents-created', (_, contents) => {
-  contents.on('will-navigate', (event, navigationUrl) => {
-    logger.warn(`Navigation to ${navigationUrl} blocked by 'will-navigate'`);
-    event.preventDefault();
+if (app.requestSingleInstanceLock()) {
+  initializeApp();
+} else {
+  // All windows will be closed immediately without asking the user,
+  // and the before-quit and will-quit events will not be emitted.
+  app.exit(1);
```
ravicious, line 21: >   // All windows will be closed immediately without asking the user,
>   // and the before-quit and will-quit events will not be emitted.

That's a very good choice here. 👍

Have you checked if we need an explicit `return` after `app.exit()`? The docs don't give a clear answer IMHO. I know that you need one after `app.quit()`, otherwise Electron just continues execution after `app.quit()`.

Technically everything's within `initializeApp` so there's nothing else to be executed after that `if`. But I'd add that `return` if it's needed, just to be explicit.

ravicious (approved):
Uh nice, good job catching that.

packages/teleterm/src/main.ts
ravicious: nvm you cannot call `return` in a top-level if

ravicious:
We could add

```
console.log('Attempted to open a second instance of the app.');
```

before `app.exit()`, then in the terminal you'll see:

```
$ electron build/app/dist/main/main.js --inspect
Starting inspector on 127.0.0.1:9229 failed: address already in use
Attempted to open a second instance of the app.
error Command failed with exit code 1.
```

ravicious:
> Current master lets me happily launch two instances of the app.

It does launch the second instance but it also outputs that line:

```
Starting inspector on 127.0.0.1:9229 failed: address already in use
```

gzdunek:
Good idea with `Attempted to open a second instance of the app, exiting`, thanks!

> I wonder where that is coming from. Current master lets me happily launch two instances of the app.

You mean `Starting inspector...` line?
It comes from the [debugging inspector](https://nodejs.org/en/docs/guides/debugging-getting-started/#enable-inspector): `electron build/app/dist/main/main.js --inspect`

ravicious:
Ah, yeah, at the time I thought that this PR triggered that line but as I explained in another commit, it was still logged on master when running two instances.

PR #1045, opened 2022-07-28
`SharedDirectoryMoveRequest`

ibeckermayer:
Requires backport to v9/v10

#### `webapps`

## What to expect

image

image

Try moving the file from one folder in the shared directory to another, and you should see something like

image

in the console


kimlisa (approved):
hm, i don't see what you see on console, but this is work in progress so I didn't worry too much about it.
 
i tried moving a file into another folder inside the shared directory but it just pauses on this (i double checked i was on the right branch this time)

Screen Shot 2022-07-28 at 12 08 51 AM

ibeckermayer:
> hm, i don't see what you see on console, but this is work in progress so I didn't worry too much about it.

@kimlisa I think you're just not viewing the debug logs in the console. You need to click on "Verbose" to see the `logger.debug()` logs:

image

The on screen behavior is correct 👍

PR #1046, opened 2022-07-28
Log shared process `stdout` and `stderr`

gzdunek:
Currently, when the shared process fails to start (for example because `node.pty` hasn't been recompiled) we can't see the reason.
This PR fixes it. The logged output will be saved in the `shared` log file.


packages/teleterm/src/mainProcess/mainProcess.ts
```diff
@@ -109,6 +109,16 @@ export default class MainProcess {
...
 
+    const sharedProcessLogger = createLoggerService({
+      dev: this.settings.dev,
+      dir: this.settings.userDataDir,
+      name: 'shared',
+      passThroughMode: true,
+    });
+
+    sharedProcessLogger.pipeProcessOutputIntoLogger(this.sharedProcess.stdout);
+    sharedProcessLogger.pipeProcessOutputIntoLogger(this.sharedProcess.stderr);
```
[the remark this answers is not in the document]
gzdunek: Unfortunately, doing this would break `resolveNetworkAddress` (it uses pipes to find grpc network address), so I would like to leave `stdio` as it is.

What comes to my mind is that shared process can work similarly to tshd, so it won't save any files on its own, but rather log everything to stdout. 

Cons:
- Stacktraces have time added to each line. 

Pros:
- We don't have to fight with duplicated logs.
- We have more control over the logs. As we have more logs sources, we can show source of each line in the terminal (it looks cool IMO):
image

Let me know guys what do you think!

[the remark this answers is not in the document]
avatus: > it looks cool IMO

absolutely agree.

In terms of the con, I don't think adding time to each stack trace is too big of a deal. Sure, it adds some noise but as long as it doesn't _remove_ information it's not the worst. I like the idea of not saving any files on its own tbh.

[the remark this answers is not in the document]
ravicious: > What comes to my mind is that shared process can work similarly to tshd, so it won't save any files on its own, but rather log everything to stdout.

That seems to be a good solution.

You mentioned that time is added to each line but that doesn't seem to be the case on your screenshot, did you purposefully turn them off there or just cut them out when making the screenshot?

packages/teleterm/src/services/logger/loggerService.ts
```diff
@@ -79,9 +120,15 @@ function stringifier(message: unknown[]): string {
...
-type Options = {
+type FileLoggerOptions = {
   dir: string;
   name: string;
+  /**
+   * ANSI display attribute (SGR parameter) for the logger name e.g. 104 for Bright Blue background color.
+   * Logger name is printed in the terminal, only in dev mode.
+   * If not specified, the logger name will not be printed.
+   */
+  loggerNamePrintCode?: string;
```
gzdunek, line 131: Any better name is welcome 😏

gzdunek:
I added changes discussed above (with slightly changed `tshd` color).

packages/teleterm/src/services/logger/loggerService.ts
avatus: I can't think of a better name. Maybe `loggerNameANSICode` but that looks weird. Idk, thats a tough one.

avatus (approved):

packages/teleterm/src/services/logger/loggerService.ts
ravicious: What about this:

Create a palette / "enum" of colors in loggerService.ts:

```typescript
const COLORS = {
  Magenta: 45,
  Cyan: 46,
  Yellow: 43,
}
```

and then rename the arg to `backgroundColor` and it'd be used like so:

```typescript
backgroundColor: COLORS.Magenta
```

I'm not sure about the capitalization of those color names but it ultimately doesn't matter that much. I suppose the capitalized version would be closer to TypeScript's enums.

packages/teleterm/src/mainProcess/mainProcess.ts
```diff
@@ -88,15 +88,16 @@ export default class MainProcess {
       },
     });
```
ravicious, line 89: Could you add comments to both instances of `stdio: 'pipe'` that stdio must be set to pipe as we read the address of the gRPC server from stdout? That's valuable insight and I think it should get written down in the code.

packages/teleterm/src/services/logger/loggerService.ts
```diff
@@ -1,11 +1,38 @@
...
+} from 'winston';
 import { isObject } from 'lodash';
 
 import split2 from 'split2';
 
-import { Logger, NodeLoggerService } from './types';
+import { Logger, LoggerService, NodeLoggerService } from './types';
 
-export default function createLoggerService(opts: Options): NodeLoggerService {
+export function createStdoutLoggerService(): LoggerService {
```
ravicious, line 12: Could you add a comment explaining why we need both `createStdoutLoggerService` and `createFileLoggerService`?

ravicious (commented):
I think we should add some comments explaining why we need to do things in a certain way, I feel like this kind of knowledge gets forgotten really fast.

packages/teleterm/src/services/logger/loggerService.ts
gzdunek: Good idea!

Do you have anything against enums? I like them more than objects in cases like that (I can elegantly type the field) and they are used in our codebase.
I used name `loggerNameColor` because it also determines if the logger name is printed. I think it makes more sense than `backgroundColor` in this case.

packages/teleterm/src/mainProcess/mainProcess.ts
gzdunek: Done

packages/teleterm/src/services/logger/loggerService.ts
gzdunek: Done

PR #1047, opened 2022-07-28
CI: Fail if an update to yarn.lock is needed

ravicious:
Normally we'd use `yarn install --frozen-lockfile` but yarn has a bug where it ignores this flag when workspaces are set up. https://github.com/yarnpkg/yarn/issues/4098

We can't use git as suggested under that issue because Docker isn't inside a git repo at that point.

* [Example of a failed run](https://console.cloud.google.com/cloud-build/builds/1d47366e-0eef-4532-a492-81c03ca51904;step=1?project=ci-account).
* [Example of a successful run](https://console.cloud.google.com/cloud-build/builds/46011d80-e02a-4921-b080-2e38b3de915e;step=2?project=ci-account) – the cache is working as expected.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,14 @@ COPY README.md packages/webapps.e/telepor[t]/package.json web-apps/packages/weba
 
 # download and install npm dependencies
 WORKDIR web-apps
-RUN yarn install
+# Install JavaScript dependencies and manually check if yarn.lock needs an update.
+# Yarn v1 doesn't respect the --frozen-lockfile flag when using workspaces.
+# https://github.com/yarnpkg/yarn/issues/4098
+RUN sha384sum yarn.lock > yarn-lock-sha \
+  && yarn install \
+  && sha384sum --check yarn-lock-sha || \
+  { echo "yarn.lock needs an update; run yarn install, verify that correct dependencies were installed \
+and commit the updated version of yarn.lock"; exit 1; }
 
 # copy the rest of the files and run yarn build command
 COPY  . .
--- a/yarn.lock
+++ b/yarn.lock
@@ -10412,22 +10412,6 @@ node-gyp@8.4.1:
     tar "^6.1.2"
     which "^2.0.2"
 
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -19,7 +19,14 @@ COPY README.md packages/webapps.e/telepor[t]/package.json web-apps/packages/weba
...
 WORKDIR web-apps
-RUN yarn install
+# Install JavaScript dependencies and manually check if yarn.lock needs an update.
+# Yarn v1 doesn't respect the --frozen-lockfile flag when using workspaces.
+# https://github.com/yarnpkg/yarn/issues/4098
+RUN sha384sum yarn.lock > yarn-lock-sha \
+  && yarn install \
+  && sha384sum --check yarn-lock-sha || \
+  { echo "yarn.lock needs an update; run yarn install, verify that correct dependencies were installed \
+and commit the updated version of yarn.lock"; exit 1; }
```
ravicious, line 29: Using a single RUN so that Docker creates only a single intermediate layer.

I used to have this on separate lines but Docker would first take about 20 seconds to update the cache after `RUN yarn install` and only then it'd fail.

ryanclark: I like the solution. Just out of curiosity, what made you choose `sha384sum`?

ryanclark (approved):

Dockerfile
ravicious: I wasn't sure if my knowledge about which SHAs are safe to use was up to date, so I went to https://en.wikipedia.org/wiki/Secure_Hash_Algorithms and picked the lowest one with both green fields. 😶‍🌫️

Though it seems that 256 is okay for checksums, we use 256 for our downloads. And in this scenario probably even md5 would be good enough.

ryanclark: Yeah that's fair, I doubt we'd end up with a collision between yarn.lock changes, even if the hashing algorithm has been proved to collide (md5 for example)

gzdunek (approved):

PR #1048, opened 2022-07-28
Set TCP protocol explicitly

gzdunek:
When protocol is not set, `resolveNetworkAddress` fails to match it.

### DIFF
--- a/packages/teleterm/src/mainProcess/runtimeSettings.ts
+++ b/packages/teleterm/src/mainProcess/runtimeSettings.ts
@@ -134,8 +134,8 @@ function requestChildProcessesAddresses(): ChildProcessAddresses {
   switch (process.platform) {
     case 'win32': {
       return {
-        tsh: 'localhost:0',
-        shared: 'localhost:0',
+        tsh: 'tcp://localhost:0',
+        shared: 'tcp://localhost:0',
       };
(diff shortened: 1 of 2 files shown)


packages/teleterm/src/mainProcess/runtimeSettings.ts
```diff
@@ -134,8 +134,8 @@ function requestChildProcessesAddresses(): ChildProcessAddresses {
   switch (process.platform) {
     case 'win32': {
       return {
-        tsh: 'localhost:0',
-        shared: 'localhost:0',
+        tsh: 'tcp://localhost:0',
+        shared: 'tcp://localhost:0',
```
ravicious, line 138: Down the line we'll have to update this comment here, right? https://github.com/gravitational/teleport/blob/fa65fd02b1a72bd9cb5fe2471b4d179ecd03fb3f/lib/teleterm/config.go#L60

ravicious (approved):

packages/teleterm/src/mainProcess/runtimeSettings.ts
gzdunek: We can, but on the other hand it still tells the truth, you can pass the address without protocol.

PR #1051, opened 2022-07-28
Checkout webassets at the right branch

zmb3:
Rather than checking out master and then switching to the appropriate
branch, we just check out the right branch from the beginning.

This works because the branches in gravitational/webassets match up
with the branches in this repository.

### DIFF
--- a/.github/workflows/update-webassets.yaml
+++ b/.github/workflows/update-webassets.yaml
@@ -28,6 +28,7 @@ jobs:
           repository: gravitational/webassets-fork # TODO(zmb3): switch off the fork
           submodules: recursive
           path: dist
+          ref: ${{ github.ref_name }}
 
       # This workflow only runs on push to a protected branch, so the code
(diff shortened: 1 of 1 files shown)


ryanclark (approved):

.github/workflows/update-webassets.yaml
```diff
@@ -28,6 +28,7 @@ jobs:
           repository: gravitational/webassets-fork # TODO(zmb3): switch off the fork
           submodules: recursive
           path: dist
+          ref: ${{ github.ref_name }}
```
hatched, line 31: Might be worth adding a comment here because I don't think this will work on `webapps` because we use different branch names there.

hatched (approved):
Nice fix!

.github/workflows/update-webassets.yaml
zmb3: Nope, we use the same branch names.

PR #1054, opened 2022-07-28
webassets: stash and pull to stay up to date

zmb3:


avatus (approved):

.github/workflows/update-webassets.yaml
```diff
@@ -43,15 +43,22 @@ jobs:
           git config --global user.email github-goteleport-core-svc@goteleport.com
           git config --global user.name "$GITHUB_ACTOR"
 
+          echo "Updating webassets.e"
           cd dist/e
+          git stash save
```
avatus, line 48: Apparently `git stash save` is deprecated in favor of `git stash push` which is equivalent to just `git stash` now. Should update to one of those instead

PR #1055, opened 2022-07-28
webassets: remove the stash workflow

zmb3:
This didn't work. I don't think it should be necessary anyway.


avatus (approved):
best code is removed code

PR #1056, opened 2022-07-29
[v10] CI: Fail if an update to yarn.lock is needed (#1047)

ravicious:
Backport #1047.


gzdunek (approved):

ravicious:
Hmm, this backport makes no sense as CI doesn't run for teleport-v* branches IIRC?

gzdunek:
Ah, you're right

PR #1058, opened 2022-07-29
Remove building native deps from `yarn build-term`

ravicious:
Calling `electron-builder install-app-deps` during `yarn build-and-package-term` is unnecessary as `electron-builder build` already does that.

The only time where you'd need to run `electron-builder install-app-deps` is when you just cloned the repo and would like to run the dev version of the app without building the prod version.

TODO:

- [x] Explain why
- [x] Make a fresh clone, run build-and-package-term, start packaged app
- [x] Make a fresh clone, run build-native-deps, start app in dev mode
- [x] Make [a tag build](https://drone.platform.teleport.sh/gravitational/teleport/14547/12/1): 11.0.0-dev.14 works just fine.


packages/teleterm/README.md
```diff
@@ -48,28 +48,34 @@ Prepare Webapps repo
...
-**Make sure to run `yarn build-term` first** (as described above) before attempting to launch the
-app in the development mode. That's because Electron is running its own version of Node. That
-command will fetch native packages that were built for that specific version of Node.
+**Make sure to run `yarn build-native-deps-for-term` first** before attempting to launch the app in
+development mode. That's because Electron is running its own version of Node. That command will
+fetch or build native packages that were made for that specific version of Node.
 
-To launch `teleterm` in the development mode:
+```
+$ cd webapps
```
avatus, line 63: what context are we expecting the users to be in when running this command?  If they are in the directory of this README, wouldn't we want them do cd ../../webapps. I don't like that either but, I'm just curious where they would be "starting" from

packages/teleterm/README.md
```diff
@@ -48,28 +48,34 @@ Prepare Webapps repo
...
+$ yarn build-native-deps-for-term
+```
+
+To launch `teleterm` in development mode:
 
 ```sh
 $ cd webapps
 
-## TELETERM_TSH_PATH is the environment variable that points to local tsh binary
+## TELETERM_TSH_PATH is the environment variable that points to the local tsh binary.
```
avatus, line 74: nit: I would change to `variable that points to a local tsh binary` since they technically could have multiple and this assumes they have a tsh binary built in the teleport repo held in a common parent folder. Super minor tho and not a blocker

avatus (approved):

packages/teleterm/README.md
ravicious: Yeah agree, thanks for correcting that! For me it's sometimes hard to nitpick on such minor things so I appreciate that.

package.json
```diff
@@ -10,6 +10,7 @@
     "tdd": "jest --watch",
     "package-term": "yarn workspace @gravitational/teleterm package",
     "build-term": "yarn workspace @gravitational/teleterm build",
+    "build-native-deps-for-term": "yarn workspace @gravitational/teleterm build",
```
gzdunek, line 13: Shouldn't it be `yarn workspace @gravitational/teleterm build-native-deps`?

packages/teleterm/README.md
ravicious: I think we expect them to be in the root of webapps, I mean at least that's where I run my commands from. `cd webapps` is stated explicitly to follow what the other snippets do. I suspect they mostly do it because some steps require you to be in the teleport repo but others need you to be in webapps.

package.json
ravicious: lmao thanks, just before pushing this PR I was like "It looks fine to me but I hope someone else will check this". 😶‍🌫️

Initially I added just `build-native-deps` in the package.json for teleterm. The readme also said "Remember to run `yarn build-native-deps`". But when I tried to actually follow the instructions from scratch, I realized that I'd have to be in `packages/teleterm` in order to run that command, hence last minute addition of `build-native-deps-for-term` which I copy pasted from `build-term` and forgot to change the last part.

gzdunek (approved):

PR #1070, opened 2022-08-02
Move focus to active document

gzdunek:
Closes https://github.com/gravitational/webapps.e/issues/323

`useRefAutoFocus` was added mainly because of the changes from this PR: https://github.com/gravitational/webapps/pull/1071
If we decide to not merge it, then probably I would remove this hook and just put the code directly in `Document`.

### DIFF
--- a/packages/shared/hooks/useRefAutoFocus/useRefAutoFocus.ts
+++ b/packages/shared/hooks/useRefAutoFocus/useRefAutoFocus.ts
@@ -1,20 +1,20 @@
 import { DependencyList, MutableRefObject, useEffect, useRef } from 'react';
 
 /**
- * Returns `ref` object that is automatically focused when `canFocus` is `true`.
+ * Returns `ref` object that is automatically focused when `shouldFocus` is `true`.
  * Focus can be also re triggered by changing any of the `refocusDeps`.
  */
 export function useRefAutoFocus<T extends { focus(): void }>(options: {
-  canFocus: boolean;
+  shouldFocus: boolean;
   refocusDeps?: DependencyList;
 }): MutableRefObject<T> {
   const ref = useRef<T>();
 
   useEffect(() => {
-    if (options.canFocus) {
+    if (options.shouldFocus) {
       ref.current?.focus();
     }
-  }, [options.canFocus, ref, ...(options.refocusDeps || [])]);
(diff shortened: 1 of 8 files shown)


packages/shared/hooks/useRefAutoFocus/useRefAutoFocus.ts
```diff
@@ -0,0 +1,20 @@
+import { DependencyList, MutableRefObject, useEffect, useRef } from 'react';
+
+/**
+ * Returns `ref` object that is automatically focused when `canFocus` is `true`.
+ * Focus can be also re triggered by changing any of the `refocusDeps`.
+ */
+export function useRefAutoFocus<T extends { focus(): void }>(options: {
```
ravicious, line 7: I wasn't sure if `useRefAutoFocus` is the right name but I guess it really is an autofocus _if_ `canFocus` is true – it'll call focus as soon as the component is rendered.

packages/shared/hooks/useRefAutoFocus/useRefAutoFocus.ts
```diff
@@ -0,0 +1,20 @@
...
+  canFocus: boolean;
+  refocusDeps?: DependencyList;
+}): MutableRefObject<T> {
+  const ref = useRef<T>();
+
+  useEffect(() => {
+    if (options.canFocus) {
```
ravicious, line 14: Linguistic nit: An option name such as `canFocus` makes sense when talking from the overall perspective of the problem you were trying to fix with this PR: a `<Document>` cannot be focused when it's not visible so we should not attempt to call `focus` in that scenario.

However, from the perspective of this hook in separation, `useRefAutoFocus` always _can_ focus a ref, it's just a matter of whether it _should_ or _shouldn't_ do that, isn't it?

ravicious (approved):
This is pretty damn cool! I like the way you decided to implement this because it takes care of like 90% of cases and provides a generalized solution to this that should be hard to break by mistake in the future.

And the tests are nice too, I didn't know about `rerender`.

I ran into an edge case with modals but I'll bring that up under gravitational/webapps.e#323.

packages/shared/hooks/useRefAutoFocus/useRefAutoFocus.ts
gzdunek: Absolutely! I changed the name to `shouldFocus`.

PR #1071, opened 2022-08-02
Refactor input focusing after transition

gzdunek:
### Motivation
In order to make some changes I in my other PR, I wanted to call `focus()` on `FieldInput`. Unfortunately, I could not do this because `ref` was not forwarded. After adding `forwardRef`, I had a problem with the operations performed on `inputRef` (I would have to merge the forwarded ref and the local `inputRef`, which is not difficult, but I didn't want to complicate things), but it led me to wonder if we really need these event handlers in `FieldInput`.

BTW, in the end it turned out that I didn't need that `call()` on `FieldInput`, but I think the changes I made here still make sense. 

### Problem
- Listening for the `transitionend` events in the `FieldInput` breaks the separation of concerns rule. The `FieldInput` shouldn't be aware of such specific events that only make sense in the context of  `StepSlider`.
It should be a responsibility of `StepSlider` to tell its children "the transition has finished, it is safe to do operations on the layout".
- We have such conditions that seem to be very error-prone:
```typescript
transitionPropertyName={willTransition ? 'height' : ''}
```
- The current solution only works for `FieldInput`, if we want to add this to another field we would have to add the same props and reuse the same logic.

### Solution
`StepSlider` can have a flag that tells the children that the transition is done (no event handlers!).
The actual input focusing can be done by a separate hook. I implemented it in `useRefAutoFocus` (which can focus any object with `focus()` function).

I tested my changes using the storybook examples (thanks for adding them!) and it seems to work as before. I hope I did not miss any edge case ;) 

What do you think @kimlisa?

### DIFF
--- a/.cloudbuild/ci/build.yaml
+++ b/.cloudbuild/ci/build.yaml
@@ -21,6 +21,8 @@ steps:
     id: install-deps
     args:
       - build
+      - '--build-arg'
+      - YARN_FROZEN_LOCKFILE=true
       - .
 
   - name: gcr.io/cloud-builders/docker
@@ -30,6 +32,8 @@ steps:
       - build
       - '--build-arg'
       - NPM_SCRIPT=lint
+      - '--build-arg'
+      - YARN_FROZEN_LOCKFILE=true
       - .
 
   - name: gcr.io/cloud-builders/docker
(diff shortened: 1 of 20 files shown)


kimlisa (approved):
wow this is lovely! 🤯  thanks for simplifying things, learned about forward ref. i integrated it with the passwordless changes too and it works awesome.

when do you think you'll be able to merge it in? I should probably wait for this to merge before I merge my passwordless PR (no rush)

gzdunek:
>I checked that the slider story works in the same before & after this change. I didn't have time to test the changes made in Welcome/NewCredentials but it'd be worthwhile to make sure those still work properly, particularly the autofocus.

I checked that and it seems to works fine.

PR #1072, opened 2022-08-02
Add a playback speed selector for Desktop Access recordings

xacrimon:
This PR builds on https://github.com/gravitational/teleport/pull/15124 to add a playback speed selector that adjusts the playback speed. A demo video of this is available in the linked PR.


packages/teleport/src/Player/ProgressBar/ProgressBar.tsx
```diff
@@ -55,8 +56,38 @@ export type ProgressBarProps = {
   toggle: () => void;
   style?: React.CSSProperties;
   id?: string;
+  playSpeed?: (speed: number) => void;
```
ibeckermayer, line 59: nit
```suggestion
  onPlaySpeedChange?: (newSpeed: number) => void;
```

ibeckermayer (approved):
Nice work, looks good

packages/teleport/src/Player/ProgressBar/ProgressBar.tsx
```diff
@@ -55,8 +56,38 @@ export type ProgressBarProps = {
...
 };
 
+function PlaySpeedSelector(props: { onChange?: (speed: number) => void }) {
+  if (!props.onChange) {
+    return null;
+  }
+
+  let handleChange = (event: any) => {
```
kimlisa, line 67:
```suggestion
  const handleChange = (event: any) => {
```
also, that `any` should be typed, i think you can hover over onChange and see the type (depending on what IDE you use), i think it would be `React.ChangeEvent<HTMLSelectElement>`

yarn.lock
```diff
@@ -2672,7 +2672,7 @@
     "@types/minimatch" "*"
     "@types/node" "*"
 
-"@types/google-protobuf@^3.10.0", "@types/google-protobuf@^3.7.2":
+"@types/google-protobuf@^3.10.0":
```
kimlisa, line 2675: this file needs to be reverted, you probably need to init your `e` submodule, i think our `webapps.e` uses the `3.7.2` version

kimlisa (approved):

Issue #615, opened 2022-02-22
Create `TdpClientEvent.CLIENT_ERROR` event in `lib/tdp/client.ts`


ibeckermayer:
Currently the `handleMfaChallenge` function emits `TdpClientEvent.TDP_ERROR` events for errors that aren't in fact tdp errors sent to us over the tdp connection, but are more accurately frontend client errors. 

dd a new `TdpClientEvent.CLIENT_ERROR` event and the necessary listeners to handle it in order to maintain the proper boundaries of our abstractions.

PR #1074, opened 2022-08-02
`SharedDirectoryMoveResponse`

ibeckermayer:
Unfortunately when doing research for this feature, I failed to notice that the browser API I was planning to use for moving files and directories was unfinished https://github.com/gravitational/webapps/issues/1064.

In the meantime, I've decided to make all error notifications show up in a modal (instead of a banner), and make a distinction between fatal and non-fatal errors, with non-fatal errors being displayed as a dismissible modal

image

and fatal errors showing up as a modal with a button that allows you to refresh the page

image

(See the video at the very bottom of this description for the modals in action).

Playback is currently unaffected by these changes, because directory sharing events are not captured or played back, the only non-fatal error is thrown by a directory sharing message. When this changes, we may need to refactor the playback system to handle non-fatal errors accurately.

Requires backport to v9/v10

closes https://github.com/gravitational/webapps/issues/615

#### `webapps`

## What to expect

image

image

Try moving the file from one folder in the shared directory to another, and you should see something like

https://user-images.githubusercontent.com/13578537/182468149-27e963e3-5433-4b87-8044-62be915411ef.mov


packages/teleport/src/DesktopSession/DesktopSession.tsx
```diff
@@ -40,42 +46,117 @@ export function DesktopSession(props: State) {
...
-        <DesktopSessionAlert my={2} mx={10} children={alertText} />
+        <Dialog
+          dialogCss={() => ({ width: '484px' })}
+          onClose={onDialogClose}
+          open={errorDialog.open}
+        >
+          <DialogHeader style={{ flexDirection: 'column' }}>
+            {errorDialog.fatal && <DialogTitle>Fatal Error</DialogTitle>}
+            {!errorDialog.fatal && (
+              <DialogTitle>Dismiss to Continue</DialogTitle>
```
kimlisa, line 131: I think it will be better to say something like `Unsupported Action` or something among those lines so at least i get the big picture before i read the message

packages/teleport/src/DesktopSession/DesktopSession.tsx
```diff
@@ -40,42 +46,117 @@ export function DesktopSession(props: State) {
...
+    if (fetchAttempt.status === 'failed') {
+      errorText = fetchAttempt.statusText || 'fetch attempt failed';
+    } else if (tdpConnection.status === 'failed') {
+      errorText = tdpConnection.statusText || 'tdp connection failed';
+    } else if (tdpConnection.status === '') {
+      errorText = tdpConnection.statusText || 'encountered a non-fatal error';
+    } else if (clipboardError) {
+      errorText = clipboardState.errorText || 'clipboard sharing failed';
+    } else if (unknownConnectionError) {
+      errorText = 'Session disconnected for an unknown reason';
```
kimlisa, line 111:
```suggestion
      errorText = 'Session disconnected for an unknown reason. Try reloading (or refreshing) the browser.';
```

and then the button could say `Reload` or `Refresh`? 

also i'm confused whether our error messages should be all lower cap or properly capitalized with punctuation. I'm just seeing a mixture of both. It's not a big deal to me but just wondering in general

packages/teleport/src/DesktopSession/DesktopSession.tsx
```diff
@@ -40,42 +46,117 @@ export function DesktopSession(props: State) {
...
+            )}
+            {errorDialog.fatal && (
+              <ButtonSecondary
+                size="large"
+                width="30%"
+                onClick={() => {
+                  window.location.reload();
+                }}
+              >
+                Retry
```
kimlisa, line 155: look at the prior comment, but if we dont' want to change the error message maybe `Reload Session` perhaps?

packages/teleport/src/DesktopSession/DesktopSession.tsx
```diff
@@ -40,42 +46,117 @@ export function DesktopSession(props: State) {
...
-    alertText = fetchAttempt.statusText || 'fetch attempt failed';
-  } else if (tdpConnection.status === 'failed') {
-    alertText = tdpConnection.statusText || 'tdp connection failed';
-  } else if (clipboardError) {
-    alertText = clipboardState.errorText || 'clipboard sharing failed';
-  } else if (unknownConnectionError) {
-    alertText = 'Session disconnected for an unknown reason';
-  }
+  // Manages the state of the error dialog.
+  const [errorDialog, setErrorDialog] = useState({
```
JanKaczmarkiewicz, line 63: This should not be state since this is only a computation result. In current form react will render twice this component
1. render with the initial value
2. render with the computed value (caused by setErrorDialog in `useEffect`)

If you want to group this logic you could use local function. In this case React will render once with the computed value;
```ts
 const computeErrorDialog = () => {
    const clipboardError = clipboardState.enabled && clipboardState.errorText;

    // Websocket is closed but we haven't
    // closed it on purpose or registered a fatal tdp error.
    const unknownConnectionError =
      wsConnection === 'closed' &&
      !disconnected &&
      (tdpConnection.status === 'success' || tdpConnection.status === '');

    let errorText = '';
    if (fetchAttempt.status === 'failed') {
      errorText = fetchAttempt.statusText || 'fetch attempt failed';
    } else if (tdpConnection.status === 'failed') {
      errorText = tdpConnection.statusText || 'tdp connection failed';
    } else if (tdpConnection.status === '') {
      errorText = tdpConnection.statusText || 'encountered a non-fatal error';
    } else if (clipboardError) {
      errorText = clipboardState.errorText || 'clipboard sharing failed';
    } else if (unknownConnectionError) {
      errorText = 'Session disconnected for an unknown reason';
    }
    const open = errorText !== '';
    const fatal = tdpConnection.status !== '';

    return { open, text: errorText, fatal };
}

const errorDialog = computeErrorDialog()
```

packages/teleport/src/lib/tdp/codec.ts
```diff
@@ -609,6 +617,23 @@ export default class Codec {
     return buffer;
   }
 
+  // | message type (24) | completion_id uint32 | err_code uint32 |
+  encodeSharedDirectoryMoveResponse(res: SharedDirectoryMoveResponse): Message {
```
JanKaczmarkiewicz, line 621: What do you think of an alternative approach for encoding similar to https://github.com/gravitational/webapps/pull/1003#discussion_r931866959

packages/teleport/src/lib/tdp/client.ts
```diff
@@ -300,16 +306,25 @@ export default class Client extends EventEmitterWebAuthnSender {
...
-    // TODO(isaiah): here's where we'll respond with a SharedDirectoryMoveResponse
+    this.sendSharedDirectoryMoveResponse({
+      completionId: req.completionId,
+      errCode: SharedDirectoryErrCode.Failed,
+    });
+    this.handleError(
+      new Error(
+        'Moving the files and directories within a shared \
+        directory on the Windows side is not supported. \
+        Try moving the files on your local machine instead.'
```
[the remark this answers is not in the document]
ibeckermayer: I made it just 'Moving files and directories within a shared directory is not supported.'

packages/teleport/src/DesktopSession/DesktopSession.tsx
ibeckermayer: > also i'm confused whether our error messages should be all lower cap or properly capitalized with punctuation.

In my opinion if we're displaying it as a message to the user, we should strive to give it proper capitalization and punctuation.

packages/teleport/src/DesktopSession/DesktopSession.tsx
ibeckermayer: Changed to "Refresh"

packages/teleport/src/DesktopSession/DesktopSession.tsx
ibeckermayer: Good catch

packages/teleport/src/lib/tdp/codec.ts
ibeckermayer: Great idea. I've added it as an issue: https://github.com/gravitational/webapps/issues/1087

kimlisa (approved):

PR #1086, opened 2022-08-04
Reorganize approach to cluster names

gzdunek:
Closes https://github.com/gravitational/webapps.e/issues/261
tshd counterpart: https://github.com/gravitational/teleport/pull/15200

`getClusterName` function no longer makes sense as it only returns the passed value.


packages/teleterm/src/ui/DocumentTerminal/useDocumentTerminal.ts
```diff
@@ -58,16 +58,16 @@ async function initState(
...
+      return cluster.name;
     }
 
     /*
      When restoring the documents, we do not always have the leaf clusters already fetched.
      In that case we can fall back to `clusterId` from a leaf cluster URI
-     (for a leaf cluster `clusterId` === `actualName`)
+     (for a leaf cluster `clusterId` === `name`)
     */
     const parsed = routing.parseClusterUri(clusterUri);
```
ravicious, line 72: That damn hack lol. But maybe it gives us some insight for the future – we should strive to make it so that `Document` holds all information that's required to connect to a resource without having to wait for data from another source.

packages/teleterm/src/ui/TopBar/Clusters/ClustersFilterableList/ClusterItem.tsx
```diff
@@ -24,7 +23,7 @@ export function ClusterItem(props: ClusterItemProps) {
   });
 
   const LabelVersion = props.isSelected ? InvertedLabel : Label;
-  const clusterName = getClusterName(props.item);
+  const clusterName = props.item.name;
```
ravicious, line 26: Good job differentiating between places that know upfront cluster is present vs those that don't. 👍

ravicious (commented):
I didn't have time to test it but the changes look alright.

What about places that parse out "root cluster name" from the URI? I think there were a few in webapps and one or two in teleport. The only thing we need to do there is to make sure the assumptions they make are not wrong.

packages/teleterm/src/ui/TopBar/Clusters/ClustersFilterableList/ClusterItem.tsx
avatus: Could we use optional chaining like `props.item?.name`. I know it _should_ always exist but might as well throw the optional chain in there to make sure a renderer doesn't break on the off chance that name doesn't end up where we expect it.

packages/teleterm/src/ui/DocumentTerminal/useDocumentTerminal.ts
```diff
@@ -81,7 +81,7 @@ async function initState(
 
   const clusterUri = routing.getClusterUri(doc);
```
ravicious, line 82: Could you update the call here so that `leafClusterId` and `rootClusterId` are explicitly read from doc so that the usage of those fields shows up when using TypeScript editor plugins?

packages/teleterm/src/ui/StatusBar/useActiveDocumentClusterBreadcrumbs.ts
```diff
@@ -31,6 +31,6 @@ export function useActiveDocumentClusterBreadcrumbs(): string {
```
ravicious, line 31: Same as in the previous comment, could you update the definition of `getResourceUri` so that it reads the fields directly?

packages/teleterm/src/ui/DocumentTerminal/useDocumentTerminal.ts
```diff
@@ -81,7 +81,7 @@ async function initState(
...
   const rootCluster = ctx.clustersService.findRootClusterByResource(clusterUri);
-  const cmd = createCmd(doc, rootCluster.proxyHost, getClusterActualName());
+  const cmd = createCmd(doc, rootCluster.proxyHost, getClusterName());
```
ravicious, line 84: Some change somewhere broke connecting to SSH servers on leaf clusters. When I'm trying to connect, I get this error in the terminal tab:

> ERROR: failed connecting to node 4bbde2c9-b805-4dd8-8032-a0bed162d85b. unable to locate node matching UUID-like target 4bbde2c9-b805-4dd8-8032-a0bed162d85b

(Attaching a comment to this line of code just because it's related, I didn't actually check if it's the one causing problems).

ravicious (commented):
Autocomplete works just fine but I found that I cannot ssh into servers on leaf clusters.

packages/teleterm/src/ui/DocumentTerminal/useDocumentTerminal.ts
gzdunek: Done

packages/teleterm/src/ui/StatusBar/useActiveDocumentClusterBreadcrumbs.ts
gzdunek: Done

packages/teleterm/src/ui/DocumentTerminal/useDocumentTerminal.ts
gzdunek: Ah, I was wrong when I removed `leafClusterId` from `DocumentTshNode`. It is should be passed to `routing.getClusterUri(doc)` (from the comment above) to get server from the correct cluster. 
I reverted that commit, because `Terminal` component should also work for kube documents, so `leafClusterId` is needed as well.

Reading `leafClusterId` and `rootClusterId` explicitly helped find this bug 😏

ravicious (approved):

packages/teleterm/src/ui/DocumentTerminal/useDocumentTerminal.ts
ravicious: I'm not sure why TypeScript can't infer that those fields are read there, after all TypeScript is okay with passing a doc there as `Params`.

PR #1088, opened 2022-08-05
[Discover] Add resource discovery polling and static OS logins

kimlisa:
#### Description

- Enabled flow (albeit hackily) for the sake of testing the flow
- Added polling during download script step, which just periodically queries for the resource given a search filter for `reference resource ID` which is returned by creation of a join token, and will be used by the script to uniquely label the added resource. 
- Changed `principal` language with `OS usernames` (can't remember if it was `OS usernames` or `OS logins`)
- Added the static logins that we retrieved by querying for the resource (currently fetching the user traits only returns the dynamic logins defined by the user). Since static logins cannot be modified it is disabled in the list.

#### Required before merge
- [ ] https://github.com/gravitational/teleport/pull/15114

#### How to test
1) You need an account from this cluster (DM me, I can't disable 2FA for cloud): https://kimlisa2.cloud.gravitational.io/web, this cluster contains @marcoandredinis backend changes
2) Check this branch out and in `packages/teleport/src/config.ts`, set the flag value `enabledDiscoverWizard` to `true`
3) Start dev server `yarn start-teleport-e --target=https://kimlisa2.cloud.gravitational.io/web`
4) Click on `Discovery Wizard` from dashboard top right corner menu and go through the flow.

Notes on testing:
- When you run the download script, you'll need to add `-k` flag to the curl command to use insecure connection
- You'll need to nuke any existing teleport data (the script will tell you how)

#### Demo

I'm kinda uncertain about the language i used for the different polling states, any suggestions to make it clearer to user would be great. I'm also uncertain what to do for the error state. Currently it provides a user a button to regenerate script and start polling again. @xinding33 

happy path demo (installer takes some time):

polling-demo2

polling state:

image

success state:

image

error state:

image


packages/teleport/src/Discover/Shared.tsx
```diff
@@ -33,13 +33,26 @@ export const Header: React.FC = ({ children }) => (
   </Text>
 );
 
-export const ActionButtons = ({ onProceed }: { onProceed(): void }) => {
+export const ActionButtons = ({
+  onProceed,
+  isProceedDisabled,
+}: {
+  onProceed?(): void;
+  isProceedDisabled?: boolean;
```
hatched, line 41: this name hints that it's a function call when it's' an argument so it should probably be something like 
```suggestion
  disableProceed?: boolean;
```

alex-kovoy-dd (commented):
See if you can use this one https://github.com/gravitational/webapps/blob/master/packages/teleport/src/components/AjaxPoller/AjaxPoller.jsx

Once done, this component can be unmounted 

```tsx
{ isPolling && <AjaxPoller ... /> }
```

packages/teleport/src/Discover/DownloadScript/useDownloadScript.tsx
```diff
@@ -22,20 +22,120 @@ import { AgentStepProps } from '../types';
...
+      })
     );
-  }, []);
+
+    return () => {
+      cleanUp();
+    };
+  }, [retry]);
+
+  function regenerateScriptAndRepoll() {
```
hatched, line 129: Using a 'flag' variable for a hook is an anti-pattern. Hook dependencies should only be values if you can help it.

In this case what I think you can do instead is to have this function call to generate the new token, then set that token into state and use that token as the dependent for the polling call. 

```ts
  function regenerateScriptAndRepoll() {
    const joinToken = fetchJoinToken();
    setJoinToken(joinToken)
  }
```

marcoandredinis:
@kimlisa Please be aware that the required API in the backend was updated
You can check the PR for the details

marcoandredinis:
@kimlisa Please be aware that the required API in the backend was updated
You can check the PR for the details

packages/teleport/src/Discover/DownloadScript/useDownloadScript.tsx
```diff
@@ -22,20 +22,120 @@ import { AgentStepProps } from '../types';
...
+          cleanUp();
+        }, FIVE_MINUTES_IN_MS);
+      })
     );
-  }, []);
+
+    return () => {
+      cleanUp();
+    };
+  }, [retry]);
```
JanKaczmarkiewicz, line 127: Are we ok with skipping hooks dependencies? Eg: currently we will only rerun this effect when `retry` boolean changes. We are ignoring: `pollState`, `ctx.nodesService.fetchNodes`, `ctx.clusterId`, `props.updateAgentMeta`, `props.agentMeta` etc.

hatched: We already have that package installed, looks like the correct config isn't enabled though, thx for the reminder.

hatched (approved):

packages/teleport/src/Discover/DownloadScript/useDownloadScript.tsx
kimlisa: so about this, i know it's an eslint rule, but from prior discussions I had with @alex-kovoy-dd, this rule is just a safeguard and a hint that there might be a bug. As long as we are intentional and know what we are doing, it was okay to ignore this rule. I also feel like providing an exhaustive dependency list is more confusing, but if folks are against this I can start providing the exhaustive list.

kimlisa:
> See if you can use this one https://github.com/gravitational/webapps/blob/master/packages/teleport/src/components/AjaxPoller/AjaxPoller.jsx
> 
> Once done, this component can be unmounted
> 
> ```tsx
> { isPolling && <AjaxPoller ... /> }
> ```

hi @alex-kovoy-dd !!! i opted not to use this component as it will require refactoring (also change to a hook), and I didn't want to spend time on refactoring for a feature that is always changing (we will revisit later though, when it's time to clean things up)

packages/teleport/src/Discover/DownloadScript/useDownloadScript.tsx
gzdunek: I agree with @JanKaczmarkiewicz it would better to not skip the dependencies (and enable the plugin in the future). I don't see much benefit from skipping them, as we risk potential bugs that may not be visible at first glance. React docs also discourage this. In addition, adding all dependencies reduces the number of decisions because we do not have to think about whether a particular dependency should be added or not.
I'd try to provide exhaustive lists of dependencies, so it will be easier to enable the plugin.
Nice art: [Don’t Lie to React About Dependencies](https://overreacted.io/a-complete-guide-to-useeffect/#dont-lie-to-react-about-dependencies)

PR #1090, opened 2022-08-07
`SharedDirectoryCreateRequest` and `SharedDirectoryCreateResponse`

ibeckermayer:
Adds the TDP messages and `SharedDirectoryManager` method necessary to allow users to copy files and directories _into_ the shared directory.

You should now be able to:
- Navigate the directory tree
- Read files
- Write files (< 16kb temporarily)
- Copy files and directories _from_ the shared directory
- Copy files and directories _into_ the shared directory **(new)**
- Move files and directories _into_ the shared directory **(new)**

Moving files and directories _from_ the shared directory still won't work, since it requires unimplemented `SharedDirectoryDelete` messages.

Corresponds with: https://github.com/gravitational/teleport/pull/15289

Requires backport to v9/v10

#### `webapps`

## What to expect

image

image


packages/teleport/src/lib/tdp/client.ts
```diff
@@ -136,6 +137,14 @@ export default class Client extends EventEmitterWebAuthnSender {
         case MessageType.SHARED_DIRECTORY_INFO_REQUEST:
           this.handleSharedDirectoryInfoRequest(buffer);
           break;
+        case MessageType.SHARED_DIRECTORY_CREATE_REQUEST:
+          // A typical sequence is that we receive a SharedDirectoryCreateRequest
+          // immediately followed by a SharedDirectoryWriteRequest. It's important
+          // that we await here so that this client doesn't field the SharedDirectoryWriteRequest
+          // until the create has successfully completed, or else we might get an error
+          // trying to write to a file that hasn't been created yet.
+          await this.handleSharedDirectoryCreateRequest(buffer);
```
hatched, line 146: nice comment, thx

packages/teleport/src/lib/tdp/client.ts
```diff
@@ -102,7 +103,7 @@ export default class Client extends EventEmitterWebAuthnSender {
     };
   }
 
-  processMessage(buffer: ArrayBuffer) {
+  async processMessage(buffer: ArrayBuffer) {
```
hatched, line 106: now that this is an `async` function you should probably type the return value here.
```suggestion
  async processMessage(buffer: ArrayBuffer): Promise<void> {
```

packages/teleport/src/lib/tdp/playerClient.ts
```diff
@@ -42,7 +42,7 @@ export class PlayerClient extends Client {
   }
 
   // Overrides Client implementation.
-  processMessage(buffer: ArrayBuffer) {
+  async processMessage(buffer: ArrayBuffer) {
```
hatched, line 45: See comment above about the return type.

packages/teleport/src/lib/tdp/sharedDirectoryManager.ts
```diff
@@ -11,6 +11,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+import { FileType } from './codec';
```
hatched, line 14: Plz be explicit when importing types.

```suggestion
import type { FileType } from './codec';
```

hatched (approved):

packages/teleport/src/lib/tdp/client.ts
ibeckermayer: Why is that particularly important for `async` functions?

packages/teleport/src/lib/tdp/sharedDirectoryManager.ts
ibeckermayer: It's used as an object here

https://github.com/gravitational/webapps/blob/189c1d4ed84ed7a282c7028239ac3ce199f4e668/packages/teleport/src/lib/tdp/sharedDirectoryManager.ts#L166

hatched: Ahh I see now, thanks sorry I missed that.

packages/teleport/src/lib/tdp/client.ts
hatched: async functions have an implicit return type of a `Promise` but the value from within that promise isn't automagically picked up by the compiler so by specifying the type in the generic there it just helps it be more explicit in the editor.

JanKaczmarkiewicz: > value from within that promise isn't automagically picked up by the compiler

it is actually:
Screenshot 2022-08-09 at 18 13 52

hatched: Oh nice, it wasn't working for me in this PR so that's why I suggested the change.

PR #1091, opened 2022-08-08
Show Connect icon in a better quality

gzdunek:
This is a very small change, but enables proper icon scaling on Windows and Linux (here the icon wasn't shown at all before).
The [documentation is quite poor](https://www.electron.build/icons.html), it take me a while to figure out that it should be a directory, not a simple file.

Before:
image

After:
image

(now I see it also fixes icon size in the title bar)


avatus (approved):
Nice!

ravicious (approved):

PR #1100, opened 2022-08-10
Update app access events

r0mant:
- Add `app.session.end` event which was introduced with TCP app access.
- Update other app events to include app name in the description instead of session ID which makes them more useful.

Refs https://github.com/gravitational/teleport/issues/15397. Needs v10 backport.


packages/teleport/src/Audit/__snapshots__/Audit.story.test.tsx.snap
```diff
@@ -4103,43 +4140,6 @@ exports[`list of all events 1`] = `
           </button>
         </td>
       </tr>
-      <tr>
```
r0mant, line 4106: Unrelated to app access events, but the tests were failing because this event seems to be missing from fixtures so I removed it from here as well.

kimlisa (approved):

PR #1104, opened 2022-08-11
Update sidebar styling

hatched:


packages/teleport/src/Discover/Discover.tsx
```diff
@@ -113,10 +118,10 @@ function TopBar(props: { onLogout: VoidFunction; username: string }) {
       height="56px"
       pl={5}
       borderColor="primary.main"
-      border="1px solid"
+      css={{ borderBottomWidth: '1px', borderBottomStyle: 'solid' }}
```
kimlisa, line 121: i guess it's preference, but you can also use actual css property names (just fyi):

```suggestion
      css={`border-bottom-width: 1px, border-bottom-style: solid`}
```

kimlisa (approved):

PR #1106, opened 2022-08-11
Add functional empty states to discover UI

hatched:
- Moved all of the add resource popups into the Discover UI.
- Clicking links in the resource lists now takes you to the Discover UI.
- SlideTab component now supports specifying a starting index.


packages/teleport/src/Discover/SelectResource/SelectResource.tsx
```diff
@@ -148,25 +171,66 @@ export function SelectResource({ nextStep }: Props) {
...
+              //   type: selectedType,
+              // });
+            }}
+          >
+            Proceed
+          </ButtonPrimary>
+          <ButtonSecondary as="a" href={cfg.routes.root} size="medium">
+            Go to dashboard
+          </ButtonSecondary>
+        </Box>
```
kimlisa, line 230: lets use the `ActionButtons` component in the `Shared` file

kimlisa:
missed a few edge cases:
- windows desktop:
  - list resources empty view: it shows a `add` button when this state is not handled for desktops yet
  - discover view: renders the incorrect list resources empty view
- we seem to have two different actions of `add XXX` (i only tested for server) depending on where the add button is clicked:
  - clicking on `add server` from the empty component goes to discover _however_,
  - clicking on `add server` standalone button does NOT go to discover but shows the old dialogue in place
  - also, do we need the logic for servers for the old dialogue? this can be removed

kimlisa:
looks like we have some updated language, might as well include it in this PR as well: https://www.figma.com/file/8DQZl1mYUPLgvKgmA6zPOM/In-Progress-Work?node-id=1492%3A71982

packages/teleport/src/Discover/SelectResource/SelectResource.tsx
```diff
@@ -35,22 +42,71 @@ import type { TabComponent } from 'design/SlideTabs/SlideTabs';
...
-  return <SelectResource nextStep={props.nextStep} />;
+  const { clusterId, isLeafCluster } = useStickyClusterId();
+  const ctxState = ctx.storeUser.state;
+  return (
+    <SelectResource
+      authType={ctxState.authType}
+      canCreate={ctx.storeUser.getTokenAccess().create}
+      clusterId={clusterId}
+      isEnterprise={ctx.isEnterprise}
+      isLeafCluster={isLeafCluster}
```
kimlisa, line 57: `isLeafCluster` required? this is only used to disable adding as users can only add new resources on root cluster

packages/teleport/src/Discover/SelectResource/SelectResource.tsx
```diff
@@ -35,22 +42,71 @@ import type { TabComponent } from 'design/SlideTabs/SlideTabs';
 import type { ResourceType, ResourceLocation } from '../resource-lists';
 import type { AgentStepProps } from '../types';
 import type { State } from '../useDiscover';
+import type { AuthType } from 'teleport/services/user';
 
 export default function Container(props: AgentStepProps) {
   const ctx = useTeleport();
-  return <SelectResource nextStep={props.nextStep} />;
+  const { clusterId, isLeafCluster } = useStickyClusterId();
```
kimlisa, line 49: we don't need this hook, we can only add new resources on the root cluster, we can get the clusterId from `cfg.proxyCluster` like it does inside the hook

kimlisa (approved):

packages/teleport/src/Kubes/Kubes.tsx
```diff
@@ -119,7 +118,6 @@ export function Kubes(props: State) {
         <Empty
           clusterId={clusterId}
           canCreate={canCreate && !isLeafCluster}
-          onClick={() => setShowAddKube(true)}
```
JanKaczmarkiewicz, line 122: From what I see there is no way for `showAddKube` to be true. Is this intended?

hatched: Yep you're right it was missed in this PR but it's being removed in another one :)

PR #1109, opened 2022-08-12
[Discovery] Add Finish Component and Tweaks

kimlisa:
#### Description

#### Screenshot
Finished component

image

Count down timer 

image

Updated test connection output

image


packages/teleport/src/Discover/Discover.tsx
```diff
@@ -167,18 +169,24 @@ function SideNavAgentConnect({ currentStep }) {
...
-          {agentStepTitles.map((step, index) => {
+          {agentStepTitles.map((stepTitle, index) => {
             let className = '';
             if (currentStep > index) {
               className = 'checked';
             } else if (currentStep === index) {
               className = 'active';
             }
 
+            // All flows will have a finished step that
```
hatched, line 180: Instead of having it be '', should we just add a "Finished" step title?

packages/teleport/src/Discover/DownloadScript/DownloadScript.tsx
```diff
@@ -69,22 +70,29 @@ export function DownloadScript({
...
+              <TextIcon
+                css={`
+                  white-space: pre;
+                `}
+              >
                 <Icons.Restore fontSize={4} />
-                Waiting for resource discovery...
+                {`Waiting for node   |   ${formatTime(
+                  countdownTime
+                )} until script expires`}
```
hatched, line 81:
```suggestion
                )} until this script expires`}
```

packages/teleport/src/Discover/DownloadScript/useDownloadScript.tsx
```diff
@@ -24,13 +29,18 @@ import { AgentStepProps } from '../types';
...
-const FIVE_MINUTES_IN_MS = 300000;
+const FIVE_MINUTES = 5;
 const THREE_SECONDS_IN_MS = 3000;
+const ONE_SECOND_IN_MS = 1000;
 
 export function useDownloadScript({ ctx, props }: Props) {
   const { attempt, run, setAttempt } = useAttempt('processing');
   const [joinToken, setJoinToken] = useState<JoinToken>();
   const [pollState, setPollState] = useState<PollState>('polling');
+  const [countdownTime, setCountdownTime] = useState<CountdownTime>({
```
hatched, line 40: The logic for this countdown is spread out all over the place. I think it would be a lot easier to maintain (and use elsewhere if needed) if it was its own separate component

`<Countdown time="00:05:00" />`

hatched (changes requested):

packages/teleport/src/Discover/Discover.tsx
kimlisa: we can discuss it next week, keeping it as is until then

packages/teleport/src/Discover/DownloadScript/useDownloadScript.tsx
kimlisa: if there is no faulty logic, i would like to keep this as is, until it's time to refactor (ill put a TODO marker)

hatched (approved):

PR #1110, opened 2022-08-13
Add warning dialog for unsupported browsers for directory sharing

ibeckermayer:
Adds logic for popping up a non-fatal dialog when the user attempts to initiate directory sharing on an unsupported browser.

Requires backport to v9/v10

#### `webapps`

## What to expect
**Due to an arbitrary choice in the ordering of a particular if-else-if statement, you must be logged in as a user with clipboard access disabled to test this change manually**

From a browser which doesn't support the File Access API (Safari/FF):

image

You should see a non-fatal warning dialog

image

### DIFF
--- a/packages/teleport/src/DesktopSession/DesktopSession.tsx
+++ b/packages/teleport/src/DesktopSession/DesktopSession.tsx
@@ -43,6 +43,8 @@ declare global {
 
 export function DesktopSession(props: State) {
   const {
+    directorySharingBrowserErr,
+    setDirectorySharingBrowserErr,
     clipboardState,
     fetchAttempt,
     tdpConnection,
@@ -62,14 +64,11 @@ export function DesktopSession(props: State) {
   // onDialogClose is called when a user
   // dismisses a non-fatal error dialog.
   const onDialogClose = () => {
-    // This setTdpConnection call will cause the useEffect below
-    // to calculate the errorDialog state.
+    // The following state-setting calls will
+    // cause the useEffect below to calculate the
+    // errorDialog state.
+
     setTdpConnection(prevState => {
-      // onDialogClose should only be called when
-      // the user dismisses a non-fatal error dialog,
-      // and prevState.status === '' means non-fatal
-      // error dialog, so the below if statement should
-      // always be true.
       if (prevState.status === '') {
         // If prevState.status was a non-fatal error,
         // we assume that the TDP connection remains open.
@@ -77,6 +76,8 @@ export function DesktopSession(props: State) {
       }
       return prevState;
     });
+
+    setDirectorySharingBrowserErr(false);
   };
 
   const computeErrorDialog = () => {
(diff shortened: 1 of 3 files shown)


hatched (approved):

packages/teleport/src/DesktopSession/DesktopSession.tsx
```diff
@@ -100,9 +101,16 @@ export function DesktopSession(props: State) {
       errorText = clipboardState.errorText || 'clipboard sharing failed';
     } else if (unknownConnectionError) {
       errorText = 'Session disconnected for an unknown reason.';
+    } else if (directorySharingBrowserErr) {
+      errorText =
+        'Your user role supports directory sharing over desktop access, \
+      however this feature is only available by default on some chromium \
```
zmb3, line 107:
```suggestion
      however this feature is only available by default on some Chromium \
```

packages/teleport/src/DesktopSession/DesktopSession.tsx
```diff
@@ -229,16 +238,20 @@ function Session(props: PropsWithChildren<State>) {
...
+        .showDirectoryPicker()
+        .then(sharedDirHandle => {
+          setIsSharingDirectory(true);
+          tdpClient.addSharedDirectory(sharedDirHandle);
+          tdpClient.sendSharedDirectoryAnnounce();
+        })
+        .catch(() => {
+          setIsSharingDirectory(false);
+        });
+    } catch (e) {
```
zmb3, line 252: I'm not used to seeing a promise's `.catch` used in conjunction with a standard catch block. Are both of them going to be called in the event of an error?

packages/teleport/src/DesktopSession/useDesktopSession.tsx
```diff
@@ -47,6 +47,8 @@ export default function useDesktopSession() {
 
   const [canShareDirectory, setCanShareDirectory] = useState(false);
   const [isSharingDirectory, setIsSharingDirectory] = useState(false);
+  const [directorySharingBrowserErr, setDirectorySharingBrowserErr] =
```
zmb3, line 50: Instead of 3 booleans, would it make more sense to have a directory sharing state? It could encode all of this info:

- has permissions to share
- is actively sharing
- failed to share due to an error

packages/teleport/src/DesktopSession/DesktopSession.tsx
ibeckermayer: > Are both of them going to be called in the event of an error?

No, what this means is if the `Promise` returned by `window.showDirectoryPicker()` is rejected (which happens if the user closes the directory picker without selecting a directory), we `setIsSharingDirectory(false);`. On the other hand, if the call to `window.showDirectoryPicker()` throws an error (as happens on browsers where this feature is not implemented, i.e.:`Uncaught TypeError: window.showDirectoryPicker is not a function`), we `setDirectorySharingBrowserErr(true);`.

Granted I'm making the assumption that those are the only two situations common enough to distinguish between, and we don't need to implement any finer grained handling within each block. In theory it's possible that there's some other reason that the promise might be rejected, and some other error that could be thrown, in which case the behavior may or may not be what we want.

packages/teleport/src/DesktopSession/useDesktopSession.tsx
ibeckermayer: 59af539a62d4d09ccc2f16b8280f185d9378d786

PR #1111, opened 2022-08-16
Move electron-builder to a JS file

ravicious:
This PR includes a few extra changes that I made when preparing the app for Touch ID. I decided to include them here since they're not directly related to Touch ID – that PR will be reviewed by Alan as well so I wanted to reduce the number of unrelated changes to a minimum. ;)

But the main purpose behind this PR is to move electron-builder config to a JS file. This way we'll be able to control the packaging process with env variables which will become necessary for the macOS build as I explained on Slack. Different environments will need to provide different paths to tsh.app and we're going to utilize env vars for that which would be impossible with the config stored in package.json.


packages/teleterm/electron-builder-config.js
```diff
@@ -0,0 +1,70 @@
+/**
+ * @type { import('electron-builder').Configuration }
+ */
```
ravicious, line 3: This doesn't seem to work (just as other typings provided through JSDoc in other files in our project) but I left it here anyway in case we manage to fix it in the future.

packages/teleterm/src/mainProcess/runtimeSettings.ts
```diff
@@ -79,28 +82,25 @@ function getTshHomeDir() {
...
 
-  return tshPath;
-}
-
-function getBinDir() {
-  if (!app.isPackaged) {
-    return;
+  const tshBinPath = env[TSH_BIN_ENV_VAR];
+  if (!fs.existsSync(tshBinPath)) {
+    throw Error(`${TSH_BIN_ENV_VAR} must point at a tsh binary`);
```
ravicious, line 100: In one of the upcoming commits in another PR I'll update that to `throw new Error` so that stacktrace is preserved.

ravicious (commented):
I tested both the macOS and Windows version and they seem to work just fine. I'm going to start a tag build on drone soon.

packages/teleterm/electron-builder-config.js
gzdunek: As we talked, it works only for autocomplete:

image (2)

packages/teleterm/electron-builder-config.js
```diff
@@ -0,0 +1,70 @@
...
+        to: './bin/tsh',
+      },
+    ],
+  },
+  directories: {
+    buildResources: 'assets',
+    output: 'build/release',
+  },
+  extraResources: ['./assets/**'],
+  publish: {
```
gzdunek, line 65: I guess we can remove it?

packages/teleterm/src/mainProcess/runtimeSettings.ts
```diff
@@ -79,28 +82,25 @@ function getTshHomeDir() {
   return tshPath;
 }
 
-function getTshBinaryPath() {
+// binDir is used in the packaged version to add tsh to PATH.
+// tshBinPath is used by Connect to call tsh directly.
+function getBinaryPaths(): { binDir?: string; tshBinPath: string } {
```
gzdunek, line 87: Nit: `getBinariesPaths()`?

gzdunek (approved):
Good job!

packages/teleterm/src/mainProcess/runtimeSettings.ts
ravicious: I've always been pluralizing only the second word but I've never actually checked if that's correct, do you have any resources about that?

If it those were paths to files and not binaries I'd probably call it `getFilePaths` and not `getFilesPaths` so I think `getBinaryPaths` is correct here. 🤔

gzdunek: Ok, your argumentation seems reasonable, let's keep it as it is :)

PR #1118, opened 2022-08-17
Enable no-unused-vars

hatched:


packages/teleport/src/Discover/SelectResource/SelectDBType.tsx
```diff
@@ -0,0 +1,96 @@
+import React, { useState } from 'react';
```
hatched, line 1: Code was moved here so that it can be used for the next step of Discover even though it's technically unused at this point.

kimlisa (approved):

PR #1131, opened 2022-08-18
[Discover] Update copy and design tweaks

kimlisa:
#### Description

Just updating the language and updating some small design changes based on figma


packages/teleport/src/Discover/Discover.tsx
```diff
@@ -110,15 +110,15 @@ export function Discover({
...
-  ];
+const agentStepTitles: string[] = [
+  'Select Resource Type',
+  'Configure Resource',
+  'Set Up Access',
+  'Test Connection',
+  '',
+];
 
+function SideNavAgentConnect({ currentStep }) {
```
ryanclark, line 121: Can we add typings here?

packages/teleport/src/Discover/DownloadScript/DownloadScript.tsx
```diff
@@ -77,35 +78,55 @@ export function DownloadScript({
                 `}
               >
                 <Icons.Restore fontSize={4} />
-                {`Waiting for node   |   ${formatTime(
+                {`Waiting for Teleport SSH Service   |   ${formatTime(
```
ryanclark, line 81: Are we re-rendering this entire component once a second to display the counter?

kimlisa: yes, do you think it could be a problem? i think what we are rendering won't cause any performance problems because it's a pretty light component, but if you feel strongly about it, i can do refactoring on another PR since this PR is mostly about language/design tweaks.

ryanclark: Nah I don't see it being an issue, I imagine we'll refactor it out at some point - I was just curious

ryanclark (approved):

PR #1137, opened 2022-08-23
[v10] Add prettier commands to `package.json`, run prettier

ryanclark:
This needs to be merged before #1136, as that PR backports the prettier check to v10.


gzdunek (commented):
We should update the `.prettierrc` (to match the one from master) and add `.prettierignore` to skip formatting protobuf files.

hatched (approved):

PR #1141, opened 2022-08-23
Turn on directory sharing by default 

ibeckermayer:
Removes `enableDirectorySharing` flag to turn on directory sharing by default

corresponds with https://github.com/gravitational/teleport/pull/15789

requires backport to v10

### DIFF
--- a/packages/teleport/src/services/user/makeAcl.ts
+++ b/packages/teleport/src/services/user/makeAcl.ts
@@ -14,8 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import cfg from 'teleport/config';
-
 import { Acl } from './types';
 
 export default function makeAcl(json): Acl {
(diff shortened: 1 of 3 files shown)


packages/teleport/src/services/user/makeAcl.ts
```diff
@@ -49,8 +47,7 @@ export default function makeAcl(json): Acl {
   // Behaves like clipboardSharingEnabled, see
   // https://github.com/gravitational/teleport/pull/12684#issue-1237830087
   const directorySharingEnabled =
-    (json.directorySharing !== undefined ? json.directorySharing : true) &&
-    cfg.enableDirectorySharing;
+    json.directorySharing !== undefined ? json.directorySharing : true;
```
zmb3, line 50: This handles undefined but not null. Is this intentional?

ibeckermayer: `json` comes [directly from the backend](https://github.com/gravitational/teleport/blob/bc19ff0e439b09f62ec86df821d69a83d285df0a/lib/web/ui/usercontext.go#L91) and is converted to a javascript object, so technically it should always be defined. This logic is just me being overly cautious (similar to the variables above), since technically `json` is type `any` and so according to Typescript, `json.directorySharing` might be `undefined`.

zmb3: I would have just done

```
const directorySharingEnabled = !!json.directorySharing
```

but I suppose it doesn't really matter.

zmb3 (approved):

PR #1153, opened 2022-08-26
Adds special handling for CapsLock on MacOS

ibeckermayer:
See the docstring on `handleCapsLock` for the explanation.

I'd like to double check my understanding that CapsLock is treated normally on Linux and Windows. @kimlisa and @avatus if you wouldn't mind running this change locally on Linux and Windows respectively like

```
yarn start-teleport --target=https://ec2-35-171-27-185.compute-1.amazonaws.com/
```

logging in with the credentials [here](https://gravitational.slack.com/archives/C02DQ1C2BMW/p1657807702085829), and opening a notepad and just checking that caps lock works like you'd expect it to.

Requires backport to v10, v9

closes https://github.com/gravitational/teleport/issues/15849


zmb3 (approved):
Code looks fine to me

packages/teleport/src/DesktopSession/useTdpClientCanvas.tsx
```diff
@@ -104,13 +110,37 @@ export default function useTdpClientCanvas(props: Props) {
     setWsConnection('open');
   };
 
+  const platform = navigator.userAgentData?.platform || navigator.platform;
+  const isMacOS = /mac/i.test(platform);
```
kimlisa, line 114: we have this function: https://github.com/gravitational/webapps/blob/master/packages/design/src/theme/utils/getPlatform.js

kimlisa (approved):
works on chrome on linux

avatus (approved):
Works in Chrome on windows 10

PR #1157, opened 2022-08-29
Update GitHub connector template

zmb3:
We deprecated teams_to_logins in favor of teams_to_roles in v10.
Starting with v11, teams_to_logins will no longer be accepted.


rudream (approved):

zmb3:
Note: backport to v10 only.

PR #1159, opened 2022-08-30
[v10] Add unsafe-eval to CSP in dev mode to make source maps work

ravicious:
Backport #1031 & #1032.

I forgot to backport this so dev mode in v10 wasn't really working but it looks like no one ran into that problem yet.


avatus (approved):

ravicious:
Added a backport for #1032 too.

PR #1169, opened 2022-08-31
Add BannerList and Banner components to display cluster alerts on load

hatched:
When loading the UI, render cluster alerts.

Currently the only supported alert is for an upgrade notification but this component should handle rendering any alerts with multiple severity levels without updates.

Screen Shot 2022-09-06 at 11 59 56 AM

## QA Notes

When dismissing a banner in the Discover UI it'll appear again in the Main app. This will be resolved in a follow-up branch where banner dismiss persistence will be added.

This code is non functional without the corresponding teleport updates: 
- [teleport pr](https://github.com/gravitational/teleport/pull/15774) 
- and a subsequent PR built on it which adds the apiserver endpoints.

Use [this teleport branch](https://github.com/gravitational/teleport/tree/hatched/cluster-alerts-apiserver) which contains all of the necessary updates and hacks to QA this branch.

Start it with `TELEPORT_UNSTABLE_VC_SYNC_ON_START=yes ./build/teleport start`


packages/teleport/src/components/BannerList/Banner.test.tsx
```diff
@@ -0,0 +1,140 @@
...
+describe('components/BannerList/Banner', () => {
+  it('displays the supplied message', () => {
+    const msg = 'I am a banner';
+    render(
+      <Banner message={msg} severity={0} id="test-banner" onClose={() => {}} />
+    );
+    expect(screen.getByText(msg)).toBeInTheDocument();
+  });
+
+  function assertComputedStyle(
```
hatched, line 31: This should probably be moved to the shared testing utils

JanKaczmarkiewicz, line 31: Have you tried to use https://github.com/styled-components/jest-styled-components#tohavestylerule?

packages/teleport/src/components/BannerList/BannerList.tsx
```diff
@@ -0,0 +1,67 @@
...
+  useEffect(() => {
+    const newList = {};
+    banners.forEach(banner => (newList[banner.id] = banner));
+    Object.assign(newList, bannerList);
+    setBannerList(newList);
+  }, [banners]);
+
+  const removeBanner = id => {
+    const newList = { ...bannerList };
+    newList[id].hidden = true;
```
JanKaczmarkiewicz, line 37: this still mutates the original bannerList[id] object reference. You need to shallow copy both the root object and target object 
```
const newList = { ...bannerList, [id]: { ...bannerList[id], hidden: true } };
```

packages/teleport/src/components/BannerList/Banner.test.tsx
```diff
@@ -0,0 +1,140 @@
...
+    styleValue
+  ) {
+    const componentClass = Component(props).type.styledComponentId;
+    const componentElement = doc.getElementsByClassName(componentClass);
+    const style = win.getComputedStyle(componentElement[0]);
+    expect(style[styleName]).toBe(styleValue);
+  }
+
+  it('renders an info banner for under severity 5', async () => {
+    await Promise.all(
```
JanKaczmarkiewicz, line 46: Why we are using Promise.all here? 
```ts
    severity => {
        const bannerProps = {
          message: 'I am steve banner',
          severity,
          id: 'test-banner',
          onClose: () => {},
        };

        const { unmount } = render(<Banner {...bannerProps} />);
        expect(screen.getByRole('icon')).toHaveClass('icon-info_outline');

        assertComputedStyle(
          window,
          document,
          Banner,
          bannerProps,
          'backgroundColor',
          'rgb(3, 155, 229)'
        );

        unmount();
      }```
is a synchronous function so the whole `Promise.all` will execute callbacks sequentially

I think we should just check the severity border values or use https://jestjs.io/docs/api#each

File: packages/teleport/src/components/BannerList/Banner.tsx [hunk 4]
```diff
@@ -0,0 +1,85 @@
...
+  if (severity < 10) {
+    return 'warning';
+  }
+  if (severity >= 10) {
+    return 'danger';
+  }
+  return 'secondary.light';
+}
+
+const Close = styled.div`
```
JanKaczmarkiewicz, line 81: Since this is a clickable element maybe we should use a button tag here to make it focusable

packages/teleport/src/components/BannerList/Banner.tsx
```diff
@@ -0,0 +1,85 @@
...
+          <Cross />
+        </Close>
+      </Flex>
+    </Box>
+  );
+}
+
+function generateIcon(severity: number): ReactNode {
+  // severity is checked using ranges to allow additional severities to be added
+  // on the backend without requiring an update to the client.
```
JanKaczmarkiewicz, line 53: I think we should move backend-specific code out of the reusable UI component.
```ts
// component props
severity: 'info' | 'warning' | 'danger';
```

```ts
// client usage
const mapSeverity = (severity: number) => {
    if (severity < 5) {
      return "info"
    }
    if (severity < 10) {
       return "warning"
    }
    return "danger"
}

const banners = props.alerts
    ? props.alerts.map(alert => ({
        message: alert.spec.message,
        severity: mapSeverity(alert.spec.severity),
        id: alert.metadata.name,
      }))
    : [];

<Banner baners={baners} />
```

packages/teleport/src/Main/Main.tsx
```diff
@@ -99,6 +107,7 @@ export function Main(props: State) {
       <RouterDOM.Switch>
         <Redirect exact={true} from={cfg.routes.root} to={indexRoute} />
       </RouterDOM.Switch>
+      <BannerList banners={banners} />
```
kimlisa, line 110: any reason why this banner isn't fixed at the top like it is for our expired license and switchback banner? without it, you can get awkward/confusing double horizontal scrollers (outer scroll and an inner scroll)

packages/teleport/src/components/BannerList/Banner.tsx
kimlisa: i'd also throw in a "on hover, button gets lighter" or some style

packages/teleport/src/components/BannerList/useAlerts.tsx
```diff
@@ -0,0 +1,49 @@
...
+  useEffect(() => {
+    fetchClusterAlerts(clusterId)
+      .then(res => {
+        if (!res) {
+          return;
+        }
+        if (Array.isArray(res)) {
+          setAlerts(res);
+          return;
+        }
```
kimlisa, line 40: lets put this check at the service level: `json.alerts || []`

packages/teleport/src/config.ts
```diff
@@ -111,6 +111,7 @@ const cfg = {
     applicationsPath:
       '/v1/webapi/sites/:clusterId/apps?searchAsRoles=:searchAsRoles?&limit=:limit?&startKey=:startKey?&query=:query?&search=:search?&sort=:sort?',
     clustersPath: '/v1/webapi/sites',
+    clusterAlertsPath: '/v1/webapi/sites/:clusterId/alerts',
```
kimlisa, line 114: do we get cluster specific alerts? like if we change a cluster, can we get a different set of alerts?

packages/teleport/src/components/BannerList/useAlerts.tsx
```diff
@@ -0,0 +1,49 @@
...
+        setAlerts([]);
+      })
+      .catch(err => {
+        logger.error(err);
+      });
+  }, []);
```
kimlisa, line 46: if we are expecting different set of alerts when cluster changes, we should re-fetch?
```suggestion
  }, [clusterId]);
```

packages/teleport/src/Main/Main.tsx
```diff
@@ -90,6 +90,14 @@ export function Main(props: State) {
     ctx.storeNav.getSideItems()[0]?.getLink(cfg.proxyCluster) ||
     cfg.routes.support;
 
+  const banners = props.alerts
```
JanKaczmarkiewicz, line 93: [nit]
You can destructure alerts from props and assign default [] there
```ts
const { status, statusText, ctx, alerts = [] } = props;
...
const banners = alerts.map(alert => ({
    message: alert.spec.message,
    severity: alert.spec.severity,
    id: alert.metadata.name,
}))
```

packages/teleport/src/components/BannerList/BannerList.tsx
```diff
@@ -0,0 +1,67 @@
...
+        .filter(banner => !banner[1].hidden)
+        .map(banner => (
+          <Banner
+            message={banner[1].message}
+            severity={banner[1].severity}
+            id={banner[1].id}
+            onClose={removeBanner}
+            key={banner[1].id}
+          />
+        ))}
```
JanKaczmarkiewicz, line 53:
```suggestion
    <Box>
      {Object.values(bannerList)
        .filter(banner => !banner.hidden)
        .map(banner => (
          <Banner
            message={banner.message}
            severity={banner.severity}
            id={banner.id}
            onClose={removeBanner}
            key={banner.id}
          />
        ))}
```

packages/teleport/src/components/BannerList/Banner.tsx
```diff
@@ -0,0 +1,85 @@
...
+  id: string;
+  onClose: (id: string) => void;
+};
+
+export function Banner({ id, message = '', severity = 0, onClose }: Props) {
+  return (
+    <Box bg={getBackgroundColor(severity)} p={3}>
+      <Flex alignItems="center">
+        {generateIcon(severity)}
+        <Text bold>{message}</Text>
```
kimlisa, line 37: lets center this text as we do for our expired license and switchback banner

kimlisa (commented):
couple questions:

- do we want to show this banner every time the user refreshes the browser?
- do we want to show this banner in the discovery screen?

we can re-use our banner by refactoring a little bit, it goes something like this:

OSS:
```ts
export default Container() {
const {alerts} = useBanner()

return <Banner banners={alerts banners}
}
```

Enterprise:
```ts
export default Container() {
const {alerts, license} = useBanner()

const banners: JSX.Element[] = [];

  if (alerts) {
    // do stuff
  }

  if (license) {
    banners.push(
      <LicenseWarning
        key="license-warning-banner"
        severity={license.severity}
        text={license.text}
      />
    );
  }

  if (ctx.storeAccessRequests.getSessionExpiry()) {
    banners.push(
      <SwitchBack key="access-request-banner" data-testid="banner" />
    );
  }

return <Banner banners={banners}
}
```

packages/teleport/src/components/BannerList/Banner.test.tsx
hatched: I did, and it mostly works but requires another dependency and you can see a bit more detailed description in the email.

packages/teleport/src/components/BannerList/BannerList.tsx
hatched: Yeah I wasn't concerned about modifying the nested object as I only needed the pointer of the container to change to trigger a re-render.

packages/teleport/src/components/BannerList/Banner.test.tsx
hatched: Yeah I wasn't concerned about the sequential nature as it's a simple test but using `each` made sense, so updated. I just wish it could update the test label per iteration 😄

packages/teleport/src/components/BannerList/Banner.tsx
hatched: I've converted it from having `role="button"` to being an actual button and added a hover style to it.

Screen Shot 2022-08-31 at 10 05 10 AM

packages/teleport/src/components/BannerList/BannerList.tsx
hatched: done

packages/teleport/src/components/BannerList/Banner.tsx
hatched: I also added a focussed state 

Screen Shot 2022-08-31 at 10 30 55 AM

packages/teleport/src/Main/Main.tsx
hatched: I'm not able to reproduce this, can you give me some repro instructions?
> awkward/confusing double horizontal scrollers (outer scroll and an inner scroll)

packages/teleport/src/Main/Main.tsx
hatched: Good catch, not sure why I did it like this 😆

packages/teleport/src/components/BannerList/Banner.tsx
hatched: Left aligned text (in English) is superior for legibility and imho is more pleasing to the eye in this situation.

packages/teleport/src/components/BannerList/Banner.tsx
hatched: I went back and forth on this myself but you've convinced me... updated 😁

packages/teleport/src/components/BannerList/useAlerts.tsx
hatched: done

packages/teleport/src/components/BannerList/useAlerts.tsx
hatched: Good catch, updated.

packages/teleport/src/config.ts
hatched: Technically the backend supports it so I felt it best to include it here and in the apiserver endpoint.

hatched:
Thanks for the reviews, it's ready for another look

@kimlisa 
> do we want to show this banner every time the user refreshes the browser?

Yes these warnings are only dismissible for the current users session. In the future we expect this to change and have different criteria depending on the message (see below).

> do we want to show this banner in the discovery screen?

I didn't think so as I felt the discover UI is a modal over the Access Manager UI.

> i think we can re-use our banner in our enterprise by refactoring a little bit...

The enterprise-only handling of the license warning will not be special in the near future as it will instead come over the same cluster alerts endpoint. This warning will also likely not be dismissible. The RFD for this is not yet finalized so we'll modify this functionality when the time comes.

packages/teleport/src/Main/Main.tsx
kimlisa: visit the audit log page? that should have sufficient data in the table to cause the inner scrolling.

packages/teleport/src/components/BannerList/Banner.tsx
kimlisa: i still think it should be centered and fixed to the top, otherwise i'll clash, or we make the other banners left aligned (the switchback banner might look weird)
image

hatched: So the license one will be moving to this new banner system but the switchback one will remain the same as the messaging is long vs short form

packages/teleport/src/components/BannerList/Banner.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+    assertComputedStyle(
+      window,
+      document,
+      Banner,
+      bannerProps,
+      'backgroundColor',
+      'rgb(255, 145, 0)'
+    );
+
+    unmount();
```
JanKaczmarkiewicz, line 95: testing-library should automatically unmount components. I think we don't need manual `unmount();`

packages/teleport/src/components/BannerList/Banner.test.tsx
JanKaczmarkiewicz: > Styled Components has a [tohavestylerule](https://styled-components.com/docs/tooling#tohavestylerule) assertion for Jest for this purpose but I found it didn't work if you had any wrappers, like a `<Router>`.

Can you provide an example? I wasn't able to reproduce this issue

packages/teleport/src/components/BannerList/BannerList.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const [bannerList, setBannerList] = useState<{ [id: string]: BannerType }>(
+    {}
+  );
+
+  useEffect(() => {
+    const newList = {};
+    banners.forEach(banner => (newList[banner.id] = banner));
+    Object.assign(newList, bannerList);
+    setBannerList(newList);
+  }, [banners]);
```
JanKaczmarkiewicz, line 35: What will happen if we fetch alerts for one cluster and then the user will switch to another cluster? Will it display the merged list of alerts from two clusters? Is it intended?

packages/teleport/src/Main/Main.tsx
hatched: Oh I see what you're referring to now. Fixed!

packages/teleport/src/components/BannerList/Banner.test.tsx
hatched: I switched to using the `toHaveStyleRule` as the aforementioned issue wasn't actually in this pr, just something from my testing with it.

packages/teleport/src/components/BannerList/BannerList.tsx
hatched: That is what would happen, yeah. 

We don't have a spec as to what's supposed to happen in all of these various notification edge cases. I'd like to see this notification system be replaced by a more fully featured notification system which could filter and categorize the notifications by location/type. For now though I just wanted it to merge any new notifications with ones that were already there.

packages/teleport/src/components/BannerList/BannerList.tsx
hatched: Right good catch, I applied your suggestion above but also properly cloned the `banners` prop as was my original intention.

packages/teleport/src/components/BannerList/Banner.test.tsx
hatched: Yeah woops this was left in from the previous test refactor. It's been removed in this latest test refactor :)

JanKaczmarkiewicz (approved):

packages/teleport/src/Main/Main.tsx
kimlisa: I am not seeing this fixed, could you test it together with either a switchback banner or an expired license banner.

hatched:
Totally agree it's a bit odd, but it's also not finished :D I need these changes to land so that I can make the necessary updates into e

packages/teleport/src/Main/Main.tsx
hatched: The switchback & license banners are in e, I need this to land so that I can make the necessary updates there.

hatched:
e now has a pr as well and will need to land before this one.

packages/teleport/src/components/BannerList/useAlerts.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+import { fetchClusterAlerts } from 'teleport/services/alerts';
+import useStickyClusterId from 'teleport/useStickyClusterId';
+
+import type { ClusterAlert } from 'teleport/services/alerts';
+
+const logger = Logger.create('ClusterAlerts');
+
+export function useAlerts() {
+  const [alerts, setAlerts] = useState<ClusterAlert[]>();
```
kimlisa, line 28: the error is here, should set it as an empty array initially to handle network errors

packages/teleport/src/components/BannerList/Banner.tsx
```diff
@@ -0,0 +1,82 @@
...
+}
+
+const Close = styled.button`
+  background: none;
+  border: 1px solid transparent;
+  box-sizing: border-box;
+  cursor: pointer;
+  display: flex;
+  margin-left: auto;
+  padding: 0.5rem;
```
kimlisa, line 74: this is not necessary if we utilize flexing features:

```
<Flex alignItems="center" justifyContent="space-between">
  <Flex> icon and text </Flex>
  <ButtonClose>icon</ButtonClose>
</Flex>
```

packages/teleport/src/components/BannerList/Banner.tsx
```diff
@@ -0,0 +1,82 @@
...
+          }}
+        >
+          <Cross />
+        </Close>
+      </Flex>
+    </Box>
+  );
+}
+
+const Close = styled.button`
```
kimlisa, line 67:
```suggestion
const ButtonClose = styled.button`
```

kimlisa (commented):
i keep finding visual discrepancy, i would suggest to do the following test per change:

Open up a cluster side by side with your dev cluster, each with same setting and I would compare the changes you made and see if it deviates from the original with the following:

- view the page with only the expired license banner, does any behavior here deviate from the original? from my testing, your changes allows the license warning to be closable, but from previous discussions we did not want this alert to be closable. without any explanation or confirmation that this was okay, i'm going to assume that this is a bug
- view the page with only the access request banner
- view the page with only the other alerts
- view the page with all 3 alerts (license, switchback, and one other alert)

test shrink the window size both horizontally and vertically in all above scenarios to catch any visual bugs. from my testing, shrinking horizontally, alerts with long text message will wrap which with your changes, increases the height of the banner, resulting in the very same issue we were trying to fix: double vertical scrolling.

now that we can get various alerts, we might not be able to assume a certain text length anymore. granted the way we did for license wasn't very fool proof either, like the license warning could get longer for whatever reason and it could overflow. neither case is good and it should be fixed.

i think the fastest and easiest way to handle this is to allow the banner to be scrollable, and i think this is okay for now b/c the alert system is still in development (and we aren't expecting long alert messages anytime soon).

the best way would probably be to redesign this whole alert system (or our whole layout) because it has one huge flaw: because our banners are fixed to the top, stacked alerts will reduce the users vertical viewing space. maybe it's not an issue since users can close most of them? but it could be better designed @xinding33 @rishibarbhaya-design

packages/teleport/src/components/BannerList/useAlerts.tsx
hatched: Fixed, good catch

packages/teleport/src/components/BannerList/BannerList.tsx
```diff
@@ -0,0 +1,92 @@
...
+      ))}
+      {customBanners}
+      {children}
+    </Wrapper>
+  );
+};
+
+const Wrapper = styled(Box)`
+  ${MainContainer} {
+    height: calc(100% - ${props => props.bannerCount * 62}px);
```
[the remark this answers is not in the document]
hatched: You're right, they are quite large....I've reduced the heights of all banners now.

packages/teleport/src/Main/Main.tsx
```diff
@@ -90,6 +93,24 @@ export function Main(props: State) {
...
+  // pre-defined values: LOW: 0; MEDIUM: 5; HIGH: 10
+  const mapSeverity = (severity: number) => {
+    if (severity < 5) {
+      return 'info';
+    }
+    if (severity < 10) {
+      return 'warning';
+    }
+    return 'danger';
+  };
```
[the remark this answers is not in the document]
hatched: While I agree with you in principal....in practice the JIT will likely optimize this function out of the stack so it won't be redefined. 

I was thinking of putting it into alerts but the string conversion of `severity` is a banner level value so I didn't want to modify the data structure of an alert.

packages/teleport/src/Main/Main.tsx
```diff
@@ -90,6 +93,24 @@ export function Main(props: State) {
...
+
+  const banners: BannerType[] = alerts.map(alert => ({
+    message: alert.spec.message,
+    severity: mapSeverity(alert.spec.severity),
+    id: alert.metadata.name,
+  }));
```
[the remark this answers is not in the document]
hatched: For now I want to keep banners and alerts separate. I could move this into a `useBanners` but moving 2 tiny functions into that seems a bit overkill for now.

packages/teleport/src/components/BannerList/Banner.tsx
```diff
@@ -0,0 +1,82 @@
...
+    info: <Info mr={3} fontSize="3" role="icon" />,
+    warning: <Info mr={3} fontSize="3" role="icon" />,
+    danger: <Warning mr={3} fontSize="3" role="icon" />,
+  }[severity];
+
+  const backgroundColor = {
+    info: 'info',
+    warning: 'warning',
+    danger: 'danger',
+  }[severity];
```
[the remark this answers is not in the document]
hatched: Yeah I'll use the `severity` value directly here. This was a holdover from using different labels. Good catch.

packages/teleport/src/components/BannerList/Banner.tsx
hatched: 🤷‍♂️

packages/teleport/src/components/BannerList/Banner.tsx
hatched: I wanted more control over the padding than was afforded by using the flex rules.

packages/teleport/src/components/BannerList/BannerList.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+import type { ReactNode } from 'react';
+import type { Severity } from './Banner';
+
+export const BannerList = ({
+  banners = [],
+  children,
+  customBanners = [],
+}: Props) => {
+  const [bannerList, setBannerList] = useState<{ [id: string]: BannerType }>(
```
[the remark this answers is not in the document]
hatched: updated to `bannerData` and `setBannerData`

packages/teleport/src/components/BannerList/useAlerts.tsx
```diff
@@ -0,0 +1,97 @@
...
+  return date + hours * 60 * 60 * 1000;
+}
+
+function getItem(key: string): string | null {
+  return window.localStorage.getItem(key);
+}
+
+function setItem(key: string, data: string) {
+  window.localStorage.setItem(key, data);
+}
```
kimlisa, line 39: we have a local storage file, uses a particular pattern such as `getXXX` returns already the parsed data: https://github.com/gravitational/webapps/blob/master/packages/teleport/src/services/localStorage/localStorage.ts

packages/teleport/src/components/BannerList/useAlerts.tsx
```diff
@@ -0,0 +1,97 @@
...
+import useStickyClusterId from 'teleport/useStickyClusterId';
+
+import type { ClusterAlert } from 'teleport/services/alerts';
+
+const logger = Logger.create('ClusterAlerts');
+
+const DISABLED_BANNERS = 'disabledAlerts';
+
+function addHours(date: number, hours: number) {
+  return date + hours * 60 * 60 * 1000;
```
kimlisa, line 30: can we make the 60 * 60 * 1000 a const? b/c i'm not sure what it means

packages/teleport/src/components/BannerList/useAlerts.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+        const newAlerts = res.filter(alert => {
+          return !dismissedNames.includes(alert.metadata.name);
+        });
+        setAlerts(newAlerts);
+      })
+      .catch(err => {
+        logger.error(err);
+      });
+  }, [clusterId, dismissedAlerts]);
```
kimlisa, line 81: i think the `dissmissedAlerts` dependency will trigger another alert fetch when we `setDismissedAlerts` from the useEffect above this one

packages/teleport/src/components/BannerList/useAlerts.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+export function useAlerts() {
+  const [alerts, setAlerts] = useState<ClusterAlert[]>([]);
+  const [dismissedAlerts, setDismissedAlerts] = useState<DismissedAlert[]>([]);
+  const { clusterId } = useStickyClusterId();
+
+  useEffect(() => {
+    const disabledAlerts = getItem(DISABLED_BANNERS);
+    if (disabledAlerts) {
+      // Loop through the existing ones and remove those that have passed 24h.
```
kimlisa, line 53: could you expand on this? b/c i'm not sure i understand what it means by 24 hours also why not just remove the alert from the list once user has dismissed?

kimlisa (commented):
in addition, lets also fix the below:

- on switching to discover view, or refershing the page, the dismissed alert flashes before disappearing
- license banner is dismissable, but is not persisted, if we don't intend to make it dismissable then perhaps make it undismissable? (don't render the x button)
- if you shrink the discover page with a banner, there is a wide gap in width on the far right. this is because the banner is smaller than the discover width. this is temporary fixed in v10 by passing in the width, ryan is working on a PR to unify the minWidth, but until then since this PR has to be backported, we should just patch it for now.

packages/teleport/src/components/BannerList/useAlerts.tsx
JanKaczmarkiewicz: Yeah, I think fetching should be separated from filtering here.

```tsx
  useEffect(() => {
    fetchClusterAlerts(clusterId)
      .then(res => {
        if (!res) {
          return;
        }
        setAlerts(newAlerts);
      })
      .catch(err => {
        logger.error(err);
      });
  }, [clusterId]);
  
  const dismissedNames = Object.keys(dismissedAlerts);

  const visibleAlerts = alerts.filter(alert => !dismissedNames.includes(alert.metadata.name));
  
  return {
    alerts: visibleAlerts,
    dismissAlert,
  };
```

packages/teleport/src/components/BannerList/useAlerts.tsx
```diff
@@ -0,0 +1,97 @@
...
+      const data = JSON.parse(disabledAlerts);
+      Object.entries(data).forEach(([name, expiry]) => {
+        if (new Date().getTime() > expiry) {
```
JanKaczmarkiewicz, line 56: Comparing string(eg `'2022-09-09T07:53:09.157Z'`) and number(eg `1662709995148`) won't work - this will always produce `false`. The reason why `ts` wasn't complaining about this line is that expiry has implicit any type (we plan to address that in https://github.com/gravitational/webapps/issues/957)
```suggestion
        if (new Date().getTime() > Date.parse(expiry)) {
```

JanKaczmarkiewicz: I think we should add test case to cover that

packages/teleport/src/components/BannerList/useAlerts.tsx
hatched: Sure.

There are 3,600,000ms in an hour and this is just a common way of writing that.

packages/teleport/src/components/BannerList/useAlerts.tsx
hatched: This file doesn't appear to have these utility methods? Or did you want me to add them?

These functions were simply created to save typing :) they don't do anything special.

packages/teleport/src/components/BannerList/useAlerts.tsx
hatched: Yep good catch! Fixed

packages/teleport/src/components/BannerList/useAlerts.tsx
hatched: So for this release we decided that the banners would be able to be snoozed for 24h.

The data sets for the `alerts` and the `dismissedAlerts` are separate so I wanted to keep them separate and then work on the result being sent back to a consumer instead of modifying the server response.

Doing it this way also means that the requests can come back in any order.

packages/teleport/src/components/BannerList/useAlerts.tsx
hatched: good catch! I actually had this originally as ` if (new Date().getTime() > +expiry) {` but the + was dropped somewhere along the line.

I was actually in process of writing tests for this just didn't get it finished by EOD yesterday so I'll be sure that there is a test case for this.

hatched:
@kimlisa 
> on switching to discover view, or refershing the page, the dismissed alert flashes before disappearing

As the data and deny list come in from different requests I worried that this might happen if they come in out of order. I'll add some gating around this.

> license banner is dismissable, but is not persisted, if we don't intend to make it dismissable then perhaps make it undismissable? (don't render the x button)

It's not persisted because it's injected after the fact so it's not being controlled by the alert system. It will be part of the alert system in the next release though so I'm hesitant to add too much special handling here that we'll have to be sure we go back and remove before the next release.

I think I'll modify the data structure to include the `dismissible` flag so that it'll line up with the structure being sent from the server in the upcoming release and then make it not dismissible. 

> if you shrink the discover page with a banner, there is a wide gap in width on the far right. this is because the banner is smaller than the discover width. this is temporary fixed in v10 by passing in the width, ryan is working on a PR to unify the minWidth, but until then since this PR has to be backported, we should just patch it for now.

Can you provide me with the PR that contains this patch?

packages/teleport/src/components/BannerList/useAlerts.tsx
JanKaczmarkiewicz:
```js
+'2022-09-09T07:53:09.157Z'
``` 
won't work since it will produce `NaN`

hatched: I'm not sure I understand? `expiry` is epoch time not a datestring.

JanKaczmarkiewicz: You are right. I thought we were sanitizing js dates (for unknown reason :) ). I think it should work as is now. Sorry for unnesessery discussion

hatched: no problem at all :)

PR #1173, opened 2022-08-31
[Discover] User menu checkmark, alert bubble, tweaks

kimlisa:

#### Description
- Adds checkmark to where the user is currently on
- Adds alert bubble to the discover nav item if user:
  - doesn't have a resource
  - has not visited the discover page
- Added allowing user to re-fetch failure points (eg network errors) so they don't need to restart the flow, just refetch the data that was required for that step
  
#### Screenshot
on resources (anything but discover or user menu nav items)

image

no resources alert bubble that goes away when a user visits the discover page (like github)

image

on discover

image

on a user menu nav item

image

Refetching from errors (eg network)

image
image

### DIFF
--- a/packages/teleport/src/components/UserMenuNav/UserMenuNav.tsx
+++ b/packages/teleport/src/components/UserMenuNav/UserMenuNav.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import styled from 'styled-components';
 import { useLocation } from 'react-router';
 import { NavLink } from 'react-router-dom';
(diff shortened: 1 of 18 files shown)


ryanclark (approved):

packages/teleport/src/components/UserMenuNav/UserMenuNav.tsx
```diff
@@ -14,32 +14,70 @@
...
+
+  const discover = localStorage.getOnboardDiscover();
+
+  // viewingXXX flags to determine what view the user is
+  // currently on to determine where the checkmark icon
+  // on the dropdown items should be rendered.
+  const viewingDiscover = pathname === cfg.routes.discover;
+  const viewingUserMenu =
+    !viewingDiscover &&
+    navItems.filter(n => pathname.startsWith(n.getLink())).length > 0;
```
JanKaczmarkiewicz, line 44:
```suggestion
    navItems.some(n => pathname.startsWith(n.getLink()));
```

packages/teleport/src/components/UserMenuNav/UserMenuNav.tsx
```diff
@@ -14,32 +14,70 @@
...
+  const viewingResources = !viewingUserMenu && !viewingDiscover;
+
+  const firstTimeDiscoverVisit =
+    discover && !discover.hasResource && !discover.hasVisited;
+  const showDiscoverAlertBubble = !viewingDiscover && firstTimeDiscoverVisit;
+
+  if (viewingDiscover && firstTimeDiscoverVisit) {
+    localStorage.setOnboardDiscover({
```
JanKaczmarkiewicz, line 52: modifying the external state inside the component render function can cause problems (eg tearing)

There are two main examples of when this can cause problems:

The component renders for the first time and then it is rerendered by some unrelated state change. The user will see `AlertBubble` and then it will disappear.
_____________________________________
The second problem can arise with react@18:
1. We schedule state updates using `startTransition` (informs react that rendering can be interrupted) that will cause `UserMenuNav` to render for the first time.
2. React at some point will call this component to get its jsx representation 
   * `localStorage` is empty so `firstTimeDiscoverVisit` will be `false`
   * `localStorage.setOnboardDiscover(..., hasVisited: true)` is called
3. After the call is complete react will yield and check if there is high priority update (like a user event). Let's say it is. React will start working on the part that scheduled high priority update.
4. After that update, react will start or resume working on the original state update. This can cause `UserMenuNav` to render again. In this case, `AlertBubble` will not show.

React expects components to be `pure`. I think we should put `localStorage.setOnboardDiscover` inside useEffect to ensure that the user actually sees the bubble (note this will only solve the second problem):
```ts
const isFirstTimeDiscoverVisited = viewingDiscover && firstTimeDiscoverVisit;

useEffect(() => {
    if (isFirstTimeDiscoverVisited) {
        const discover = localStorage.getOnboardDiscover();
        localStorage.setOnboardDiscover({
            ...discover,
            hasVisited: true,
        })
    }
}, [isFirstTimeDiscoverVisited])
```

I think we can try using https://reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects to help find potential places which can cause problems when migrating to react 18

kimlisa: i changed to how you proposed, for the strict mode, i left that alone for now

JanKaczmarkiewicz (approved):

PR #1178, opened 2022-09-03
Fix double scroll bars, make everything have min width 1250px

ryanclark:
This makes it so `MainContainer` no longer scrolls and `HorizontalSplit` is instead the element that scrolls when the screen width is too small.

To do this, I've made it so the contents inside `HorizontalSplit` are wrapped with a min width, to force `HorizontalSplit` to scroll when its children's size (which is 100% minus the sidebar width) exceed its width. This then makes the `MainContainer` never have to scroll, so double scroll bars won't happen.

This shows the correct scrollbar appearing when the browser width is ~1000px
image

Everything's fine at wide resolutions
image

This also changes discover to have a max content width of 1450px (so 1450px + 281px for the sidebar = it'll stop expanding past a browser being 1731px wide).
image

At 1300px wide
image

At 1000px wide
image

### DIFF
--- a/packages/teleport/src/Main/Main.tsx
+++ b/packages/teleport/src/Main/Main.tsx
@@ -156,7 +156,6 @@ export const ContentMinWidth = styled.div`
 export const HorizontalSplit = styled.div`
   display: flex;
   flex-direction: column;
-  height: 100%;
   flex: 1;
   overflow-x: auto;
 `;
--- a/packages/teleport/src/Main/Main.tsx
+++ b/packages/teleport/src/Main/Main.tsx
@@ -150,8 +150,7 @@ export function Main(props: State) {
 }
 
 export const ContentMinWidth = styled.div`
-  // minus 1px for the sidebar right border
-  min-width: calc(1250px - var(--sidebar-width) - 1px);
+  min-width: calc(1250px - var(--sidebar-width));
 `;
 
 export const HorizontalSplit = styled.div`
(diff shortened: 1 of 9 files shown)


packages/teleport/src/Main/Main.tsx
```diff
@@ -119,17 +121,22 @@ export const MainContainer = styled.div`
   display: flex;
   flex: 1;
   position: absolute;
-  min-width: 1000px;
+`;
+
+export const ContentMinWidth = styled.div`
+  // sidebar is 240px + 1px right border
```
JanKaczmarkiewicz, line 127: this is a little confusing
```
// sidebar is 240px + 1px right border
```
I can see that in another place it is
```
// sidebar is 280px + 1px right border
```
Is this intended?

ryanclark: Yeah, the sidebar for discover and the main sidebar are different widths

ryanclark: Thoughts on moving the value into a CSS variable instead? https://caniuse.com/css-variables

kimlisa:
i'm curious where did you see the double scrollbar? also did you test this with a banner? (i would test with an expired license, as it has the longest text, i have one if you need one). the banner as it is right now, will probably cause a double horizontal scrollbar. 

i feel like double scrollbars is inevitable at this point 😮‍💨

ryanclark:
This also works correctly with the new banners addition

image
image
image
image
image

hatched (approved):

PR #1181, opened 2022-09-06
[v10] Discover Bug fix appending index number to login trait names (#1180)

kimlisa:
backport of #1180 to teleport-v10


rudream (approved):
bot

PR #1186, opened 2022-09-08
Type the design system

ryanclark:
This adds initial typings for the design system.

I didn't cover everything but this is the bare minimum PR required to have the theme, `styled-components`, `styled-system` and a decent amount of our commonly used elements (`Box`, `Flex`, etc) all have typings.

The theme is now typed:
image

Our elements are typed:
image

When missing required props we now have errors
image

This also has fixed a few bugs because of this - incorrect prop usages, incorrect theme property lookups, props being accepted into elements and nothing done with their value, etc.

As a result of typings being added, any styled component that accepts custom props has to have typings for it. To do this, I've put an interface above each styled component like so -

```typescript jsx
interface StyledItemProps {
  active: boolean;
}

const StyledItem = styled.div<StyledItemProps>`
  background-color: ${(p) => p.active ? 'red' : 'blue'};
`;
```

This also cleans up the default imports/exports for any component I've moved to TypeScript. This means that it's easy to go to the definition of the component by cmd + clicking in your IDE.

This also changes the usage of `styled-components`, to allow for better type safety when we compose styled components with modifiers from `styled-system`. You can take a read of the docs here - https://github.com/gravitational/webapps/blob/ryan/type-the-design-system/docs/styled-components.md


packages/design/src/Button/Button.tsx
```diff
@@ -164,46 +187,22 @@ const StyledButton = styled.button`
...
 
-export default Button;
-export const ButtonPrimary = props => <Button kind="primary" {...props} />;
-export const ButtonSecondary = props => <Button kind="secondary" {...props} />;
-export const ButtonBorder = props => <Button kind="border" {...props} />;
-export const ButtonWarning = props => <Button kind="warning" {...props} />;
-export const ButtonText = props => <Button kind="text" {...props} />;
+export const ButtonText = <T extends ElementType>(
+  props: React.PropsWithChildren<Omit<ButtonProps<T>, 'kind'>>
+) => <Button kind="text" {...props} />;
```
JanKaczmarkiewicz, line 208: [nit]
type logic can be extracted and reused in all variants. 
```tsx
type ButtonVariant = <T extends ElementType>(
  props: Omit<ButtonProps<T>, 'kind'>
) => ReactElement;

export const ButtonPrimary: ButtonVariant = props => (
  <Button kind="primary" {...props} />
);

export const ButtonSecondary: ButtonVariant = props => (
  <Button kind="secondary" {...props} />
);
```

packages/design/src/LabelState/LabelState.test.tsx
```diff
@@ -47,12 +48,12 @@ describe('design/LabelState', () => {
       background: expected,
     });
 
-    expect(getComputedStyle(container.firstChild).boxShadow).toBe('');
+    expect(getComputedStyle(container.firstElementChild).boxShadow).toBe('');
```
JanKaczmarkiewicz, line 51: [nit]
```ts
expect(container.firstElementChild).not.toHaveStyleRole('boxShadow');
```

packages/design/src/Popover/Transition.tsx
```diff
@@ -17,7 +17,13 @@ limitations under the License.
 import React from 'react';
 import ReactDOM from 'react-dom';
 
-class Transition extends React.Component {
+interface TransitionProps {
+  onEntering: (node: any) => void;
```
JanKaczmarkiewicz, line 21: try installing `@types/react-dom` and the type here will be
```ts
onEntering: (node: Element | Text | null) => void;
```

packages/design/src/ButtonOutlined/ButtonOutlined.tsx
```diff
@@ -14,23 +14,44 @@ See the License for the specific language governing permissions and
...
 
-const ButtonOutlined = (
-  { children, setRef, ...props } = { setRef: undefined }
-) => {
+import type { SpaceProps, WidthProps } from 'design/system';
+
+import type { PropsWithTheme } from 'design/theme';
+
+interface ButtonOutlinedBaseProps<T extends ElementType> {
+  setRef?: React.Ref<HTMLButtonElement>;
```
JanKaczmarkiewicz, line 27: I think more idiomatic way would be to use `React.forwardRef` https://reactjs.org/docs/forwarding-refs.html. But thats not connected to this PR

packages/design/src/Popover/Transition.tsx
```diff
@@ -28,8 +34,6 @@ class Transition extends React.Component {
     delete childProps.onEntering;
 
     const child = React.Children.only(children);
-    return React.cloneElement(child, childProps);
+    return React.cloneElement(child as React.ReactElement<any>, childProps);
```
JanKaczmarkiewicz, line 37: Instead of casting this component should have children prop with type ReactElement
```tsx
interface TransitionProps {
  onEntering: (node: Element | Text | null) => void;
  children: React.ReactElement;
}

export class Transition extends React.Component<TransitionProps> {
  componentDidMount() {
...
```

packages/shared/components/Select/Select.tsx
```diff
@@ -59,7 +63,13 @@ export function SelectAsync(props: AsyncProps) {
...
 }
 
-export const StyledSelect = styled.div`
+interface StyledSelectBaseProps {
+  hasError?: boolean;
+}
+
+export type StyledSelectProps = StyledSelectBaseProps & SpaceProps & WidthProps;
+
+export const StyledSelect = styled.div<PropsWithTheme<StyledSelectProps>>`
```
JanKaczmarkiewicz, line 72:
```suggestion
export const StyledSelect = styled.div<StyledSelectProps>`
```
Theme is injected by `styled.div`.

packages/teleport/src/components/Tabs/TabIcon.tsx
```diff
@@ -38,10 +39,14 @@ type Props = {
   active: boolean;
   onClick(): void;
   title: string;
-  Icon(): JSX.Element;
+  Icon: (props: any) => JSX.Element;
```
JanKaczmarkiewicz, line 42:
```suggestion
  Icon: React.ComponentType;
```

packages/teleterm/src/ui/components/VirtualScroll/VirtualScroll.tsx
```diff
@@ -42,15 +42,21 @@ export function VirtualScroll<T>(props: VirtualScrollProps<T>) {
...
-  style: {
-    transform: `translateY(${props.moveBy + 'px'})`,
-  },
-}))``;
+interface OffsetProps {
+  moveBy: number;
+}
+
+const Offset = styled.div<React.PropsWithChildren<OffsetProps>>`
+  transform: translateY(${p => p.moveBy + 'px'});
```
JanKaczmarkiewicz, line 58:
```suggestion
  transform: translateY(${p => p.moveBy}px);
```

packages/teleterm/src/ui/Tabs/TabItem.tsx
```diff
@@ -87,40 +87,48 @@ export function TabItem(props: TabItemProps) {
...
+}
 
-  if (active) {
-    styles['backgroundColor'] = theme.colors.primary.darker;
-    styles['color'] = theme.colors.secondary.contrastText;
-    styles['transition'] = 'none';
-  }
+const StyledTabItem = styled.div<StyledTabItemProps>(
+  ({ theme, active, dragging, canDrag }) => {
+    const styles: any = {
```
JanKaczmarkiewicz, line 98:
```suggestion
    const styles: CSSObject = {
```

JanKaczmarkiewicz:
This is huge improvement to DX and type-safety. Thanks!

gzdunek (commented):
I agree, it's a huge improvement, thanks!

alex-kovoy:
There is a pending PR that aims to address this issue https://github.com/gravitational/webapps/pull/918

packages/design/src/LabelState/LabelState.test.tsx
ryanclark: tried this, but you have to provide a full CSS value (`box-shadow: whatever`) 

The value of this is computed, so I cant easily use `toHaveStyle` (I can't find any info about `toHaveStyleRole` anywhere) - I'm just going to leave it as it is

kimlisa:
@ryanclark can you see if you can address @gzdunek comment left here: https://github.com/gravitational/webapps/pull/918#pullrequestreview-1024532611, that was the first thing i saw running app where the table border wasn't respected

kimlisa:
this is going to require side by side comparison (with another cluster and with storybook), we lose styling and wonky route behavior:

discover page (top is `before` and bottom is `after`), we lose styling (apparent in other places too):

image

this is from `view resources` going to `discover`, it puts the other view at the bottom:

image

ryanclark:
@kimlisa the dual routes displaying is a different issue - https://github.com/gravitational/webapps.e/pull/442

packages/design/src/styled/styled.ts
```diff
@@ -0,0 +1,201 @@
...
+  $as?: KnownTarget;
+  $forwardedAs?: KnownTarget;
+  as?: KnownTarget;
+  forwardedAs?: KnownTarget;
+  theme?: DefaultTheme;
+}
+
+export interface ExecutionContext extends ExtensibleObject {
+  theme: DefaultTheme;
+}
```
JanKaczmarkiewicz, line 113:
```suggestion
```
I think following types can be removed

packages/design/src/styled/styled.ts
```diff
@@ -0,0 +1,201 @@
...
+export const styled = customStyled as StyledInterface;
+
+export type WebTarget = string | keyof JSX.IntrinsicElements;
+
+const baseStyled = <Target extends WebTarget>(tag: Target) =>
+  styled.bind(null, tag);
+
+for (const element of domElements) {
+  styled[element] = baseStyled(element);
+}
```
JanKaczmarkiewicz, line 201:
```suggestion
export const styled = customStyled as StyledInterface;

export type WebTarget = string | keyof JSX.IntrinsicElements;

const baseStyled = <Target extends WebTarget>(tag: Target) =>
  styled.bind(null, tag);

for (const element of domElements) {
  styled[element] = baseStyled(element);
}
```

packages/design/src/theme/theme.tsx
```diff
@@ -126,8 +198,6 @@ const theme = {
   breakpoints: [],
 };
```
JanKaczmarkiewicz, line 199: Alternatively we can infer theme type from value:
```tsx
const space = [0, 4, 8, 16, 24, 32, 40, 48, 56, 64, 72, 80] as const;
const contrastThreshold = 3;

export type PropsWithTheme<T> = { theme: Theme } & T;

const colors = {
  accent: '#651FFF',

  dark: '#000',

  light: '#FFFFFF',

  primary: {
    main: '#1C254D',
    light: '#222C59',
    lighter: '#2C3A73',
    dark: '#111B48',
    contrastText: '#FFFFFF',
  },

  secondary: {
    main: '#512FC9',
    light: '#651FFF',
    dark: '#354AA4',
    contrastText: '#FFFFFF',
  },

  text: {
    // The most important text.
    primary: 'rgba(255,255,255,0.87)',
    // Secondary text.
    secondary: 'rgba(255, 255, 255, 0.56)',
    // Placeholder text for forms.
    placeholder: 'rgba(255, 255, 255, 0.24)',
    // Disabled text have even lower visual prominence.
    disabled: 'rgba(0, 0, 0, 0.24)',
    // Text hints.
    hint: 'rgba(0, 0, 0, 0.24)',
    // On light backgrounds
    onLight: '#324148',
    // On dark backgrounds
    onDark: 'rgba(255, 255, 255, 0.87)',
  },

  grey: blueGrey,

  error: {
    light: red['A200'],
    dark: red['A700'],
    main: red['A400'],
  },

  action: {
    active: '#FFFFFF',
    hover: 'rgba(255, 255, 255, 0.1)',
    hoverOpacity: 0.1,
    selected: 'rgba(255, 255, 255, 0.2)',
    disabled: 'rgba(255, 255, 255, 0.3)',
    disabledBackground: 'rgba(255, 255, 255, 0.12)',
  },

  subtle: blueGrey[50],
  link: lightBlue[500],
  bgTerminal: '#010B1C',
  danger: pink.A400,
  highlight: yellow[50],
  disabled: blueGrey[500],
  info: lightBlue[600],
  warning: orange.A400,
  success: teal.A700,
} as const;

const borders = [
  0,
  '1px solid',
  '2px solid',
  '4px solid',
  '8px solid',
  '16px solid',
  '32px solid',
] as const;

export const theme = {
  colors,
  typography,
  font: fonts.sansSerif,
  fonts: fonts,
  fontWeights,
  fontSizes,
  space,
  borders,
  radii: [0, 2, 4, 8, 16, 9999, '100%'],
  regular: fontWeights.regular,
  bold: fontWeights.bold,
  // disabled media queries for styled-system
  breakpoints: [],
} as const;

export type Theme = typeof theme;
```

ryanclark: I'd like to, but they differ slightly between enterprise and OSS.

To be honest, I hate the entire theming code and think we should refactor this in a PR afterwards - thoughts?

alex-kovoy: @ryanclark this PR looks too complex for a relatively simple problem. For example, you can look at the `next` (or website) repo how it's done there. Basically the latest styled-system provides a good typescript support. You can start refactoring the code to match the latest styled-system (theme and lookup methods), unifying OSS/E themes and the rest will come alone. If that for some reason is still not enough then building a custom theme would be an option but merging this PR is not great as this is a lot of [unmaintainable code](https://github.com/gravitational/webapps/blob/ryan/type-the-design-system/packages/design/src/styled/styled.ts) that creates more issues than solves.

ryanclark: This is also introducing automatic typing for composed styled elements from `styled-system`, without having to keep the typed props and the usages inline. This is taking advantage of the latest `styled-systems` typings.

I don't see how the code you linked is unmaintainable, as it is from styled-components typings themselves and adapted to work for us. If they update their types, we can do the same.

gzdunek:
I tested Connect, the layout is broken (top and bottom parts disappeared).
image

vs
image

packages/design/src/theme/theme.tsx
alex-kovoy: Yeah, my concern that it relies on internal styled-component implementation and that maintaining it is now Teleport responsibility. Which is both risky and time consuming as it's hard to maintain something you have no control over. Even a question "Is it styled component bug/limitation or ours" is no longer a simple answer.  But I appreciate the intent, i just hopped that it could've been done with less hacky approach.

packages/design/src/LabelInput/LabelInput.tsx
```diff
@@ -14,32 +14,27 @@ See the License for the specific language governing permissions and
...
-  ${space}
 `;
 
-LabelInput.propTypes = {
-  hasError: PropTypes.bool,
-};
-
 LabelInput.defaultProps = {
-  hasError: false,
   fontSize: 0,
```
gzdunek, line 38: I noticed the `font-size` for labels has changed from 11px to 10px, removing this line should fix it.

packages/design/src/SideNav/SideNavItem.tsx
```diff
@@ -14,13 +14,11 @@ See the License for the specific language governing permissions and
...
-import styled from 'styled-components';
+import styled, { borderColor } from 'design/styled';
 
-import { borderColor } from './../system';
-import defaultTheme from './../theme';
-import Flex from './../Flex';
+import { Flex } from 'design/Flex';
 
-const fromTheme = ({ theme = defaultTheme }) => {
+const fromTheme = ({ theme }) => {
```
gzdunek, line 21: Type?

packages/design/src/SideNav/SideNavItem.tsx
```diff
@@ -48,17 +46,13 @@ const SideNavItem = styled(Flex)`
...
-
 SideNavItem.defaultProps = {
   pl: 9,
   pr: 5,
   bg: 'primary.main',
   color: 'text.primary',
-  theme: defaultTheme,
 };
 
 export default SideNavItem;
```
gzdunek, line 58: Default export can be removed.

packages/design/src/utils/copyToClipboard.ts
```diff
@@ -20,8 +20,36 @@
...
   });
 }
+
+/**
+ * fallbackCopyToClipboard is used when navigator.clipboard is not supported.
+ * Note: document.execCommand is marked obselete.
+ *
+ * @param textToCopy the text to copy to clipboard
+ */
+function fallbackCopyToClipboard(textToCopy: string): boolean {
```
gzdunek, line 40: Could we move this change to the other PR? It is not related to the types.

packages/shared/components/FileTransfer/FileTransferStateless/FileList/FileListItem.tsx
```diff
@@ -102,7 +102,7 @@ const ProgressPercentage = styled(Text)`
   width: 36px;
 `;
 
-const Li = styled.li`
+const Li = styled('li')`
```
gzdunek, line 105: `styled.li`?

packages/teleport/src/Discover/Desktop/DiscoverDesktops/NodeLine.tsx
```diff
@@ -41,7 +41,7 @@ const line = keyframes`
   }
 `;
 
-export const StyledSVG = styled.svg`
+export const StyledSVG = styled('svg')`
```
gzdunek, line 44: `styled.svg`?

gzdunek:
I found a few more regressions:
how it looks like vs how it should look like

1. Button color
image

image

2. Bottom border 
image

image

3. Font color on the list 
image
image

packages/design/src/theme/theme.tsx
gzdunek: I'm a bit unsure about `styled.ts` too :( 
If I would have to update this file I wouldn't know what exactly should I do. What part of it is just copied from `styled-components` and what is changed to work for us?

Also, it would be worth to leave a comment why we need to do this.

ryanclark: I wrote a doc about what this brings us here - https://github.com/gravitational/webapps/blob/ryan/type-the-design-system/docs/styled-components.md

I can document `styled.ts` to show exactly what is ours vs theirs, to make any changes easier on future developers. I get that it looks like a large scary typescript file, but most of it is just copied from `@types/styled-components` and adapted where we need a different constructor, or type inference.

gzdunek: >I can document `styled.ts` to show exactly what is ours vs theirs, to make any changes easier on future developers.

Yeah, I think it would be helpful.

packages/shared/components/FileTransfer/FileTransferStateless/FileList/FileListItem.tsx
ryanclark: Originally the styled implementation didn't work with `styled.element`, but I had fixed that, so I'll revert this

ryanclark:
@gzdunek 

all fixed
image

packages/design/src/theme/theme.tsx
ryanclark: Done! I've added comments about what we've copied, why we copied specific things as well as what we've added and what was their original typings.

kimlisa:
left is `before`, right is `after`, suggest comparing two storybooks after fixing

including session recordings
image

image

image

image

image


image

if it adds additional space here, will it add additional spaces elsewhere? is why i pointed this one out
image

image

think we removed a theme color, it's also missing in our theme color examples
image

ryanclark:
Thanks @kimlisa, I'll address these tomorrow

ryanclark:
Closing in favour of #1439

PR #1188, opened 2022-09-09
[v10] Add events to audit log

kimlisa:
backport #968 and #1089 to teleport-v10


packages/teleport/src/services/audit/types.ts
```diff
@@ -908,6 +1016,15 @@ type RawEventAuthFailure<T extends EventCode> = RawEvent<
   }
 >;
 
+type RawEventSFTP<T extends EventCode> = RawEvent<
+  T,
+  {
+    path: string;
+    error: string;
+    ['addr.local']: string;
```
capnspacehook, line 1024: Does `['server_hostname']: string;` need to be added as well?

capnspacehook (approved):

packages/teleport/src/services/audit/types.ts
kimlisa: oh you're right, i was wondering how it bypassed the type check but it got bypassed since we used bracket string notation. that's frustrating b/c what is the point of the type-check then :laughing: 

i'll put a fix out, but it's not urgent and doesn't affect production code

PR #1193, opened 2022-09-14
add cluster alert links to UX

michellescripts:
Supports: https://github.com/gravitational/teleport/issues/16425
Teleport PR: https://github.com/gravitational/teleport/pull/16426

Renders "teleport.internal/link" label as link in UX

### DIFF
--- a/packages/teleport/src/components/BannerList/Banner.tsx
+++ b/packages/teleport/src/components/BannerList/Banner.tsx
@@ -14,10 +14,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import React from 'react';
+import React, { useMemo } from 'react';
 import styled from 'styled-components';
 
-import { Box, Flex, Text } from 'design';
+import { Box, Flex, Link, Text } from 'design';
 import { Cross, Info, Warning } from 'design/Icon';
 
 export type Severity = 'info' | 'warning' | 'danger';
@@ -26,6 +26,7 @@ export type Props = {
   message: string;
   severity: Severity;
   id: string;
+  link?: string;
   onClose: (id: string) => void;
 };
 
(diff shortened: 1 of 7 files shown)


packages/teleport/src/components/BannerList/Banner.tsx
```diff
@@ -41,11 +43,34 @@ export function Banner({
     danger: <Warning mr={3} fontSize="3" role="icon" />,
   }[severity];
 
+  const getLink = useMemo(() => {
```
JanKaczmarkiewicz, line 46: [nit]
We can separate logic from jsx here:
```ts
const isValidTeleportLink = (link: string) => {
    try {
      const url = new URL(link);
      return url.hostname === 'goteleport.com';
    } catch (_) {
      return false;
    }
}
...
     // inside component
     {isValidTeleportLink(link) ? (
        <Link
          href={link}
          target="_blank"
          color="light"
          style={{ fontWeight: 'bold' }}
        >
          {message}
        </Link>
      ) : (
        <Text bold>{message}</Text>
      )}
```

JanKaczmarkiewicz (approved):

packages/teleport/src/services/alerts/alerts.tsx
```diff
@@ -17,6 +17,8 @@ limitations under the License.
 import api from 'teleport/services/api';
 import cfg from 'teleport/config';
 
+export const LinkLabel = 'teleport.internal/link';
```
JanKaczmarkiewicz, line 20: [nit]
Pascal case is reserved for react component types, typescript types/enums, and classes. Please use uppercase or camelCase
```suggestion
export const LINK_LABEL = 'teleport.internal/link';
```
I think we should enforce it by https://typescript-eslint.io/rules/naming-convention/

JanKaczmarkiewicz (approved):

mcbattirola (approved):

PR #1194, opened 2022-09-15
Create TextSelectCopyMulti that allows multi lines and adding comments

kimlisa:
#### Description 
Create a `TextSelectCopyMulti` matching figma design (below) and allows multi text selects and adding comments

Note: Will update and delete the original single text copy `TextSelectCopy` after this PR is approved and merged

#### Figma design
https://www.figma.com/file/8DQZl1mYUPLgvKgmA6zPOM/In-Progress-Work?node-id=444%3A49313
Screen Shot 2022-09-15 at 1 55 30 PM

image

#### Screenshots
image

image

image

scrollbar appearances

image

image

### DIFF
--- a/packages/teleport/src/components/TextSelectCopy/TextSelectCopyMulti.test.tsx
+++ b/packages/teleport/src/components/TextSelectCopy/TextSelectCopyMulti.test.tsx
@@ -16,24 +16,14 @@
 
 import React from 'react';
 
-import { render, screen, fireEvent } from 'design/utils/testing';
+import { render, screen, userEvent } from 'design/utils/testing';
 
 import { TextSelectCopyMulti } from './TextSelectCopyMulti';
 
-beforeEach(() => {
-  jest.useFakeTimers();
-  Object.assign(window.navigator, {
-    clipboard: {
-      writeText: jest.fn().mockImplementation(() => Promise.resolve()),
-    },
-  });
-});
-
-afterEach(() => {
-  jest.resetAllMocks();
-});
+jest.useFakeTimers();
 
 test('changing of icon when button is clicked', async () => {
+  const user = userEvent.setup({ delay: null });
   render(<TextSelectCopyMulti lines={[{ text: 'some text to copy' }]} />);
 
   // Init button states.
@@ -41,12 +31,12 @@ test('changing of icon when button is clicked', async () => {
   expect(screen.queryByTestId('btn-check')).not.toBeVisible();
 
   // Clicking copy button should change the button icon to "check".
-  fireEvent.click(screen.getByTestId('btn-copy'));
-  await expect(screen.findByTestId('btn-check')).resolves.toBeVisible();
-  expect(screen.queryByTestId('btn-copy')).not.toBeVisible();
-  expect(window.navigator.clipboard.writeText).toHaveBeenCalledWith(
-    'some text to copy'
-  );
+  await user.click(screen.getByTestId('btn-copy'));
+  expect(screen.getByTestId('btn-check')).toBeVisible();
+  expect(screen.getByTestId('btn-copy')).not.toBeVisible();
+
+  const clipboardText = await navigator.clipboard.readText();
+  expect(clipboardText).toBe('some text to copy');
 
   // After set time out, the buttons should return to its initial state.
   jest.runAllTimers();
@@ -55,6 +45,7 @@ test('changing of icon when button is clicked', async () => {
 });
 
 test('correct copying of texts', async () => {
+  const user = userEvent.setup({ delay: null });
   render(
     <TextSelectCopyMulti
       lines={[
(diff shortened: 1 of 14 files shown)


packages/teleport/src/components/TextSelectCopy/TextSelectCopyMulti.tsx
```diff
@@ -0,0 +1,179 @@
...
+        })}
+      </Lines>
+      <Box>
+        {lineEls.map((el, index) => {
+          const isLastText = index === lines.length - 1;
+          let spacer;
+          if (lines[index].comment) {
+            spacer = (
+              <Box
+                height={(el.firstElementChild as HTMLElement).offsetHeight}
```
JanKaczmarkiewicz, line 99: I think this could be computed without rendering elements. we know:
- a number of lines: `lineText.split("\n").length`
- line hight: `20px` 
- spacing

An alternative solution would be to use pure CSS. something with position absolute/sticky.

packages/teleport/src/components/TextSelectCopy/TextSelectCopyMulti.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+  Object.assign(window.navigator, {
+    clipboard: {
+      writeText: jest.fn().mockImplementation(() => Promise.resolve()),
+    },
+  });
+});
+
+afterEach(() => {
+  jest.resetAllMocks();
+  Object.assign(window.navigator, originalNavigator);
```
JanKaczmarkiewicz, line 36: This won't do anything since `window.navigator` and `originalNavigator` points to the same object reference.
```
const originalNavigator = window.navigator; // this will alias window.navigator not clone it
```
In result `window.navigator` will have mocked method even after test is finished

packages/teleport/src/components/TextSelectCopy/TextSelectCopyMulti.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+  fireEvent.click(screen.getByTestId('btn-copy'));
+  await waitFor(() => expect(screen.getByTestId('btn-check')).toBeVisible());
+  expect(screen.queryByTestId('btn-copy')).not.toBeVisible();
+  expect(window.navigator.clipboard.writeText).toHaveBeenCalledWith(
+    'some text to copy'
+  );
+
+  // After set time out, the buttons should return to its initial state.
+  jest.runAllTimers();
+  expect(screen.queryByTestId('btn-copy')).toBeVisible();
```
JanKaczmarkiewicz, line 56: Have you tried to use 
```tsx
expect(await screen.findByTestId('btn-copy')).toBeVisible();
```
instead of 
```tsx
jest.useFakeTimers();
...
jest.runAllTimers();
expect(screen.queryByTestId('btn-copy')).toBeVisible();
```

kimlisa: i think it's better to use the fake timers here, b/c from my understanding we can execute these timers right away instead waiting for the real time that we programmed it for (granted in this case the real time is only 1 second, so might not make much difference)

packages/teleport/src/components/TextSelectCopy/TextSelectCopyMulti.test.tsx
kimlisa: ah you are correct. though i was also thinking the resetting is unnecessary because if i am not mistaken, each test file is sandboxed and will not affect other test files. so I think it's fine to not handle resetting it.

also the `spyOn` does not work because the clipboard is not defined? there is a library that handles this stubbing for us: https://github.com/testing-library/user-event, though atm it feels like an overkill, let me know what you think

packages/teleport/src/components/TextSelectCopy/TextSelectCopyMulti.tsx
kimlisa: thanks for giving me the confidence for the position absolute, i didn't think it will work but it did 🎉

kimlisa:
@JanKaczmarkiewicz friendly ping

packages/teleport/src/components/TextSelectCopy/TextSelectCopyMulti.test.tsx
JanKaczmarkiewicz: user-event is an official testing library extension. It provides API that better emulates the user behavior. I think we could use it anyway.

The alternative would be to add `setupTests.js` file:
with
```
Object.assign(navigator, {
  clipboard: {
    writeText: () => {},
  },
});
```
+ add `setupFilesAfterEnv: ['<rootDir>/setupTests.js'],` to `jest.config.js`

packages/teleport/src/components/TextSelectCopy/TextSelectCopyMulti.test.tsx
JanKaczmarkiewicz: I don't have a strong opinion here.

JanKaczmarkiewicz (approved):

PR #1195, opened 2022-09-19
Minor kube fixes

gzdunek:
I didn't want to add new things to the previous PR, so I'm starting a new one.

It fixes two issues:
- connection resources are now correctly cleaned up when user logs out
- `insecure` flag is passed to `kube login` when you run the app in that mode

### DIFF
--- a/packages/teleterm/src/ui/ClusterLogout/useClusterLogout.ts
+++ b/packages/teleterm/src/ui/ClusterLogout/useClusterLogout.ts
@@ -6,9 +6,7 @@ import { useAppContext } from '../appContextProvider';
 export function useClusterLogout({ clusterUri, onClose, clusterTitle }: Props) {
   const ctx = useAppContext();
   const [{ status, statusText }, removeCluster] = useAsync(async () => {
-    // TODO(gzdunek): logout and removeCluster should be combined into a single acton in tshd
     await ctx.clustersService.logout(clusterUri);
-    await ctx.clustersService.removeCluster(clusterUri);
 
     if (ctx.workspacesService.getRootClusterUri() === clusterUri) {
       const [firstConnectedWorkspace] =
@@ -19,7 +17,7 @@ export function useClusterLogout({ clusterUri, onClose, clusterTitle }: Props) {
         await ctx.workspacesService.setActiveWorkspace(null);
       }
     }
-    await ctx.connectionTracker.removeItemsBelongingToRootCluster(clusterUri);
+    ctx.connectionTracker.removeItemsBelongingToRootCluster(clusterUri);
     ctx.workspacesService.removeWorkspace(clusterUri);
   });
 
--- a/packages/teleterm/src/ui/ClusterLogout/useClusterLogout.ts
+++ b/packages/teleterm/src/ui/ClusterLogout/useClusterLogout.ts
@@ -19,8 +19,8 @@ export function useClusterLogout({ clusterUri, onClose, clusterTitle }: Props) {
         await ctx.workspacesService.setActiveWorkspace(null);
       }
     }
+    await ctx.connectionTracker.removeItemsBelongingToRootCluster(clusterUri);
     ctx.workspacesService.removeWorkspace(clusterUri);
-    ctx.connectionTracker.removeItemsBelongingToRootCluster(clusterUri);
   });
 
   useEffect(() => {
--- a/packages/teleterm/src/ui/ClusterLogout/useClusterLogout.ts
+++ b/packages/teleterm/src/ui/ClusterLogout/useClusterLogout.ts
@@ -17,8 +17,8 @@ export function useClusterLogout({ clusterUri, onClose, clusterTitle }: Props) {
         await ctx.workspacesService.setActiveWorkspace(null);
       }
     }
-    ctx.connectionTracker.removeItemsBelongingToRootCluster(clusterUri);
     ctx.workspacesService.removeWorkspace(clusterUri);
+    ctx.connectionTracker.removeItemsBelongingToRootCluster(clusterUri);
   });
 
   useEffect(() => {
--- a/packages/teleterm/src/ui/services/clusters/clustersService.ts
+++ b/packages/teleterm/src/ui/services/clusters/clustersService.ts
@@ -61,9 +61,12 @@ export class ClustersService extends ImmutableStore<ClustersServiceState> {
   }
 
   async logout(clusterUri: string) {
+    // TODO(gzdunek): logout and removeCluster should be combined into a single acton in tshd
     await this.client.logout(clusterUri);
     await this.syncClusterInfo(clusterUri);
     this.removeResources(clusterUri);
+    await this.removeCluster(clusterUri);
+    await this.removeClusterKubeConfigs(clusterUri);
   }
 
   async loginLocal(params: LoginLocalParams, abortSignal: tsh.TshAbortSignal) {
@@ -591,8 +594,25 @@ export class ClustersService extends ImmutableStore<ClustersServiceState> {
     return await this.client.listDatabaseUsers(dbUri);
   }
 
+  async removeClusterKubeConfigs(clusterUri: string): Promise<void> {
+    const {
+      params: { rootClusterId },
+    } = routing.parseClusterUri(clusterUri);
+    if (!rootClusterId) {
+      throw new Error(
+        'Could not remove kube configs, `rootClusterId` is missing.'
+      );
+    }
+    return this.mainProcessClient.removeKubeConfig({
+      relativePath: rootClusterId,
+      isDirectory: true,
+    });
+  }
+
   async removeKubeConfig(kubeConfigName: string): Promise<void> {
-    return this.mainProcessClient.removeKubeConfig(kubeConfigName);
+    return this.mainProcessClient.removeKubeConfig({
+      relativePath: kubeConfigName,
+    });
   }
 
   useState() {
--- a/packages/teleterm/src/ui/services/clusters/clustersService.ts
+++ b/packages/teleterm/src/ui/services/clusters/clustersService.ts
@@ -604,9 +604,9 @@ export class ClustersService extends ImmutableStore<ClustersServiceState> {
     });
   }
 
-  async removeKubeConfig(kubeConfigName: string): Promise<void> {
+  async removeKubeConfig(kubeConfigRelativePath: string): Promise<void> {
     return this.mainProcessClient.removeKubeConfig({
-      relativePath: kubeConfigName,
+      relativePath: kubeConfigRelativePath,
     });
   }
 
--- a/packages/teleterm/src/ui/services/clusters/clustersService.ts
+++ b/packages/teleterm/src/ui/services/clusters/clustersService.ts
@@ -598,11 +598,6 @@ export class ClustersService extends ImmutableStore<ClustersServiceState> {
     const {
       params: { rootClusterId },
     } = routing.parseClusterUri(clusterUri);
-    if (!rootClusterId) {
-      throw new Error(
-        'Could not remove kube configs, `rootClusterId` is missing.'
-      );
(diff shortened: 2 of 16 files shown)


packages/teleterm/src/ui/services/connectionTracker/connectionTrackerService.ts
```diff
@@ -129,29 +129,32 @@ export class ConnectionTrackerService extends ImmutableStore<ConnectionTrackerSt
...
+  async removeItem(id: string): Promise<void> {
     const connection = this.state.connections.find(c => c.id === id);
     if (!connection) {
       return;
     }
 
-    this._trackedConnectionOperationsFactory.create(connection).remove?.();
+    await this._trackedConnectionOperationsFactory
+      .create(connection)
+      .remove?.();
```
ravicious, line 140: Adding a noop `remove` to non-k8s connections would be a valid approach here too IMHO.

packages/teleterm/src/ui/services/connectionTracker/connectionTrackerService.ts
```diff
@@ -129,29 +129,32 @@ export class ConnectionTrackerService extends ImmutableStore<ConnectionTrackerSt
...
-        return rootClusterUri !== clusterUri;
-      });
-    });
+          this._trackedConnectionOperationsFactory.create(connection);
+        if (rootClusterUri === clusterUri) {
+          await this.disconnectItem(connection.id);
+          await this.removeItem(connection.id);
+        }
+      })
+    );
```
ravicious, line 157: I'm not fond of this change to be honest. Before this change, when the user was logging out, `connectionTrackerService` was responsible only for cleaning up its own state related to the cluster that the user is logging out from. It barely had any side effects, other than updating the state of the connection tracker.

Now `removeItemsBelongingToRootCluster` orchestrates the disconnection and removal of every active cluster item.

It also used to be a single update of connection tracker's state, while now every time an item is removed, it updates the workspace service which has repercussions on the whole app but it also runs `ConnectionTrackerService._refreshState` again on every removed item.

I realize it's convenient to use as `TrackedConnectionOperations` appears to include everything that we need to accomplish this goal. However, did you consider orchestrating that within `ClustersService`? It seems that on logout we need to remove cluster gateways  and kube configs. gravitational/webapps.e#311 is about removing gateways on tshd's side. Perhaps removing those kube configs could happen there as well?

gzdunek: I agree with your points, calling `refreshState` on every removed item is not nice, but we can try to optimize it. 

However, I'm not sure how do you see orchestrating the resource removal in `ClustersService`. For example, `kubeConfigName` lives only within a connection, how the service will know about it?

Changing the way how `removeItemsBelongingToRootCluster` works seemed natural to me. When user clicks on disconnect/remove buttons we remove the resources, so why not do it when more than one connection is removed? Thanks to this we keep the cleanup logic in one place.

ravicious: > When user clicks on disconnect/remove buttons we remove the resources, so why not do it when more than one connection is removed?

Because it ultimately muddies up the logic behind logging out, I think. The act of removing a single connection is different from logging out. Before the change, during logout `ConnectionTrackerService` was responsible for merely updating its own state to reflect the changes from other services. Now it's responsible not only for its own state, but also the shutdown of whole resources (gateways and k8s).

I don't think such shift in responsibility is necessary. `ConnectionTrackerService` already does a lot, `_refreshState` is a complex function and I don't think we should put even more logic into the service. In my eyes, that service works well as a simple reflection of the "history" of added documents. Currently it knows that some resources exists (gateways, k8s) but ultimately it doesn't manage their lifetimes – this is done by the logic in `Document`-type components, in `ClustersService` and so on.

The `disconnect` function does more than what's necessary during logout. It closes and removes documents from the state which is something that's already being done by `ctx.workspacesService.removeWorkspace(clusterUri)`. It removes gateways which is something that we'd like to do on the backend when calling `ctx.clustersService.logout`.

> However, I'm not sure how do you see orchestrating the resource removal in `ClustersService`. For example, `kubeConfigName` lives only within a connection, how the service will know about it?

This could be handled by putting all kube configs belonging to a particular profile into a single directory or giving their filenames the same prefix so that during logout we know already which files we need to remove. If that's not possible due to time constraints, I think I'd still prefer `ConnectionTrackerService` to merely provide the names of the kube configs that should be removed and put the logic behind actually removing them somewhere else.

`useClusterLogout` already has access to the app context, so it can execute methods directly on the main process client. `ClustersService` seems to accept the main process client as a dependency only so that we don't have to pass the main process client as another dep to `TrackedConnectionOperationsFactory`, doesn't it?

Ultimately, `useClusterLogout` is the place that keeps the cleanup logic. I'd rather keep it explicitly expressed there. With `ConnectionTrackerService` calling `disconnect` on each item belonging to a root cluster, we bounce around between a couple of different services, cause a couple of updates because of the state changes. It was not the case before this change.

packages/teleterm/src/ui/services/connectionTracker/connectionTrackerService.ts
gzdunek: Done.

packages/teleterm/src/ui/services/connectionTracker/connectionTrackerService.ts
gzdunek: Ok, I think I understand it finally, thanks for spending your time on writing things down!  
I decided to go with placing all kube configs belonging to a particular profile into a single directory. It looks like the simplest solution. 

>useClusterLogout already has access to the app context, so it can execute methods directly on the main process client. ClustersService seems to accept the main process client as a dependency only so that we don't have to pass the main process client as another dep to TrackedConnectionOperationsFactory, doesn't it?

Yes, so maybe the best place to put the clean up logic is `ClustersService` (as you proposed in the first comment)?
I expanded the `logout` method to remove the cluster and kube configs. 
However, I can move the cleanup logic to `useClusterLogout` if you are conceived that is better.

Do you have any opinion about the naming? I changed `kubeConfigName`param to `relativePath` in `removeKubeConfig()`, because now we can remove either `localhost/my-kube-abc` or `localhost` as a dir. 
For a kube document I think I will leave it as it is (`kubeConfigName`).

packages/teleterm/src/ui/services/workspacesService/documentsService/documentsService.ts
```diff
@@ -74,7 +74,9 @@ export class DocumentsService {
       leafClusterId: params.leafClusterId,
       kubeId: params.kubeId,
       kubeUri: options.kubeUri,
-      kubeConfigName: options.kubeConfigName || `${params.kubeId}-${unique(5)}`,
+      kubeConfigName:
+        options.kubeConfigName ||
+        `${params.rootClusterId}/${params.kubeId}-${unique(5)}`,
```
ravicious, line 79: I think it'd be a good idea to note here why we're prepending this with `rootClusterId`.

packages/teleterm/src/ui/services/workspacesService/documentsService/documentsService.ts
```diff
@@ -74,7 +74,9 @@ export class DocumentsService {
       leafClusterId: params.leafClusterId,
       kubeId: params.kubeId,
       kubeUri: options.kubeUri,
-      kubeConfigName: options.kubeConfigName || `${params.kubeId}-${unique(5)}`,
+      kubeConfigName:
```
ravicious, line 77: This could become `kubeConfigPath` perhaps, what do you think?

packages/teleterm/src/ui/ClusterLogout/useClusterLogout.ts
```diff
@@ -19,8 +17,8 @@ export function useClusterLogout({ clusterUri, onClose, clusterTitle }: Props) {
         await ctx.workspacesService.setActiveWorkspace(null);
       }
     }
-    ctx.workspacesService.removeWorkspace(clusterUri);
     ctx.connectionTracker.removeItemsBelongingToRootCluster(clusterUri);
+    ctx.workspacesService.removeWorkspace(clusterUri);
```
ravicious, line 21: I suppose we could bring the original order of these two calls to, couldn't we? So that `removeWorkspace` is called before `removeItemsBelongingToRootCluster` – if it's not necessary then there's no reason to change it. ;)

packages/teleterm/src/ui/ClusterLogout/useClusterLogout.ts
```diff
@@ -6,9 +6,7 @@ import { useAppContext } from '../appContextProvider';
 export function useClusterLogout({ clusterUri, onClose, clusterTitle }: Props) {
   const ctx = useAppContext();
   const [{ status, statusText }, removeCluster] = useAsync(async () => {
-    // TODO(gzdunek): logout and removeCluster should be combined into a single acton in tshd
     await ctx.clustersService.logout(clusterUri);
-    await ctx.clustersService.removeCluster(clusterUri);
```
ravicious, line 11: Good change!

packages/teleterm/src/ui/services/clusters/clustersService.ts
```diff
@@ -591,8 +594,25 @@ export class ClustersService extends ImmutableStore<ClustersServiceState> {
...
+        'Could not remove kube configs, `rootClusterId` is missing.'
+      );
+    }
+    return this.mainProcessClient.removeKubeConfig({
+      relativePath: rootClusterId,
+      isDirectory: true,
+    });
+  }
+
   async removeKubeConfig(kubeConfigName: string): Promise<void> {
```
ravicious, line 612: It's hard to tell if those two methods belong here or not. Technically they don't have anything to do with the state that ClustersService manages, I could imagine a completely separate KubeConfigService which accomplishes the same thing.

For now it's fine, but let's keep that in mind for this feature and any future ones.

ravicious (approved):
Left some minor comments but looks good overall now!

ravicious:
I tested it on my Mac, could you also do some tests on Windows? Just to be sure. ;f

packages/teleterm/src/ui/services/clusters/clustersService.ts
```diff
@@ -591,8 +594,25 @@ export class ClustersService extends ImmutableStore<ClustersServiceState> {
...
 
+  async removeClusterKubeConfigs(clusterUri: string): Promise<void> {
+    const {
+      params: { rootClusterId },
+    } = routing.parseClusterUri(clusterUri);
+    if (!rootClusterId) {
+      throw new Error(
+        'Could not remove kube configs, `rootClusterId` is missing.'
+      );
+    }
```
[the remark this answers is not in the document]
gzdunek: Right, no need to be so defensive here. 
Btw I just tested `path.join` and it throws an error when one of the arguments is `undefined`.

packages/teleterm/src/ui/services/workspacesService/documentsService/documentsService.ts
gzdunek: What would you say for `kubeConfigRelativePath`? It would match `relativePath` in `removeKubeConfig()`. `path` alone sound like a full path to the file 🤔

packages/teleterm/src/ui/ClusterLogout/useClusterLogout.ts
gzdunek: Done

packages/teleterm/src/ui/services/workspacesService/documentsService/documentsService.ts
ravicious: `kubeConfigRelativePath` sounds fine. 👍

gzdunek:
Tested on Windows, works perfectly

avatus (approved):

PR #1203, opened 2022-09-23
[Connect] Add Document Access Requests

avatus:
[teleport PR](https://github.com/gravitational/teleport/pull/16694)
Original description can be found [here](https://github.com/gravitational/webapps.e/pull/396).

> My thought process to give you some idea of the decisions I made
> 1. Most of the work was in clustersService. It seemed like any actual 'api calls' were happening there with the rest of the services being tied to frontend state. 
> 2. accessRequestsService was based on how documentsService was created since it is workspace specific. 
> 3.  although you can only create a single access request at a time, we wanted to 'state' to persist across tabs to allow for multiple searches and to support the 'Proceed to Checkout' button across the entire workspace. It added quite a bit of overhead but I think it works nicely. 
> 4. I tried to reuse components from webapps.e Workflow however some would break due to a `Link` being present. I would generally recreate the 'page'-ish level and import components and then add my own logic with a `useComponentName`. Similar pattern to what I saw on web but without the 'Container' (basically created my own "Container").
> 5. One thing I'm not quite happy with is the RequestCheckout decision. Its a large component and I didnt want to copy all of it to avoid using a `Link` in the successful checkout flow so I decided to add an prop to pass that component in. It's a bit ugly solution but allows us to reuse 90% of the code and just take out the Buttons with `Link`s. Would love feedback and maybe its a solution if we're happy.
> 6. adding more tests but wanted to get the review started, those will come shortly. 

This PR has all the proto file changes so the files changed number is a bit bloated but still quite large. 

and a neat little video to show off the basic flow/features

https://user-images.githubusercontent.com/5201977/192059421-0b9d73f9-7e82-43b8-99a1-962986d5b802.mov

Fixes gravitational/teleport#15169


avatus:
Sorry for the messy commit history. Had some weird submodule issues and merge conflicts so ended up nuking my local and bringing changes over.
I realize now I didn't properly bring over my leaf cluster logic so I will get that in on Monday

packages/teleterm/src/ui/services/workspacesService/workspacesService.ts
```diff
@@ -1,13 +1,25 @@
 import { useStore } from 'shared/libs/stores';
 
 import { isEqual } from 'lodash';
+import isPast from 'date-fns/isPast';
+
+// adding requests to assumed adds already 'made' requests that fit our tables
+import { AccessRequest as UiAccessRequest } from 'e-teleport/services/workflow';
```
gzdunek, line 7: Is this rename to `UiAccessRequest` needed?

packages/teleterm/src/ui/services/workspacesService/workspacesService.ts
```diff
@@ -20,14 +32,22 @@ export interface Workspace {
   localClusterUri: string;
   documents: Document[];
   location: string;
+  isAccessRequestsBarCollapsed: boolean;
```
gzdunek, line 35: Maybe we can keep related state together?
```typescript
accessRequests: {
  assumed,
  pending,
  isRequestBarCollapsed
}
```

packages/teleterm/src/ui/TopBar/Clusters/ConfirmClusterChangeDialog.tsx
```diff
@@ -0,0 +1,52 @@
...
+  DialogContent,
+  DialogFooter,
+  DialogHeader,
+} from 'design/DialogConfirmation';
+import { Close } from 'design/Icon';
+
+const changeSelectedClusterWarning =
+  'Resources from different clusters cannot be combined in an access request. Current items selected will be cleared. Are you sure you want to continue?';
+
+export default function ChangeResourceDialog({
```
gzdunek, line 14: This component is used with a name `<ConfirmChangeDialog>` and sits inside  `ConfirmClusterChangeDialog.tsx`, that's a bit confusing :p 
Using one name will improve readability.

packages/teleterm/src/ui/TopBar/Clusters/useClusters.ts
```diff
@@ -13,6 +14,41 @@ export function useClusters() {
...
+    if (!pendingAccessRequest) {
+      return false;
+    }
+
+    const count =
+      Object.keys(pendingAccessRequest.node).length +
+      Object.keys(pendingAccessRequest.db).length +
+      Object.keys(pendingAccessRequest.app).length +
+      Object.keys(pendingAccessRequest.kube_cluster).length +
+      Object.keys(pendingAccessRequest.windows_desktop).length;
```
gzdunek, line 37: The same code we have in `useAccessRequestsButton`. I'd see it in the `accessRequestsService`.

packages/teleterm/src/ui/TopBar/NavigationMenu/NavigationMenu.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { MoreVert, OpenBox, Add } from 'design/Icon';
+import { Box, Text, Flex } from 'design';
+
+import { AccessRequestDocumentState } from 'teleterm/ui/services/workspacesService';
+
+import { useIdentity } from '../Identity/useIdentity';
+
+import { NavigationItem } from './NavigationItem';
+
+export function NavigationMenuContainer() {
```
gzdunek, line 14: Is this component needed? I'd use early return in `NavigationMenu` if `activeRootCluster` is empty.

packages/teleterm/src/ui/TopBar/NavigationMenu/NavigationItem.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const ctx = useAppContext();
+  const docService = ctx.workspacesService.getActiveWorkspaceDocumentService();
+
+  const onClick = () => {
+    if (!docService) {
+      return;
+    }
+    const doc = docService.createAccessRequestDocument({ clusterUri, state });
+    docService.add(doc);
+    docService.open(doc.uri);
```
gzdunek, line 25: This logic should live in `navigationItems`: 
```typescript
{
   title: 'New Access Request',
   Icon: <Add fontSize={2} />
   onNavigate: () => {...}
},
```

`NavigationItem` doesn't need to know about `docService` or anything like that.

packages/teleterm/src/services/tshd/types.ts
```diff
@@ -125,3 +162,28 @@ export type CreateGatewayParams = {
...
+  search?: string;
+  searchAsRoles?: string;
+  sort?: SortType;
+  startKey?: string;
+  limit?: number;
+  query?: string;
+};
+
+export type ReviewAccessRequestParams = {
+  state: string;
```
gzdunek, line 177: What are the possible states?

packages/teleterm/src/ui/services/workspacesService/accessRequestsService.ts
```diff
@@ -0,0 +1,118 @@
...
+    private getState: () => {
+      isAccessRequestsBarCollapsed: boolean;
+      pendingAccessRequest: PendingAccessRequest;
+      assumed: Record<string, AccessRequest>;
+    },
+    private setState: (
+      draftState: (draft: {
+        isAccessRequestsBarCollapsed: boolean;
+        pendingAccessRequest: PendingAccessRequest;
+        assumed: Record<string, AccessRequest>;
```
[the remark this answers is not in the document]
avatus: I thought about this same thing tbh. I believe its because when adding a resource, we must check the type + name since the name isn't guaranteed to be unique (imagine a node named `teleport` and a db named `teleport`), however, I am storing by the ID anyway and I'm not sure it has the same issue? Sort of a relic of the past but, either either Object.keys in places, or array.filter/map in others to get counts+types in other places.

packages/teleterm/src/ui/services/workspacesService/accessRequestsService.ts
```diff
@@ -0,0 +1,118 @@
...
+  ) {}
+
+  getCollapsed() {
+    return this.getState().isAccessRequestsBarCollapsed;
+  }
+
+  toggleBar() {
+    this.setState(draftState => {
+      draftState.isAccessRequestsBarCollapsed =
+        !this.getState().isAccessRequestsBarCollapsed;
```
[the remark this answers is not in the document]
avatus: Thank you!

packages/teleterm/src/ui/services/workspacesService/workspacesService.ts
avatus: it was... until a refactor. Thank you for catching

packages/teleterm/src/ui/TopBar/NavigationMenu/NavigationItem.tsx
avatus: I agree, because we may need to do something special per item eventually. How is `navigationItems` supposed to be aware of the doc service? I'm a bit lost there, can you point to an existing pattern similar to this?

packages/teleterm/src/ui/TopBar/NavigationMenu/NavigationMenu.tsx
```diff
@@ -0,0 +1,111 @@
...
+  {
+    title: 'New Access Request',
+    Icon: <Add fontSize={2} />,
+    state: 'creating',
+  },
+  {
+    title: 'Review Access Requests',
+    Icon: <OpenBox fontSize={2} />,
+    state: 'reviewing',
+  },
```
ravicious, line 38: Re: feature detection in Connect, do we plan to do that before the release of access requests?

avatus: I believe that is to be a fast-follow

packages/teleterm/src/ui/services/clusters/clustersService.ts
```diff
@@ -383,6 +393,95 @@ export class ClustersService extends ImmutableStore<ClustersServiceState> {
...
+
+  async deleteAccessRequest(clusterUri: string, requestId: string) {
+    const cluster = this.state.clusters.get(clusterUri);
+    if (!cluster.connected) {
+      return;
+    }
+    return this.client.deleteAccessRequest(clusterUri, requestId);
+  }
+
+  async assumeRole(
```
ravicious, line 446: Assuming resource access requests doesn't seem to work for me. After assuming such request, I don't see the requested ssh servers in Connect.

Are you able to reproduce that? Maybe it's a problem with my role setup. But I just double checked it against the docs and it seems alright.


https://user-images.githubusercontent.com/27113/192784482-29a4a335-822e-4ddc-a6e2-6f992af4ea98.mov

avatus: Can you check if the issue persists on the web-ui as well? That is usually how I check if the issue is Connect related or my roles setup.

avatus: I believe a refactor has messed up the _creation_ of the resource request actually after inspecting. Which breaks assuming. Will get a fix in shortly

ravicious: Yeah, I just checked it. I went to the Web UI and assumed the request that was created through Connect and I didn't see the resource in the Web UI.

I created another resource request through the Web UI, assumed it through the Web UI and I saw the resource.

Then I went back to Connect, assumed the request made through the Web UI and I could see the resource too.

packages/teleterm/src/ui/services/workspacesService/accessRequestsService.ts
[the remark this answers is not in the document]
gzdunek: So the key has structure `type-resourceName`?

packages/teleterm/src/ui/services/clusters/clustersService.ts
avatus: Just pushed a fix that should address this. I was accidentally passing in node name instead of cluster name during the create. You can see the difference by checking your incorrect resource request's "cluster name" in the review page and comparing to a newly created resource request after the fix. Thank you!

ravicious: Yup, it works now!

packages/teleterm/src/ui/TopBar/NavigationMenu/NavigationItem.tsx
avatus: This is awesome thank you. yes, absolutely the best path to go with.

packages/teleterm/src/services/tshd/types.ts
avatus: Updated

packages/teleterm/src/ui/TopBar/NavigationMenu/NavigationItem.tsx
avatus: This came out great! Wonderful suggestion

packages/teleterm/src/ui/services/clusters/clustersService.ts
```diff
@@ -383,6 +393,95 @@ export class ClustersService extends ImmutableStore<ClustersServiceState> {
     }
   }
 
+  async fetchServers(params: ServerSideParams) {
```
ravicious, line 396: The callsites that call these new methods from `clustersService` need to wrap the calls to those methods in `retryWithRelogin`.

When I came back to the running dev version of Electron, I saw this:

cert has expired

I think this couldn't happen by itself because AFAIK Connect just doesn't decide to make a request by itself, I must've went from the view request "page" to the list of all requests.

In any case, the general rule is that the backend handlers that use `RetryWithRelogin` need to be wrapped with `retryWithRelogin` on the frontend for the whole system to take effect.

Look up how other places in the frontend app do this. `retryWithRelogin` needs to tightly wrap the call to ClustersService.

You might have a problem with passing a meaningful value as the `resourceUri` argument to `retryWithRelogin`. In case of, for example, listing all resources, I don't think you'll have a meaningful resource URI to pass there. In that case, that argument is used only to get the root cluster URI, so you can just pass the cluster URI (if it's a leaf cluster URI, `retryWithRelogin` will extract the root cluster URI from it automatically).

avatus: Ok, i wrapped the requests. Funnily enough, that `fetchServers` seems to have been vestigial from before the `getAllServers' and 'getServers' distinction as it isn't used anywhere in the codebase atm.

packages/teleterm/src/ui/services/workspacesService/accessRequestsService.ts
```diff
@@ -0,0 +1,128 @@
...
+  getAssumedRoles() {
+    const assumed = this.getState().assumed;
+    let roles = [];
+    Object.keys(assumed).forEach(key => {
+      const request = assumed[key];
+      const newRoles = request.roles.filter(r => !roles.includes(r));
+      roles = [...roles, ...newRoles];
+    });
+
+    return roles;
```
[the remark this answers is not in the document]
avatus: I believe this would always return an empty array because the `list` in this case is the array being iterated over, not the one being built. I do like the flatmap part. I think this would work well
```js
    return [
      ...new Set(
        Object.values(this.getAssumed()).flatMap(request => request.roles)
      ),
    ];
```

packages/teleterm/src/ui/services/workspacesService/workspacesService.ts
```diff
@@ -1,13 +1,24 @@
 import { useStore } from 'shared/libs/stores';
 
 import { isEqual } from 'lodash';
+import isPast from 'date-fns/isPast';
+
+import { AccessRequest } from 'e-teleport/services/workflow';
```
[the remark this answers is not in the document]
avatus: In general, any part of the UI that uses an `AccessRequest` is pulling from the e-teleport type, any response to us is our gRPC type. The difference being our types append `List` to the end of any repeated field sent back in gRPC and this was a bandaid I never ripped off. I think its ugly, however was a simpler fix than going down the grpc rabbithole and why that `List` was being appended. I couldn't find it but maybe you have an idea of where its at? Also, my thought process was "if I remove list for this request, it would probably destroy the rest of the app that relied on that append".

also, fwiw, created/expires are used in `makeAccessRequest` in e-teleport to convert json to Date objects
https://github.com/gravitational/webapps.e/blob/9498eaa8c9893853d72c09e604940600386853d5/teleport/src/services/workflow/makeAccessRequest.ts#L20-L22

Conversion happens here with an input type I named `TshdAccessRequest` 
https://github.com/gravitational/webapps.e/blob/36b095b107e85f20782f72cd1723e511cbfe534c/teleterm/src/ui/DocumentAccessRequests/useAccessRequests.tsx#L124-L133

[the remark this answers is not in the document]
ravicious: > The difference being our types append `List` to the end of any repeated field sent back in gRPC and this was a bandaid I never ripped off. I think its ugly, however was a simpler fix than going down the grpc rabbithole and why that `List` was being appended. I couldn't find it but maybe you have an idea of where its at?

It looks like that's how the code which gets autogenerated works. Maybe there's a config option to change that but yeah, at this point it's better to work around that.

It might make sense though to manually copy over fields rather than using a spread there, just to make finding fields by reference easier. This way someone trying to find usage of AccessRequest.id or AccessRequest.resolveReason will see where and how they're actually used.

gzdunek (commented):
A few more comments from me

PR #1207, opened 2022-09-26
Add caching to Webpack during development

ryanclark:
This adds file system based caching to Webpack during development, so subsequent re-runs of `webpack-dev-server` will start up really quickly as it'll reuse the cache from when it last ran.

I've added environment variables to disable this behaviour and to configure the path if wanted.

Initial compile time: 10317ms.

Stopping and then starting Webpack again, compile time: 697ms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,17 @@ requests to the given target.
 > Keep in mind that you have to use a local user because social
 > logins (google/github) are not supported by development server.
 
+#### Caching
+
+By default, Webpack will store a cache in `node_modules/.cache/webpack` during development. This
+makes starting `webpack-dev-server` really quick after having ran it once, as it will re-use the
+cache from the last time it was running.
+
+If you want to change the location of the cache, you can set `WEBPACK_CACHE_DIRECTORY` to an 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -68,6 +68,17 @@ requests to the given target.
...
+#### Caching
+
+By default, Webpack will store a cache in `node_modules/.cache/webpack` during development. This
+makes starting `webpack-dev-server` really quick after having ran it once, as it will re-use the
+cache from the last time it was running.
+
+If you want to change the location of the cache, you can set `WEBPACK_CACHE_DIRECTORY` to an 
+absolute file path of the folder where you want to store Webpack's cache.
+
+If you wish to disable the cache, you can set `WEBPACK_DISABLE_CACHE` to `yes`.
```
zmb3, line 80: In the config below, you use `process.env.WEBPACK_CACHE_DISABLED`. I don't care which you use, but let's make sure the docs and code refer to the same env var.

ryanclark: Oooops, fixed.

hatched (approved):

avatus (approved):

PR #1172, opened 2022-08-31
[Discover] Desktop Setup Flow

ryanclark:
This refactors the navigation to allow for infinite nesting of steps. The navigation will update to the relevant navigation when selecting a resource, as well as updating the icon next to "Add Resource"

Existing server flow:
image

Desktop flow:
image

image
image
image
image
image

PR #1211, opened 2022-09-27
Change `getFeatures()` to use a context and use `useFeatures()` instead

ryanclark:
#1172 incorrectly assumed that `useFeatures()` was agnostic like `useTeleport()` which returns the correct context based on OSS vs enterprise. As a result, the enterprise version is now being passed the OSS feature set, instead of the correct enterprise feature set. This caused access requests to go missing from the UI.

This fixes that by actually making it a context that will have its value change depending on if entered from the enterprise side or OSS side.

The feature set will always default to OSS if not set, but enterprise overrides this value with its own feature set.

### DIFF
--- a/packages/teleport/src/Main/Main.tsx
+++ b/packages/teleport/src/Main/Main.tsx
@@ -31,8 +31,6 @@ import history from 'teleport/services/history';
 
 import { ClusterAlert, LINK_LABEL } from 'teleport/services/alerts/alerts';
 
-import { Feature } from 'teleport/types';
-
 import { MainContainer } from './MainContainer';
 import { OnboardDiscover } from './OnboardDiscover';
 import useMain from './useMain';
@@ -42,13 +40,11 @@ import type { BannerType } from 'teleport/components/BannerList/BannerList';
 interface MainProps {
   initialAlerts?: ClusterAlert[];
   customBanners?: React.ReactNode[];
-  features: Feature[]; // TODO(ryan): move this into a context so we can use `useFeatures()` here
 }
 
 export function Main(props: MainProps) {
   const { alerts, ctx, customBanners, dismissAlert, status, statusText } =
     useMain({
-      features: props.features,
       initialAlerts: props.initialAlerts,
       customBanners: props.customBanners,
     });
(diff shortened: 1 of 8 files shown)


packages/teleport/src/Main/Main.tsx
```diff
@@ -41,14 +42,13 @@ import type { BannerType } from 'teleport/components/BannerList/BannerList';
 interface MainProps {
   initialAlerts?: ClusterAlert[];
   customBanners?: React.ReactNode[];
+  features: Feature[]; // TODO(ryan): move this into a context so we can use `useFeatures()` here
 }
 
 export function Main(props: MainProps) {
-  const [features] = useState(() => getFeatures());
```
kimlisa, line 47: we will need this right? for the open source to get its features (unless i'm missing something) i'm fine if we did something like this for a quick patch (this was how i was doing it):

```ts
 const [features] = useState(() => {
    if (!props.customFeatures) {
      return getFeatures();
    }

    return props.customFeatures;
  });
```

packages/teleport/src/Main/Main.tsx
```diff
@@ -26,12 +26,13 @@ import cfg from 'teleport/config';
 import SideNav from 'teleport/SideNav';
 import TopBar from 'teleport/TopBar';
 import { BannerList } from 'teleport/components/BannerList';
-import getFeatures from 'teleport/features';
```
kimlisa, line 29: unless i'm missing something, we have two `getFeatures`, this one returns OSS features, the `e-teleport/features` returns OSS features + enterprise features

ryanclark: I've changed these to named exports, so we don't confuse the two of them.

packages/teleport/src/Main/Main.tsx
ryanclark: I've changed this to use a context now, and the correct value is passed in at the same time the Teleport context is created.

packages/teleport/src/Teleport.tsx
```diff
@@ -17,10 +17,14 @@ limitations under the License.
 import React from 'react';
 import ThemeProvider from 'design/ThemeProvider';
 
+import { getEnterpriseFeatures } from 'e-teleport/features';
```
kimlisa, line 20: this would pull `e-teleport` as a dependency when we build OSS?

ryanclark: Aha I just noticed this whilst rebuilding to test the OSS version. Cheers.

kimlisa (approved):
this is great!

PR #1213, opened 2022-09-27
Elasticsearch audit events.

Tener:
Add support for Elasticsearch audit events

See: https://github.com/gravitational/teleport/pull/16191


packages/teleport/src/services/audit/makeEvent.ts
```diff
@@ -889,6 +889,36 @@ export const formatters: Formatters = {
...
+          categoryString = 'SEARCH';
+          break;
+        case ElasticsearchCategory.SECURITY:
+          categoryString = 'SECURITY';
+          break;
+        case ElasticsearchCategory.SQL:
+          categoryString = 'SQL';
+          break;
+      }
+      return `User [${user}] has ran a query [${query}] in category [${categoryString}] against [${target}] on [${db_service}], request path: [${path}]`;
```
smallinsky, line 919: Can we use `truncateStr` function to ensure that logged query will be limited to 80 chars ? 
Also for most of cases query  param will be empty so wonder if we can return a different message 
```
if (query == "") {
  return ...
}
return `User [${user}] has ran a query [${query}]
```

packages/teleport/src/services/audit/makeEvent.ts
```diff
@@ -889,6 +889,36 @@ export const formatters: Formatters = {
     format: ({ user, db_service, db_name, proc_name }) =>
       `User [${user}] has send RPC Request [${proc_name}] in database [${db_name}] on [${db_service}]`,
   },
+  [eventCodes.ELASTICSEARCH_REQUEST]: {
+    type: 'db.session.elasticsearch.request""',
+    desc: 'Elasticsearch Request',
+    format: ({ user, db_service, db_name, category, target, query, path }) => {
```
smallinsky, line 895: `db_name` seems to be unused: 
```suggestion
    format: ({ user, db_service, category, target, query, path }) => {
```

packages/teleport/src/services/audit/makeEvent.ts
```diff
@@ -889,6 +889,36 @@ export const formatters: Formatters = {
     format: ({ user, db_service, db_name, proc_name }) =>
       `User [${user}] has send RPC Request [${proc_name}] in database [${db_name}] on [${db_service}]`,
   },
+  [eventCodes.ELASTICSEARCH_REQUEST]: {
+    type: 'db.session.elasticsearch.request""',
```
gzdunek, line 893: What is the purpose of `""`?

Tener: None, removed.

packages/teleport/src/services/audit/makeEvent.ts
Tener: Done, `target` got the same treatment as `query` as it can also be empty.

packages/teleport/src/services/audit/makeEvent.ts
Tener: Done.

packages/teleport/src/services/audit/makeEvent.ts
```diff
@@ -916,7 +916,18 @@ export const formatters: Formatters = {
           categoryString = 'SQL';
           break;
       }
-      return `User [${user}] has ran a query [${query}] in category [${categoryString}] against [${target}] on [${db_service}], request path: [${path}]`;
+
+      let message = `User [${user}] has ran a [${categoryString}] query in [${db_service}], request path: [${path}]`;
+
+      if (query != '') {
+        message = message + `, query string: [${truncateStr(query, 80)}]`;
```
gzdunek, line 923: Nit: 
```typescript
if (query) {
  message += `, query string: [${truncateStr(query, 80)}]`;
}
```
(same below)

gzdunek: `if` can be also shortened :)

Tener: All right, I can see this is the idiomatic way of checking for empty string, done.

smallinsky (approved):

PR #1216, opened 2022-09-27
Add application CRUD audit events.

mdwn:
Application audit events for CRUD events are now recognized in the frontend as opposed to being displayed as 'unknown.'

This will address #9490.


packages/teleport/src/services/audit/makeEvent.ts
```diff
@@ -752,6 +752,24 @@ export const formatters: Formatters = {
     desc: 'Database Deleted',
     format: ({ user, name }) => `User [${user}] deleted database [${name}]`,
   },
+  [eventCodes.APPLICATION_CREATED]: {
+    type: 'app.create',
```
zmb3, line 756: Here we have `app.create` but in the fixture above we have `app.created`. What's the difference?

mdwn: Whoops. That would be a typo from me, good catch.

packages/teleport/src/services/audit/types.ts
```diff
@@ -215,6 +218,24 @@ export type RawEvents = {
...
+    }
+  >;
+  [eventCodes.APPLICATION_UPDATED]: RawEvent<
+    typeof eventCodes.APPLICATION_UPDATED,
+    {
+      name: string;
+    }
+  >;
+  [eventCodes.APPLICATION_DELETED]: RawEvent<
+    typeof eventCodes.DATABASE_DELETED,
```
r0mant, line 234: Should this be `APPLICATION_DELETED`?

packages/teleport/src/services/audit/makeEvent.ts
mdwn: This is fixed, but I'm sort of confused why it worked -- is the `type` not used for anything outside of the details page?

packages/teleport/src/services/audit/types.ts
```diff
@@ -49,6 +49,9 @@ export const eventCodes = {
   APP_SESSION_CHUNK: 'T2008I',
   APP_SESSION_START: 'T2007I',
   APP_SESSION_END: 'T2011I',
+  APPLICATION_CREATED: 'TAP03I',
+  APPLICATION_UPDATED: 'TAP04I',
+  APPLICATION_DELETED: 'TAP05I',
```
r0mant, line 54: This is definitely a nitpick, but can we name these `APP_CREATED`, `APP_UPDATED`, etc. so they're consistent with event codes that are already defined here?

packages/teleport/src/services/audit/types.ts
mdwn: Yep, that's correct. Bad copy-paste.

packages/teleport/src/services/audit/types.ts
mdwn: Sure, no prob!

r0mant (approved):

zmb3 (approved):

zmb3:
@mdwn Please reach out if you need help with the backport process. We want to make sure this fix makes it into all supported versions.

mdwn:
> @mdwn Please reach out if you need help with the backport process. We want to make sure this fix makes it into all supported versions. 

Ack, definitely will do -- I'll reach out in the morning!

PR #1220, opened 2022-09-28
Connect: Accommodate for making gRPC server creds in shared process

ravicious:
Equivalent Teleport PR: gravitational/teleport#16782.

On Windows, we use gRPC over TCP with mTLS since Node.js doesn't support UDS on Windows. Each process creates its own keypair and saves the public key to a predetermined location.

Up until now, these were the client-server pairs we had to support:

|   Server                    |   Client   |
|-----------------------------|------------|
|   tsh daemon                |   renderer |
|   shared (Electron process) |   renderer |

For tshd-initiated communication, the tshd process will need to create a client that will connect to a gRPC server operated by the renderer process of the Electron app.

## `grpcCredentials` refactor

`packages/teleterm/src/services/grpcCredentials` used to look like this:

```
packages/teleterm/src/services/grpcCredentials/
├── clientCredentials.ts
├── helpers.ts
├── index.ts
├── makeCert
├── serverCredentials.ts
└── types.ts
```

Now it looks like this:

```
packages/teleterm/src/services/grpcCredentials/
├── credentials.ts
├── files.ts
├── index.ts
├── makeCert
└── types.ts
```

`clientCredentials.ts` and `serverCredentials.ts` contained logic that was specific to a concrete process (renderer for clientCredentials). I extracted generic functions from those files. The orchestration of credentials for each process happens now in that process' main file.

Instead of having a catch-all `helpers.ts` or `utils.ts` file, I split the generic functions into functions that operate primarily on files (`files.ts`) and functions that work with credentials (`credentials.ts`). IMHO it's very important to do that – having a `utils.ts` file is almost always a bad choice as it signals that we didn't spend enough time thinking how the contents of that file fit into the general structure of the project.

* https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common
* https://breadcrumbscollector.tech/stop-naming-your-python-modules-utils/
* https://www.adam-bien.com/roller/abien/entry/class_naming_suffix_util_is


packages/teleterm/src/services/grpcCredentials/types.ts
```diff
@@ -1,6 +1,9 @@
-// `Client` and `TshServer` file names are also used on the tshd side
+// Each process creates its own key pair. The public key is saved to disk under the specified
+// filename, the private key stays in the memory.
+//
+// `Renderer` and `Tshd` file names are also used on the tshd side.
 export enum GrpcCertName {
-  Client = 'client.crt',
```
ravicious, line 3: Now that we're going to have both a gRPC client and a gRPC server in the renderer process, the name "client" stopped being sufficient. I refactored this so that it closer reflects what we want to achieve with gRPC certs – each process owns its own key pair, keeps the private key in memory and the public key is saved to the file.

packages/teleterm/src/sharedProcess/sharedProcess.ts
```diff
@@ -1,13 +1,24 @@
...
 import { RuntimeSettings } from 'teleterm/mainProcess/types';
 import Logger from 'teleterm/logger';
 
 import { PtyHostService } from './api/protogen/ptyHostService_grpc_pb';
 import { createPtyHostService } from './ptyHost/ptyHostService';
 
+const runtimeSettings = getRuntimeSettings();
+initializeLogger(runtimeSettings);
+initializeServer(runtimeSettings);
+
```
ravicious, line 21: This was just moved from the bottom of the file to the top, before function declarations so that you can immediately see what's going to be executed when this file is loaded.

packages/teleterm/src/services/grpcCredentials/makeCert/makeCert.ts
```diff
@@ -69,6 +72,15 @@ export async function makeCert({
...
     },
+    {
+      name: 'subjectAltName',
+      altNames: [
+        {
+          type: 2, // DNS type
+          value: commonName,
+        },
+      ],
+    },
```
ravicious, line 83: To my surprise, Go didn't complain when a cert without this was used as a client cert, but it was a problem when I tried to use it as a server cert for the client credentials.

packages/teleterm/src/preload.ts
```diff
@@ -38,6 +48,38 @@ async function getElectronGlobals(): Promise<ElectronGlobals> {
...
+}> {
+  if (!shouldEncryptConnection(runtimeSettings)) {
+    return {
+      tshd: createInsecureClientCredentials(),
+      shared: createInsecureClientCredentials(),
+    };
+  }
+
+  const { certsDir } = runtimeSettings;
+  const [rendererKeyPair, tshdCert, sharedCert] = await Promise.all([
```
avatus, line 75: What happens if any of these promises fail?

avatus (commented):
I like it and I'm excited for what it leads to. Also, side note: what a succinct and informative PR.

packages/teleterm/src/preload.ts
ravicious: `createGrpcCredentials` will throw, then after it `getElectronGlobals` will throw. `getElectronGlobals` gets exposed through contextBridge as `window.electron`:

https://github.com/gravitational/webapps/blob/be9639ef963520f678665d8479549b1739f27fc0/packages/teleterm/src/preload.ts#L33

It's used in `boot.tsx`:

https://github.com/gravitational/webapps/blob/be9639ef963520f678665d8479549b1739f27fc0/packages/teleterm/src/ui/boot.tsx#L13-L14

`getElectronGlobals` in `boot.tsx` calls `window.electron` underneath, so it will fail as well, triggering the `catch` branch:

https://github.com/gravitational/webapps/blob/be9639ef963520f678665d8479549b1739f27fc0/packages/teleterm/src/ui/boot.tsx#L28-L30

This will show the error to the user in the actual browser window.

gzdunek (commented):
Well done @ravicious! I don't have any comments to the code, I have to try it out (I'll do it on Monday).

avatus (approved):

ravicious:
I’m going to backport this once v11.0.0 is released.

PR #1222, opened 2022-09-28
Create Tabs Component

kimlisa:
#### Description
Create a general tabs component based on figma design

#### Figma Design
https://www.figma.com/file/8DQZl1mYUPLgvKgmA6zPOM/In-Progress-Work?node-id=1296%3A66853
image

#### Screenshot
image

image

image


packages/teleport/src/components/TabIcon/index.ts
```diff
@@ -0,0 +1,18 @@
+/*
+Copyright 2019 Gravitational, Inc.
```
avatus, line 2: 2022 here as well?

avatus (approved):

packages/teleport/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,81 @@
...
+          // overlay scrollbars.
+          ::-webkit-scrollbar-track {
+            background: rgba(255, 255, 255, 0.05);
+          }
+        `}
+      >
+        {contents.map((content, index) => {
+          const isActive = index === currContentIndex;
+          return (
+            <Box
```
JanKaczmarkiewicz, line 39: [nit]
those should be button elements since they are interactive and can benefit from tab focusing

packages/teleport/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,81 @@
...
+    </Box>
+  );
+}
+
+type Content = {
+  // title is the tab title.
+  title: string;
+  // component is the component to render as children in
+  // tabs container.
+  component: JSX.Element;
```
JanKaczmarkiewicz, line 76: [nit]
I would rename this property to "content", "children", or "element". 

Component (in react context) means function or class that will return jsx whereas we deal with the component result in this case.

JanKaczmarkiewicz (approved):

packages/teleport/src/components/TabIcon/index.ts
```diff
@@ -0,0 +1,18 @@
+/*
```
JanKaczmarkiewicz, line 1: I think it's a broader topic but do we consider placing only reference to license file as eg Meta does? eg. https://github.com/facebook/react/blob/main/packages/react-devtools-core/src/standalone.js
What do you think?

packages/teleport/src/components/TabIcon/index.ts
kimlisa: it was only the file name that changed, opting to leave as is

packages/teleport/src/components/Tabs/Tabs.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+import React from 'react';
+
+import { render, screen, fireEvent } from 'design/utils/testing';
+
+import { Tabs } from './Tabs';
+
+test('init tab highlight and content', async () => {
+  const { container } = render(<TestTabs />);
+  expect(container).toMatchSnapshot();
```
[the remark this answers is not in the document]
kimlisa: i was originally going to do this and class name testing to ensure the correct styling was applied on the active tab, but felt that snapshotting did it better, what do you think?

packages/teleport/src/components/TabIcon/index.ts
kimlisa: sure, i'm okay doing this if the others are okay with it

PR #1225, opened 2022-09-30
Add file transfer to Connect

gzdunek:
Closes https://github.com/gravitational/webapps.e/issues/179
This PR refactors `FileTransfer` so it can be used in both WebUI and Connect. 

The biggest change from the previous version is that `FileTransfer` no longer makes HTTP requests on its own. Instead, it allows you to provide your own implementations for upload/download functions. 
There's also a refreshed UI and, as we discussed at the Design Meeting, there's no longer an Upload button. Uploading starts as soon as you select files.  
Additionally, `FileTransfer` has been moved from `teleport` to the `shared` package. 

#### How Web UI works 
Upload/download in the WebUI works as before, they use HTTP (XHR) to transfer the files. When a user downloads a file, it is stored in the browser's memory, and when it is finished, the browser's download is triggered.

#### How Connect works 
The upload/download for Connect is quite different. It doesn't physically move the files, but delegates this task to tshd (SFTP). We just specify the source/destination paths via gRPC server-side stream, and all the dirty work is done for us :) Server-side streaming is used to send the transfer progress back to the UI. To specify the destination path to download the file, we use the native OS save dialog.

### Info for the reviewers
Review commit-by-commit.

At first I didn't plan to do a refactor, but then I realized that the changes would be significant anyway, so this is a great opportunity to refactor the whole module to _modern React_, add full TS support and simplify the final usage. I also added tests and better stories. 

Although the PR added/updated/removed many files (sorry 😞), I tried to keep the logic in one place. The main logic of `FileTransfer`  is in `useFilesStore.tsx` and `FileTransfer.tsx`. 
Similarly `DocumentSsh`, `DocumentTerminal`, `getHttpFileTransferHandlers` (the XHR orchestration here is 1:1 to the previous `Transfer` service) are the places where most of the callers work is done.


packages/shared/components/FileTransfer/FileTransferStateless/UploadForm/UploadForm.tsx
```diff
@@ -0,0 +1,142 @@
...
+    fileSelectorRef.current.click();
+  }
+
+  function handleDrop(e: React.DragEvent<HTMLDivElement>): void {
+    removeDropzoneStyle(e);
+
+    const { files } = e.dataTransfer;
+    e.preventDefault();
+    e.stopPropagation();
+    upload(extractFilesFromFileList(files));
```
JanKaczmarkiewicz, line 63:
```suggestion
    upload(Array.from(files));
```

packages/shared/components/FileTransfer/useFilesStore.ts
```diff
@@ -0,0 +1,174 @@
...
+    [updateTransferState]
+  );
+
+  const cancel = useCallback((id: string) => {
+    fileTransferControllers.current?.get(id).abortController.abort();
+  }, []);
+
+  const files = useMemo(
+    () => state.ids.map(id => state.filesById[id]),
+    [state.filesById]
```
JanKaczmarkiewicz, line 159:
```suggestion
    [state.filesById, state.ids]
```

packages/shared/components/FileTransfer/FileTransferStateless/UploadForm/UploadForm.tsx
gzdunek: Thanks, I didn't know about that :P

packages/shared/components/FileTransfer/FileTransfer.tsx
```diff
@@ -0,0 +1,136 @@
...
+    };
+
+    if (!isAnyTransferInProgress) {
+      runCloseCallbacks();
+      return;
+    }
+
+    if (props.beforeClose ? await props.beforeClose() : true) {
+      runCloseCallbacks();
+    }
```
ravicious, line 67: If there's no `beforeClose`, we want to always just run the close callbacks, right?

How about this:

```suggestion
    if (!isAnyTransferInProgress || !props.beforeClose) {
      runCloseCallbacks();
      return;
    }

    if (await props.beforeClose()) {
      runCloseCallbacks();
    }
```

This way we don't have to deal with that ternary branch that always returns true.

packages/shared/components/FileTransfer/useFilesStore.ts
```diff
@@ -0,0 +1,174 @@
...
+            ...action.payload,
+            transferState: { type: 'processing', progress: 0 },
+          },
+        },
+      };
+    }
+    case 'updateTransferState': {
+      const getTransferState = (): TransferState => {
+        if (action.payload.transferState.type === 'error') {
+          const { transferState } = state.filesById[action.payload.id];
```
ravicious, line 72:
```suggestion
          const { transferState: currentTransferState } = state.filesById[action.payload.id];
```

Otherwise it's unclear from `transferUsage` below if it comes from `action.payload` or from the state.

packages/shared/components/FileTransfer/useFilesStore.ts
```diff
@@ -0,0 +1,174 @@
...
+          ...state.filesById,
+          [action.payload.id]: {
+            ...action.payload,
+            transferState: { type: 'processing', progress: 0 },
+          },
+        },
+      };
+    }
+    case 'updateTransferState': {
+      const getTransferState = (): TransferState => {
```
ravicious, line 70: I'd suggest adding "next" or "updated" to the name of this function to signify that it's responsible for calculating the next state that's supposed to be saved rather than getting the current value.

```suggestion
      const getNextTransferState = (): TransferState => {
```

packages/shared/components/FileTransfer/useFilesStore.ts
```diff
@@ -0,0 +1,174 @@
...
+          ...state.filesById,
+          [action.payload.id]: {
+            ...state.filesById[action.payload.id],
+            transferState: getTransferState(),
+          },
+        },
+      };
+    }
+    default:
+      throw new Error();
```
ravicious, line 94: Could you add an error message here?

packages/shared/components/FileTransfer/useFilesStore.ts
```diff
@@ -0,0 +1,174 @@
...
+  const [state, dispatch] = useReducer(reducer, initialState);
+  const fileTransferControllers = useRef(
+    new Map<
+      string,
+      { runFileTransfer: RunFileTransfer; abortController: AbortController }
+    >()
+  );
+
+  const add = (file: Omit<NewTransferredFile, 'id'>) => {
+    const id = new Date().getTime() + file.name;
```
ravicious, line 108: Is it a remnant of the previous implementation? If not, then what's the advantage of doing this over using `unique()`?

ravicious (commented):
I'll take a look at the rest of the UI code and the backend code tomorrow.

packages/shared/components/FileTransfer/FileTransfer.test.tsx
```diff
@@ -0,0 +1,173 @@
...
+  );
+  fireEvent.change(screen.getByLabelText('File Path'), {
+    target: { value: '/Users/g/file.txt' },
+  });
+  fireEvent.click(screen.getByText('Download'));
+  fireEvent.click(await screen.findByTitle('Cancel'));
+  expect(abortControllerMock.signal.aborted).toBeTruthy();
+});
+
+test('file is not added when  transferHandler does not return anything', async () => {
```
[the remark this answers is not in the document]
gzdunek: Done

packages/shared/components/FileTransfer/FileTransfer.test.tsx
```diff
@@ -0,0 +1,173 @@
...
+  const handler: TransferHandlers = {
+    getDownloader: async () => undefined,
+    getUploader: async () => undefined,
```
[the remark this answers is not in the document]
gzdunek: The handler returning nothing is my way to handle a situation where user [cancels selecting the path for a file to download](https://github.com/gravitational/webapps/blob/aa8afe81f2bad2b32646128e3e03c20397fb5d92/packages/teleterm/src/ui/DocumentTerminal/DocumentTerminal.tsx#L78). If nothing is returned I skip adding file to the list. We could theoretically throw an error, do you think this will make it easier to use the API?

packages/shared/components/FileTransfer/FileTransfer.test.tsx
```diff
@@ -0,0 +1,173 @@
...
+    </FileTransferContextProvider>
+  );
+  fireEvent.change(screen.getByLabelText('File Path'), {
+    target: { value: filePath },
+  });
+  fireEvent.click(screen.getByText('Download'));
+  expect(screen.queryByText('/Users/g/file.txt')).not.toBeInTheDocument();
+});
+
+describe('*close callbacks are called when dialog is closed while transferring', () => {
```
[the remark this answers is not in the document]
gzdunek: Thanks, done

packages/teleport/src/Console/DocumentSsh/ActionBar/ActionBar.tsx
```diff
@@ -1,58 +0,0 @@
-/*
```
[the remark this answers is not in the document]
gzdunek: Yes, I should have done it that way, I will remember about it next time.

packages/shared/components/FileTransfer/FileTransfer.tsx
gzdunek: Thx, I forgot to simplify this logic :) Done

packages/shared/components/FileTransfer/useFilesStore.ts
gzdunek: Done

packages/shared/components/FileTransfer/useFilesStore.ts
gzdunek: Done

packages/shared/components/FileTransfer/useFilesStore.ts
gzdunek: It's a remnant, changed to `unique()`.

packages/shared/components/FileTransfer/useFilesStore.ts
```diff
@@ -0,0 +1,174 @@
...
+      runFileTransfer: file.runFileTransfer,
+      abortController: new AbortController(),
+    });
+  };
+
+  const updateTransferState = useCallback(
+    (id: string, transferState: TransferState) => {
+      dispatch({ type: 'updateTransferState', payload: { id, transferState } });
+    },
+    [dispatch]
```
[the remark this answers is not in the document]
gzdunek: Right, removed.

packages/shared/components/FileTransfer/FileTransfer.test.tsx
```diff
@@ -0,0 +1,173 @@
...
+    handleBeforeClose.mockReturnValue(Promise.resolve(true));
+    fireEvent.click(screen.getByTitle('Close'));
+    expect(handleBeforeClose).toHaveBeenCalled();
+
+    // wait for dialog to close
+    await waitFor(() => {
+      expect(
+        screen.queryByTestId('file-transfer-container')
+      ).not.toBeInTheDocument();
+    });
```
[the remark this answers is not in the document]
gzdunek: Thx, done

JanKaczmarkiewicz (approved):

packages/shared/components/FileTransfer/FileTransfer.test.tsx
```diff
@@ -0,0 +1,173 @@
...
+      target: { value: '~/abc' },
+    });
+
+    fireEvent.click(screen.getByText('Download'));
+    await screen.findByRole('listitem');
+
+    return { handleBeforeClose, handleAfterClose };
+  };
+
+  test('handleAfterClose is not called when close is cancelled', async () => {
```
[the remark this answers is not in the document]
ravicious: "is not called when closing the dialog has been aborted" would be fine IMHO. 👍

packages/design/src/ButtonIcon/ButtonIcon.jsx
```diff
@@ -50,6 +50,7 @@ const fromProps = props => {
   return {
     '&:disabled': {
       color: theme.colors.action.disabled,
+      pointerEvents: 'none',
```
ravicious, line 53: We might be better off solving it this way:

```js
    '&:disabled': {
      color: theme.colors.action.disabled,
    },
    '&:hover:enabled, &:focus:enabled': {
      background: theme.colors.action.hover,
    },
```

I just tested this and it still works, I suppose you wanted to remove the hover effect when the file transfer box is shown?

`pointerEvents: none` strikes me as a very substantial change and I'm worried that there are some places in the code which depend on `ButtonIcon` being a target of pointer events, even when disabled.

ravicious (approved):
The keyboard navigation works fine now, thanks for fixing this.

ravicious:
I didn't get around to testing if the Web UI file transfer still works correctly but I did test it for Connect.

Also, did you compare performance of file transfer through Connect vs Web UI and tsh? I won't be able to do that today but it'd be good to make sure we're not sluggish compared to them for some reason.

packages/shared/components/FileTransfer/FileTransfer.test.tsx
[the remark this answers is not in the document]
gzdunek: Done

packages/design/src/ButtonIcon/ButtonIcon.jsx
gzdunek: It works great, thanks!
I also added `cursor: default` when disabled.

packages/teleterm/src/services/tshd/createClient.ts
```diff
@@ -413,8 +415,37 @@ export default function createClient(
...
+        listeners.onProgress(data.getPercentage())
+      );
+      call.on('end', () => {
+        listeners.onComplete();
+        call.removeAllListeners();
+      });
+      call.on('error', error => {
+        listeners.onError(error);
+        call.removeAllListeners();
+      });
```
ravicious, line 446: I wonder if perhaps a better idea would be to make the tshd client return just the `call` object and then contain the rest of the logic within `FileTransferClient`.

Right now all `FileTransferClient` does is just pass some stuff further down – at this point one might wonder why we even bother creating this class at all.

See how I organized the stream stuff in [my initial draft PR for tshd-initiated communication](https://github.com/gravitational/webapps/pull/1176/files#diff-f59f2f399d4ba6b8395ee8cacb38f0d2ac18fb12050554e9249747f04efb2371). I'm talking about the files in `services/tshd`. Though here you probably don't want to call `FileTransferClient` from within tshd client, but rather as I suggested, tsdh client would return the call object and then `FileTransferClient` could operate on it.

File transfer doesn't need to follow it 1:1. I think the files you created so far are alright it's just a matter of moving the logic around a bit.

ravicious (commented):
As I mentioned in DMs, I somehow totally forgot to take a look at the non-UI code. I might have more followups soon, right now I'm looking at the problem with server ID on the backend side of this.

packages/teleterm/src/services/tshd/createClient.ts
gzdunek: Unfortunately, context bridge doesn't allow us to return `call` object, but I followed the structure from linked PR, so now we have a function `createFileTransferStream` that is returned from tshd client.

Also, as discussed in DM's, we decided not to pass `FileTransferListeners` as an argument, but instead we return it from the tshd/http calls. It makes the code easier to understand, especially `getDownloader`/`getUploader` parts (less arguments, one callback instead of a callback returning another callback).
I also added `createFileTransferEventsEmitter` utility function that makes working with the events/listeners more convenient.

packages/teleterm/src/ui/services/fileTransferClient/fileTransferClient.ts
```diff
@@ -0,0 +1,23 @@
+import { FileTransferListeners } from 'shared/components/FileTransfer';
+
+import { TshClient } from 'teleterm/services/tshd/types';
+import { FileTransferRequest } from 'teleterm/services/tshd/v1/service_pb';
+
+export class FileTransferClient {
+  constructor(private tshClient: TshClient) {}
+
+  transferFile(
+    options: FileTransferRequest.AsObject,
```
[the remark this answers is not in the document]
gzdunek: Done

packages/teleterm/src/ui/DocumentTerminal/useTshFileTransferHandlers.ts
```diff
@@ -47,14 +46,14 @@ function transferFile(
   appContext: IAppContext,
   originatingDocumentUri: string,
   file: FileTransferRequestObject,
-  fileTransferListeners: FileTransferListeners,
   abortController: AbortController,
   direction: FileTransferDirection
-): void {
+): FileTransferListeners {
   const server = appContext.clustersService.getServer(file.serverUri);
+  const eventsEmitter = createFileTransferEventsEmitter();
```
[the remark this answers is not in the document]
gzdunek: Yes, `appContext.fileTransferClient.transferFile` already returns an object that we can use, but I need the event emitter to create a layer between `tshClient` and `FileTransfer`. Events returned from the `tshClient` are used inside `useTshFileTransferHandlers` and `FileTransfer` uses these returned from `useTshFileTransferHandlers`.
It is needed, because the errors needs to go through `retryWithRelogin` which decides what to do with them first.
For example, when there is `ssh: cert has expired` error should show to the user only when he reject to log in. 

I wanted to avoid this, but I don't see any other way not to create an event emitter.

packages/teleterm/src/ui/services/fileTransferClient/fileTransferClient.ts
```diff
@@ -19,7 +18,6 @@ export class FileTransferClient {
         abortController.signal.removeEventListener('abort', cb);
       },
     };
-
-    return this.tshClient.transferFile(options, listeners, abortSignal);
+    return this.tshClient.transferFile(options, abortSignal);
```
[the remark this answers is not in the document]
gzdunek: Totally agree, we should avoid using `tshClient` directly.

packages/teleterm/src/ui/services/fileTransferClient/fileTransferClient.ts
```diff
@@ -0,0 +1,23 @@
+import { FileTransferListeners } from 'shared/components/FileTransfer';
+
+import { TshClient } from 'teleterm/services/tshd/types';
+import { FileTransferRequest } from 'teleterm/services/tshd/v1/service_pb';
+
+export class FileTransferClient {
```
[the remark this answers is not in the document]
gzdunek: Yes, the name comes from the first iteration and doesn't have much sense now, changed.

PR #1226, opened 2022-10-01
Remove database reference when undefined in session start, end for Audit Log

stevenGravy:
Connecting to some DB sessions does not have a db name.  Then you get a undefined database in the log.

Modifies 
image
image

To show as

image

image

When available it will still show as in this postgres

image

### DIFF
--- a/packages/teleport/src/services/audit/makeEvent.ts
+++ b/packages/teleport/src/services/audit/makeEvent.ts
@@ -699,7 +699,9 @@ export const formatters: Formatters = {
     type: 'db.session.start',
     desc: 'Database Session Started',
     format: ({ user, db_service, db_name, db_user }) =>
-      `User [${user}] has connected to database [${db_name}] as [${db_user}] on [${db_service}]`,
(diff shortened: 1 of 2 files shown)


packages/teleport/src/services/audit/makeEvent.ts
```diff
@@ -699,7 +699,9 @@ export const formatters: Formatters = {
     type: 'db.session.start',
     desc: 'Database Session Started',
     format: ({ user, db_service, db_name, db_user }) =>
-      `User [${user}] has connected to database [${db_name}] as [${db_user}] on [${db_service}]`,
+      `User [${user}] has connected ${
+        db_name ? 'to database [' + db_name + '] ' : ''
```
zmb3, line 703:
```suggestion
        db_name ? `to database [${db_name}] ` : ''
```
Here and below.

stevenGravy:
updated @zmb3

zmb3 (approved):

PR #1238, opened 2022-10-06
retryWithRelogin: Enable use outside of document context

ravicious:
webapps.e counterpart: gravitational/webapps.e#427

Update: See this discussion https://github.com/gravitational/webapps.e/pull/427#discussion_r1013864097. `retryWithRelogin` will now check just if the workspace of the given resource is still active.

So far retryWithRelogin has assumed that it's used within a document. This was manifested by the fact that in order to check if the relevant UI is still active when the response comes in, retryWithRelogin accepted originatingDocumentUri to check if that document is still visible on the screen.

In access request PRs, we want to use retryWithRelogin in places outside of the context of a document. The question of whether the relevant UI is still active won't be answered by checking the status of a document but rather through other means.

This commit makes it possible to express that in the code. Instead of taking a document URI as an argument, retryWithRelogin now takes a function which answers said question.

A helper function was added for convenience for places which still use retryWithRelogin within the context of a document.


avatus:
I accidentally merged master into this PR. Had the wrong tab open. Will revert, Sorry!

ravicious:
@avatus Did you use this approach in the end or does `retryWithRelogin` in the context of the access requests use some kind of a dummy doc URI?

ravicious:
Ah, next time there's no need to revert such a merge. ;) I've been there before.

avatus:
@ravicious I need to come back to this one and add in the changes to access requests before we merge it in, but in local testing this will work fine. I'll push up some changes to this branch to include access requests changes as well.

packages/teleterm/src/ui/services/clusters/clustersService.ts
```diff
@@ -413,6 +413,10 @@ export class ClustersService extends ImmutableStore<ClustersServiceState> {
     return cluster.loggedInUser?.assumedRequests || {};
   }
 
+  getAssumedRequest(clusterUri: string, requestId: string) {
```
ravicious, line 416: @avatus Do methods such as `getAssumedRequests`, `getRequestableRoles`, `getAccessRequests` and so on work only with root clusters? If so, it'd be worthwhile to update the argument name to reflect that.

gzdunek (approved):

packages/teleterm/src/ui/services/clusters/clustersService.ts
avatus: I believe these do not rely on root cluster as they either defer to the access request id (from whatever connected auth server) and then role mapping for the others so we should be good here.

avatus (approved):

packages/teleterm/src/ui/services/clusters/clustersService.ts
ravicious: I'm afraid I don't understand, are we good here to change the arg name or do those methods need to work with leaf clusters too, so the arg name should stay as `clusterUri`? 😶

I just looked into this and:

* `getAssumedRequests` does `cluster.loggedInUser?.assumedRequests`. `assumedRequests` is present only for root clusters so here we should definitely change the arg name to reflect that.
* `getRequestableRoles` and `getAccessRequests` reach out to the auth server. So I think it's for root clusters only?

I think the overarching question is: could the user ever perform any operations on access requests on a leaf cluster? If not, then all methods in `ClustersService` that accept `clusterUri` as an arg should have that arg renamed to `rootClusterUri` IMHO.

gzdunek (approved):

ravicious:
Reminder to myself to merge the webapps.e PR first and then update the e ref before merging this PR.

packages/teleterm/src/ui/services/clusters/clustersService.ts
avatus: I think we are saying the same thing but from two different perspectives. I'm saying "it all goes to the root cluster auth server anyway so it doesn't matter what clusterUri is sent (root or leaf)" and you're saying "because it goes to the root cluster auth server, we should name that variable to rootClusterUri so they know". Now that I understand what you're saying, I agree with your perspective more.

PR #1249, opened 2022-10-11
CatchError and Discover related feature extensions

kimlisa:
part of https://github.com/gravitational/teleport/issues/16856

Thought i'd break it so it doesn't look so huge.

#### Description
- Fixed a bug where if you go to the desktop flow, go to the token fetch step, then go to a different route, come back to it, results were cached so it was using the previous token that may or may not be expired or already used
- Moved some shareable files to the `Discovery/Shared/*` directory and extended them to allow for fetching for resources based on service kinds (fyi we are moving away from the term `agents` to `services`, stated on our discovery discussions)
- refactors `CatchError` component to a typescript file and extends it to allow custom fallback UI. It will be used to catch suspense errors later

#### TODO
- [ ] backport to v11
- [ ] backport to v10


packages/teleport/src/Discover/Shared/PingTeleportContext.tsx
```diff
@@ -30,27 +32,41 @@ export function PingTeleportProvider(props: {
...
   );
 
+  function servicesFetchFn(signal: AbortSignal) {
+    const clusterId = ctx.storeUser.getClusterId();
+    const request = {
+      search: `${INTERNAL_RESOURCE_ID_LABEL_KEY} ${joinToken.internalResourceId}`,
+      limit: 1,
+    };
+    switch (props.serviceKind) {
+      case 'node':
```
ryanclark, line 54: Can we make use of the shared `ResourceKind` enum here? https://github.com/gravitational/webapps/blob/5e80f7f88fce2628a88629c103f71da3da229fd0/packages/teleport/src/Discover/Shared/ResourceKind.ts#L17-L23

packages/teleport/src/components/CatchError/index.ts
```diff
@@ -14,5 +14,5 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import CatchError from './CatchError';
-export default CatchError;
+import { CatchError } from './CatchError';
+export { CatchError };
```
ryanclark, line 18:
```suggestion
export { CatchError } from './CatchError';
```

packages/teleport/src/Discover/Shared/JoinTokenContext.tsx
```diff
@@ -60,15 +64,22 @@ interface PromiseResult {
 }
 
 let abortController: AbortController;
-let result: PromiseResult;
+let cachedJoinTokenResult: PromiseResult;
+
+export function clearCachedJoinTokenResult() {
```
ryanclark, line 69:
```suggestion
function clearCachedJoinTokenResult() {
```

I don't think we need to export this

packages/teleport/src/Discover/Shared/Timeout.tsx
```diff
@@ -1,37 +1,37 @@
 import React, { useEffect, useState } from 'react';
-import styled from 'styled-components';
 
 interface TimeoutProps {
   timeout: number; // ms
+  message?: string;
+  // mt is to specify margin top in pixels.
+  mt?: number;
```
ryanclark, line 7: This confused me as you had it set to `100` and I thought that was the regular `mt={100}`, which would've been huge. Maybe we should wrap the usage of `<Timeout />` in a `<Box mt={4}>` or something?

ryanclark: We could probably change `<Timeout />` to return a `<span>` instead then too. It's only really meant to give the text, so having the usage wrapped in a box would be a good approach.

packages/teleport/src/Discover/Shared/JoinTokenContext.tsx
kimlisa: i require it in several places, in storybook with msw (results get cached so i need to clear it between different story states that makes http requests), and when i provide a CatchError, `onRetry` i also need to clear the cache, it looks something like this (in my next PR). 

```ts
    <CatchError
      onRetry={clearCachedJoinTokenResult}
      fallbackFn={props => (
        <Box>
          <Heading />
          <ScriptBox pollState="error">
            <TextIcon mt={2} mb={3}>
              <Icons.Warning ml={1} color="danger" />
              Encountered Error: {props.error.message}
            </TextIcon>
            <ButtonBlueText ml={2} onClick={props.retry}>
              Refetch a command
            </ButtonBlueText>
          </ScriptBox>
        </Box>
      )}
    >
      <Suspense
        fallback={
          <Box>
            <Heading />
            <ScriptBox height="144px">
              <Box textAlign="center" height="108px">
                <Indicator />
              </Box>
            </ScriptBox>
          </Box>
        }
      >
        <DownloadScript {...props} />
      </Suspense>
    </CatchError>
```

packages/teleport/src/Discover/Shared/JoinTokenContext.tsx
```diff
@@ -60,15 +64,22 @@ interface PromiseResult {
...
 
 export function useJoinTokenValue() {
   const tokenContext = useContext(joinTokenContext);
 
   return tokenContext.joinToken;
 }
 
-export function useJoinToken(): {
+export function useJoinToken(
+  joinRole: JoinRole,
```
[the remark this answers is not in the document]
kimlisa: hm, we could, but the `JoinRole` is the actual values that the backend expects. Should we do like use enum here, then on the `services/joinToken.ts` in the fetch definition convert it to the actual value required by backend?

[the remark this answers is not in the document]
ryanclark: I'd have a function helper next to the enum to convert the enum value in the specific format you want

```typescript
export function resourceKindToRole(kind: ResourceKind) {
  switch (kind) {
    case ResourceKind.Server:
      return 'Node';
    // etc
  }
}
```

packages/teleport/src/Discover/Shared/JoinTokenContext.tsx
ryanclark: Ah sorry I missed those usages - I must not have loaded all of the diff

kimlisa: no worries! it's because these usages are going to be on my next PR (it's not in this PR) i wanted to make my PR small

packages/teleport/src/Discover/Desktop/ConnectTeleport/RunConfigureScript.tsx
```diff
@@ -28,7 +28,8 @@ interface RunConfigureScriptProps {
 export function RunConfigureScript(
   props: React.PropsWithChildren<RunConfigureScriptProps>
 ) {
-  const { joinToken, reloadJoinToken, timeout, timedOut } = useJoinToken();
+  const { joinToken, reloadJoinToken, timeout, timedOut } =
+    useJoinToken('WindowsDesktop');
```
ryanclark, line 32:
```suggestion
    useJoinToken(ResourceKind.Desktop);
```

packages/teleport/src/Discover/Shared/JoinTokenContext.tsx
```diff
@@ -91,40 +108,42 @@ export function useJoinToken(): {
...
-        joinToken: result.response,
+        joinToken: cachedJoinTokenResult.response,
         reloadJoinToken: run,
         timedOut: tokenContext.timedOut,
         timeout: tokenContext.timeout,
       };
     }
 
-    throw result.promise;
+    throw cachedJoinTokenResult.promise;
```
JanKaczmarkiewicz, line 146: Just a side note: Suspense for data fetching is still experimental even in the latest v18.2. Our current version is 16.8.4. Personally, I think we should use it only after this API is stabilized.
https://17.reactjs.org/docs/concurrent-mode-suspense.html

packages/teleport/src/components/CatchError/CatchError.tsx
```diff
@@ -43,3 +57,18 @@ export default class CatchError extends React.Component {
...
+};
+
+type State = {
+  error: Error;
+};
+
+type Props = {
+  children: React.ReactNode;
+  onRetry?(): void;
+  fallbackFn?(props: FallbackFnProp): React.ReactNode;
```
JanKaczmarkiewicz, line 73: Do we have a use case where we need fallbackFn, onRetry?

packages/teleport/src/Discover/Shared/JoinTokenContext.tsx
ryanclark: It's already in the codebase as it's working fine for our simplified needs. If this was more complicated and was throwing bugs, I'd remove it - we'll be on React 18 soon, so I'd rather keep it in (it's already in the codebase) than refactor it out

JanKaczmarkiewicz: > Also we're already using Suspense here (albeit with a null fallback, but I'm not entirely sure why) - https://github.com/gravitational/webapps/blob/master/packages/teleport/src/Main/Main.tsx#L88-L90

I think it was only used to support react.lazy components which are fine since suspense for code splitting was added in 16.6 https://github.com/facebook/react/blob/main/CHANGELOG.md#1660-october-23-2018

JanKaczmarkiewicz: I mean my concern here is that this place could break even on a patch react update

packages/teleport/src/components/CatchError/CatchError.tsx
kimlisa: yes sorry the usage is in my next PR but the usage looks like this:

```ts
    <CatchError
      onRetry={clearCachedJoinTokenResult}
      fallbackFn={props => (
        <Box>
          <Heading />
          <ScriptBox pollState="error">
            <TextIcon mt={2} mb={3}>
              <Icons.Warning ml={1} color="danger" />
              Encountered Error: {props.error.message}
            </TextIcon>
            <ButtonBlueText ml={2} onClick={props.retry}>
              Refetch a command
            </ButtonBlueText>
          </ScriptBox>
        </Box>
      )}
    >
      <Suspense
        fallback={
          <Box>
            <Heading />
            <ScriptBox height="144px">
              <Box textAlign="center" height="108px">
                <Indicator />
              </Box>
            </ScriptBox>
          </Box>
        }
      >
        <DownloadScript {...props} />
      </Suspense>
    </CatchError>
```

packages/teleport/src/Discover/Shared/Timeout.tsx
```diff
@@ -1,37 +1,35 @@
 import React, { useEffect, useState } from 'react';
-import styled from 'styled-components';
 
 interface TimeoutProps {
   timeout: number; // ms
+  message?: string;
```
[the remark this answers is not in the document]
kimlisa: yes, it will be used in my next PR (a mistake on my part as it's causing confusion), but i needed a way to send a custom timeout message

packages/teleport/src/Discover/Shared/JoinTokenContext.tsx
kimlisa: i have no strong opinions, but if we are going to the refactoring route, can we do it later? (since it's working fine atm). or when we upgrade react, we will be mindful of the suspense feature. our team isn't that big so i'm pretty positive one of us will remember it (:

JanKaczmarkiewicz: Sure

packages/teleport/src/Discover/Shared/JoinTokenContext.tsx
```diff
@@ -15,11 +21,10 @@ interface JoinTokenContextState {
 const joinTokenContext = React.createContext<JoinTokenContextState>(null);
 
 export function JoinTokenProvider(props: {
-  timeout: number;
+  timeout?: number;
```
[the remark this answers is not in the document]
kimlisa: yes it should be required, good catch

hatched (approved):

packages/teleport/src/Discover/Shared/Timeout.tsx
```diff
@@ -1,37 +1,35 @@
...
 
   return (
-    <Container>
-      This script is valid for another {minutes}:{seconds}.
-    </Container>
+    <span>
+      {message
+        ? `${message} ${minutes}:${seconds}`
+        : `This script is valid for another ${minutes}:${seconds}`}
+    </span>
```
ryanclark, line 32: What do you think about adding
```typescript
message = message || 'This script is valid for another';
```

and making this 

```suggestion
    <span>
      {message} {minutes}:{seconds}
    </span>
```

packages/teleport/src/components/CatchError/CatchError.tsx
```diff
@@ -20,21 +20,34 @@ import Logger from 'shared/libs/logger';
...
 
-export default class CatchError extends React.Component {
+export class CatchError extends React.PureComponent<Props, State> {
+  state: State = { error: null };
+
+  private retry = () => {
+    this.setState({ error: null });
+    if (this.props.onRetry) {
+      this.props.onRetry();
+    }
```
JanKaczmarkiewicz, line 30: [nit]
```suggestion
    this.props.onRetry?.();
```

kimlisa: woah! i like it, thanks!

packages/teleport/src/Discover/Shared/Timeout.tsx
kimlisa: much better, thanks

PR #1251, opened 2022-10-11
caches latest clipboard text to avoid unnecessary clipboard syncs

ibeckermayer:


packages/teleport/src/DesktopSession/useTdpClientCanvas.tsx
```diff
@@ -83,6 +84,7 @@ export default function useTdpClientCanvas(props: Props) {
   const onClipboardData = (clipboardData: ClipboardData) => {
     if (clipboardSharingEnabled && document.hasFocus() && clipboardData.data) {
       navigator.clipboard.writeText(clipboardData.data);
+      latestClipboardText.current = clipboardData.data;
```
zmb3, line 87: Would computing a checksum make more sense here? I'm thinking about copying large amounts of data - this requires the browser to hold an entire copy of the data in memory just to tell if the data changes in the future.

ibeckermayer: Right, ok I added a Sha256 digest 46abb46df277209401fc36f69fbea3f11b4ed8fd

packages/teleport/src/lib/util.ts
```diff
@@ -32,3 +32,15 @@ export function pluralize(num: number, word: string) {
 
   return word;
 }
+
+// Adapted from https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest#converting_a_digest_to_a_hex_string
+export async function Sha256Digest(
```
hatched, line 37: Could you add a test for this plz.

hatched (approved):

packages/teleport/src/lib/util.ts
ibeckermayer: It's difficult because `window.crypto` is not currently supported by `jsdom`. I tried the node and external library suggested in [this thread](https://github.com/jsdom/jsdom/issues/1612), however neither solution had `crypto.subtle.digest`

e.g. I created a `packages/teleport/src/lib/util.test.ts` and tried both of
```typescript
import { Sha256Digest } from './util';

const crypto = require('crypto');
// eslint-disable-next-line jest/require-hook
Object.defineProperty(global.self, 'crypto', {
  value: Object.setPrototypeOf({ subtle: crypto.subtle }, crypto),
});

// This is needed for tests until jsdom adds support for TextEncoder (https://github.com/jsdom/jsdom/issues/2524)
const { TextEncoder: TestTextEncoder } = require('util');
window.TextEncoder = window.TextEncoder || TestTextEncoder;

test('the Sha256Digest function does something sensible', () => {
  let digest = Sha256Digest('digest me');
  expect(digest).toBe('d');
});

```

```typescript
import { Crypto } from '@peculiar/webcrypto';

import { Sha256Digest } from './util';

global.crypto = new Crypto();

const crypto = require('crypto');
// eslint-disable-next-line jest/require-hook
Object.defineProperty(global.self, 'crypto', {
  value: Object.setPrototypeOf({ subtle: crypto.subtle }, crypto),
});

// This is needed for tests until jsdom adds support for TextEncoder (https://github.com/jsdom/jsdom/issues/2524)
const { TextEncoder: TestTextEncoder } = require('util');
window.TextEncoder = window.TextEncoder || TestTextEncoder;

test('the Sha256Digest function does something sensible', () => {
  let digest = Sha256Digest('digest me');
  expect(digest).toBe('d');
});

```

PR #1254, opened 2022-10-12
Do not keep assumed roles in `app_state.json`

gzdunek:
[[webapps.e](https://github.com/gravitational/webapps.e/pull/408/files)]
This PR removes assumed requests from `app_state.json`, we shouldn't keep them there, because it means duplicating `tsh` state. 

I added `assumedRolesRequests` to the current `LoggedInUser` type, I treat it as a temporary solution. In the future we'll only be able to assume 1 access request at a time, it will be a good opportunity to change the backend to just return `cluster.loggedInUser.assumedRequest`(containing `id`, `expires` and `roles`) instead of `cluster.loggedInUser.activeRequestsList` (which is a list containing only IDs).

### DIFF
--- a/packages/teleterm/src/ui/services/workspacesService/workspacesService.ts
+++ b/packages/teleterm/src/ui/services/workspacesService/workspacesService.ts
@@ -1,19 +1,18 @@
 import { useStore } from 'shared/libs/stores';
 
 import { isEqual } from 'lodash';
-import isPast from 'date-fns/isPast';
-
-import { AccessRequest } from 'e-teleport/services/workflow';
 
 import { ResourceKind } from 'e-teleport/Workflow/NewRequest/useNewRequest';
 
 import { ModalsService } from 'teleterm/ui/services/modals';
 import { ClustersService } from 'teleterm/ui/services/clusters';
-import { StatePersistenceService } from 'teleterm/ui/services/statePersistence';
+import {
+  StatePersistenceService,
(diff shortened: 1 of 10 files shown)


packages/teleterm/src/ui/services/workspacesService/workspacesService.ts
```diff
@@ -289,22 +279,6 @@ export class WorkspacesService extends ImmutableStore<WorkspacesState> {
...
-    requests.forEach(request => {
-      const expired = isPast(new Date(request.expires));
-      // only add any assumed that still exist on the loggedInUser cert
-      if (user.activeRequestsList.includes(request.id) && !expired) {
-        validRequests[request.id] = request;
-      }
-    });
-    return validRequests;
-  }
-
```
gzdunek, line 307: Can I safely remove it? It seems to me that `tshd` shouldn't return expired requests.

avatus (approved):
This works great and I think it's a much better solution than the app_state way. Letting the certificate do as much work for us as we can is ideal. 

As far as the shape of `loggedInUser` in the future, it may be slightly different depending on how we depend to implement it, but this is a fine short term solution for now. I'll create an issue to track this.

gzdunek:
>small non-blocking question, but what do you think about changing assumedRolesRequests to just assumedRequests

Changed, I was unsure about the naming, good that you caught this :)

PR #1256, opened 2022-10-13
Stop `FeatureBox` from adding a scrollbar

ryanclark:
`FeatureBox` allowed it's content to scroll, causing a permanent scrollbar visible on Safari

image

`FeatureBox` shouldn't be able to scroll - it's up to the `HorizontalSplit` to scroll, with content that's controlled via the `ContentMinWidth` direct child.

By removing this overflow, it now displays correctly

#### Normal screen size

image

#### Small screen size

image

#### Normal screen size

image

#### Small screen size

image

Fixes https://github.com/gravitational/teleport/issues/17227

This will need backporting to v11 & v10


zmb3 (approved):

ibeckermayer (commented):
It looks like we can just nuke the unused `packages/design/src/Layout/Layout.jsx`, which we may as well do here. [(See discussion in Slack for reference).](https://gravitational.slack.com/archives/C02DQ1C2BMW/p1665684709628999)

gzdunek (approved):

Issue #1264, opened 2022-10-13
Expand width of share link modal


hatched:
The url wraps, we should increase the modals width by a bit to accommodate this.

<img width="595" alt="Screen Shot 2022-10-13 at 2 09 50 PM" src="https://user-images.githubusercontent.com/532033/195699536-d2751151-7e0a-44c1-8bd0-904bdad478e4.png">

PR #1271, opened 2022-10-14
Increase the maximum width of the Share Link dialog.

hatched:
Fixes https://github.com/gravitational/webapps/issues/1264

Before:
Screen Shot 2022-10-13 at 2 09 50 PM

After:
Screen Shot 2022-10-14 at 9 57 50 AM


kimlisa (approved):

kimlisa:
actually, this will depend on the cluster URL, i don't think its worth changing

PR #1273, opened 2022-10-17
[Connect] Conditionally render Access Request navigation menu

avatus:
Relies on https://github.com/gravitational/teleport/pull/17497

Right now, this menu will show on any cluster. This could lead to bad UX if a user thinks they have access to creating requests just to receive a permission error when they try to submit a new request. This change will not display the menu if the active root cluster does not have the access request feature enabled. Only one file was changed, `NavigationMenu`, the rest are protobufs from the linked PR.


packages/teleterm/src/ui/services/clusters/clustersService.ts
```diff
@@ -72,7 +72,6 @@ export class ClustersService extends ImmutableStore<ClustersServiceState> {
   async logout(clusterUri: string) {
     // TODO(gzdunek): logout and removeCluster should be combined into a single acton in tshd
     await this.client.logout(clusterUri);
-    await this.syncClusterInfo(clusterUri);
```
avatus, line 75: Why would we need to try and `syncClusterInfo` _after_ logging out? I removed it because now that `GetCluster` has a `addMetadataToRetryableError`, logging out would fail here. "no SSH auth methods found. are you logged in?". Of course we aren't we just logged out the line above!  I think this line might have just been an error from the start (git blame says back in febuary) and it wasn't never noticed since it was essentially ignoring the error.

avatus:
I forgot to request review from you both @ravicious and @gzdunek yesterday when I posted the backend portion of this. This might have given a better context of my backend change.

gzdunek (approved):
Looks good!

packages/teleterm/src/ui/services/clusters/clustersService.ts
ravicious: Just for posterity, I'm linking my comment from another PR https://github.com/gravitational/teleport/pull/17497#discussion_r998037506 which explains that this is probably a remnant of times when logging out and removing a cluster from the list were two separate actions.

packages/teleterm/src/ui/TopBar/NavigationMenu/NavigationMenu.tsx
```diff
@@ -59,7 +59,10 @@ export function NavigationMenu() {
   const [isPopoverOpened, setIsPopoverOpened] = useState(false);
   const selectorRef = useRef<HTMLButtonElement>();
 
-  if (!activeRootCluster) {
+  if (
```
ravicious, line 62: Would it make sense to put the access requests menu to the left of profile selector? Otherwise the profile selector switches position when switching between a cluster that has access requests and one that doesn't.

packages/teleterm/src/ui/TopBar/NavigationMenu/NavigationMenu.tsx
```diff
@@ -59,7 +59,10 @@ export function NavigationMenu() {
...
+    !activeRootCluster ||
+    !activeRootCluster.features.advancedAccessWorkflows
```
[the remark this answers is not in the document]
avatus: I think this is bad design. I've updated the backend to only return the `features` on the relevant rpcs (such as `GetCluster`). Although I think in the future we may want to rename `ListRootClusters` to something like `ListRootClustersInProfile` since that more clearly describes what it does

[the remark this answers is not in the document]
ravicious: > Although I think in the future we may want to rename `ListRootClusters` to something like `ListRootClustersInProfile` since that more clearly describes what it does

Another idea for the future would be to utilize [the render view pattern](https://cloud.google.com/apis/design/design_patterns#resource_view) that Alan mentioned. But yeah, we're free to completely change the RPCs around getting root cluster details, there's nothing holding us back. They work the way they work right now because we've never had to do more than just read stuff from the disk.

PR #1275, opened 2022-10-18
Use sessionStorage for Authentication Bearer Token

rudream:
## Purpose

This PR resolves https://github.com/gravitational/teleport-private/issues/158

Resolves a security concern due to using localStorage to store the session token. This is how it was initially but was changed to use localStorage in [this PR](https://github.com/gravitational/webapps/pull/215) due to a Firefox bug involving access request waiting rooms, this bug has been addressed in the `e` counterpart to this PR. 

TODO:
- [x] Fix failing unit tests

### DIFF
--- a/packages/build/package.json
+++ b/packages/build/package.json
@@ -56,6 +56,7 @@
     "@typescript-eslint/parser": "^5.4.0",
     "babel-loader": "^8.2.5",
     "babel-plugin-lodash": "^3.3.4",
+    "broadcast-channel": "4.18.1",
     "clean-webpack-plugin": "4.0.0",
     "core-js": "^3",
     "cross-env": "5.0.5",
--- a/packages/teleport/src/Console/Console.tsx
+++ b/packages/teleport/src/Console/Console.tsx
@@ -18,6 +18,7 @@ import React from 'react';
 import styled from 'styled-components';
 import { Flex } from 'design';
 
+import useWebSession from 'teleport/useWebSession';
 import AjaxPoller from 'teleport/components/AjaxPoller';
 
 import { useConsoleContext, useStoreDocs } from './consoleContextProvider';
@@ -37,7 +38,8 @@ const POLL_INTERVAL = 5000; // every 5 sec
 
 export default function Console() {
   const consoleCtx = useConsoleContext();
-  const { verifyAndConfirm } = useOnExitConfirmation(consoleCtx);
+  const webSession = useWebSession();
+  const { verifyAndConfirm } = useOnExitConfirmation(consoleCtx, webSession);
   const { clusterId, activeDocId } = useTabRouting(consoleCtx);
 
   const storeDocs = consoleCtx.storeDocs;
@@ -68,7 +70,7 @@ export default function Console() {
   }
 
   function onLogout() {
-    consoleCtx.logout();
+    webSession.logout();
   }
 
   const disableNewTab = storeDocs.getNodeDocuments().length > 0;
--- a/packages/teleport/src/Console/useOnExitConfirmation.test.ts
+++ b/packages/teleport/src/Console/useOnExitConfirmation.test.ts
@@ -16,14 +16,18 @@
 
 import renderHook from 'design/utils/renderHook';
 
-import session from 'teleport/services/websession';
+import { getMockWebSession } from 'teleport/services/websession/test-utils';
 
 import ConsoleContext from './consoleContext';
 import useOnExitConfirmation from './useOnExitConfirmation';
 
 test('confirmation dialog before terminating an active ssh session', () => {
   const ctx = new ConsoleContext();
-  const { current } = renderHook(() => useOnExitConfirmation(ctx));
+
+  const mockWebSession = getMockWebSession();
+  const { current } = renderHook(() =>
+    useOnExitConfirmation(ctx, mockWebSession)
+  );
   const event = new Event('beforeunload');
 
   // two prompts that can be called before closing session/window
@@ -79,7 +83,7 @@ test('confirmation dialog before terminating an active ssh session', () => {
   docTerminal.created = new Date('2019-04-01');
 
   // test that expired session does not prompt
-  jest.spyOn(session, '_timeLeft').mockReturnValue(0);
+  jest.spyOn(mockWebSession, '_timeLeft').mockReturnValue(0);
   window.dispatchEvent(event);
   expect(event.preventDefault).not.toHaveBeenCalled();
 
@@ -90,7 +94,7 @@ test('confirmation dialog before terminating an active ssh session', () => {
   expect(window.confirm).toHaveReturnedWith(false);
 
   // test aged terminal doc triggers prompt
-  jest.spyOn(session, '_timeLeft').mockReturnValue(5);
+  jest.spyOn(mockWebSession, '_timeLeft').mockReturnValue(5);
   window.dispatchEvent(event);
   expect(event.preventDefault).toHaveBeenCalledTimes(1);
 });
--- a/packages/teleport/src/WebSessionContext.tsx
+++ b/packages/teleport/src/WebSessionContext.tsx
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 Gravitational, Inc.
+Copyright 2022 Gravitational, Inc.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -14,6 +14,20 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-export const KeysEnum = {
-  DISCOVER: 'grv_teleport_discover',
+import React from 'react';
+
+import { WebSession } from './services/websession';
+
+export const SessionContext = React.createContext<WebSession>(null);
+
+export const SessionContextProvider: React.FC<
+  React.PropsWithChildren<Props>
+> = props => {
+  return (
+    <SessionContext.Provider value={props.session} children={props.children} />
+  );
+};
+
+type Props = {
+  session: WebSession;
 };
--- a/packages/teleport/src/components/Authenticated/Authenticated.tsx
+++ b/packages/teleport/src/components/Authenticated/Authenticated.tsx
@@ -18,9 +18,10 @@ import React from 'react';
 import { throttle } from 'lodash';
 import Logger from 'shared/libs/logger';
 
-import session from 'teleport/services/websession';
+import { WebSession } from 'teleport/services/websession';
 import history from 'teleport/services/history';
-import localStorage from 'teleport/services/localStorage';
+
+import { SessionContextProvider } from 'teleport/WebSessionContext';
 
 const logger = Logger.create('/components/Authenticated');
 const ACTIVITY_CHECKER_INTERVAL_MS = 30 * 1000;
@@ -38,7 +39,18 @@ const events = [
   'pointerdown',
 ];
 
-const Authenticated: React.FC = ({ children }) => {
+const Container: React.FC<React.PropsWithChildren<unknown>> = ({
+  children,
+}) => {
+  const session = new WebSession();
+
+  return <Authenticated session={session}>{children}</Authenticated>;
+};
+
+const Authenticated: React.FC<{
+  session: WebSession;
+  children: React.ReactNode;
+}> = ({ session, children }) => {
   React.useEffect(() => {
     if (!session.isValid()) {
       logger.warn('invalid session');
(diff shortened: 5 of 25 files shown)


alex-kovoy:
> Opening multiple tabs/windows with the same URL creates sessionStorage for each tab/window.

It will trigger automatic logout in existing tabs when opening an URL in the new tab.

[More here](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage)

rudream:
> It will trigger automatic logout in existing tabs when opening an URL in the new tab.

Nice to see you @alex-kovoy 🙂

The behaviour you described isn't occurring, the bearer token is being obtained on page load and placed in each tab/window's `sessionStorage` individually.

alex-kovoy:
There is a logic that refreshes a token every 9 min and updates it for all open tabs (thus `localStorage`). With sessionStorage, that logic needs to be revisited as 2 tabs (each with an instance of sessionStorage) would attempt to refresh a token at the same time. This might be OK though because we do not automatically invalidate `used` tokens (i believe) because of 10min TTL but nevertheless this is something we want to avoid.

Ideally it would be better to refactor websession file by moving the token-refresh code to a shared web-worker (or something along these lines).

alex-kovoy:
> Resolves a security concern due to using localStorage to store the session token

Aside from _general_ recommendations, it's totally safe to use localStorage as long as we know what we do :) If this is the main reason for the change then i would close is as by design.

packages/teleport/src/services/websession/storeWebSession.ts
```diff
@@ -0,0 +1,163 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import history from 'teleport/services/history';
+
+import {
```
hatched, line 19:
```suggestion
import type {
```

packages/teleport/src/services/websession/storeWebSession.ts
```diff
@@ -0,0 +1,163 @@
...
+  KeysEnum,
+  EventKeys,
+  BroadcastChannelMessage,
+  BearerToken,
+} from './types';
+
+const isTest = process.env.NODE_ENV === 'test';
```
hatched, line 26: Generally you want to try and void adding test specific code, especially toggling on a 'test environment' as any resulting tests will not be testing the true execution of the code.

I also don't see a test file for this?

packages/teleport/src/services/websession/storeWebSession.ts
```diff
@@ -0,0 +1,163 @@
...
+
+export class StoreWebSession {
+  bcBroadcaster: BroadcastChannel;
+  bcReceiver: BroadcastChannel;
+
+  constructor() {
+    this.initBroadcastChannel();
```
hatched, line 33: In interest of removing the test specific code, the constructor can be expanded to take a configuration object which would contain the broadcast channel.

packages/teleport/src/services/websession/storeWebSession.ts
```diff
@@ -0,0 +1,163 @@
...
+    const bearerToken = getBearerToken();
+    const time = Number(bearerToken.sessionInactiveTimeout);
+    return time ? time : 0;
+  }
+
+  initBroadcastChannel() {
+    this.bcBroadcaster = new BroadcastChannel('websession_store');
+    this.bcReceiver = new BroadcastChannel('websession_store');
+
+    this.bcReceiver.onmessage = ({
```
hatched, line 83: To support passing in a custom configuration this can be moved to a static method `attachBroadcastHandlers`. So  in a test this class would be called using something like...

```ts
const broadcastChannel = new BroadcastChannel('websession_store');
StoreWebSession.attachBroadcastHandlers(broadcastChannel)
const webSession = new StoreWebSession({broadcastChannel});
// setup initial messages for test
webSession.bcBroadcaster.postMessage({
  key: EventKeys.GET_TOKEN,
});
```

packages/teleport/src/services/websession/storeWebSession.ts
```diff
@@ -0,0 +1,163 @@
...
+      });
+      this.bcBroadcaster.postMessage({
+        key: EventKeys.GET_LAST_ACTIVE,
+      });
+    }
+  }
+}
+
+export function getBearerToken(): BearerToken {
+  return JSON.parse(window.sessionStorage.getItem(KeysEnum.TOKEN));
```
hatched, line 162: As this is grabbing data from the user space this can throw and because TS doesn't give you a way to indicate that something throws it's best to handle such an eventuality here.

packages/teleport/src/services/localStorage/localStorage.ts
```diff
@@ -24,47 +23,6 @@ const storage = {
     window.localStorage.clear();
   },
 
-  subscribe(fn) {
```
hatched, line 27: With all this removed I question the usefulness of this file for getting and setting a single key. 

It looks like we interact with `localStorage` directly in most places, in interest of consistency throughout the application those should maybe be all moved into this file or this file removed. But that's probably out of scope of this PR, thoughts on removing this in favour of an update?

packages/teleport/src/Console/Console.tsx
```diff
@@ -37,7 +38,8 @@ const POLL_INTERVAL = 5000; // every 5 sec
 
 export default function Console() {
   const consoleCtx = useConsoleContext();
-  const { verifyAndConfirm } = useOnExitConfirmation(consoleCtx);
+  const webSession = useWebSession();
```
hatched, line 41: For consistency sake, it seems that for components with their own `use<component>Context` functions the session logout is handled there instead of in the component. See `useDiscover`, `useTopBar`, and `Player` below.

hatched (commented):
Thanks for this @rudream

packages/teleport/src/services/websession/storeWebSession.ts
ryanclark: A static method wouldn't do anything to `webSession` in this context, I don't think `attachBroadcastHandlers` is needed - I agree that it should be accepted in the constructor for mocking during tests, however.

packages/teleport/src/Console/Console.tsx
ryanclark: I'd argue for using `useWebSession` _more_ in components than hiding it in these magic single hook methods we've created for the state management of entire components and their subtrees.

`useDiscover` shouldn't know anything about logging the user out - why would it? It just needs to know about discover and anything discover related. The component that actually displays the logout button should be the one calling `useWebSession`.

Our abstraction of everything into single hooks is confusing and bloated, and causes inefficient re-renders of our UI.

packages/teleport/src/services/websession/websession.ts
```diff
@@ -30,28 +31,28 @@ const RENEW_TOKEN_TIME = 180 * 1000;
 const TOKEN_CHECKER_INTERVAL = 15 * 1000; //  every 15 sec
 const logger = Logger.create('services/session');
 
-let sesstionCheckerTimerId = null;
+let sessionCheckerTimerId = null;
+
+export default class WebSession extends StoreWebSession {
```
ryanclark, line 36:
```suggestion
export class WebSession extends StoreWebSession {
```

packages/teleport/src/services/websession/index.ts
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import websession from './websession';
+import WebSession from './websession';
 
 export * from './types';
-export default websession;
+export default WebSession;
```
ryanclark, line 20:
```typescript
export { WebSession } from './websession';
```

packages/teleport/src/services/websession/storeWebSession.ts
ryanclark: Agreed - we're introducing side effects where things run differently in our test environment vs web app

packages/teleport/src/services/websession/storeWebSession.ts
hatched: Oops sorry I had a typo in my pseudo code :D (updated). Yeah it's entirely possible that it's not required. I just wanted to illustrate a way that similar handlers could be added outside of the class and passed in. Lots of different ways to accomplish this.

packages/teleport/src/Console/Console.tsx
hatched: I'm happy with what @ryanclark has said....my concern was just one of consistency.

jentfoo:
Hell @rudream, thank you for your work on this issue so far.  This was reported by Visa and so we want to make sure we get this fixed.  Is there anything I can do to help get this across the line?  Do you have any ETA for when this might be released?  Thank you!

rudream:
Hi @jentfoo, this is on track to be released next week.

rudream:
The `broadcast-channel` yarn package has been added in order to allow us to mock it in tests. That `BroadcastChannel` module's type differs slightly from the native `BroadcastChannel` type, hence this type conversion had to be done to allow us to continue using the native `BroadcastChannel` in the app while being able to use the yarn package in unit tests:

https://github.com/gravitational/webapps/blob/b8186241143d911cc2c1e328ca42d1bb5beef967/packages/teleport/src/Discover/Discover.test.tsx#L87-L93

packages/build/package.json
```diff
@@ -95,4 +96,4 @@
     "webpack-cli": "^4.10.0",
     "webpack-dev-server": "^4.9.3"
   }
-}
+}
```
ryanclark, line 99: nit: can we keep the last empty line please

packages/teleport/src/WebSessionContext.tsx
```diff
@@ -0,0 +1,31 @@
...
+limitations under the License.
+*/
+
+import React from 'react';
+
+import { WebSession } from './services/websession';
+
+export const SessionContext = React.createContext<WebSession>(null);
+
+export const SessionContextProvider: React.FC<Props> = props => {
```
ryanclark, line 23:
```suggestion
export const SessionContextProvider: React.FC<React.PropsWithChildren<Props>> = props => {
```

packages/teleport/src/useWebSession.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import { SessionContext } from './WebSessionContext';
+
+export default function useWebSession() {
+  const sessionContext = React.useContext(SessionContext);
+  if (!sessionContext) {
+    throw new Error('Unable to retrieve WebSession Context');
+  }
+
+  return (window['sessionContext'] = sessionContext);
```
ryanclark, line 27: Does this need to be on the window?

packages/teleport/src/Console/useOnExitConfirmation.test.ts
```diff
@@ -13,17 +13,29 @@
...
+import { WebSession } from 'teleport/services/websession';
 
 import ConsoleContext from './consoleContext';
 import useOnExitConfirmation from './useOnExitConfirmation';
 
 test('confirmation dialog before terminating an active ssh session', () => {
   const ctx = new ConsoleContext();
-  const { current } = renderHook(() => useOnExitConfirmation(ctx));
+
+  jest.mock('broadcast-channel');
```
hatched, line 28: Can you move this mock into a shared test util as the same mock needs to happen across multiple files. That will also be a good spot to put the comment as to why this is necessary.

packages/teleport/src/useWebSession.ts
rudream: Good catch. I was using this to make it easier to debug from the console and forgot to remove it.

packages/teleport/src/components/Authenticated/Authenticated.tsx
```diff
@@ -38,7 +39,16 @@ const events = [
   'pointerdown',
 ];
 
-const Authenticated: React.FC = ({ children }) => {
+const Container: React.FC = ({ children }) => {
```
ryanclark, line 42:
```suggestion
const Container: React.FC<React.PropsWithChildren<unknown>> = ({ children }) => {
```

packages/teleport/src/components/Authenticated/Authenticated.tsx
```diff
@@ -107,7 +121,7 @@ function startActivityChecker(ttl = 0) {
   return stop;
 }
 
-function isInactive(ttl = 0) {
-  const lastActive = localStorage.getLastActive();
+function isInactive(ttl = 0, getLastActive: () => number) {
```
ryanclark, line 124: Can `getLastActive` be `lastActive: number` instead? As it's invoked immedIately I don't think it'll be stale by the time this function compares the values

ryanclark (approved):
LGTM, two minor comments

hatched (approved):
Thanks for keeping up on these changes!

zmb3 (approved):

PR #1277, opened 2022-10-19
Code split routes, reduce the initial bundle size + loading times 

ryanclark:
This adds the ability to run Webpack's bundle analyzer via setting the environment variable `WEBPACK_ANALYZE_BUNDLE` to `yes`. This runs a separate process with an interactive tree map that shows exactly what files are being included in what bundle that Webpack outputs.

This also code splits out all our routes and features, so the user doesn't download unnecessary JS code on first launch, and will only download what's needed for each route.

I've also removed the maximum size for a font before it's put into the JS bundle - it was adding nearly 400kb to the main `app.js` bundle.

Here's some results on the page load, JS bundle size and lighthouse performance after each change. (note these are all production builds)

|                                    | App JS Size | Transferred | Finish | DOMContentLoaded | Loaded | Lighthouse Performance |
|:----------------------------------:|:-----------:|:-----------:|:------:|:----------------:|--------|:------------------------:|
| Initial                            |    624kb    |    2.2mb    |  382ms |       270ms      |  281ms |           88           |
| Code split private routes          |    561kb    |    2.0mb    |  336ms |       224ms      |  235ms |           90           |
| Code split private + public routes |    558kb    |    2.0mb    |  332ms |       221ms      |  231ms |           90           |
| Code split all the routes          |    520kb    |    1.9mb    |  311ms |       196ms      |  206ms |           92           |
| Code split all the routes, remove fonts from JS bundle        |    151kb    |    1.5mb    |  290ms |       179ms      |  189ms |           94           |

Before this change, we had two bundles being served to the user - `vendor` and `app`. 

image

This means the user would get all the `node_modules` on initial load, and all of application code.

We also were including all of `lodash.js`, which is huge. This adds `lodash-webpack-plugin` and `babel-lodash-plugin` to create smaller bundles and only include the needed parts of lodash.

After this change, the output looks like:

image

Which has `node_modules` distributed into the relevant bundles, that are only loaded when needed (navigating to a route for example).

This also changes Webpack's `optimization` configuration to only happen in prod builds (speeds up development cycles) and have a more fine tuned config for production builds.

We still have further code splitting to do after this, as we're still producing bundles bigger than the recommended.

image


hatched:
So it looks like pulling the fonts out and tree-shaking lodash was a significant savings 👍 Lets definitely keep that! 

However for our use case I don't think we should be splitting the route assets. Our application is typically loaded once, then cached in perpetuity until a new version is released. This means that we're not as sensitive to the very first load time vs a website where that all mighty initial load time is important. This means that to save 100ms on the initial load the user will potentially run into a UX issue where navigating between routes within the application could hang while it's pulling in the assets for that route.

ryanclark:
Code splitting routes means we're not loading all our node dependencies at once. We could probably chill on the amount of code split routes if it matters, but we don't want to be loading `xterm.js` for example, unless the user needs to create an SSH session for example.

There's effort to reduce the initial latency for SSH connections in the Web UI alongside the efforts for `tsh`, so creating small minimal bundles (that, like the app and vendor bundles we have at the moment, are cached after having loaded once) is better for page load performance. Even if it's just shaving off `x` milliseconds now, it avoids the page load from getting slower and slower as the codebase grows and more external modules are pulled in.

It's more likely for an API call to fail or take too long than an asset failing to load from the proxy, so I'm not concerned about that. We could probably use something a bit better than `null` for the `Suspense` fallback, but I was just copying what we use in our current `Suspense` usage when rendering the features.

hatched:
I expect that making this change will actually increase the latency for the first ssh connection in the UI as we will be shifting the loading of the assets (like xterm) to the first load of the ssh UI instead of on the first load of the app. 

The browser will have less JS to parse though which may decrease the TTI when going directly to a ssh window. I'd be interested to know that once the cache is primed if this decreases the time to `Loaded` on a direct-to-ssh pageview.

ryanclark:
Here's a comparison of before/after code splitting for a new SSH session

|        | Transferred | Finish | DOMContentLoaded | Loaded | Lighthouse Performance |
|--------|:-----------:|:--------:|:------------------:|:--------:|:------------------------:|
| Before |    1.8mb    | 332ms  |       263ms      | 280ms  |           92           |
| After  |    1.3mb    | 287ms  |       172ms      | 181ms  |           94           |

alex-kovoy:
(my 5cents) I would suggest to have code splitting for _public_ routes such as login and for the rest it seems nice to have.

One thing to check is `fast-refresh` if it still works as expected (like it should not _always_ trigger page reloads even for minor JSX changes and instead update on-the-fly). Previously lazy-loading used to break old `hot-reloads`.

But generally, i suggest to favor code simplicity over minor optimizations if possible.

ryanclark:
Yeah hot reloading still works fine with code splitting, no full page reloads when editing a file that's been split out from the main bundle

hatched:
It looks like the data has proven me wrong 😁 shipit!

hatched (approved):

PR #1278, opened 2022-10-19
Change the application access authentication flow

ryanclark:
This is the frontend changes for this PR - https://github.com/gravitational/teleport/pull/17592

This sets a custom `X-Cookie-Value` header on the request to `/x-teleport-auth` instead of putting it in the `POST` body.

We also remove a lot of the redirects (and the code that goes with it) as it's no longer required with the new authentication flow in the Teleport PR.


ikkisoft:
LGMT. Comments left in [#17592](https://github.com/gravitational/teleport/pull/17592)

packages/teleport/src/AppLauncher/AppLauncher.tsx
```diff
@@ -14,25 +14,77 @@ See the License for the specific language governing permissions and
...
+        method: 'POST',
+        credentials: 'include',
+        headers: {
+          'X-Cookie-Value': session.cookieValue,
+          'X-Subject-Cookie-Value': session.subjectCookieValue,
+        },
+      });
+
+      window.location.replace(`https://${fqdn}${port}`);
+    } catch (err) {
```
kimlisa, line 57:
```suggestion
    } catch ((err: Error)) {
```

kimlisa (approved):
> We also remove a lot of the redirects (and the code that goes with it) as it's no longer required with the new authentication flow in the Teleport PR.

that means like url params like `awsrole` and `state` right? can we also delete `useAppLaunder` file since it doesn't look like it's being used. also does the backend PR delete the code that does the redirecting? eg `redirectToLauncher`, before I saw the above comment, i thought we missed these params

packages/teleport/src/AppLauncher/AppLauncher.tsx
ryanclark: It's never guaranteed what is being thrown is of type `Error` -
image

kimlisa: ah ok, then should we do this (create a helper func perhaps): https://github.com/gravitational/webapps/blob/5458f66bccbf7e3df879e98e7ee8096bb0f7665b/packages/teleport/src/Discover/Database/CreateDatabase/useCreateDatabase.ts#L188

ryanclark:
> that means like url params like awsrole and state right? can we also delete useAppLaunder file since it doesn't look like it's being used. also does the backend PR delete the code that does the redirecting? eg redirectToLauncher, before I saw the above comment, i thought we missed these params

`state` is no longer needed - the backend changes remove the `?state=` dance

`arn` is preserved and passed through to `createAppSession` if it's present

image

I forgot about `?path=` and have updated the PR to redirect to the given path if it's in the query params - nice catch!

packages/teleport/src/AppLauncher/AppLauncher.tsx
ryanclark: I don't love that approach as it's assuming if we don't get an `Error`, that the thing thrown can be converted to a string. If it's an object that's thrown, that would put `[object Object]` into the message.

I've updated this catch to check for an instance of an error and use the error message provided if so. If not, it'll just fallback to a generic "Something went wrong"

hatched (approved):

PR #1297, opened 2022-10-25
Warn user when desktop is active

ibeckermayer:
Follow on from @xacrimon's work in https://github.com/gravitational/webapps/pull/1065

Corresponds with https://github.com/gravitational/teleport/pull/17770

When a user attempts to start a desktop session for which another user already has an active session, they will be shown a warning dialog asking them if they'd like to continue

image

This is in response to Dyonsec review that recognized that without this, a user could be tricked into clicking a link that would DOS another user's active session.

**NOTE TO SELF**: Do not merge/update-webassets for this until https://github.com/gravitational/teleport/pull/17770 is merged, or else clients will break. Don't forget this on backports as well.


packages/teleport/src/DesktopSession/DesktopSession.tsx
```diff
@@ -162,6 +171,43 @@ export function DesktopSession(props: State) {
...
+  if (showAnotherSessionActiveDialog) {
+    return (
+      <Session {...props}>
+        <Dialog
+          dialogCss={() => ({ width: '484px' })}
+          onClose={() => {}}
+          open={true}
+        >
+          <DialogHeader style={{ flexDirection: 'column' }}>
+            Another Session Is Active
```
kimlisa, line 183:
```suggestion
            <DialogTitle>Another Session Is Active</DialogTitle>
```
it will give more emphasis, and make it consistent with our other dialogs

kimlisa (approved):

hatched (approved):

ryanclark:
What's your thoughts on aligning the buttons to the right?

xacrimon (approved):

ibeckermayer:
@kimlisa @hatched 

Mind giving this one more run through before I merge? I had to change things a bit to get everything to work with MFA. The primary changes since you reviewed are in `packages/teleport/src/components/TdpClientCanvas/TdpClientCanvas.tsx` and the related changes to `packages/teleport/src/DesktopSession/DesktopSession.tsx`.

packages/teleport/src/DesktopSession/DesktopSession.tsx
```diff
@@ -173,31 +232,32 @@ export function DesktopSession(props: State) {
   }
 
   if (processing) {
+    // We don't know whether another session for this desktop is active while the
+    // fetchAttempt is still processing, so hold off on starting a TDP connection
+    // until that information is available.
+    const initTdpCli = !(fetchAttempt.status === 'processing');
```
hatched, line 238: nit :) 
```suggestion
    const initTdpCli = fetchAttempt.status !== 'processing';
```

packages/teleport/src/DesktopSession/DesktopSession.tsx
```diff
@@ -173,31 +232,32 @@ export function DesktopSession(props: State) {
...
+
     return (
-      <Session {...props}>
+      <Session initTdpCli={initTdpCli} displayCanvas={false} {...props}>
```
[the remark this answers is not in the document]
ibeckermayer: I added that, and I also added an error state for if `fetchAttempt.status === 'processing' &&  tdpAttempt.status === 'success'`

168a47528509b676647fb7b90c7d969c4df8c51f

packages/teleport/src/DesktopSession/DesktopSession.tsx
```diff
@@ -102,6 +111,14 @@ export function DesktopSession(props: State) {
       based browsers like Google Chrome or Microsoft Edge. Brave users can \
       use the feature by navigating to brave://flags/#file-system-access-api \
       and selecting "Enable". Please switch to a supported browser.';
+    } else if (
+      fetchAttempt.status === 'processing' &&
+      tdpConnection.status === 'success'
+    ) {
+      errorText =
```
hatched, line 118: The snapshot output for the `invalid` test doesn't appear to be correct as it doesn't have this content.

ibeckermayer: Good catch: a47b801008c7ae5621425b79327a96bc73d7c55e

hatched (approved):
Thanks for the continued updates!

packages/teleport/src/DesktopSession/DesktopSession.tsx
```diff
@@ -173,31 +240,32 @@ export function DesktopSession(props: State) {
...
 
   if (processing) {
+    // We don't know whether another session for this desktop is active while the
+    // fetchAttempt is still processing, so hold off on starting a TDP connection
+    // until that information is available.
+    const initTdpCli = fetchAttempt.status !== 'processing';
+
     return (
-      <Session {...props}>
+      <Session initTdpCli={initTdpCli} displayCanvas={false} {...props}>
```
kimlisa, line 249: opt nit, just makes it look like we may overwrite props
```suggestion
      <Session {...props} initTdpCli={initTdpCli} displayCanvas={false}>
```

kimlisa (approved):

PR #1298, opened 2022-10-25
Handle private key policy errors

kimlisa:
resolves https://github.com/gravitational/teleport/issues/16168

#### Description
There are three ways we determine if use of private key is enabled:
  1) Cluster config flag: this setting won't allow the user to login to the web, it will just direct user to login with other supported methods
  2) `privateKeyPolicyEnabled` flag: only returned with a 200 for `resets and invites`. This flag was required so that we can additionally render recovery codes if any (only for Teleport Cloud)
  3) checking if the error returned contains the backend hard coded keywords: `hardware_key` or `hardware_key_touch`, this is used for:
      - logins with role config that requires private keys
      - assuming a role (from access requests in waiting room or in dashboard) that requires private keys (enterprise feature that's handled here https://github.com/gravitational/webapps.e/pull/422)

#### TODO
- [x] requires https://github.com/gravitational/teleport/pull/17651
- [x] update e-ref after merge (reason why test is failing, but other wise locally passes) https://github.com/gravitational/webapps.e/pull/422
- [ ] backport to `v11`

#### Test cases to add to test plan (added screenshots for UI and copy review)
Enforce use of private keys by cluster config or by role config.
```yaml
# Cluster wide config
auth_service:
    require_session_mfa: hardware_key_touch  # or `hardware_key`

# Role config
kind: role
spec:
  options:
    require_session_mfa: hardware_key_touch  # or `hardware_key`
```
Private key policy card should render for:
- [x] With cluster setting on login page
- [ ] With cluster setting on login page for Teleport Cloud (should also render option to recover authn)
- [x] With role setting, after successfully logging in with password only, with 2fa's, and passwordless

image

image

With a access request role that requires private key, a private key required dialogue pops up for:
- [x] Waiting room
- [x] Assuming an approved access request through the `access request list`
- [x] Assuming an approved access request through the `view access request` screen

(For waiting room, action button says `Logout`, but when logged in and you get this error from trying to assume requests, the action button says `Okay`)
image

Below tests, it shouldn't matter if private key enforcement is set through cluster or role setting.
Private key policy card should render for Invites and Resets:
- [x] After successfully registering with password only, with 2fa's, and passwordless
- [x] After successfully resetting with password only, with 2fa's, and passwordless)
- [x] After successfully registering/resetting with Teleport Cloud and successfully receiving recovery codes

image

image


xinding33:
@kimlisa suggested copy:

Web UI Login Disabled - This Teleport Cluster requires that user private keys be stored on hardware authentication devices. Since these keys are not accessible by web browsers, Web UI login has been disabled. Please use Teleport Connect or tsh to log in.

kimlisa:
friendly ping @JanKaczmarkiewicz @ryanclark

JanKaczmarkiewicz (approved):

ryanclark (approved):

PR #1299, opened 2022-10-25
[UI]: Make roles render as labels

dev-slatto:
Hey folks! 

Could we make roles render as labels in stead of a comma-separated list? Makes it much more readable when having lots of roles, and it also follows the overall design of the Teleport UI 😄 

Se screenshot for how it will look:
image

Would love to have it merged! 🚀


packages/teleterm/src/ui/TopBar/Identity/IdentityList/IdentityList.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 
 import styled from 'styled-components';
 
-import { Box, Flex, Text } from 'design';
+import { Box, Flex, Text, Label as SingleLabel } from 'design';
```
avatus, line 5: This import doesn't need to be renamed. We can keep it as `Label`

packages/teleterm/src/ui/TopBar/Identity/IdentityList/IdentityList.tsx
```diff
@@ -33,7 +33,12 @@ export function IdentityList(props: IdentityListProps) {
             <Box>
               <Text bold>{props.loggedInUser.name}</Text>
               <Text typography="body2" color="text.secondary">
-                {props.loggedInUser.rolesList.join(', ')}
+                {props.loggedInUser.rolesList.map(label =>
```
avatus, line 36: This map argument should be named `role` instead of `label` since we are mapping through a `rolesList`

packages/teleterm/src/ui/TopBar/Identity/IdentityList/IdentityList.tsx
```diff
@@ -33,7 +33,12 @@ export function IdentityList(props: IdentityListProps) {
...
+                  <SingleLabel mb="1" mr="1" key={label} kind="secondary">
+                    {label}
+                  </SingleLabel>
+
```
avatus, line 40:
```suggestion
```

dev-slatto:
@avatus Thanks for the feedback! It should all be resolved in the latest commit to the PR.

gzdunek (approved):

ravicious:
@dev-slatto I was experimenting with this PR locally and I'm pretty sure you need to run `yarn prettier-write`, otherwise the Webapps-Build check will fail.

ravicious (approved):
Great change, this makes long roles definitely easier to read.

BTW, we have a story in our storybook (`yarn storybook`) written to inspect the profile selector with long names, it's under Teleterm > Identity > Long Names With Many Roles. I'm glad I wrote it back in the day as it was quite useful here. ;)

dev-slatto:
There we go! Sorry about that, prettier is always a pre-commit hook in the other repos I contribute in so it didn't cross my mind. Should be fine now 😉 Thanks for the Storybook tip! No more manually creating long role names to check if it breaks the styling 😄

avatus:
Awesome change. Thanks for much! Merged 👍

PR #1304, opened 2022-10-26
Change Session Recording created date to UTC timestamp

avatus:
Fixes https://github.com/gravitational/teleport/issues/17623

The current session recordings log would display localized timestamps when using `new Date()`. In order to keep these recording logs consistent with the audit logs, we should display the UTC timestamp sent back from the server.


hatched (approved):
Thanks for making this consistent with the Audit log. I wonder if we shouldn't do a follow-up to keep these in UTC but prettify them back to a date time.

`2022-07-11 14:39:33` instead of `2022-07-11T14:39:33.373Z`

avatus:
> Thanks for making this consistent with the Audit log. I wonder if we shouldn't do a follow-up to keep these in UTC but prettify them back to a date time.

I like this idea. I'll create a separate PR to solicit feedback on it.

PR #1307, opened 2022-10-27
[Connect] Use server side search in resource tables (Advanced Search)

avatus:
Hey all.

I'm opening up a draft with this work in progress because I want to solicit some feedback before going too far down a path. I've redone this maybe three times now because I can't make up my mind. So I'd like to get your opinions on some things.

1. One of the pieces of feedback from the Access Requests PR was that quite a bit of code was duplicated from the web's version of Access Requests. While in that case there was enough minor changes to warrant the _mostly_ duplicated code (or at least reused aspects of it), this change for Serverside stuff in the resources table will have a lot duplicated between not only Access Requests tables but also the individual resource tables themselves (servers, dbs, kubes, etc). To try and combat this, I've moved some common functions to `shared/utils` as well as two components to the `shared/components`. I tried to follow convention for how some other things have been abstracted in the codebase but let me know if this is a good change. Also, I tried updating the 'fetch' logic to not have to have a `fetch`, `fetchNext` and `fetchPrev` but... my solution might have been just as weird. I don't mind either one (you can find an example in `useNewRequest` of the thing I'm trying to avoid).

2. As of now, we sync the cluster and just clientside pagniate/search through it. My original idea was to keep the current clientside searching for "normal" search and then do a network request for advanced search. This ended up being too complicated and added a bunch of unneeded code complexity.  it's easier to use serverside search for normal/advanced. This lets `useServers` completely control what data is shown in the `Servers` table. That being said, the `clustersService` map of servers is important for connecting to the selected resource (and eventually for spotlight search), so the idea is to just add something like `addIfMissing(resourceMap, fetchedData)`. This isn't included in this PR because I wanted to keep it mostly focused on the search implementation but I think its fairly straight forward. Cluster will sync on init and then update if needed during a resource search (normal OR advanced). 

4. Finally, because the way we render each resource table, the table will unmount when selecting a new resource which means, the usual behavior of "fetch on mount" will happen in between every switch of resource. It's not too big of a deal but in order to keep the data there, we'd have to move this data one level up and I don't know I want to abstract it away. I like it being close to the component and this might just be something we deal with? Thoughts?

Anyway, I apologize for the rambling above but I think I'd rather get feedback on it now before duplicating potentially bad code into `Databases` and `Kubes`. I'll fix it once and then throw it into those tables when we agree on something :) 

This needs to go in as well https://github.com/gravitational/webapps.e/pull/436


avatus:
A note on the table "loading state". It's currently using the loading state that `Table` has built in. This is different from the Access Requests `NewRequest` table loading state. I left them different on purpose so we can decide on which one to use. I will update both tables to match in this PR once we decide.  My vote is the line darkening of the `NewRequest` table but I'm open to either.

avatus:
After talking with @ravicious we decided to move the serverside logic into its own file so that all the resource tables could reuse the fetching logic. It's way cooler now! I left some comments on a part I'm not quite happy with but I'm unsure if there is a better way.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts
```diff
@@ -0,0 +1,153 @@
...
+    switch (resource) {
+      case ResourceKind.Server:
+        return retryWithRelogin(ctx, documentUri, clusterUri, () =>
+          ctx.clustersService.fetchServers(params)
+        );
+      default: {
+        throw new Error(`Fetch not implemented for: ${resource}`);
+      }
+    }
+  }
```
avatus, line 38: Originally I was planning to just send in the clustersService function to the call of this function in `useServers` but was getting some big typescript errors. If I typed the callback going into `useServers` I would get errors that `useServers` didn't know what type was coming back from the `fetchAttempt`, and if I typed it, then `retryWithRelogin` was throwing typescript errors of `Cannot return Promise<unknown>` or something along those lines. 

This looks a bit ugly but, it pleased the typescript gods and isn't too bad imo. 

Ideally I would have it the first suggested way of `useServers(ctx.clustersService.fetchServers)` but I hit my limit of typescript arcane knowledge I think.

Secondly, we couldn't have the `retryWithRelogin` wrapper outside of this file (example like moving it to `useServers(retryWithRelogin....)` because the `params` needed for the internal callback are held within this file instead.

If i've missed something simple let me know and I'm happy to update!

packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts
```diff
@@ -0,0 +1,153 @@
...
+  // "why don't we just name it nextKey then?"
+  // Mostly because it's called startKey almost everywhere else in the UI, and also because we'd have the same issue
+  // for prevPage if we swapped named, and this comment would be explaining "this page's startKey".
+  useEffect(() => {
+    const newKeys = [...keys];
+    newKeys[pageIndex] = fetchAttempt.data?.startKey;
+    setKeys(newKeys);
+
+    setPageCount(getPageCount());
+  }, [fetchAttempt.data]);
```
avatus, line 102: This is where I'd imagine we would put something like `addDataIfMissing` type of call to the clustersService cache for the current resource

packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts
```diff
@@ -0,0 +1,153 @@
...
+import { useClusterContext } from '../clusterContext';
+import { AgentFilter, AgentLabel } from 'teleport/services/agents';
+import { addAgentLabelToQuery } from 'shared/utils/addAgentLabelToQuery';
+import { ResourceKind } from 'teleport/Discover/Shared';
+
+export function useServerSideResources(resource: ResourceKind) {
+  const ctx = useAppContext();
+  const { clusterUri, documentUri } = useClusterContext();
+  const [pageIndex, setPageIndex] = useState(0);
+  const [keys, setKeys] = useState<string[]>([]);
```
avatus, line 16: Previously, these two values were held in a `Page` object, but these two values serve different purposes now and are accessed/used at different times. the `pageIndex` is now used to (obviously) update our current page, which will trigger a fetch effect. and `keys` is used to store data returned from that effect. If they were to be kept separate, the state would get all wonky trying to update only half of itself at "opposite" times. This smooths out the experience.

gzdunek:
> [...] To try and combat this, I've moved some common functions to shared/utils as well as two components to the shared/components. I tried to follow convention for how some other things have been abstracted in the codebase but let me know if this is a good change. [...]

Generally, it is a good direction, but IMO we should approach it differently. The UI of server side search in Web and Connect is identical, but is implemented in quite different ways. Web uses `serversideProps` which can do some work for us (like checking if next page is available), but we added our own `<SearchPanel />` and `<SearchPagination />`. Is there any reason to do that? It seems to me that is shouldn't be very hard to reuse the existing `<ServersideSearchPanel />`. Probably the main thing that needs to be changed is that `<ServersideSearchPanel />` should no longer work directly on URL query params, but instead should have a function like `onSearchChange()` that would:
- update URL for Web
- run `updateQuery` or `updateSearch` from `useServerSideResources` in Connect

I would like to avoid maintaining two implementations for basically the same UI if possible.

gzdunek:
BTW, I don't remember, did we have any idea how the integration between spolight search and advanced search is going to look like? Will we have a full copy of the resources locally, but also perform network queries, because of the advanced search?

gzdunek (commented):
I left my first thoughts

packages/shared/components/Search/SearchPagination.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import { Flex } from 'design';
+import { StyledPanel } from 'design/DataTable/StyledTable';
+import { StyledArrowBtn } from 'design/DataTable/Pager/StyledPager';
```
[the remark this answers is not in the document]
avatus: This file (and `SearchPanel`) were just abstracted away from the table used in `NewRequest` so that it could be reused here. If I'm going to change the implementation then these probably wouldn't need to exist anymore

avatus:
> but we added our own `<SearchPanel />` and `<SearchPagination />`. Is there any reason to do that?

This implementation follows more closely how the `NewRequest` table works which (in both Connect and Web) created their own `SearchPanel` and `SearchPagination`. Because of this, I moved those two components into the shared folder so they could be used in both places (now 3 places) easier. `SearchPanel` was moved because it was currently in the `webapps.e` directory and didn't need to be.

I agree that two different implementations is not a good way to maintain. There were already two to chose from (the way the resources work, and the way the access requests tables work) since they both basically do the same thing. The reason I chose the way I did was it seemed easier to abstract the component for a search panel and not make any changes to them (`SearchPanel` and `SearchPagination` were taken directly from `NewRequest` than it did to try and get away from updating URL things and using `serversideProps`. IMO, `serversideProps` has a strong opinion towards URL based routing.

avatus:
> BTW, I don't remember, did we have any idea how the integration between spolight search and advanced search is going to look like? Will we have a full copy of the resources locally, but also perform network queries, because of the advanced search?

The idea is for now, we have a full copy as things like autocomplete depend on it. Hopefully, when we have a "live cache" setup, we would end up querying the live cache instead but this is the path to get advanced search integrated now until spotlight search is back on the roadmap

gzdunek:
>This implementation follows more closely how the NewRequest table works which (in both Connect and Web) created their own SearchPanel and SearchPagination. Because of this, I moved those two components into the shared folder so they could be used in both places (now 3 places) easier.

Ah, I was not aware that the file was moved, not added.

In that case it looks like we have two competing options ;) I don't have any strong opinion right now because making them consistent is quite a lot of work. 
If we would like to do it, we could try to use `serversideProps` everywhere or create a new component like `ServerSideTable` (probably you will have to create such component anyway because repeating the same code - `SearchPanel`, `Table`, `SearchPagination` for each resource type is not practical). 

Anyway, I think we can wait with the decision for @ravicious.

avatus:
I'm ok with creating `ServersideTable` (funnily enough, thats the name of the component that is returned when `serversideProps` is passed to a `Table`). Although, adding `<ServersideProps ...allThePropsHere />` and `<SearchPanel ...allThePropsHere />` won't be much different.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts
ravicious: The attached patch makes passing the function work as expected. I used generic type to extract the parts of the response that change between calls to `useServerSideResources`.

Unfortunately, we cannot simply do `useServerSideResources(appContext.clustersService.fetchServers)` because in JS instance methods are not bound to the instance. We'd have to do

```typescript
useServerSideResources(appContext.clustersService.fetchServers.bind(appContext.clustersService))
```

but Function.bind loses all type information so we'd also need to add

```typescript
useServerSideResources<types.Server>(appContext.clustersService.fetchServers.bind(appContext.clustersService))
```

at which point it's no longer worth it I think.

<details>
<summary>Patch</summary>

```diff
diff --git a/packages/teleterm/src/ui/DocumentCluster/ClusterResources/Servers/useServers.ts b/packages/teleterm/src/ui/DocumentCluster/ClusterResources/Servers/useServers.ts
index 45bfe80d..4b29c15c 100644
--- a/packages/teleterm/src/ui/DocumentCluster/ClusterResources/Servers/useServers.ts
+++ b/packages/teleterm/src/ui/DocumentCluster/ClusterResources/Servers/useServers.ts
@@ -13,8 +13,8 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-import { ResourceKind } from 'teleport/Discover/Shared';
 import { useAppContext } from 'teleterm/ui/appContextProvider';
+
 import { useServerSideResources } from '../useServerSideResources';
 
 export function useServers() {
@@ -30,7 +30,9 @@ export function useServers() {
     onAgentLabelClick,
     updateSort,
     pageCount,
-  } = useServerSideResources(ResourceKind.Server);
+  } = useServerSideResources(params =>
+    appContext.clustersService.fetchServers(params)
+  );
 
   function getSshLogins(serverUri: string): string[] {
     const cluster = appContext.clustersService.findClusterByResource(serverUri);
diff --git a/packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts b/packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts
index 30c535f1..6fb55a89 100644
--- a/packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts
+++ b/packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts
@@ -1,15 +1,21 @@
 import { useState, useEffect } from 'react';
+
 import { SortType } from 'design/DataTable/types';
 import { useAsync } from 'shared/hooks/useAsync';
+import { AgentFilter, AgentLabel } from 'teleport/services/agents';
+import { addAgentLabelToQuery } from 'shared/utils/addAgentLabelToQuery';
+
 import { ServerSideParams } from 'teleterm/services/tshd/types';
 import { useAppContext } from 'teleterm/ui/appContextProvider';
 import { retryWithRelogin } from 'teleterm/ui/utils';
+
 import { useClusterContext } from '../clusterContext';
-import { AgentFilter, AgentLabel } from 'teleport/services/agents';
-import { addAgentLabelToQuery } from 'shared/utils/addAgentLabelToQuery';
-import { ResourceKind } from 'teleport/Discover/Shared';
 
-export function useServerSideResources(resource: ResourceKind) {
+export function useServerSideResources<T>(
+  fetchFunction: (
+    params: ServerSideParams
+  ) => Promise<{ agentsList: Array<T>; totalCount: number; startKey: string }>
+) {
   const ctx = useAppContext();
   const { clusterUri, documentUri } = useClusterContext();
   const [pageIndex, setPageIndex] = useState(0);
@@ -25,28 +31,18 @@ export function useServerSideResources(resource: ResourceKind) {
 
   const limit = 15;
 
-  function getFetchCallback(params: ServerSideParams) {
-    switch (resource) {
-      case ResourceKind.Server:
-        return retryWithRelogin(ctx, documentUri, clusterUri, () =>
-          ctx.clustersService.fetchServers(params)
-        );
-      default: {
-        throw new Error(`Fetch not implemented for: ${resource}`);
-      }
-    }
-  }
-
   // startKey is used here as a way to paginate through agents returned from
   // their respective rpcs.
-  const [fetchAttempt, fetch] = useAsync(async (startKey: string) => {
-    return getFetchCallback({
-      ...agentFilter,
-      limit,
-      clusterUri,
-      startKey,
-    });
-  });
+  const [fetchAttempt, fetch] = useAsync(async (startKey: string) =>
+    retryWithRelogin(ctx, documentUri, clusterUri, () =>
+      fetchFunction({
+        ...agentFilter,
+        limit,
+        clusterUri,
+        startKey,
+      })
+    )
+  );
 
   // If there is no startKey at the current page's index, there is no more data to get
   function nextPage() {
```
</details>

ravicious (commented):
Just an initial look, I noticed some UI issues while testing my patch, like the UI briefly showing "No servers found" or something like that but this can wait for later.

packages/teleterm/src/ui/services/clusters/clustersService.ts
```diff
@@ -13,6 +13,7 @@ import {
   Cluster,
```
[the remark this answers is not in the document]
avatus: First of all, love pair programming so sign me up. 
I think making autocomplete async is fine although I'd rather not leave it broken. If anything, I'd rather just leave it inefficient by updating the map with new entries. I may be naive here, but it seems as simple as making a new function similar to `updateMap`
```js
const helpers = {
  updateMap<T extends { uri: string }>(
    parentUri = '',
    map: Map<string, T>,
    received: T[]
  ) {
    // delete all entries under given uri
    for (let k of map.keys()) {
      if (routing.ensureClusterUri(k) === parentUri) {
        map.delete(k);
      }
    }

    received.forEach(s => map.set(s.uri, s));
  },
};
```

but just skipping the delete phase. This would make it work and be very little code to remove in the autocomplete async PR. Either way, I don't mind.

Good news I think is, if we implement everything using the same async functions (`fetchServers`, `fetchKubes`, etc), once livecache comes into play, we _should_ be able to just update those async functions to pull from the cache instead and the rest of the app would work.


> remove resource counts in DocumentCluster (they're going to be present at the bottom of the table anyway)

Yes. Agreed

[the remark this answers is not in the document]
ravicious: The problems with just updating the cache when server-side requests come in are:

* The autocomplete might miss some entries if there was no server-side search which would return those entries.
   * That's what I was referring to when I said we'd leave the autocomplete slightly broken, I should've been more clear here.
* The autocomplete is still going to be unusable on clusters with a lot of resources as it does all filtering within the main browser thread.

> I think making autocomplete async is fine although I'd rather not leave it broken. If anything, I'd rather just leave it inefficient by updating the map with new entries.

I don't understand the distinction between broken and inefficient that you used here, what you wrote seems to imply that making the autocomplete async == leaving it broken? ;)

When I said that we'd leave it slightly broken, I referred to a situation where we would leave it sync, depending on `ClustersService` cache.

> Good news I think is, if we implement everything using the same async functions (`fetchServers`, `fetchKubes`, etc), once livecache comes into play, we _should_ be able to just update those async functions to pull from the cache instead and the rest of the app would work.

Right. And I imagine, though I'm not 100% sure, that whatever solution we're going to come up with for the live cache might look significantly different than what we have now so that's another argument for ripping out resource management from `ClustersService` for now.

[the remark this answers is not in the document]
avatus: I misunderstood broken. I didn't apply "doesn't work with large clusters" to the broken adjective and only thought of it missing entries. What you spoke of originally makes more sense to me now and I'm onboard 👍

packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts
ravicious: You might want to use a less vague name for the generic type instead of `T`. `Agent` would be good.

avatus:
> Just an initial look, I noticed some UI issues while testing my patch, like the UI briefly showing "No servers found" or something like that but this can wait for later.

I added a [small change](https://github.com/gravitational/webapps/pull/1307/commits/e29150c6da1b7810d530ea4500b4ecda97d91081) for this. I think the change is all that is needed to remedy this tbh.

avatus:
For now, the `resourcesService` looks a bit sparse but obviously adding in the other resources with the real PR will fill it up a bit. And perhaps more things can move into here that are currently in `useServersideResources` but I haven't fiddled with that yet. 

Side note, it was quite easy to remove the need for `clustersService` in the connect cell for servers (and also databases, but not included in these changes yet).

packages/shared/components/Search/SearchPanel.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { Text, Flex } from 'design';
+import { StyledPanel } from 'design/DataTable';
+import { PageIndicatorText } from 'design/DataTable/Pager/Pager';
+import InputSearch from 'design/DataTable/InputSearch';
+import { AgentFilter } from 'teleport/services/agents';
+import Toggle from 'teleport/components/Toggle';
+import Tooltip from 'teleport/components/ServersideSearchPanel/Tooltip';
+import { PredicateDoc } from 'teleport/components/ServersideSearchPanel/ServersideSearchPanel';
+
+export function SearchPanel({
```
ravicious, line 12: Excuse me but I don't understand where the issue with `<SearchPanel/>` and `<SearchPagination/>` is that you discussed above.

`packages/shared/components/Search/SearchPanel.tsx` looks almost identical to `packages/teleport/src/components/ServersideSearchPanel/ServersideSearchPanel.tsx`. The only difference is in the callbacks passed to the particular children + the one from `packages/shared` supports `showSearchBar` as a prop. But couldn't both components reuse the same visual component and pass different props?

The situation looks similar when it comes to `packages/shared/components/Search/SearchPagination.tsx` and `packages/design/src/DataTable/Pager/Pager.tsx`. `Pager` seems to support a bunch of additional props but couldn't we reuse the visual component instead of copying it?

ravicious (commented):
Took another cursory look today, mostly because I couldn't figure out what the issue with those shared search components is. ;f I'll try to make a more comprehensive review tomorrow.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Servers/Servers.tsx
```diff
@@ -103,7 +103,11 @@ function ServerList(props: State) {
             },
           ]}
           customSort={customSort}
-          emptyText="No Nodes Found"
+          emptyText={
+            fetchAttempt.status === 'processing'
+              ? 'Searching...'
+              : 'No servers found'
+          }
```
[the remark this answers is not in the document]
avatus: I didn't even know that was a thing lol. Sure!

packages/shared/components/Search/SearchPanel.tsx
avatus: The issue we spoke of was using these as components to "build" the table rather than using `Table`'s prop of `serversideProps`. I don't care where we pull the components from tbh, these two components `SearchPanel` and `SearchPagination` I just created to abstract from the files they were originally in (`NewRequest` and a non-shared version of `SearchPanel`). If they exist elsewhere, I'm happy to use those too.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Servers/Servers.tsx
```diff
@@ -114,3 +153,10 @@ const renderAddressCell = ({ addr, tunnel }: types.Server) => (
     {!tunnel && addr}
   </Cell>
 );
+
+const Wrapper = styled(Box)`
+  &.disabled {
+    pointer-events: none;
+    opacity: 0.5;
+  }
```
ravicious, line 161: I finally got around to testing actual pagination, the following patch gives me "infinite" pagination to play around with.

<details>
<summary>patch</summary>

```diff
diff --git a/lib/teleterm/apiserver/handler/handler_servers.go b/lib/teleterm/apiserver/handler/handler_servers.go
index 685f6ee6d..c37fd70e1 100644
--- a/lib/teleterm/apiserver/handler/handler_servers.go
+++ b/lib/teleterm/apiserver/handler/handler_servers.go
@@ -16,7 +16,9 @@ package handler
 
 import (
 	"context"
+	"fmt"
 	"sort"
+	"strconv"
 
 	"github.com/gravitational/trace"
 
@@ -41,18 +43,43 @@ func (s *Handler) GetAllServers(ctx context.Context, req *api.GetAllServersReque
 
 // GetServers accepts parameterized input to enable searching, sorting, and pagination
 func (s *Handler) GetServers(ctx context.Context, req *api.GetServersRequest) (*api.GetServersResponse, error) {
+	originalStartKey := req.StartKey
+	var base int
+	var err error
+	if originalStartKey == "" {
+		base = 0
+	} else {
+		base, err = strconv.Atoi(originalStartKey)
+		if err != nil {
+			return nil, trace.Wrap(err)
+		}
+		base++
+	}
+	req.StartKey = ""
 	resp, err := s.DaemonService.GetServers(ctx, req)
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
 
+	var startKey int
+
 	response := &api.GetServersResponse{
-		TotalCount: int32(resp.TotalCount),
+		TotalCount: int32(base + 30),
 		StartKey:   resp.StartKey,
 	}
 	for _, srv := range resp.Servers {
-		response.Agents = append(response.Agents, newAPIServer(srv))
+		for i := 0; i < 15; i++ {
+			k := i + base
+			apiSrv := newAPIServer(srv)
+			apiSrv.Uri = fmt.Sprintf("%v%d", apiSrv.Uri, k)
+			apiSrv.Name = fmt.Sprintf("%v%d", apiSrv.Name, k)
+			apiSrv.Hostname = fmt.Sprintf("%v%d", apiSrv.Hostname, k)
+			apiSrv.Addr = fmt.Sprintf("%v%d", apiSrv.Addr, k)
+			response.Agents = append(response.Agents, apiSrv)
+			startKey = k
+		}
 	}
+	response.StartKey = strconv.Itoa(startKey)
 
 	return response, nil
 }
```
</details>

I think the transition here from no opacity to 0.5 is a bit too sudden. It causes heavy flashing when testing locally but that's also partly because I'm running the cluster locally so the requests come back almost immediately. But even after adding 250–850ms of artificial delay to the requests, it was a bit too jarring because the change in brightness is quite big and happens too fast.

What would you say about this? This needs to be repeated for `<SearchPanel>` as well.

```suggestion
  // The timing functions of transitions have been chosen so that the element loses opacity slowly
  // when entering the disabled state but gains it quickly when going out of the disabled state.
  transition: opacity 150ms ease-out;

  &.disabled {
    pointer-events: none;
    opacity: 0.7;
    transition: opacity 150ms ease-in;
  }

```

packages/teleterm/src/ui/services/clusters/clustersService.ts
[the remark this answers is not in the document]
avatus: New working hours are fine, I'll make it work no problem at all. I think that although priorities have changed for the quarter, since this (to me) is pretty close to done (advanced search), might as well get it out. If we think that changing autocomplete is required for this to go out (which to be clear, I do), I think the that work is worth it. A couple hours in the grand scheme of the quarter shouldn't affect the goals. I'll setup some time next week on our calendar.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Servers/Servers.tsx
```diff
@@ -65,46 +69,45 @@ function ServerList(props: State) {
...
-              />
-            ),
-          },
-          {
-            altKey: 'connect-btn',
-            render: server =>
-              renderConnectCell(
-                () => getSshLogins(server.uri),
-                login => connect(server.uri, login)
+      <Wrapper className={disabledRows ? 'disabled' : ''}>
```
ravicious, line 72: I suppose renaming `Wrapper` to something that more directly expresses the role that this element plays would be a good idea!

packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts
```diff
@@ -0,0 +1,149 @@
...
+  function updateQuery(query: string) {
+    setAgentFilter({ ...agentFilter, search: '', query });
+  }
+
+  function onAgentLabelClick(label: AgentLabel) {
+    const query = addAgentLabelToQuery(agentFilter, label);
+    setAgentFilter({ ...agentFilter, search: '', query });
+  }
+
+  function getPageCount() {
```
ravicious, line 117: If a function can accept an additional argument or two and be defined at the top level, outside of the hook, then I generally prefer to do so. I think `getPageCount` and the hardcoded `limit` are great candidates for top-level definitions.

This impacts not only performance (functions and vars defined within the hook are redefined on each render) but most importantly they make reading the hook body much easier.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts
```diff
@@ -0,0 +1,149 @@
...
+        // we have to go back 2. which is why we use newPageIndex instead
+
+        // example after going from pageIndex = 0 to pageIndex = 1:
+        // ['startKeyFromInitialFetch', 'goEvenFurtherStartKey']
+        //     ^- just came from here          ^- we are here
+        //
+        // so in order to get the start key for the page BEFORE this one, we go back twice.
+        // this could also be written/read as keys[pageIndex - 2]
+        fetch(keys[newPageIndex - 1]);
+        setPageIndex(newPageIndex);
```
ravicious, line 74: The new implementation is better than the one from bd7fc4bcd28b. However, I remember we talked about keeping the implementation of `prevPage` and `nextPage` as lean as possible where they could merely update the pagination metadata, which would cause some `useEffect` to re-run which would actually fetch the data. Did that turn out to not be a good approach here?

packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts
```diff
@@ -0,0 +1,149 @@
...
+    }
+    return null;
+  }
+
+  useEffect(() => {
+    // when agent filter changes we consider it a 'new' search so we should pass an empty
+    // startKey to get the beginning of the list with new filters, and set pageIndex back to zero
+    fetch('');
+    setPageIndex(0);
+  }, [agentFilter]);
```
ravicious, line 85: If the page index needs to be set to 0 when the agent filter changes, we should do so in functions which update the agent filter.

See:

* https://beta.reactjs.org/learn/you-might-not-need-an-effect#chains-of-computations
* https://beta.reactjs.org/learn/you-might-not-need-an-effect#sharing-logic-between-event-handlers

packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts
```diff
@@ -0,0 +1,149 @@
...
+  // "why don't we just name it nextKey then?"
+  // Mostly because it's called startKey almost everywhere else in the UI, and also because we'd have the same issue
+  // for prevPage if we swapped named, and this comment would be explaining "this page's startKey".
+  useEffect(() => {
+    const newKeys = [...keys];
+    newKeys[pageIndex] = fetchAttempt.data?.startKey;
+    setKeys(newKeys);
+
+    setPageCount(getPageCount());
+  }, [fetchAttempt.data]);
```
ravicious, line 98: This falls under chains of computations that I mentioned above.

https://beta.reactjs.org/learn/you-might-not-need-an-effect#chains-of-computations

packages/shared/components/Search/SearchPanel.tsx
ravicious: I think the least we could do here is avoid introducing additional duplication by reusing those visual components. I didn't use those table components that much so without investing more time into it, it's hard for me to say what's the best course regarding the duplication of logic between different uses of data tables.

gzdunek:
I was trying to integrate with `serversideProps` and it's not that easy, so I think we can live with "building" the table from smaller components.

However, there is also code duplication in the actual fetching logic (I'm mean the `fetch status`, functions like `fetch`, `fetch prev`, `fetch next`, `page count`, etc.). Currently, each serverside table has its own implementation (looks like we have 10+ implementations for _fetch next page_!), which are scattered in the places like:
1. teleport nodes/dbs/kubes/apps/desktops
2. teleport access requests
3. connect servers
4. connect access requests  

If we would polish our implementation then maybe we could create a hook like `useServerSideFetch()` which we would use in other places? It doesn't need to happen in this PR, but I'd like to unify this non-trivial logic as much as possible and remove the duplicated code.

avatus:
> However, there is also code duplication in the actual fetching logic

I agree, it's a bit messy now. I spoke with @ravicious a bit about this on a call before the change. Before, it was even more duplicated but we decided to try and 'clean it up' this way. So, yes, this is another NEW way, with the idea of "hey, once this is done, we can take this new, better way and port it back to the needed spots" (at least in Connect). I believe all our resource tables and access request tables can use this one `useServerSideResources` hook and we'd be golden.

gzdunek:
> [...] we can take this new, better way and port it back to the needed spots" (at least in Connect). I believe all our resource tables and access request tables can use this one `useServerSideResources` hook and we'd be golden.

Ok, makes sense!

packages/shared/components/Search/SearchPanel.tsx
avatus: technically, this PR didn't introduce this duplication. This component was being used in the teleport package and I just moved it to the shared package. I can dig into why they used two different panels but you can see the [changes](https://github.com/gravitational/webapps.e/compare/74d3be697bfcc31b4c69ffe024083902c516be55...31757a460bc3f73c79c0ba301d86120d4120e051) on webappse

I wish it was a bit more clear in this PR that those files were just moved, not added.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Servers/Servers.tsx
avatus: This does look really nice

packages/shared/utils/addAgentLabelToQuery.ts
```diff
@@ -0,0 +1,21 @@
+import { AgentFilter, AgentLabel } from 'teleport/services/agents';
+
+export function addAgentLabelToQuery(filter: AgentFilter, label: AgentLabel) {
```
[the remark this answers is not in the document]
avatus: Sure, we can move it to top level

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Servers/Servers.tsx
avatus: `FetchStatusBox` ?? Can't really think of a good name here. I'm open to whatever. It's purpose is to darken things while fetching lol
`FetchDarkener` sounds kinda fantasy and thats neat

packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts
avatus: `limit` for sure, but we'd have to pass in `pageIndex` as well as `fetchAttempt` which then we'd have to pass around the `Agent` type of `fetchAttempt` as well I believe. We can do it for sure but I think that it might be more convoluted removing getPageCount from the hook. Thoughts?

packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts
avatus: Updated 👍

packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts
```diff
@@ -0,0 +1,149 @@
...
+  // page index. think of this as "this page's nextKey".
+  // "why don't we just name it nextKey then?"
+  // Mostly because it's called startKey almost everywhere else in the UI, and also because we'd have the same issue
+  // for prevPage if we swapped named, and this comment would be explaining "this page's startKey".
+  useEffect(() => {
+    const newKeys = [...keys];
+    newKeys[pageIndex] = fetchAttempt.data?.startKey;
+    setKeys(newKeys);
+
+    setPageCount(getPageCount());
```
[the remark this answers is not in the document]
avatus: I think you're right! Updated.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts
avatus: I think my brain got a bit stuck in an old flow and when I finally figured out how to get it 'lean' it stayed with the old. Updated the `fetch` useEffect to react to agentFilter and page. (page will always update with agentFilter after your feedback above, but agentFilter won't always update with page).

packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts
ravicious: > These two pieces of data can only be updated after the data has been received as that returned data contains the information from the server regarding page metadata and keys.

Right, so we not update them when you receive that data?

The docs say:

> Sometimes you might feel tempted to chain Effects that each adjust a piece of state based on other state.

And I think this is what we do here: we wait for `fetchAttempt` to get updated so that we can adjust keys but we could do it after we receive a response from `fetch`.

In that effect where you execute `fetch`, you could do something like

```typescript
useEffect(() => {
  const fetchAndUpdateKeys = async () => {
    const [response, err] = await fetch(keys[pageIndex - 1]);

    if (err) {
      return;
    }

    const startKey = response.startKey;

    // Do something with keys.
  };

  fetchAndUpdateKeys();
}, [agentFilter, pageIndex]);
```

It'd be nice to pass `agentFilter` directly to `fetch` too so that the dependency between them is clearly visible when one calls `fetch`.

ravicious: But in general I have to say this hook looks better and better with every change! The current version is much, much less intimidating than the initial one.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Servers/Servers.tsx
avatus: I like that!

packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts
avatus: I see the forest through the trees now and understand what you mean. Great change and even easier to read

packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts
```diff
@@ -0,0 +1,159 @@
...
+      // page index. think of this as "this page's nextKey".
+      // "why don't we just name it nextKey then?"
+      // Mostly because it's called startKey almost everywhere else in the UI, and also because we'd have the same issue
+      // for prevPage if we swapped named, and this comment would be explaining "this page's startKey".
+      const newKeys = [...keys];
+      newKeys[pageIndex] = response.startKey;
+      setKeys(newKeys);
+    };
+    fetchAndUpdateKeys();
+  }, [agentFilter, pageIndex]);
```
ravicious, line 97: Yeah, that looks good, one last suggestion here that I have is to pass `agentFilter` as an arg to `fetch` to explicitly show that `fetch` depends on that value.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts
```diff
@@ -0,0 +1,159 @@
...
+      };
+    }
+    return null;
+  }
+
+  useEffect(() => {
+    const fetchAndUpdateKeys = async () => {
+      const [response, err] = await fetch(keys[pageIndex - 1]);
+      // the error will be handled via the fetchAttempt outside
+      // return early here as there are no keys to update
```
ravicious, line 82:
```suggestion
      // The error will be handled via the fetchAttempt outside.
      // Return early here as there are no keys to update.
```

packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts
```diff
@@ -0,0 +1,159 @@
...
+  const { clusterUri, documentUri } = useClusterContext();
+  const [pageIndex, setPageIndex] = useState(0);
+  const [keys, setKeys] = useState<string[]>([]);
+  const [agentFilter, setAgentFilter] = useState<AgentFilter>({
+    sort: getDefaultSort(),
+  });
+
+  // startKey is used here as a way to paginate through agents returned from
+  // their respective rpcs.
+  const [fetchAttempt, fetch] = useAsync(async (startKey: string) =>
```
ravicious, line 45: `async` is not necessary here.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Servers/Servers.tsx
```diff
@@ -36,43 +41,77 @@ export default function Container() {
...
+                  () => getSshLogins(server.uri),
+                  login => connect(server, login)
+                ),
+            },
+          ]}
+          customSort={customSort}
+          emptyText={
+            fetchAttempt.status === 'success'
+              ? 'No servers found'
+              : 'Searching…'
```
ravicious, line 109: If the fetch attempt fails, the tab looks like this:

failed fetch

We could change this to `(!fetchAttempt.status || fetchAttempt.status === 'loading') ? 'Searching…' : 'No servers found'` but then it'd say "No servers found" on error too. In that situation it'd be better if it just said "Failed to fetch servers".

And since it means the conditional has to be much bigger, IMHO it'd be better to define it above as a variable, let's say `tableEmptyText`, and pass it as a prop here. FWIW, I like to use the switch statement when operation on types such as `fetchAttempt.status`.

ravicious, line 109: BTW, is sorting by address something that you plan to implement or should this column just not be sortable?

avatus: The web ui doesn't support address sort so it seemed to have snuck in there. Disabling.

avatus: I made a switch statement inside `useServers` instead of `useServersideResources` because I think it will give us greater flexibility in the text we return "No _servers_ found" as an example. Also, `useServersideResources` probably shouldn't care how the data is used and `useServers` is more like "ok lets fill a Table specifically with this info and here is the empty text for said table". 

What I'm getting at is: Great idea!! 👍

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Servers/useServers.ts
```diff
@@ -18,6 +18,19 @@ import { useAppContext } from 'teleterm/ui/appContextProvider';
 
 import { useServerSideResources } from '../useServerSideResources';
 
+function getEmptyTableText(status: '' | 'error' | 'processing' | 'success') {
```
ravicious, line 21: I'd probably reexport `status` from `Attempt` from `useAsync` as `AttemptStatus` instead of copying it over. If we ever refactor `useAsync`, it'll be more clear which places need changing (or even TypeScript will be able to automatically substitute some strings for us).

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Servers/Servers.tsx
ravicious: > I made a switch statement inside `useServers` instead of `useServersideResources` because I think it will give us greater flexibility in the text we return "No _servers_ found" as an example. Also, `useServersideResources` probably shouldn't care how the data is used and `useServers` is more like "ok lets fill a Table specifically with this info and here is the empty text for said table".

Good move and the rationale sounds very good too. 👍

avatus:
Ok, opened it as ready to review. Added the `useServersideResources` flow to `Kubes` and `Databases`. Removed the sync button from `DocumentCluster` since that won't be used anymore.

avatus:
webapps.e ref https://github.com/gravitational/webapps.e/pull/436

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Kubes/Kubes.tsx
```diff
@@ -80,3 +113,14 @@ export const renderConnectButtonCell = (
     </Cell>
   );
 };
+
+const DarkenWhileDisabled = styled(Box)`
+      // The timing functions of transitions have been chosen so that the element loses opacity slowly
+      // when entering the disabled state but gains it quickly when going out of the disabled state.
+      transition: opacity 150ms ease-out;
+      &.disabled {
+        pointer - events: none;
```
gzdunek, line 122: It should be replaced by a common `DarkenWhileDisabled` anyway, but something weird happened in this line (I mean `pointer - events`).

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Servers/Servers.tsx
```diff
@@ -16,15 +16,20 @@ limitations under the License.
 
 import React from 'react';
 
-import Table, { Cell } from 'design/DataTable';
+import styled from 'styled-components';
```
gzdunek, line 19: nit: we don't need so many empty lines between imports, something like this is enough:
```typescript
import React from 'react';
import styled from 'styled-components';
import { Box } from 'design';
import { Danger } from 'design/Alert';
import Table, { Cell, ClickableLabelCell } from 'design/DataTable';
import { MenuLogin } from 'shared/components/MenuLogin';
import { SearchPanel } from 'shared/components/Search';
import { SearchPagination } from 'shared/components/Search/SearchPagination';

import * as types from 'teleterm/ui/services/clusters/types';

import { MenuLoginTheme } from '../MenuLoginTheme';

import { useServers, State } from './useServers';
```

packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts
```diff
@@ -0,0 +1,167 @@
...
+import { AgentFilter, AgentLabel } from 'teleport/services/agents';
+import { ResourceKind } from 'teleport/Discover/Shared';
+
+import { ServerSideParams } from 'teleterm/services/tshd/types';
+import { useAppContext } from 'teleterm/ui/appContextProvider';
+import { retryWithRelogin } from 'teleterm/ui/utils';
+
+import { useClusterContext } from '../clusterContext';
+
+export function addAgentLabelToQuery(filter: AgentFilter, label: AgentLabel) {
```
gzdunek, line 13: No need to export this.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts
```diff
@@ -0,0 +1,167 @@
...
+    updateSearch,
+    updateSort,
+    updateQuery,
+    agentFilter,
+    prevPage: prevPage(),
+    nextPage: nextPage(),
+    onAgentLabelClick,
+    pageCount,
+  };
+}
```
gzdunek, line 154: I noticed a small issue: when the page is changed, the `page count` becomes **SHOWING 0 - 0 of 0** for a while. It can be fixed this way:
```suggestion 
const pageCount = useMemo(() => {
    const emptyPageCount = {
      from: 0,
      to: 0,
      total: 0,
    };

    if (!(fetchAttempt.data && fetchAttempt.data.totalCount)) {
      return emptyPageCount;
    }

    const startKeyIndex = keys.indexOf(fetchAttempt.data.startKey);
    if (startKeyIndex < 0) {
      return emptyPageCount;
    }

    const from = startKeyIndex * limit + 1;
    return {
      from,
      to: from + fetchAttempt.data.agentsList.length - 1,
      total: fetchAttempt.data.totalCount,
    };
  }, [fetchAttempt, keys]);
```

packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts
```diff
@@ -30,14 +31,15 @@ export function addAgentLabelToQuery(filter: AgentFilter, label: AgentLabel) {
 const limit = 15;
 
 export function useServerSideResources<Agent>(
+  resourceKind: ResourceKind,
```
gzdunek, line 36: What would you say to have here `defaultSort: SortType` instead of `resourceKind`?
We wanted the _fetch function_ to be passed as a parameter, it makes sense to treat _sorting_ the same way.

gzdunek (commented):
Looks good! 

Question about `ClusterContext`: it has a few methods that are no loner used, are you going to remove them in this PR, or in another one?

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/Databases.tsx
```diff
@@ -37,38 +35,80 @@ export default function Container() {
...
+              headerText: 'Name',
+              isSortable: true,
+            },
+            {
+              key: 'desc',
+              headerText: 'Description',
+              isSortable: true,
+            },
+            {
+              key: 'type',
```
gzdunek, line 82: Thanks for adding this, but it needs to be formatted before displaying. Please take a look at https://github.com/gravitational/webapps/blob/master/packages/teleport/src/services/databases/makeDatabase.ts

gzdunek (approved):
One minor comment left :)

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/Databases.tsx
avatus: Done 👍

avatus:
Thanks for all the work on this @gzdunek and @ravicious !! Any final comments?

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/Databases.tsx
```diff
@@ -37,38 +37,88 @@ export default function Container() {
...
+          columns={[
+            {
+              key: 'name',
+              headerText: 'Name',
+              isSortable: true,
+            },
+            {
+              key: 'desc',
+              headerText: 'Description',
+              isSortable: true,
```
ravicious, line 81: Sorting by description doesn't appear to work, I get the following error:

```
12 UNIMPLEMENTED: sorting by field "desc" for resource "db_server" is not supported
```

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/useDatabases.ts
```diff
@@ -14,19 +14,51 @@ See the License for the specific language governing permissions and
...
+    case 'error':
+      return 'Failed to fetch databases.';
+    case '':
+      return 'Searching…';
+    case 'processing':
+      return 'Searching…';
+    case 'success':
+      return 'No databases found.';
+  }
+}
```
ravicious, line 38: Any reason this is repeated three times instead of having one function which handles all three tables?

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/useDatabases.ts
```diff
@@ -66,8 +98,23 @@ export function useDatabases() {
 
   return {
     connect,
-    dbs,
-    syncStatus,
+    fetchAttempt,
+    dbs: fetchAttempt.data?.agentsList,
```
ravicious, line 102: Sorry that I didn't catch that while I was reviewing `useServers` but I don't think extracting stuff out of `fetchAttempt` is worthwhile long term.

Ideally the hook would expose `fetchAttempt` and then any UI code that needs to work on that data would look at `fetchAttempt`.

At the moment, we return four different values that are based on `fetchAttempt`:

* `fetchAttempt`
* `dbs` - the view can read `fetchAttempt.data?.agentsList || []` instead.
   * Ideally, the `Table` component would work with `Attempt<T>` out of the box. But I understand if we decided not to do it right now.
* `disabledRows` - this can be "calculated" within the component that calls `useDatabases`.
* `emptyTableText` - same as above.

I think the idea is that the hook returns the minimal amount of data required to display the UI and then the component that calls `useDatabases` works with that data to render the UI.

Theoretically, a custom hook like that could be reused in another page but right now it's also preoccupied with returning data that's related only to one specific component that uses it.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Kubes/useKubes.ts
```diff
@@ -14,18 +14,63 @@ See the License for the specific language governing permissions and
...
+    prevPage,
+    nextPage,
+    updateSearch,
+    onAgentLabelClick,
+    updateSort,
+    pageCount,
+  } = useServerSideResources<Kube>(
+    { fieldName: 'name', dir: 'ASC' }, // default sort
+    (params: ServerSideParams) => appContext.resourcesService.fetchKubes(params)
+  );
```
ravicious, line 54: We do a lot of destructuring only to return most of those values from `useKubes` without touching them.

I experimented with this a bit and arrived at this:

<details>

```diff
diff --git a/packages/teleterm/src/ui/DocumentCluster/ClusterResources/Kubes/useKubes.ts b/packages/teleterm/src/ui/DocumentCluster/ClusterResources/Kubes/useKubes.ts
index 77ef9469..a8dd125d 100644
--- a/packages/teleterm/src/ui/DocumentCluster/ClusterResources/Kubes/useKubes.ts
+++ b/packages/teleterm/src/ui/DocumentCluster/ClusterResources/Kubes/useKubes.ts
@@ -38,33 +38,17 @@ function getEmptyTableText(status: AttemptStatus) {
 export function useKubes() {
   const appContext = useAppContext();
   const ctx = useClusterContext();
-  const {
-    fetchAttempt,
-    updateQuery,
-    agentFilter,
-    prevPage,
-    nextPage,
-    updateSearch,
-    onAgentLabelClick,
-    updateSort,
-    pageCount,
-  } = useServerSideResources<Kube>(
+  const { updateSort, ...serverSideResources } = useServerSideResources<Kube>(
     { fieldName: 'name', dir: 'ASC' }, // default sort
     (params: ServerSideParams) => appContext.resourcesService.fetchKubes(params)
   );
+  const { fetchAttempt, agentFilter } = serverSideResources;
 
   return {
+    ...serverSideResources,
     connect: ctx.connectKube,
-    fetchAttempt,
     kubes: fetchAttempt.data?.agentsList,
-    agentFilter,
-    updateQuery,
-    updateSearch,
-    onAgentLabelClick,
-    pageCount,
     disabledRows: fetchAttempt.status === 'processing',
-    nextPage,
-    prevPage,
     emptyTableText: getEmptyTableText(fetchAttempt.status),
     customSort: {
       dir: agentFilter.sort?.dir,
```
</details>

This will become even cleaner if we refactor `customSort` so that it's returned by `useServerSideResources` instead, then we wouldn't have to make special accommodations for it. From what I see, `customSort` is being constructed in the same way in all three hooks.

I checked and it doesn't impair Typescript's "Go to definition" feature. You can go to definition of `updateQuery` in Kubes.tsx and it takes you straight to `useServerSideResources`.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Kubes/useKubes.ts
```diff
@@ -14,18 +14,63 @@ See the License for the specific language governing permissions and
...
+    onAgentLabelClick,
+    pageCount,
+    disabledRows: fetchAttempt.status === 'processing',
+    nextPage,
+    prevPage,
+    emptyTableText: getEmptyTableText(fetchAttempt.status),
+    customSort: {
+      dir: agentFilter.sort?.dir,
+      fieldName: agentFilter.sort?.fieldName,
+      onSort: updateSort,
```
ravicious, line 72: As I mentioned above, this is repeated three times across the hooks, I'd say let's move this into `useServerSideResources`.

ravicious (commented):
I managed to take a look only at the new hooks for kubes and databases and do some light testing of the databases table. I'll continue the review tomorrow.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/Databases.tsx
avatus: These shouldn't be sortable, I'll update

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/useDatabases.ts
avatus: Each table displays slightly different text. I figured it'd be easier to just have the functions be in each table rather than extract

avatus:
@ravicious Updated based on your comments and also propagated those changes to `useDatabases` and `useServers` as well. Made those hooks much cleaner!

packages/shared/services/types.ts
```diff
@@ -48,3 +48,11 @@ export type AuthProvider = {
...
   | 'none'
   | typeof privateKeyEnablingPolicies[number];
+
+export type DbType = 'redshift' | 'rds' | 'gcp' | 'self-hosted';
+export type DbProtocol =
+  | 'postgres'
+  | 'mysql'
+  | 'mongodb'
+  | 'sqlserver'
+  | 'redis';
```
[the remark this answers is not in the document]
avatus: Oh oops. Ended up refactoring and forgot to make the switch. thanks

gzdunek (approved):

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/useDatabases.ts
```diff
@@ -14,51 +14,24 @@ See the License for the specific language governing permissions and
...
-    updateSearch,
-    onAgentLabelClick,
-    updateSort,
-    pageCount,
-  } = useServerSideResources<Database>(
-    { fieldName: 'name', dir: 'ASC' }, // default sort
-    (params: ServerSideParams) =>
-      appContext.resourcesService.fetchDatabases(params)
-  );
+  const { fetchAttempt, documentUri, ...serverSideResources } =
```
ravicious, line 27: Minor nit but `documentUri` doesn't seem to be used anywhere, right? It's used within `useServerSideResources` but it doesn't seem that other places need it at the moment so it probably could not be exposed in the first place.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/useDatabases.ts
```diff
@@ -14,51 +14,24 @@ See the License for the specific language governing permissions and
...
+    useServerSideResources<Database>(
+      { fieldName: 'name', dir: 'ASC' }, // default sort
+      (params: ServerSideParams) =>
+        appContext.resourcesService.fetchDatabases(params)
+    );
 
-  function connect(db: Database, dbUser: string): void {
+  function connect(db: ReturnType<typeof makeDatabase>, dbUser: string): void {
```
ravicious, line 34: Edit: Oh, I misunderstood what's happening here. I suppose for now it's a necessary evil.

Another minor nit but I'd probably use `Database` here. For me `ReturnType` seems like one of those utility types that should be reserved for functions that accept other functions. Most type systems that I've seen don't really use the concept of "I want to use the return type of this other function", I feel like TypeScript has it just because it needs to support all those dynamic JS acrobatics.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Servers/Servers.tsx
```diff
@@ -31,9 +32,21 @@ export default function Container() {
...
+    case 'error':
+      return 'Failed to fetch servers.';
+    case '':
+      return 'Searching…';
+    case 'processing':
+      return 'Searching…';
+    case 'success':
+      return 'No servers found.';
+  }
+}
```
ravicious, line 46: To be honest, I still don't understand why this function needs to be copied three times instead of once and accepting the parts that change through args. 😶‍🌫️ From what I can tell, it's only the noun that's different.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/useServerSideResources.ts
```diff
@@ -0,0 +1,172 @@
...
+    const newPageIndex = pageIndex - 1;
+    if (newPageIndex > -1) {
+      return () => {
+        setPageIndex(newPageIndex);
+      };
+    }
+    return null;
+  }
+
+  useEffect(() => {
```
ravicious, line 82: I think this hook needs to provide a cleanup function too. While we're waiting for the results for servers, the user can switch to databases and thus unmount the servers component which will attempt to update its state after the response comes in.

When this gets triggered, I see this warning in the console:

> react-dom.development.js?a814:88 Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
>     in Container (created by ClusterResources)

gravitational/webapps#1377 needs to be merged in first though.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Servers/Servers.tsx
```diff
@@ -35,44 +32,89 @@ export default function Container() {
...
+            {
+              key: 'hostname',
+              headerText: 'Hostname',
+              isSortable: true,
+            },
+            {
+              key: 'addr',
+              headerText: 'Address',
+              isSortable: false,
+              render: renderAddressCell,
```
ravicious, line 91: Idk what has changed exactly, but the address column breaks up that arrow next to tunnel now.

address column comparison

packages/teleterm/src/ui/DocumentCluster/DocumentCluster.tsx
```diff
@@ -91,12 +90,6 @@ export function Cluster() {
             {`${clusterCtx.state.clusterName}`}
           </Text>
         </Text>
-        <Flex alignItems="center">
-          <ClusterSearch onChange={clusterCtx.changeSearchValue} />
-          <ButtonPrimary ml={2} size="small" onClick={clusterCtx.sync}>
-            Sync
```
ravicious, line 97: Should we replace this button with some kind of "Refresh" button? Right now there's no way to simply repeat the request with the current params. It could use the same params sans the page – idk if it's feasible to do refresh while keeping page position given how pagination works on our clusters.

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/Databases.tsx
```diff
@@ -36,39 +35,93 @@ export default function Container() {
...
+    case '':
+      return 'Searching…';
+    case 'processing':
+      return 'Searching…';
+    case 'success':
+      return 'No databases found.';
+  }
+}
+
 function DatabaseList(props: State) {
```
ravicious, line 51: This component seems to fetch the databases of the root cluster after I switch to a leaf cluster. Servers and kubes seem to work fine.

packages/shared/components/Search/SearchPanel.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const [query, setQuery] = useState(filter.search || filter.query || '');
+  const [isAdvancedSearch, setIsAdvancedSearch] = useState(!!filter.query);
+
+  useEffect(() => {
+    setIsAdvancedSearch(!!filter.query);
+    setQuery(filter.search || filter.query || '');
+  }, [filter]);
+
+  function onToggle() {
+    setQuery('');
```
ravicious, line 37: Do we need to reset the query here? The Web UI doesn't do this, if I start typing an advanced query only to realize I forgot to turn the toggle, it's pretty disruptive for the app to clear the whole query.

packages/shared/components/Search/SearchPanel.tsx
```diff
@@ -0,0 +1,110 @@
...
+import { StyledPanel } from 'design/DataTable';
+import { PageIndicatorText } from 'design/DataTable/Pager/Pager';
+import InputSearch from 'design/DataTable/InputSearch';
+import { AgentFilter } from 'teleport/services/agents';
+import Toggle from 'teleport/components/Toggle';
+import Tooltip from 'teleport/components/ServersideSearchPanel/Tooltip';
+
+import { PredicateDoc } from './PredicateDoc';
+
+export function SearchPanel({
```
ravicious, line 13: This component seems to get obliterated if you make the window narrow enough, but I suppose we can fix that at some later time.

narrow search

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Databases/useDatabases.ts
avatus: I can remove it. I think we originally used it for the old `retryWithRelogin` and it has become vestigial

packages/teleterm/src/ui/DocumentCluster/ClusterResources/Servers/Servers.tsx
avatus: I don't see the same thing on the current branch. 
Screenshot 2022-11-23 at 3 22 21 PM

packages/teleterm/src/ui/DocumentCluster/DocumentCluster.tsx
avatus: A refresh button outside of the table would prove to be difficult as the state would need to be propped up outside of its current context. If you press 'Enter' in the search bar, it will resend the same query and 'refresh' the page with the most recent info which I think is probably the best/easiest solution right now. Also, going back into the bar and pressing 'Enter' to me is even more intuitive than a refresh button. Thoughts?

packages/shared/components/Search/SearchPanel.tsx
avatus: While I do agree it's not optimal, it isn't quite as bad on my end. I wonder if something was up with the styling in your build? I can't replicate these style errors you've had exactly. 

Screenshot 2022-11-23 at 3 37 14 PM

packages/shared/components/Search/SearchPanel.tsx
avatus: Updated

packages/shared/components/Search/SearchPanel.tsx
ravicious: You're right, something must have been wrong with my setup. I cannot reproduce this issue and the issue with the address now.

packages/teleterm/src/ui/DocumentCluster/DocumentCluster.tsx
ravicious: > A refresh button outside of the table would prove to be difficult as the state would need to be propped up outside of its current context.

I think it should be doable without moving the state, perhaps through `useImperativeHandle` or a callback of some kind, no? The parent component doesn't need to use any of that state, it just so happens that the way the components are organized doesn't allow us to have the button be managed by the appropriate component itself. So IMHO it's fine to use such escape hatch in that case.

> Also, going back into the bar and pressing 'Enter' to me is even more intuitive than a refresh button.

I'd say it's far less intuitive than clicking on a button that says "Refresh" which refreshes the list. 😏

Anyway, it seems that we don't have time for that right now and it's perhaps not the most important thing in the world. The Web UI doesn't offer that functionality as well but at least the browser has a persistent reload button.

ravicious (approved):
🚢

packages/teleterm/src/ui/DocumentCluster/DocumentCluster.tsx
gzdunek: I'm not sure if we need the refresh button. In what situations would it be useful? I think the resource lists change very rarely. For example, Slack and GitHub don't have a separate button to refresh the search results. 

Also if we decide to add a button, it seems to me that it should be placed at the table level, not above it (it would disrupt the hierarchy IMO).

gzdunek: Yeah, it would be really nice to have automatic UI updates. It especially hurts us in access requests, where you don't know if your request has been accepted until you refresh the list.

PR #1309, opened 2022-10-27
[v10] [UI]: Make roles render as labels (#1299)

avatus:
Backport #1299


gzdunek (approved):
Not sure if we need v10 backport :p

PR #1313, opened 2022-10-29
[v10] Warn user when desktop is active (#1297)

ibeckermayer:
Backports #1297 to v10

I found a stumper of a webpack dev server error when manually testing which was fixed by b27c1b831ac2a403519aef2c640899317d21b930

Do not merge until https://github.com/gravitational/teleport/pull/17936 is merged or desktop sessions will break.

### DIFF
--- a/packages/teleport/src/DesktopSession/DesktopSession.story.test.tsx
+++ b/packages/teleport/src/DesktopSession/DesktopSession.story.test.tsx
@@ -28,8 +28,8 @@ test('tdp processing', () => {
 });
 
 test('invalid processing', () => {
-  render(<InvalidProcessingState />);
-  expect(screen.getByTestId('Modal')).toMatchSnapshot();
+  const { getByTestId } = render(<InvalidProcessingState />);
+  expect(getByTestId('Modal')).toMatchSnapshot();
 });
 
 test('connected settings false', () => {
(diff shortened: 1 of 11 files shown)


packages/teleport/src/Apps/AddApp/AddApp.tsx
```diff
@@ -22,7 +22,7 @@ import * as Icons from 'design/Icon';
 
 import useTeleport from 'teleport/useTeleport';
 
-import { TabIcon } from 'teleport/components/Tabs';
+import TabIcon from 'teleport/components/Tabs/TabIcon';
```
ravicious, line 25: Importing the named export doesn't appear to be an issue on my machine.

But [I've run into a similar problem](https://gravitational.slack.com/archives/C02DQ1C2BMW/p1667223568113269) with webpack today. Could you check if removing the `packages/teleport/node_modules/.cache` directory helps?

ravicious (approved):
Leaving an approval but please check if that problem with the import was an issue with cache. If so, we should use the named export.

hatched (approved):
Approved with confirmation of @ravicious's comment.

packages/teleport/src/Apps/AddApp/AddApp.tsx
ibeckermayer: Yeah that fixed it.

PR #1317, opened 2022-10-31
[v11] Connect: Use gap instead of margins for <Label> groups

ravicious:
Backport #1316.

@avatus Could you merge it to v11 after reviewing? I won't be able to do it tomorrow and I'd like this change to be released together with the other change to the role list that was submitted.


avatus:
Sure thing @ravicious

avatus (approved):

PR #1318, opened 2022-11-01
Change `teleport configure` to use `--proxy`

ryanclark:
As part of config v3, `teleport configure` was changed to allow `--proxy`, which will set `proxy_server` to the correct proxy address. This also changed the previous use of `--auth-server` to set `auth_server`, which doesn't work with a proxy URL.

This changes the manual step's command for adding an application to use `--proxy`. Will also need a backport to v11.


kimlisa (approved):
i would always reference the teleport PR to help out reviewer (i had to go dig b/c i was confused by thsi change ;)

hatched (approved):

PR #1333, opened 2022-11-03
Adds support for TDP Notifications, shows "warning" level notifications in a dismissible dropdown rather than a fullscreen modal

ibeckermayer:
Corresponds with https://github.com/gravitational/teleport/pull/18131

Adds support for the TDP Notification message. In the UI, when we receive a warning notification, it will appear as a dismissible item in a dropdown menu hidden behind a button.

image

This change completely removes the non-fatal modal, making it difficult for a bad client or server to DoS a session by inducing warning Notifications.


hatched (approved):

ravicious (approved):

zmb3:
I'd like to call this new message "notice" and keep the original error message only for fatal errors.

zmb3:
Given that we can't get this out until Teleport 12 for compatibility reasons, I'd rather not merge the behavior where we show the model with a non-fatal message. Can we just get the other design implemented?

ibeckermayer:
> Given that we can't get this out until Teleport 12 for compatibility reasons, I'd rather not merge the behavior where we show the model with a non-fatal message. Can we just get the other design implemented?

Sure, I can just add those changes to this PR. Will make it a draft for the time being.

ibeckermayer:
@ravicious @hatched this has been significantly altered since you reviewed.

@gzdunek please double check that my movement of `Notification.tsx` from `teleterm` to `shared` didn't break any of your work.

packages/teleterm/src/ui/services/notifications/notificationsService.ts
```diff
@@ -1,10 +1,11 @@
 import { useStore } from 'shared/libs/stores';
 
-import { ImmutableStore } from 'teleterm/ui/services/immutableStore';
 import {
   NotificationItem,
   NotificationItemContent,
-} from 'teleterm/ui/components/Notifcations';
+} from '@gravitational/shared/components/Notification';
```
ravicious, line 6: The `@gravitational/` part isn't necessary, I think we usually just use the alias to import straight from `shared/something`. But AFAIR VSCode's refactoring tools do insert that prefix automatically.

packages/shared/components/Notification/Notification.tsx
```diff
@@ -1,51 +1,49 @@
...
-    isAutoRemovable: true,
-  },
-};
-
-const autoRemoveDurationMs = 10_000; // 10s
-
-export function Notification(props: NotificationProps) {
+const defaultAutoRemoveDurationMs = 10_000; // 10s
+
+export function Notification(props: NotificationProps & propTypes) {
```
ravicious, line 36: Since this component is now in the shared package, I think it'd be worthwhile to add some stories for it.

There's already some stories for the bigger `Notifications` component in Connect but I think here we could have one simple notification and one with `isAutoRemovable` set to `true`. Just simple enough so that if someone wants to use this shared component, they'll see that there are stories for it.

packages/teleport/src/DesktopSession/WarningDropdown.tsx
```diff
@@ -0,0 +1,148 @@
...
+  const toggleDropdown = () => {
+    if (warnings.length > 0) {
+      setShowDropdown(prevState => !prevState);
+    }
+  };
+
+  // Dropdown is always closed if there are no errors to show
+  useEffect(() => {
+    if (warnings.length === 0) setShowDropdown(false);
+  }, [warnings]);
```
ravicious, line 39: I think this falls under [the "Adjusting some state when a prop changes" section](https://beta.reactjs.org/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes) from You Might Not Need an Effect. There's also some advice on how to address this, I suppose in this case we could move this conditional outside of effect and it would just work.

Here the effect runs on every render anyway because `useEffect` uses `Object.is` to compare effect dependencies.

packages/teleport/src/DesktopSession/useDesktopSession.tsx
```diff
@@ -100,9 +106,52 @@ export default function useDesktopSession() {
...
+        .then(sharedDirHandle => {
+          setDirectorySharingState(prevState => ({
+            ...prevState,
+            isSharing: true,
+          }));
+          tdpClient.addSharedDirectory(sharedDirHandle);
+          tdpClient.sendSharedDirectoryAnnounce();
+        })
+        .catch(() => {
+          setDirectorySharingState(prevState => ({
```
hatched, line 128: Did we want to set a warning in this case?

packages/teleport/src/DesktopSession/WarningDropdown.tsx
```diff
@@ -0,0 +1,148 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import React, { useRef, useState, useEffect } from 'react';
+import { Text, Flex, Button, Card, ButtonIcon } from 'design';
+import styled from 'styled-components';
+import { Notification, NotificationItem } from 'shared/components/Notification';
```
hatched, line 20: `import type`

packages/teleport/src/DesktopSession/WarningDropdown.tsx
```diff
@@ -0,0 +1,148 @@
...
+import { Warning, Close } from 'design/Icon';
+import { orange } from 'design/theme/palette';
+
+export function WarningDropdown(props: Props) {
+  const { warnings, onRemoveWarning } = props;
```
hatched, line 25: personal preference but I prefer to have the destructuring in the function declaration. 

```ts
export function WarningDropdown({ warnings, onRemoveWarning }: Props) {
```

packages/teleport/src/DesktopSession/WarningDropdown.tsx
```diff
@@ -0,0 +1,148 @@
...
+  };
+
+  // Dropdown is always closed if there are no errors to show
+  useEffect(() => {
+    if (warnings.length === 0) setShowDropdown(false);
+  }, [warnings]);
+
+  // Close the dropdown if it's open and the user clicks outside of it
+  useEffect(() => {
+    const handleClickOutside = e => {
```
hatched, line 43: You can use an existing hook `useClickOutside.ts` to handle this.

packages/teleport/src/DesktopSession/TopBar.tsx
```diff
@@ -18,9 +18,12 @@ import styled, { useTheme } from 'styled-components';
 import { Text, TopNav, Flex } from 'design';
 import { Clipboard, FolderShared } from 'design/Icon';
 
+import { NotificationItem } from 'shared/components/Notification';
```
hatched, line 21: `import type`

packages/teleport/src/DesktopSession/DesktopSession.tsx
```diff
@@ -325,10 +255,9 @@ function Session(props: PropsWithChildren<Props>) {
           tdpClient.nuke();
         }}
         userHost={`${username}@${hostname}`}
-        clipboardSharingEnabled={clipboardSharingActive}
         canShareDirectory={directorySharingState.canShare}
         isSharingDirectory={directorySharingState.isSharing}
-        onShareDirectory={onShareDirectory}
+        {...props}
```
hatched, line 260: `TopBar` isn't actually using any of these props that are being passed in from `DesktopSession` so I think this can be removed.

packages/teleport/src/DesktopSession/DesktopSession.tsx
```diff
@@ -325,10 +255,9 @@ function Session(props: PropsWithChildren<Props>) {
           tdpClient.nuke();
         }}
         userHost={`${username}@${hostname}`}
-        clipboardSharingEnabled={clipboardSharingActive}
```
hatched, line 328: I'm not sure anything is passing this into `TopBar` so it can maybe be removed from `TopBar` as well.

packages/teleport/src/DesktopSession/DesktopSession.tsx
```diff
@@ -264,13 +222,13 @@ function Session(props: PropsWithChildren<Props>) {
     tdpClient,
     username,
     hostname,
-    clipboardSharingEnabled,
     setClipboardSharingEnabled,
```
hatched, line 225: I'm not sure this is used any longer?

packages/teleport/src/DesktopSession/WarningDropdown.tsx
ibeckermayer: > I think this falls under [the "Adjusting some state when a prop changes" section](https://beta.reactjs.org/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes) from You Might Not Need an Effect. There's also some advice on how to address this, I suppose in this case we could move this conditional outside of effect and it would just work.

👍

> Here the effect runs on every render anyway because `useEffect` uses `Object.is` to compare effect dependencies.

I don't understand why using `Object.is` means this runs on every render. Wouldn't it only run on renders induced by a call to `setWarnings`?

https://github.com/gravitational/webapps/blob/b22a8a7fa943e71e643e1ca95e1cb8e7f80d7114/packages/teleport/src/DesktopSession/useDesktopSession.tsx#L95

ibeckermayer: Right, but won't it be precisely the same object unless I call `setWarnings`? Or is that not guaranteed by React's `useState`?

packages/teleport/src/DesktopSession/useDesktopSession.tsx
```diff
@@ -16,7 +16,9 @@ limitations under the License.
 
 import { useEffect, useState, useMemo } from 'react';
 import { useParams } from 'react-router';
+
 import useAttempt from 'shared/hooks/useAttemptNext';
+import { NotificationItem } from 'shared/components/Notification';
```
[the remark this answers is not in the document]
ibeckermayer: How about I ignore all of these in this PR and then I'll add [`consistent-type-imports`](https://typescript-eslint.io/rules/consistent-type-imports/) and fix it for the whole codebase in an upcoming PR?

packages/teleport/src/DesktopSession/DesktopSession.tsx
ibeckermayer: It's using `clipboardSharingEnabled, onShareDirectory, warnings, onRemoveWarning`

packages/teleport/src/DesktopSession/DesktopSession.tsx
ibeckermayer: https://github.com/gravitational/webapps/pull/1333#discussion_r1022069855

packages/teleport/src/DesktopSession/DesktopSession.tsx
ibeckermayer: https://github.com/gravitational/webapps/pull/1333#discussion_r1022069855

hatched (approved):

packages/teleport/src/DesktopSession/WarningDropdown.tsx
ravicious: > I don't understand why using `Object.is` means this runs on every render. Wouldn't it only run on renders induced by a call to `setWarnings`?

Oh sorry, I didn't follow how `warnings` is created back to its source. I think you might be right here.

packages/shared/hooks/useClickOutside.ts
```diff
@@ -14,9 +14,12 @@ See the License for the specific language governing permissions and
...
+import { MutableRefObject, useEffect } from 'react';
 
-function useClickOutside(ref, handler) {
+function useClickOutside(
+  ref: MutableRefObject<any>,
+  handler: (e: MouseEvent) => void
+) {
   useEffect(() => {
     const listener = (e: MouseEvent) => {
       if (!ref.current || ref.current.contains(e.target)) {
```
ravicious, line 25: Better types:

```suggestion
  ref: MutableRefObject<Node>,
  handler: (e: MouseEvent) => void
) {
  useEffect(() => {
    const listener = (e: MouseEvent) => {
      if (!ref.current || ref.current.contains(e.target as Node)) {
```

packages/teleport/src/DesktopSession/WarningDropdown.tsx
```diff
@@ -0,0 +1,140 @@
...
+      setShowDropdown(prevState => !prevState);
+    }
+  };
+
+  // Dropdown is always closed if there are no errors to show
+  if (warnings.length === 0 && showDropdown) setShowDropdown(false);
+
+  // Close the dropdown if it's open and the user clicks outside of it
+  const handleClickOutside = (e: MouseEvent) => {
+    if (ref.current && !ref.current.contains(e.target)) {
```
ravicious, line 42: The conditional doesn't seem to be necessary now that we're using `useClickOutside`. We can probably just inline this function.

ravicious (approved):
Added some minor comments but looks good overall!

packages/teleport/src/DesktopSession/WarningDropdown.tsx
```diff
@@ -0,0 +1,140 @@
...
+              {warnings.length} {warnings.length > 1 ? 'Warnings' : 'Warning'}
+            </Text>
+            <ButtonIcon size={1} ml={1} mr={2} onClick={toggleDropdown}>
+              <Close />
+            </ButtonIcon>
+          </Flex>
+          <StyledOverflow>
+            {warnings.map(warning => (
+              <Notification
+                my={2}
```
gzdunek, line 83: Nit: if you want, you could try to replace Y margins with the flex box `gap`. 
Example from Connect [here](https://github.com/gravitational/webapps/pull/1316).
It makes margins more consistent:
`margins`
image
vs
`gap`
image

gzdunek (commented):
Looks good, I left some minor nits ;)

packages/teleport/src/DesktopSession/WarningDropdown.tsx
ibeckermayer: TIL, thanks

packages/shared/components/Notification/Notification.tsx
```diff
@@ -92,11 +90,10 @@ export function Notification(props: NotificationProps) {
         }
       }}
       onClick={toggleIsExpanded}
+      {...props}
```
ravicious, line 93: This emits a warning because it passes design system props as well as `NotificationProps` to a `<div>`:

```
Warning: Unknown event handler property `onRemove`. It will be ignored.
    in div (created by Notification__Container)
```

The solution would be to destructure props into individual NotificationProps and then capture the rest of them into a separate var that can be spread here.

I don't have time to fix this right now but I can try to do it next week.

ibeckermayer: https://github.com/gravitational/webapps/pull/1387

ravicious:
Can this PR be backported to v11?

I have #1474 with some fixes to notifications which depend on the shared component from this PR. I wanted to backport #1474 to v11 but it turned out that v11 doesn't have this component yet.

PR #1334, opened 2022-11-04
Fix table cell styles and correctly sort time column

sielakos:
This PR fixes issue [#17795 ](https://github.com/gravitational/teleport/issues/17795)
There is one minor styling fix and added support for ISO Date Strings sorting in standard sorting support.

There is sibling [PR](https://github.com/gravitational/webapps.e/pull/428) that updates enterprise snapshots


packages/design/src/DataTable/DataTable.story.tsx
```diff
@@ -187,6 +221,27 @@ const extraData: DummyDataType[] = [
...
+    removedDate: new Date(1635322403).toISOString(),
+    tags: ['test12: test12'],
+    bool: false,
+  },
+  {
+    name: 'j-test',
+    desc: 'and another',
+    amount: -20,
+    createdDate: new Date(1635364173).toISOString(),
+    removedDate: new Date(1635322403).toISOString(),
```
michellescripts, line 239: Nit: the two removed dates are actually the same, so you lose the ability to test sorting by removed date in storybook

packages/design/src/DataTable/useTable.ts
```diff
@@ -200,6 +200,10 @@ function sortAndFilter<T>(
         const aValue = a[sort.key];
         const bValue = b[sort.key];
 
+        if (isISODate(aValue) && isISODate(bValue)) {
+          return new Date(aValue).getTime() - new Date(bValue).getTime();
+        }
+
```
michellescripts, line 206: Thoughts on adding a test case for this to `Table.test.tsx`?

kimlisa:
thanks for doing this, while we are at this, could we also fix this. it's very subtle but the `labels` is not middle aligned, and it's because we set a bottom margin to [renderLabelCell](https://github.com/gravitational/webapps/blob/de34c584be8f4fd32300664936d16b89433b5dbb/packages/design/src/DataTable/Cells.tsx#L85) and [ClickableLabelCell](https://github.com/gravitational/webapps/blob/de34c584be8f4fd32300664936d16b89433b5dbb/packages/design/src/DataTable/Cells.tsx#L105), so that when it wraps there are spaces between top and bottom labels.

We should remove this bottom margin and let `line-height` take care of the space between, so i was thinking something like this (or if you have something better):

```ts
// current
const $labels = labels.map(label => (
    <Label mb="1" mr="1" key={label} kind="secondary">
        {label}
    </Label>
return <Cell>{$labels}</Cell>;
));

// with the fix
const $labels = labels.map(label => (
    <Label mr="1" key={label} kind="secondary">
        <div css={{lineHeight: "20px"}}>{label}</div>
    </Label>
return <Cell><div css={{lineHeight: "20px"}}>{$labels}</div></Cell>;
));
```

image

packages/design/src/DataTable/useTable.ts
sielakos: Yeah, that's a good idea

sielakos:
@kimlisa @michellescripts Thanks for review and suggestions! Already applied :)

packages/design/src/DataTable/useTable.ts
michellescripts: I see the new test... but, if I comment out your logic on lines 203-206, the tests still pass

packages/design/src/DataTable/Table.test.tsx
```diff
@@ -319,3 +319,49 @@ describe('sorting by field defined in key and altSortKey', () => {
...
+      />
+    );
+
+    const { rows } = getTableRows();
+
+    expect(rows[0]).toHaveTextContent('a1');
+    expect(rows[1]).toHaveTextContent('a2');
+    expect(rows[2]).toHaveTextContent('a3');
+  });
+});
```
michellescripts, line 367: Nit
line 339: ESLint: Trailing spaces not allowed.(no-trailing-spaces)
EOF: no new line

packages/design/src/DataTable/useTable.ts
sielakos: That's actually a pretty good catch. I had a mistake with my code. I assumed ISO dates are formatted in the same manner as  JavaScript does it. It's not true. Some are smaller even if they include a "Z" postfix. I have changed the code to remove that assumption and pasted explicit ISO strings as generated by our backend to properly test it.

packages/design/src/DataTable/DataTable.story.tsx
```diff
@@ -42,6 +42,65 @@ export const DataTable = () => {
   );
 };
 
+export const DataTableWrappedLabels = () => {
```
michellescripts, line 45: Nit: This isn't really a "new" feature that needs to be documented separately in our storybook. I would update `DataTable` to have a long tag value.

michellescripts:
I think the long labels look a lot better now, but they still feel off WRT to pill style. IK this ticket wasn't looking at labels so, at a certain point, this is probably out of scope. I'll leave it to you @sielakos & @kimlisa 

Screenshot 2022-11-08 at 9 06 03 AM

michellescripts (approved):

packages/design/src/DataTable/Cells.tsx
```diff
@@ -82,12 +82,16 @@ export const DateCell = ({ data }: { data: Date }) => (
...
-    <Label mb="1" mr="1" key={label} kind="secondary">
+    <Label mr="1" key={label} kind="secondary">
       {label}
     </Label>
   ));
 
-  return <Cell>{$labels}</Cell>;
+  return (
+    <Cell>
+      <div style={{ lineHeight: '20px' }}>{$labels}</div>
```
kimlisa, line 92: style uses inline styling, i think from prior discussions we preferred css
```suggestion
      <div css={{ lineHeight: '20px' }}>{$labels}</div>
```

packages/design/src/DataTable/Cells.tsx
```diff
@@ -82,12 +82,16 @@ export const DateCell = ({ data }: { data: Date }) => (
...
+    </Cell>
+  );
```
kimlisa, line 94: also needs same line height styling for the below `ClickableLabelCell` line `115`

kimlisa:
> I think the long labels look a lot better now, but they still feel off WRT to pill style. IK this ticket wasn't looking at labels so, at a certain point, this is probably out of scope. I'll leave it to you @sielakos & @kimlisa
> 

good catch, might as well also fix it, @sielakos see https://github.com/gravitational/webapps/pull/1334#discussion_r1016901157, in adding that line height to the clickable cells that should handle the vertical space between.

And I was playing around, and I think we should just set the border radius to `4px` and have more squarish look, it still looks nice i think? @michellescripts @sielakos what do you guys think?

image

image

image

michellescripts:
@kimlisa I don't have a strong preference for the radius value (anything but 100%). Poking through figma, I see a few pills in the 30-40 range. 

Screenshot 2022-11-08 at 10 24 50 AM

Screenshot 2022-11-08 at 10 22 50 AM

sielakos:
@michellescripts @kimlisa It looks like the issue with labels isn't exactly clear. I would leave it as it is in the current PR.  And if needed there can be new a issue created for labels. Maybe after some consultation with the designer.

packages/design/src/DataTable/Cells.tsx
```diff
@@ -102,13 +106,12 @@ export const ClickableLabelCell = ({
...
       key={`${label.name}:${label.value}`}
       mr="1"
-      mb="1"
       kind="secondary"
       css={`
         cursor: pointer;
       `}
     >
-      {`${label.name}: ${label.value}`}
+      <div css={{ lineHeight: '20px' }}>{`${label.name}: ${label.value}`}</div>
```
kimlisa, line 114: this applies line height to each label, which won't have the affect we want. we need this style wrapping the entire labels 

```
return <Cell><div css={{ lineHeight: '20px' }}>{$labels}</div></Cell>;
```

sielakos:
@michellescripts @kimlisa 

Ok, I have updated labels to account for really big labels :)
It's kinda edge case, but it looks better now.

image

michellescripts (approved):

packages/design/src/DataTable/Cells.tsx
kimlisa: bump. to see this in storybook to visually test it, is to look at resource lists like `nodes`, `apps`, `desktops`, etc.

packages/design/src/Label/Label.jsx
```diff
@@ -59,15 +59,16 @@ const kind = ({ kind, theme }) => {
...
   display: inline-flex;
   align-items: center;
   justify-content: center;
   min-height: 16px;
   line-height: 1.4;
   font-size: 10px;
   font-weight: 500;
-  padding: 0 8px;
+  padding: 6px 16px;
+  margin: 1px 0;
```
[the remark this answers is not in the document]
sielakos: Honestly, it's such an edge case that I don't know if there is a sense in fixing it. Also maybe we don't want multiline labels anyway. So just cutting it with ``some text...`` might be better at a specific length. Right now it feels like I'm trying to fix something that should not have a place. 

@xinding33 what do you think?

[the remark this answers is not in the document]
sielakos: @kimlisa 

I'm trying to focus a little bit more on this issue today. I haven't much time to focus on it today. Please help me to understand the issue with the margin making labels off-centered in the table. 

As far as I can see it only adds one pixel both to the top and bottom without making it off-centered. But I might be missing something here or might have other conditions. 

Without margin:
image

With margin:
image

sielakos:
As far as I can see we have 2 label-related problems here. 

1. Too many labels
2. Too long labels

Both can occur at the same time. 

**As for the 1st problem, I see 2 possible solutions:**
1. Wrap rows with labels as currently done.
2. Set a limit on the number of displayed labels.

I think of those setting the limit is rather a bad solution as it still might not solve the problem and there is no easy way to tell what this limit should be. We can have many short labels that fit perfectly inside a single row while at the same time 2 labels with long values with be displayed on top of each other. 

So I propose to leave it as it is right now. 

**As for 2nd problem, I see 3 possible solutions**
1. Current one with a smaller border radius.  
2. Setting a limit to the length of the label and adding ``...`` when the text is longer. With an optional tooltip to display the full length of the label. 
3. Doing nothing, as it might be not really a scenario that occurs often or even one that we want to support. _Will need to check the data on this one._ 

My personal preference is a second solution. Especially with a tooltip, as we don't need to support multi-row labels there, but still provide a way to see full label value to the user if needed. Might be even good to copy the label value on click, so it can be used somewhere.

xinding33:
@sielakos @kimlisa Hey folks, this issue is getting scope creeped. Let's revert the label margin changes and either leave it as it was originally or set `border-radius` to `10px` and call it a day. We can solve the "too many labels" or "labels too long" issues in the future if it makes sense.

sielakos:
> @sielakos @kimlisa Hey folks, this issue is getting scope creeped. Let's revert the label margin changes and either leave it as it was originally or set `border-radius` to `10px` and call it a day. We can solve the "too many labels" or "labels too long" issues in the future if it makes sense.

@xinding33 I'm not against that, but I think we could solve it also. Just need to decide how.

@jimbishopp what do you think about?

sielakos:
So I have added the changes mentioned above and agreed on the team weekly meeting.

In my opinion, it looks much better now:
image

The only thing I worry about is the tooltip, I have used the default browser, as one ``Popover`` component provided by our design package feels very heavy for such purposes. 

Browser tooltip:
image

Popover component:
image

It would also require adding a bit of JS logic to make it work on hover, which is easy enough to do, but not needed if the browser provides functionality that works out of the box.

The only thing that is issue is it might look a bit different based on the browser and system:

Safari:
image

Edge windows:
Screenshot 2022-11-23 131827

michellescripts:
I wasn't at the meeting, but a heads up that the current iteration no longer responsive. The table has a width, and at smaller window sizes, requires a scroll bar to adjust:

Before:
https://www.loom.com/i/6aba39b71e3f4c7a9ce6d72d2480ef5c

After:
https://www.loom.com/share/a2478d4ddd66489eadea525a0ece1e98

sielakos:
> I wasn't at the meeting, but a heads up that the current iteration is no longer responsive. The table at smaller window sizes requires a scroll bar to adjust:
> 
> Before: https://www.loom.com/i/6aba39b71e3f4c7a9ce6d72d2480ef5c
> 
> After: https://www.loom.com/share/a2478d4ddd66489eadea525a0ece1e98

@michellescripts Thank you for noticing it. You are right about it. I have tried a pure CSS solution, but it will not work correctly.  So I reworked it to be more responsive: 

image

But as a whole, the app is not responsible. Currently, the main issue is the top part above the table is not wrapping correctly:
image

But I think the re-design of the whole page is for sure out of scope for changes in this PR.

But the table by itself is responsive:
image

packages/design/src/DataTable/Cells.tsx
```diff
@@ -80,14 +81,39 @@ export const DateCell = ({ data }: { data: Date }) => (
...
-      {label}
+    <Label mr="1" key={label} kind="secondary">
+      <TextEllipsisLabelContent text={label} maxLength={100} />
     </Label>
   ));
 
-  return <Cell>{$labels}</Cell>;
+  return (
+    <Cell>
+      <div css={{ lineHeight: '20px' }}>{$labels}</div>
```
kimlisa, line 114: since we opted for margin
```suggestion
      <div>{$labels}</div>
```

packages/design/src/Label/Label.jsx
```diff
@@ -59,15 +59,16 @@ const kind = ({ kind, theme }) => {
...
+  border-radius: 10px;
   display: inline-flex;
   align-items: center;
   justify-content: center;
   min-height: 16px;
   line-height: 1.4;
   font-size: 10px;
   font-weight: 500;
-  padding: 0 8px;
+  padding: 0 6px;
```
kimlisa, line 70: revert padding, think the original padding is fine

packages/design/src/DataTable/StyledTable.tsx
```diff
@@ -115,3 +115,41 @@ export const StyledEmptyIndicator = styled.div(
...
+
+export const LabelWrapper = styled.div`
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  max-width: 100%;
+  align-items: center;
+
+  ${labelWidths}
+`;
```
kimlisa, line 155: unless i'm missing something, this is not necessary. CSS has an [ellipsis](https://www.w3schools.com/cssref/tryit.php?filename=trycss3_text-overflow) feature

packages/design/src/Label/Label.jsx
```diff
@@ -59,15 +59,16 @@ const kind = ({ kind, theme }) => {
 
 const Label = styled.div`
```
kimlisa, line 60: we can probably achieve the ellipsis affect here, all with css (this block of code is old, we can change it, for example not sure why we are even using flex here, `inline-block` probably makes more sense

packages/design/src/DataTable/Cells.tsx
```diff
@@ -80,14 +81,39 @@ export const DateCell = ({ data }: { data: Date }) => (
...
+  if (isLonger) {
+    displayText = text.substring(0, 100);
+  }
+
+  return (
+    <LabelContent title={text} dots={isLonger}>
+      {displayText}
+    </LabelContent>
+  );
+};
```
kimlisa, line 103: not necessary (minus the `title` feature, we probably need that), see my other comment about the CSS ellipsis feature

kimlisa (changes requested):
Lets slow down a bit and focus on this comment.

Just to drive it home, we have two types of Labels (screenshots do not contain changes from this PR):

1): [Default Label](https://github.com/gravitational/webapps/blob/99bd5ddfe3a141c469ad5ce93e618c1da222d669/packages/design/src/DataTable/Cells.tsx#L76): these are just simple and don't have special formatting, it just renders an array of string, as array of string, eg: [Users](https://github.com/gravitational/webapps/blob/99bd5ddfe3a141c469ad5ce93e618c1da222d669/packages/teleport/src/Users/UserList/UserList.tsx#L114)

(though it looks like we need to update the code to use the generic Label we created)

image

2): [Clickable Label Cell](https://github.com/gravitational/webapps/blob/99bd5ddfe3a141c469ad5ce93e618c1da222d669/packages/design/src/DataTable/Cells.tsx#L93): this formats an array of string into a `<key>: <value>` format and adds onclick funcs etc. eg: used in any of our resource lists that supports server side filtering and pagination eg: [Desktops](https://github.com/gravitational/webapps/blob/master/packages/teleport/src/Desktops/DesktopList/DesktopList.tsx#L75), `Servers`, `Kubernetes`, etc.

image

The issue I keep finding is inconsistent styling between these two types of labels (eg. our clickable labels atm, the text in the label is not vertically aligned). So to help with visual aid, lets create two data table story components named `DefaultLabels` and `ClickableLabels` (that just has two columns for `name` and `labels`). OR maybe better is to have one story component `DefaultAndClickableLabels`, and render both tables. 

This way any changes, you can easily visually see changes. Also these two types of labels should `look` the same. The examples above links to usage of these labels. 

We should also better name our stories:

image

The folder is already called `DataTable`, we don't need to prefix this, i know it was existing, but lets make it clearer, suggestion:

`DataTable/VariousColumns`
`DataTable/ClientSidePagination` (our server side paginated tables look different)
`DataTable/ISODateStrings`

packages/design/src/DataTable/StyledTable.tsx
sielakos: @kimlisa Thanks for noticing it. I agree and I tried to use it at first, but I found difficulties that I have described below. 

This feature does not work correctly here. We would have to set the dimensions on the label for it to work here. 
And given the specificity of the usage of labels:
1. It is used in different tables that have a different amount of columns and content in them
2. Number of labels and text length of them can vary even in the same table. 
3. They can be used on many different screens and devices.

I found it difficult to use that feature to correctly work in a lot of different use cases. Also given that we want to display more or less 100 chars of the label, I found it was just easier to shorten the label with JS and add dots with CSS after the element (so it cannot be selected). 

It is not a perfect solution, but it works a lot better because:
1. Labels can adjust their dimensions to match the available space and length of their content. 
2. It is simple in implementation compared to juggling a lot of CSS to make the ``ellipsis`` feature work as we want it to work.

I just took an approach that looked simpler. I always try to stick to the KISS principle when possible. I have tried to use the ellipsis CSS feature before and found the difficulties described above. 

I hope that answers your question.

packages/design/src/DataTable/Cells.tsx
sielakos: I have answered your previous comments about why actually using CSS ellipsis is not a good move here in my opinion, and why I made the decision not to do that. 

``LabelConent`` component is actually quite needed because CSS ellipsis does not simplify the code and styling for tags.

packages/design/src/DataTable/StyledTable.tsx
sielakos: To better illustrate the issue I made a [pen](https://codepen.io/sielakos/pen/dyKjqMe) 

Those are basically some divs with an ellipsis inside a table. Which is our use case.

1. So the red case of not setting a dimension for the div does not allow for the ellipsis feature to work. 

2. The green case with ``max-width`` set to some value enables ellipsis, but on a smaller screen it still does not work the table blows horizontally and we observe the horizontal scroll bar of death. 
Also, it would be difficult to set ``max-width`` that works well with multiple different tables.

3. The blue case with text wrapping enabled does not cause the horizontal bar to appear, but also ellipsis does not work and we see full text. We would have to set ``max-height`` here for it to work. And getting it to display 100 chars of the label and work correctly in multiple conditions would be a challenge. 

image

That's why just cutting text is an easier (and therefore better) solution in my opinion. 

All that said, I welcome all better solutions. It's just that I have experimented a lot to make it match 2 criteria:
1. Be responsive
2. Show at max 100 chars of the label, but not less than that if the label is longer. 

And came to the conclusion that a bit of fairly simple JS (it is just the ``.substring`` call after all) is better than a large block of magical CSS.

sielakos:
@kimlisa Thank you for the great explanation, suggestions, and review! That's really helpful. 

I have answered your comments about the ``ellipsis`` feature in detail. So I hope that explains my reasoning and motivation for doing it a bit non-standard. But as always I do welcome feedback and better idea if presented. 

I have refactored the story for the Table and added a new entry for both kinds of labels.
In my opinion, the Clickable and Default labels to look the same:
image

The only thing I worry about is the vertical alignment of the labels. I have tried to make it look good, but honestly, those are 1px wide differences and it's extremely hard for me to tell the difference. I just hope I got it right as I have tried to pay a lot of attention to this.

packages/design/src/DataTable/StyledTable.tsx
kimlisa: can you try this out https://github.com/gravitational/webapps/commit/454c6aec11928e00dc80c06c67bdea6419b3980b, and let me know what you think.

keep in mind our app is not designed for really small screens like mobile phone

kimlisa (approved):
:shipit:

michellescripts (approved):

sielakos:
@michellescripts @kimlisa Thank you both for the reviews! I greatly appreciate the feedback!

I have been talking about this issue with @jimbishopp yesterday and we both agreed that this issue shows us that issue with long labels is difficult to solve in a universal and satisfactory way. So for now I'm removing all ellipsis-related code and hope to merge this PR. 

And the issue related to the number and length of the labels will have to be handled separately. 

So I ask you to please once again look at the code without an ellipsis and see if it's okay :)

Sorry for the very long process. It was very exploratory issue.

kimlisa (approved):

michellescripts:
Sounds good!

PR #1335, opened 2022-11-04
[Discover] Refactor DownloadScript Screens

kimlisa:

Note: suggest reviewing by commit

#### Description
- Tentatively implemented the database download script screen (needs language update)
- Refactored kube's screen with a less hacky version (i had a belated aha moment)
- Refactored server's screen using the new suspense and fallback method
- Add new db script URL and add new field for token request (https://github.com/gravitational/teleport/pull/17653)

#### TODO
- [ ] backport to v11

#### Figma design
https://www.figma.com/file/8DQZl1mYUPLgvKgmA6zPOM/In-Progress-Work?node-id=1299%3A68951

#### Screenshot
- All states available in storybook `Teleport/Discover/Database/*`

Init

image

polling

image


packages/teleport/src/Discover/Kubernetes/HelmChart/HelmChart.story.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+export const PollingSuccess = () => {
+  const { worker, rest } = window.msw;
+  // Use default fetch token handler defined in mocks/handlers
+
+  worker.use(
+    rest.get(cfg.api.kubernetesPath, (req, res, ctx) => {
+      return res(ctx.json({ items: [{}] }));
+    })
+  );
```
alex-kovoy, line 76: Hm...any reasons not to use a high-level hook (useHelpChart.ts) here? The purpose of these hooks is to provide separation between rendering/behavior so it's easy to write stories for different states. You would also want to snapshot some of the stores and it's harder to do without access to the state. Yeah, I would keep treating stories as first-class citizens as that was **the best source of documentation.** With this we not only made it harder to storybook/shopshot-test but also introduced mocks into the environment which is not suited for that (like jest).

to summarize:  treat your storybook as documentation and do the actual testing in `jest`.

packages/teleport/src/Discover/Kubernetes/HelmChart/HelmChart.tsx
```diff
@@ -46,12 +45,14 @@ import type { AgentStepProps } from '../../types';
...
 import type { AgentMeta, KubeMeta } from 'teleport/Discover/useDiscover';
 import type { Kube } from 'teleport/services/kube';
+import type { Poll } from 'teleport/Discover/Shared/CommandWithTimer';
 
 export default function Container(
   props: AgentStepProps & { runJoinTokenPromise?: boolean }
 ) {
   const [namespace, setNamespace] = useState('');
   const [clusterName, setClusterName] = useState('');
+  const [showHelmChart, setShowHelmChart] = useState(props.runJoinTokenPromise);
```
alex-kovoy, line 55: Alternatively, instead of exporting a Container as default export you may consider this naming convention

```ts
// initializes state hooks and inject it to the view
export const HelmChart = (props) => {
  const state = useHelmChart()
  return <HelmChartView {...state} />
}

// renders the state passed via props
export const HelmChartView = props => {
}
```

packages/teleport/src/services/joinToken/joinToken.ts
```diff
@@ -17,35 +17,32 @@ limitations under the License.
...
       .post(
         cfg.getJoinTokenUrl(),
         {
-          roles,
-          join_method: joinMethod,
-          allow: makeAllowField(rules),
+          roles: req.roles,
+          join_method: req.method || 'token',
+          allow: makeAllowField(req.rules || []),
+          agent_matcher_labels: makeLabelMapOfStrArrs(req.agentMatcherLabel),
```
hatched, line 37: as `agentMatcherLabel` is optional can you define a default plz.

packages/teleport/src/services/joinToken/joinToken.test.ts
```diff
@@ -0,0 +1,59 @@
...
+  );
+
+  // Test all fields gets set as requested.
+  const mock: JoinTokenRequest = {
+    roles: ['Node'],
+    rules: [{ awsAccountId: '1234', awsArn: 'xxxx' }],
+    method: 'iam',
+    agentMatcherLabel: [{ name: 'env', value: 'dev' }],
+  };
+  svc.fetchJoinToken(mock);
```
hatched, line 48: Can you split these up into disparate tests plz.

packages/teleport/src/services/agents/make.ts
```diff
@@ -40,3 +44,19 @@ function makeTraces(traces: any): ConnectionDiagnosticTrace[] {
     error: t.error,
   }));
 }
+
+// makeLabelMapOfStrArrs converst an array of type AgentLabel into
```
hatched, line 48:
```suggestion
// makeLabelMapOfStrArrs converts an array of type AgentLabel into
```

packages/teleport/src/Discover/Database/DownloadScript/DownloadScript.tsx
```diff
@@ -0,0 +1,378 @@
...
+  if (pollingTimedOut) {
+    poll = {
+      state: 'error',
+      error: {
+        reasonContents: [
+          <>
+            The command was not run on the server you were trying to add,
+            regenerate command and try again.
+          </>,
+          // TODO (lisa): not sure what this message should be.
```
hatched, line 179: Can you create an issue to track this update as this will likely get forgotten about otherwise.

hatched (commented):
Thanks for this @kimlisa 

I'd also like you to take a look into the comments by @alex-kovoy around moving the state out of the helm chart component.

packages/teleport/src/Discover/Kubernetes/HelmChart/HelmChart.story.tsx
kimlisa: hi @alex-kovoy!!! are you ready to come back yet? :)

You're right in that we need to write more tests, i've just been lax because the discover feature goes through so much  changes and refactoring and we also wanted it out fast. Probably a bad reason to not write tests.

This specific screen (`download/helm`) uses the [suspense](https://17.reactjs.org/docs/concurrent-mode-suspense.html) technique, which means I have to also use [error boundary](https://17.reactjs.org/docs/concurrent-mode-suspense.html#handling-errors) to gracefully handle network errors and allow for retries in the middle of the flow. 

The suspense technique doesn't use the `useAttempt` method and is a very different way to compose components, but suspense is equivalent to `attempt.status === 'loading'` and error boundary is equivalent to `attempt.status === 'failed'`. In order to see what these state looks like, i had to use `msw` and setup contexts for it. And since I was using this setup anyways, I decided to just use them for all states instead of spending extra time making dummy components. I am unsure if there is a better way to do it.

The `msw` lib allowed me to create all states possible and then the stories were used for design review.

This is the only screen that uses this technique... and I'm still on the fence about it (it makes it harder to mock and write stories), but since it was already done we decided to just go with it with the possibility of refactoring it out later or just scope usage of suspense only to this screen? Not sure b/c some devs were leery on using an experimental feature.

Other screens (that I personally create) uses the `smart/dummy` approach as you suggested. But with that said... I don't think this approach is being enforced as there are devs who does not like this approach (and I don't review all PRs). There was a whole thread of discussion on it with the result that looked like half who likes this approach (me included) and half who doesn't. And I honestly don't know what the result of that discussion was... For this particular feature, I decided that getting it out was more important.

kimlisa: I am honestly unsure what our process is anymore. Are we enforcing writing stories? Are we enforcing smart/dummy approach? Right now it feels like anything goes. Maybe that's ok, as long as you can write stories for it? @hatched @ryanclark @ravicious @gzdunek @rudream @ibeckermayer @JanKaczmarkiewicz @michellescripts

kimlisa: I'll try to mock out the important bits of the `download/helm` so we can snapshot it (i'm also unsure of the status of using snapshots), so then `msw` is just used to see what failed/loading state looks like and these states aren't important to snapshot.

alex-kovoy: @kimlisa you summarize it very well why to avoid using `Suspense`. The short answer is that it's harder to test/storytell. If you want to treat your tests and stories as first-class citizens it's better to avoid Suspense. Also this App is not a static website where the basics of Suspense would suffice (hard to imagine vscode using suspense to show a loading indicator)

> The msw lib allowed me to create all states possible and then the stories were used for design review.

 Reminded me of [this commercial ](https://www.youtube.com/watch?v=g5nMbnxeQjA&ab_channel=NJMInsuranceGroup)

> There devs who does not like this approach (and I don't review all PRs). 

Devs can write tests against smart `containers` and also create stories and snapshots (rendering) against dummy `views` if needed. This separation just makes it easier (less code) to do so. Not sure what would be the reason to couple `state` and `view` together while writing tests/stories shows an opposite (basic CS concept)

alex-kovoy: @kimlisa  
> I am honestly unsure what our process is anymore. Are we enforcing writing stories? Are we enforcing smart/dummy approach? Right now it feels like anything goes. Maybe that's ok, as long as you can write stories for it?

I would enforce writing tests + a few stories (where it makes sense) to document behavior/view (helps a lot when regression/design changes). Just the fact of treating tests and stories as _first-class citizens_ would automatically drive your decision making towards more friendly composition/architecture/less code.

@gzdunek  `term` is still missing many tests!! :) and no stories.

packages/teleport/src/services/joinToken/joinToken.ts
kimlisa: it does provide a default (the `makeLabelMapOfStrArrs`), it's also tested with empty values and comes out as expected

packages/teleport/src/Discover/Database/DownloadScript/DownloadScript.tsx
kimlisa: tracked here https://github.com/gravitational/teleport/issues/16858

packages/teleport/src/Discover/Kubernetes/HelmChart/HelmChart.story.tsx
kimlisa: > I'll try to mock out the important bits of the download/helm so we can snapshot it (i'm also unsure of the status of using snapshots), so then msw is just used to see what failed/loading state looks like and these states aren't important to snapshot.

@hatched i've decided to do this later, i'm planning to come back to refactor more stuff anyway

packages/teleport/src/Discover/Database/DownloadScript/DownloadScript.tsx
```diff
@@ -0,0 +1,378 @@
...
+  }
+
+  function removeLabel(index: number) {
+    if (labels.length === 1) {
+      // Since at least one label is required
+      // instead of removing the row, clear
+      // the input and turn on error.
+      const newList = [...labels];
+      newList[index].name = '';
+      newList[index].value = '';
```
JanKaczmarkiewicz, line 261: [nit]
This still mutates orginal `labels[index]` (since `[...labels]` creates only shallow copy all items still points to the same items from label ). I think react components should not mutate props passed to it. To avoid this: 
```suggestion
      const newList = [...labels];
      newList[index] = { name: '', value: '' };
```

packages/teleport/src/Discover/Database/DownloadScript/DownloadScript.tsx
```diff
@@ -0,0 +1,378 @@
...
+    event: React.ChangeEvent<HTMLInputElement>,
+    index: number,
+    labelField: 'key' | 'value'
+  ) => {
+    const { value } = event.target;
+    const newList = [...labels];
+    if (labelField === 'key') {
+      newList[index].name = value;
+    } else {
+      newList[index].value = value;
```
JanKaczmarkiewicz, line 282: [nit]
`labelField` can be `keyof AgentLabel` so we don't need extra layer of mapping: `'key' | 'value'`
```suggestion
    labelField: keyof AgentLabel
  ) => {
    const { value } = event.target;
    const newList = [...labels];
    newList[index] = {...newList[index], [labelField]: value}
```

packages/teleport/src/Discover/Shared/CommandWithTimer/CommandWithTimer.tsx
```diff
@@ -0,0 +1,123 @@
...
+  border: 2px solid
+    ${props => {
+      switch (props.pollState) {
+        case 'error':
+          return props.theme.colors.danger;
+        case 'success':
+          return props.theme.colors.success;
+        default:
+          // polling
+          return '#2F3659';
```
JanKaczmarkiewicz, line 113: This magic token is used in other places. Maybe we should move it to theme?

packages/teleport/src/Discover/Kubernetes/HelmChart/HelmChart.story.tsx
```diff
@@ -0,0 +1,154 @@
...
+  return (
+    <Provider timeout={20}>
+      <HelmChart runJoinTokenPromise={true} />
+    </Provider>
+  );
+};
+export const Processing = () => {
+  const { worker, rest } = window.msw;
```
JanKaczmarkiewicz, line 100: [nit]
```suggestion
const { worker, rest } = window.msw;
export const Processing = () => {
```
can be moved at module level to avoid repetition in other stories. 

Whats the reason behind assigning `msw` to `window`? I think it could work nicely as module:
```
import { worker, rest } 'utils/msw'
```

JanKaczmarkiewicz (approved):

russjones (approved):
Looks good to me.

packages/teleport/src/Discover/Kubernetes/HelmChart/HelmChart.story.tsx
ravicious: > Are we enforcing writing stories? Are we enforcing smart/dummy approach? Right now it feels like anything goes. Maybe that's ok, as long as you can write stories for it?

What I was advocating for in discussions under gravitational/teleport#15813 pretty much aligns with what Alexey said here.

packages/teleport/src/Discover/Kubernetes/HelmChart/HelmChart.story.tsx
kimlisa: i guess i didn't want the worker to be specific to teleport (like if teleterm wanted to define their handlers, though not sure how that works with grpc) 

.storybook/preview.js
```
// Checks we are running non-node environment (browser)
if (typeof global.process === 'undefined') {
  const worker = setupWorker(...handlersTeleport);  <--- can join more handlers
  worker.start();

  // So it can be accessed in stories more easily.
  window.msw = { worker, rest };
}
```

packages/teleport/src/Discover/Shared/CommandWithTimer/CommandWithTimer.tsx
JanKaczmarkiewicz: You are right, lets wait

PR #1340, opened 2022-11-08
[v9] Add support for leaf cluster that does not support new api `ListResources`

kimlisa:
fixes https://github.com/gravitational/teleport/issues/18171

Note: easier to review by commit

#### Description
Starting v9.1, changes were made to support backend pagination/filtering. But fallbacks were not provided for the web UI, resulting in user's not being able to see a list of resources from v8 clusters. This PR adds fallback support by falling back to a client side filtering table when serverside pagination is not supported. 

The backend will return a `null` (to mean field not set, different from `undefined`) for both the `startKey` and `totalCount` and this will be the flag to determine if pagination is not supported.

#### TODO
- [ ] requires backend changes to merge first https://github.com/gravitational/teleport/pull/18262
- [ ] forward port to v10

#### Screenshot
the server side paginating table

image

the fallback table

image

### DIFF
--- a/packages/teleport/src/Console/DocumentNodes/DocumentNodes.story.test.tsx
+++ b/packages/teleport/src/Console/DocumentNodes/DocumentNodes.story.test.tsx
@@ -15,7 +15,11 @@ limitations under the License.
 */
 
 import React from 'react';
-import { Document, createContext } from './DocumentNodes.story';
+import {
+  Document,
+  PaginationUnsupported,
+  createContext,
+} from './DocumentNodes.story';
 import { waitFor, render } from 'design/utils/testing';
 
 test('render DocumentNodes', async () => {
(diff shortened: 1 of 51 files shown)


kimlisa:
should i add a info that says the following:
image

packages/teleport/src/Console/DocumentNodes/DocumentNodes.story.test.tsx
```diff
@@ -28,3 +32,14 @@ test('render DocumentNodes', async () => {
   await waitFor(() => expect(ctx.fetchNodes).toHaveBeenCalledTimes(1));
   expect(container.firstChild).toMatchSnapshot();
 });
+
+test('render DocumentNodes pagination unsupported', async () => {
+  const ctx = createContext(true /* paginationUnsupported */);
```
ravicious, line 37: In this situation I'd consider passing an object with options instead of having to document what the boolean means.

packages/teleport/src/services/apps/apps.ts
```diff
@@ -31,6 +31,8 @@ const service = {
         apps: items.map(makeApp),
         startKey: json?.startKey,
         totalCount: json?.totalCount,
+        paginationUnsupported:
+          json?.startKey === null && json?.totalCount === null,
```
ravicious, line 35: I left a comment on the backend PR regarding how we detect if pagination is supported or not. In my experience, in most cases accepting only null and not undefined generates more headaches than structuring the code so that both null and undefined result in the same behavior.

ravicious (approved):
It's a bit unfortunate that we have to copy the same code for each resource, but this PR is obviously not the place to fix that.

BTW, why do we need to forward port it to v10? Is it in order to support v9.0 leaf clusters which might not support pagination as well when talking with a v10 root cluster?

packages/teleport/src/services/apps/apps.ts
kimlisa: replied https://github.com/gravitational/teleport/pull/18262#discussion_r1018214854

kimlisa:
> BTW, why do we need to forward port it to v10? Is it in order to support v9.0 leaf clusters which might not support pagination as well when talking with a v10 root cluster?

Yes you're correct. Just fyi:

```
Marking for deletion:

Major release: +1
 ie if 7.0.0, mark DELETE IN 8.0.0

Minor/Path release: +2
ie if 7.1.0, mark DELETE IN 9.0.0
```

kimlisa:
Adds fallback support by falling back to a client side 
filtering table when serverside pagination is not 
supported.

fixes https://github.com/gravitational/teleport/issues/18171

PR #1343, opened 2022-11-10
Prevent non-https protocol from opening external windows

avatus:
Fixes https://github.com/gravitational/teleport-private/issues/173

As far as I'm aware, we only open two external links and both of them are `https` protocol. This only allows that protocol and prevents any links such as `window.open("ftp://badurl.here")` from being opened.

### DIFF
--- a/packages/teleterm/src/main.ts
+++ b/packages/teleterm/src/main.ts
@@ -118,6 +118,9 @@ function initializeApp(): void {
       const url = new URL(details.url);
 
       function isUrlSafe(): boolean {
+        if (url.protocol !== 'https:') {
+          return false;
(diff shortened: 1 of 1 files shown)


gzdunek (approved):

zmb3 (approved):

ravicious (approved):
> As far as I'm aware, we only open two external links and both of them are https protocol.

There's one in `DocumentDatabase` leading to database docs, one in the menu bar leading to Connect docs and two in the feedback form leading to our GitHub and Cloud signup form. The feeback form can also render a link to discussions on GitHub if your feedback is too long.

PR #1344, opened 2022-11-11
capture events from webapps

michellescripts:
[Supports reporting work](https://github.com/gravitational/cloud/issues/2520)
- [teleport.e PR](https://github.com/gravitational/teleport.e/pull/593)
- [teleport PR](https://github.com/gravitational/teleport/pull/18142)

tracks user events during the onboarding flow, including banner clicks to upgrade to paid plan

Screenshot 2022-12-02 at 11 02 13 AM


packages/teleport/src/Main/OnboardDiscover/OnboardDiscover.tsx
```diff
@@ -55,10 +72,10 @@ export function OnboardDiscover({
         </Text>
       </DialogContent>
       <DialogFooter>
-        <ButtonPrimary width="100%" size="large" onClick={() => onOnboard()}>
+        <ButtonPrimary width="100%" size="large" onClick={proceed}>
           add my first resource
         </ButtonPrimary>
-        <ButtonText pt={2} width="100%" size="large" onClick={onClose}>
+        <ButtonText pt={2} width="100%" size="large" onClick={delay}>
```
kimlisa, line 78: nit: `handleClose` (i won't nit the others :B)

packages/teleport/src/Welcome/NewCredentials/NewMfaDevice.tsx
```diff
@@ -66,6 +70,11 @@ export function NewMfaDevice(props: Props) {
   ) {
     e.preventDefault(); // prevent form submit default
 
+    userEventService.capturePreUserEvent({
+      event: userEvents.preUser.onboard.registerChallengeSubmitEvent,
```
kimlisa, line 74: should we also capture this for `NewPasswordlessDevice`?

packages/teleport/src/Welcome/Welcome.tsx
```diff
@@ -21,13 +21,24 @@ import history from 'teleport/services/history';
...
+import userEvents from 'teleport/services/userEvent/UserEvents/userEvents';
+
+import useToken from 'teleport/Welcome/useToken';
+
 import { NewCredentials } from './NewCredentials';
 import { CardWelcome } from './CardWelcome';
 
 export default function Welcome() {
   const { tokenId } = useParams<{ tokenId: string }>();
+  const { resetToken } = useToken(tokenId);
```
kimlisa, line 35: not sure about this, the `useToken` fetches token data from the back, and doing this will fetch it here only to get the username, and fetch it again when user click "next" (form screen, line 70 and 73)

if the get started event is important, perhaps we can on [fetch token on init](https://github.com/gravitational/webapps/blob/master/packages/teleport/src/Welcome/useToken.ts#L38), also send the capture event? This fetch on init happens in [NewCredentials](https://github.com/gravitational/webapps/blob/5b3a2dfa63faf2e11872a2805b9eabaac0731ffa/packages/teleport/src/Welcome/NewCredentials/NewCredentials.tsx) component

if it's important to capture the user event like "click" (ensure user performs an actual action) before we send an event, maybe we can send a state with the route similar to something we do here: https://github.com/gravitational/webapps/blob/master/packages/teleport/src/components/AgentButtonAdd/AgentButtonAdd.tsx#L45

packages/teleport/src/Welcome/Welcome.tsx
```diff
@@ -21,13 +21,21 @@ import history from 'teleport/services/history';
...
 
+import { CaptureEvent, userEventService } from 'teleport/services/userEvent';
+import useToken from 'teleport/Welcome/useToken';
+
 import { NewCredentials } from './NewCredentials';
 import { CardWelcome } from './CardWelcome';
 
 export default function Welcome() {
   const { tokenId } = useParams<{ tokenId: string }>();
+  const tokenProps = useToken(tokenId);
```
kimlisa, line 32: sorry I should've explained the reason behind this whole welcome screen. if my understanding is correct, we had problems with crawlers clicking on the invite/reset link before the user did (email link), which would read the token and generate the QR code. By the time user got to it, the QR code will expire. This welcome screen prevented the crawler from reading the token (and generation of the QR code). This was an issue for users who opted for second factor `otp`. @rudream can you double check please?

history: https://github.com/gravitational/webapps/pull/479

michellescripts: Oh yes great context

kimlisa: i think it might be a good idea to leave a comment on this component explaining why it exists (with a link to PR perhaps)

michellescripts: I will remove this event for the first iteration and rethink how to solve this once we go live next week with the first round of events. I'll make a note to add a comment to the component at that time! Thank you for the history!

rudream: That's correct @kimlisa, there was an issue where users who received invite links sent by email (ie for Cloud) would get `totp code invalid` errors when trying to sign up. The problem was caused by mail web crawlers who would open the link and trigger the TOTP QR code to be generated before the user did. Moving it to a second page and only allowing the web crawler to see a welcome screen solved the issue.

packages/teleport/src/Welcome/NewCredentials/NewMfaDevice.tsx
michellescripts: Added to the list for a follow-up PR!

packages/teleport/src/components/RecoveryCodes/RecoveryCodes.tsx
```diff
@@ -40,6 +42,14 @@ export default function RecoveryCodesDialog({
     });
   };
 
+  const handleContinue = () => {
+    userEventService.capturePreUserEvent({
+      event: CaptureEvent.PreUserOnboardRecoveryCodesContinueClickEvent,
+      username: username,
+    });
+    onContinue();
+  };
```
kimlisa, line 51: sorry i didn't catch this earlier, components that are defined in `teleport/src/components` means it's shared in multiple places

looks like this use of component will send out an event in the following flows:

1) reset/invite (what our target is)
2) user is logged in -> goes to account settings -> create new recovery codes (`preuser` capture would be incorrect then)
3) going through the recovery flow

not sure if it was okay to be sending out an event for the other cases?

michellescripts: Yes, you're right. And I think that is acceptable. If anything, we should rename this action to remove the onboarding prefix since that's misleading. It's a step in onboarding, but not particular to it. I will chat with the team to see if we should rename it now or follow up (since it'll need to change at every layer: webapps, teleport, prehog).

michellescripts: Renamed for Clarity:
[prehog](https://github.com/gravitational/prehog/pull/51/files)
[teleport](https://github.com/gravitational/teleport/pull/18142/commits/1335f114146bcb67053b865f70e718606e2ad18b)
[webapps](https://github.com/gravitational/webapps/pull/1344/commits/09c1e887eb7f70dfa10e60d4bebd579ac39f42cd)

packages/teleport/src/components/RecoveryCodes/RecoveryCodes.tsx
```diff
@@ -40,6 +42,14 @@ export default function RecoveryCodesDialog({
     });
   };
 
+  const handleContinue = () => {
+    userEventService.capturePreUserEvent({
+      event: CaptureEvent.PreUserRecoveryCodesContinueClickEvent,
+      username: username,
+    });
+    onContinue();
```
kimlisa, line 50: should we put a comment that says that this event is only successfully captured with reset/invite for the moment? it looks like if we don't send in the username with the `capturePreUserEvent` request it fails silently to capture

kimlisa (approved):

mcbattirola (approved):

rudream (approved):

PR #1351, opened 2022-11-14
Update to Electron 21

gzdunek:
Closes https://github.com/gravitational/webapps.e/issues/369 
Closes https://github.com/gravitational/teleport-private/issues/175

Drone build https://drone.platform.teleport.sh/gravitational/teleport/17585 

I quickly tested the updated app and didn't see any issues.
  
I removed custom `node-gyp` resolution, it doesn't have any effect.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:16.18.1-slim
+FROM node:16.18-slim
 RUN apt-get update && apt-get install git g++ make python3 tree -y
 
 RUN mkdir -p web-apps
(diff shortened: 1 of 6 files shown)


avatus (approved):
Worked for me, didn't notice anything

Dockerfile
```diff
@@ -1,5 +1,5 @@
-FROM node:16.3-slim
-RUN apt-get update && apt-get install git g++ make python tree -y
+FROM node:16.18.1-slim
```
ravicious, line 1: Do we need it to be pinned to a specific patch version or can we keep it pinned to a minor version like it was before?

gzdunek: No, we don't have to. Changed to `16.18`.

ravicious (approved):
I did some tests on the dev version: window restoration, file transfer, terminal. Everything appears to be working fine.

PR #1356, opened 2022-11-15
Fix null array

kimlisa:
fixes https://github.com/gravitational/teleport/issues/18369

#### Description
fixes bug where we try to iterate over a `null` array. the source was from `services/nodes/makeNodes`. I also placed checks for `arrayStrDiff` thinking this was the source of bug initially but it wasn't but why not.

#### TODO
- [ ] backport to v11
- [ ] backport to v10

### DIFF
--- a/packages/teleport/src/services/nodes/makeNode.ts
+++ b/packages/teleport/src/services/nodes/makeNode.ts
@@ -16,24 +16,17 @@ limitations under the License.
 
 import { Node } from './types';
 
-export default function makeNode(json): Node {
-  const {
-    id,
-    siteId,
-    hostname,
-    addr,
-    tunnel,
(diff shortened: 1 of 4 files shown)


packages/teleport/src/services/nodes/makeNode.ts
```diff
@@ -16,24 +16,17 @@ limitations under the License.
...
-  const {
-    id,
-    siteId,
-    hostname,
-    addr,
-    tunnel,
-    tags = [],
-    sshLogins = [],
-  } = json;
+export default function makeNode(json: any): Node {
```
hatched, line 19: It would be nice if we could type these instead of using `any` :)

hatched (approved):

PR #1361, opened 2022-11-16
Use a single websocket for SSH connections

hatched:
Instead of having a HTTP call to get or create a new ssh session. All of that is handled via the websocket connection. 

In addition this branch adds OpenTelemetry support to allow us to collect metrics on the performance implications of this change. (Note: they were dramatically improved)

Metrics are off by default.

To enable metrics run `localStorage.setItem('enable-telemetry', true)` and then refresh the UI.

Needs to land with: https://github.com/gravitational/teleport/pull/18523


packages/teleport/src/telemetry-boot.ts
```diff
@@ -0,0 +1,60 @@
...
+      new UserInteractionInstrumentation(),
+      new XMLHttpRequestInstrumentation(),
+      new FetchInstrumentation(),
+    ],
+  });
+
+  const resource = Resource.default().merge(
+    new Resource({
+      [SemanticResourceAttributes.SERVICE_NAME]: 'web-ui',
+      [SemanticResourceAttributes.SERVICE_VERSION]: '0.1.0',
```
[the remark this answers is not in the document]
hatched: Yeah that would be nice. Unfortunately at this point we haven't yet created the connection to get the version of teleport yet. (The UI doesn't have a version # separate from teleport). I'll make a note to look into how we might be able to update this.

packages/teleport/src/telemetry-boot.ts
```diff
@@ -0,0 +1,60 @@
...
+import { ZoneContextManager } from '@opentelemetry/context-zone';
+
+export function instantiateTelemetry() {
+  // Optionally register automatic instrumentation libraries
+  registerInstrumentations({
+    instrumentations: [
+      new DocumentLoadInstrumentation(),
+      new UserInteractionInstrumentation(),
+      new XMLHttpRequestInstrumentation(),
+      new FetchInstrumentation(),
```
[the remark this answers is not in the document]
hatched: I think we can leave these as-is for now and then revisit them a bit later when we have spent some additional time using the metrics.

packages/teleport/src/telemetry-boot.ts
```diff
@@ -0,0 +1,60 @@
...
+    resource: resource,
+  });
+
+  provider.addSpanProcessor(new BatchSpanProcessor(new ConsoleSpanExporter()));
+  provider.addSpanProcessor(
+    new BatchSpanProcessor(
+      new OTLPTraceExporter({
+        timeoutMillis: 15000,
+        url: `${window.location.origin}/v1/webapi/traces`,
+        concurrencyLimit: 10, // an optional limit on pending requests
```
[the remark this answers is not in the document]
hatched: As were pinging Teleport itself these values can probably be lowered as there is limited latency and we don't ever have more than 1 ping in flight at a time but I don't really see any reason to reduce them at this time. (famous last words 😉 )

packages/teleport/src/boot.tsx
```diff
@@ -29,6 +30,13 @@ cfg.init(window['GRV_CONFIG']);
 // use browser history
 history.init();
 
+const enableTelemetry =
+  localStorage.getItem('enable-telemetry') === 'true' ? true : false;
+
+if (enableTelemetry) {
+  instantiateTelemetry();
+}
```
ryanclark, line 38: Can you do something like (no idea if this is the correct syntax) 

```suggestion
if (enableTelemetry) {
  import('./telementry-boot').then((module) => module.instantiateTelemetry());
}
```

So we're not including 13 telemetry packages in the main JS bundle if it's not needed

hatched: Interesting idea, however we need these in the package as we want the user to be able to enable it as they see fit without having to rebuild the UI.

packages/teleport/src/boot.tsx
```diff
@@ -29,6 +30,13 @@ cfg.init(window['GRV_CONFIG']);
 // use browser history
 history.init();
 
+const enableTelemetry =
+  localStorage.getItem('enable-telemetry') === 'true' ? true : false;
```
ryanclark, line 34:
```suggestion
const enableTelemetry =
  localStorage.getItem('enable-telemetry') === 'true';
```

packages/teleport/src/boot.tsx
ryanclark: How is it enabled? You could also import and instantiate it when toggled in the UI

hatched: Oh sorry you meant to not have it in the main bundle not not included at all in the bundle :) I can give this a try.

It's enabled by running `localStorage.setItem('enable-telemetry', true)` in the console then refreshing the UI.

kimlisa (commented):
did not look at code yet, just went through some test plan related to terminal and found the following:

- a user who types `exit` in the terminal, for both user the session should end, but it remains open for the other user (after some time, doing it immediately works as expected), may be related to the second bullet point

- after some time, we prevent the user from `accidentally` closing the tab, and require confirmation from user before closing, it throws an error when we meet that limit and try to exit the tab:
image

- scp upload and download does not work:
image

hatched:
Thanks for the QA @kimlisa I've addressed the last 2 points but the first one also appears to be an issue on `master`.

packages/teleport/src/Console/consoleContext.tsx
```diff
@@ -115,34 +120,37 @@ export default class ConsoleContext {
...
+
+      return Promise.all(requests).then(results => {
+        let parties: ParticipantList = {};
+        for (let i = 0; i < results.length; i++) {
+          parties = {
+            ...results[i],
+          };
+        }
+
+        this.storeParties.setParties(parties);
```
JanKaczmarkiewicz, line 151: This logic is equivalent to the following suggestion:
Only the last element of the results is assigned to the parties
```suggestion
        this.storeParties.setParties(results[results.length-1]);
```
Or should this merge results into one object? :
```suggestion
        const parties: ParticipantList = {};
        for (const result in results) {
            Object.assign(parties, result);
        }
        this.storeParties.setParties(parties);
```

packages/teleport/src/boot.tsx
hatched: Heh woops, changed here and in `e`.

packages/teleport/src/telemetry-boot.ts
```diff
@@ -0,0 +1,60 @@
...
+import { B3Propagator } from '@opentelemetry/propagator-b3';
+import { UserInteractionInstrumentation } from '@opentelemetry/instrumentation-user-interaction';
+import {
+  CompositePropagator,
+  W3CTraceContextPropagator,
+} from '@opentelemetry/core';
+import { ZoneContextManager } from '@opentelemetry/context-zone';
+
+export function instantiateTelemetry() {
+  // Optionally register automatic instrumentation libraries
```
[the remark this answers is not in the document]
hatched: Comment was from the docs, removing.

packages/teleport/src/Console/consoleContext.tsx
hatched: Thanks @JanKaczmarkiewicz but that was just lift & shifted by adding the telemetry to it, so at this time I don't want to introduce any more changes 😄 but I'll make a note to revisit it.

hatched:
All issues have now been addressed. @kimlisa the `exit` issue has also been resolved and works as expected.

packages/teleport/src/boot.tsx
hatched: I've left this change out of this PR because I don't want to introduce further changes at this time. I'll mark it down for a follow-up though 👍

packages/teleport/src/Console/DocumentSsh/DocumentSsh.tsx
```diff
@@ -64,14 +62,6 @@ export default function DocumentSsh({ doc, visible }: PropTypes) {
...
           </Box>
         )}
-        {status === 'error' && (
-          <Alerts.Danger mx="10" mt="5">
-            Connection error: {statusText}
-          </Alerts.Danger>
-        )}
-        {status === 'notfound' && (
-          <SidNotFoundError sid={doc.sid} clusterId={doc.clusterId} />
-        )}
```
kimlisa, line 74: if i enter an invalid SID in the URL, my terminal just says `disconnected` (which is not helpful to the user). Before we use to query for the existence of session ID first and if an error returns, we would assume the session is no longer active, which may or may not be true (which can cause confusion if it wasn't true, bad wording). If it was true though, then user can be directed to the recording.

Can we send back a more user friendly error message from the back, since we no longer query if the session exists before attempting wss cn?

packages/teleport/src/Console/DocumentSsh/useSshSession.ts
```diff
@@ -39,74 +43,80 @@ export default function useSshSession(doc: DocumentSsh) {
...
-      // subscribe to tty events to handle connect/disconnects events
-      tty.on(TermEventEnum.CLOSE, () => ctx.closeTab(doc));
-
-      tty.on(TermEventEnum.CONN_CLOSE, () =>
-        ctx.updateSshDocument(doc.id, { status: 'disconnected' })
-      );
-
-      tty.on('open', () => handleTtyConnect(ctx, session, doc.id));
+    function initTty(session) {
+      tracer.startActiveSpan('initTTY', undefined, context.active(), span => {
```
kimlisa, line 47: can you leave an inline comment on `undefined`, seems to be a pattern in the backend where if we use values like `true` or `false` we add what it is to be clear and i think it's helpful

```suggestion
      tracer.startActiveSpan('initTTY', undefined /* variable name */, context.active(), span => {
```

packages/teleport/src/Sessions/useSessions.ts
```diff
@@ -17,15 +17,31 @@ limitations under the License.
...
+
 export default function useSessions(ctx: Ctx, clusterId: string) {
   const [sessions, setSessions] = useState<Session[]>([]);
   const [attempt, attemptActions] = useAttempt({ isProcessing: true });
 
   function onRefresh() {
-    return ctx.sshService.fetchSessions(clusterId).then(setSessions);
+    return tracer.startActiveSpan(
+      'onRefresh',
+      undefined,
```
kimlisa, line 35: // comment

packages/teleport/src/telemetry-boot.ts
```diff
@@ -0,0 +1,60 @@
...
+      new DocumentLoadInstrumentation(),
+      new UserInteractionInstrumentation(),
+      new XMLHttpRequestInstrumentation(),
+      new FetchInstrumentation(),
+    ],
+  });
+
+  const resource = Resource.default().merge(
+    new Resource({
+      [SemanticResourceAttributes.SERVICE_NAME]: 'web-ui',
```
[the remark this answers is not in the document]
kimlisa: bump

packages/teleport/src/telemetry-boot.ts
```diff
@@ -0,0 +1,59 @@
...
+  const provider = new WebTracerProvider({
+    resource: resource,
+  });
+
+  provider.addSpanProcessor(new BatchSpanProcessor(new ConsoleSpanExporter()));
+  provider.addSpanProcessor(
+    new BatchSpanProcessor(
+      new OTLPTraceExporter({
+        timeoutMillis: 15000,
+        url: `${window.location.origin}/v1/webapi/traces`,
```
kimlisa, line 47: this doesn't work for me, i get 403 `need auth` error

kimlisa (approved):
this PR is good to go after addressing this: https://github.com/gravitational/webapps/pull/1361/files#r1026877794

the ambiguous error problem i think we can punt that to another PR

packages/teleport/src/Sessions/useSessions.ts
hatched: added

packages/teleport/src/Console/DocumentSsh/useSshSession.ts
hatched: I've added a comment as requested but I'm not really sure that this is a pattern that's necessary in TS. In JS sure, but in TS the type is clearly defined.

packages/teleport/src/Console/DocumentSsh/DocumentSsh.tsx
hatched: We definitely can do. I added a `WebsocketError` [message type](https://github.com/gravitational/teleport/pull/18523/files#diff-cc34e8b8bef47626c4518b46eb5a56458acedd961d290d5a79b5fe5ff63fa832R701) in the backend for other failures but didn't send one for this case. I'll make a note to add one in a follow-up.

packages/teleport/src/telemetry-boot.ts
hatched: Right, sorry, I forgot about that. I believe that this will also not work if the UI is being served from webpack as webpack has not yet been configured to proxy the trace pings.

PR #1366, opened 2022-11-17
Forward SSH agent

gzdunek:
Closes https://github.com/gravitational/teleport/issues/18320 

This PR makes Connect to always try to forward the agent. 
Default behavior for other clients:
- in `tsh` it is disabled; `-A`/`--forward-agent` option must be used explicitly,
- in `web` [teleport agent forwarding is always enabled](https://github.com/gravitational/teleport/blob/e17bcb13014e7b71bf787b3fcab5e8bd754d3b95/lib/web/terminal.go#L327). 
 
Forwarding is controlled by user's role. If it contains `forward_agent: true` then forwarding is enabled.  

In the future we would like to have an option (perhaps in a settings panel?) to configure the forwarding. Possible options would be:
- disabled
- forward all keys
- forward only teleport key 

You can read more about the options here https://github.com/gravitational/teleport/blob/master/rfd/0022-ssh-agent-forwarding.md


avatus (approved):
I like the idea of options for this in the future!

ravicious:
> * in `tsh` it is disabled; `-A`/`--forward-agent` option must be used explicitly,
> 
> * in `web` [teleport agent forwarding is always enabled](https://github.com/gravitational/teleport/blob/e17bcb13014e7b71bf787b3fcab5e8bd754d3b95/lib/web/terminal.go#L327).

Just a note to my future self: `tsh ssh --forward-agent` sets `config.ForwardAgent` to `client.ForwardAgentYes`, web sets it to `client.ForwardAgentLocal`. Making `tsh ssh` use `client.ForwardAgentLocal` involves passing an OpenSSH option like so:

```
$ tsh ssh -o "ForwardAgent local" root@example.com
```

ravicious (approved):
We should probably backport this to v10 too, right?

gzdunek:
Yes, I think so.

zmb3 (approved):

Issue #1081, opened 2022-08-03
Hide `teleport.internal` labels from the UI


hatched:
`teleport.internal/` labels are used to power various Teleport features. Modifying them or relying on them for RBAC purposes can break things, so we should hide these labels from the UI.

sielakos:
Hi @hatched 

I have a few questions about it:
1. Exactly what entity?
2. How to reproduce it?

hatched:
So this issue may not be necessary any longer. The new cluster alerts feature was at one point going to have labels on Servers containing the alert message but ultimately it went another direction. 

We probably should still filter those labels but at the moment I don't know of a situation where it would be problematic.

zmb3:
We should still hide `teleport.internal` labels from the UI.

You can reproduce this by creating a any resource (an SSH node, for example) with a label key prefixed by `teleport.internal/`. For example: `teleport.internal: test`.

sielakos:
@zmb3 @jimbishopp 
Should we always hide ``teleport.internal`` for all methods including ``tsh`` and ``tctl``?
I have a quick fix, but it's rather universal:

Before fix:

```bash
root@go:/app# tctl nodes ls
Host         UUID                                 Public Address  Labels                 Version
------------ ------------------------------------ --------------- ---------------------- ----------
749df7b60ade 0760c6af-a54e-49f4-abd8-e5ad1dbacb6a 172.20.0.7:3022 teleport.internal=test 11.0.1
node         bf77f03d-76ac-4f31-ad87-427f3ce31126                 hostname=node          12.0.0-dev
```

After fix:

```bash
root@go:/app# tctl nodes ls
Host UUID                                 Public Address Labels        Version
---- ------------------------------------ -------------- ------------- ----------
node bf77f03d-76ac-4f31-ad87-427f3ce31126                hostname=node 12.0.0-dev
```

Not sure if we want to go this way? If not I will need to add a new optional filter to ListResourceRequest.

zmb3:
Yes, you can hide them from tsh and from the web UI. Remember, it's any label with a `teleport.internal/` key prefix. So `teleport.internal/foo` should be hidden as well.

sielakos:
@zmb3 Based on your comment I have prepared a new approach on the client side (gRPC client side). I have described my thinking in a comment to PR. Do you think it is the right way to go?

PR #1369, opened 2022-11-18
Hide teleport internal labels

sielakos:
Solves issue #1081

### DIFF
--- a/packages/design/src/DataTable/Cells.tsx
+++ b/packages/design/src/DataTable/Cells.tsx
@@ -9,7 +9,6 @@ import {
   ServersideProps,
   SortDir,
   TableColumn,
-  LabelDescription,
   ClickableLabelCellProps,
 } from './types';
 
--- a/packages/design/src/DataTable/Cells.tsx
+++ b/packages/design/src/DataTable/Cells.tsx
@@ -81,11 +81,13 @@ export const DateCell = ({ data }: { data: Date }) => (
 );
 
 const renderLabelCell = (labels: string[] = []) => {
-  const $labels = labels.map(label => (
-    <Label mb="1" mr="1" key={label} kind="secondary">
-      {label}
-    </Label>
-  ));
+  const $labels = labels
+    .filter(label => !label.startsWith('teleport.internal'))
+    .map(label => (
+      <Label mb="1" mr="1" key={label} kind="secondary">
+        {label}
+      </Label>
+    ));
 
   return <Cell>{$labels}</Cell>;
 };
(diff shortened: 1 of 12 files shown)


packages/design/src/DataTable/Cells.tsx
```diff
@@ -97,20 +99,22 @@ export const ClickableLabelCell = ({
...
-      kind="secondary"
-      css={`
-        cursor: pointer;
-      `}
-    >
-      {`${label.name}: ${label.value}`}
-    </Label>
-  ));
+  const $labels = labels
+    .filter(label => !label.name.startsWith('teleport.internal'))
```
JanKaczmarkiewicz, line 103: Since this component is part of the design package we may don't want to put teleport-specific business logic inside. The design only provides UI components and the consumer package should shape the data to the correct form. 
On the other hand, this place will ensure that all cases are handled. What do you think?

JanKaczmarkiewicz (approved):

packages/design/src/DataTable/Cells.tsx
sielakos: Another option would be to filter it inside data fetching services. Which of course can be done. But would require a bit of duplication to have it applied for each service. Which is fine, but I don't think it is necessary.

JanKaczmarkiewicz: Alternatively, we could create `TeleportLabelCell` wrapper for `ClickableLabelCell` and put it in the `shared` package:
```tsx
export const TeleportLabelCell = ({
  labels,
  ...props
}: ClickableLabelCellProps) => (
   <ClickableLabelCell
      {...props}
      labels={labels.filter(({ name }) => !name.startsWith('teleport.internal'))}
  />
);
```
And use it instead of ClickableLabelCell where needed.

sielakos: That's actually a great idea. I would put it into teleport/components as it is actually used and needed only inside teleport package

zmb3:
@sielakos I actually liked your solution of doing this at the backend, because it would also help us with gravitational/teleport#17041. I think it's worth looking at that approach again, since you wouldn't have to put any special logic in here.

I think your previous attempt was just doing the filtering at the wrong place. Resources are converted to UI objects [here](https://github.com/gravitational/teleport/blob/2868069924786c8ec0754fa176fce86081ae4dc8/lib/web/ui/server.go#L72), so maybe we can just filter out labels at that point (which is well after RBAC)?

sielakos:
@zmb3 the link you sent applies only to nodes. Wouldn't it be better to modify listResources backend function used by all endpoints to remove the labels from ResourceWithLabels items?

zmb3:
@sielakos I highlighted the nodes function, but that file has the conversion for all resources for the web UI. It should be done here instead of in `listResources` because `listResources` may need that info.

sielakos:
@zmb3 Thanks for the clarification, I have prepared [an alternative PR](https://github.com/gravitational/teleport/pull/18638) is that what you had in mind?

PR #1103, opened 2022-08-11
ci: Implement code scanning with CodeQL

reedloden:
Use CodeQL to scan all pull requests and pushes for known security issues.

https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/about-code-scanning-with-codeql

PR #1373, opened 2022-11-18
Create initial codeql.yml configuration

jentfoo:
This PR adds a new CodeQL configuration.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,46 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches:
+      - master
+      - branch/*
+  pull_request:
+    branches:
+      - master
+      - branch/*
+    paths-ignore:
+      - 'scripts/**'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'javascript' ]
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+
+    # Initializes the CodeQL tools for scanning.
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v2
+      with:
+        languages: ${{ matrix.language }}
+
+    # - run: |
+    #   echo "Run, Build Application using script"
+    #   ./location_of_script_within_repo/buildscript.sh
+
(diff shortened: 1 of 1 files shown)


adaadb6 (approved):

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,46 @@
...
+
+on:
+  push:
+    branches:
+      - master
+      - branch/*
+  pull_request:
+    branches:
+      - master
+      - branch/*
```
reedloden, line 11: See above.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,46 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches:
+      - master
+      - branch/*
```
reedloden, line 7: The branch names on this repository are different from `teleport`. It's actually `teleport-v*`.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,46 @@
...
+    - name: Checkout repository
+      uses: actions/checkout@v3
+
+    # Initializes the CodeQL tools for scanning.
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v2
+      with:
+        languages: ${{ matrix.language }}
+
+    # - run: |
```
reedloden, line 39: Can we remove this? Do we need to include autobuild at all?

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,46 @@
+name: "CodeQL"
```
reedloden, line 1: We use `codeql-analysis.yml` on other repositories. I don't have a strong preference as to which to use, but should standardize, imho.

reedloden (commented):
This replaces https://github.com/gravitational/webapps/pull/1103.

This will need to be backported to all supported branches.

.github/workflows/codeql.yml
jentfoo: Which standard would be easier to adopt?  GitHub choose the `codeql.yml` name as a default

jentfoo:
I will address these recommendations shortly, thank you for the comments @reedloden!

.github/workflows/codeql.yml
jentfoo: I only see three cases of `codeql-analysis.yml`.  If it sounds good I suggest we transition to the default naming to reduce the manual steps of on-boarding a project.

I can submit PR's for the 3 projects we have today

PR #1377, opened 2022-11-22
useAsync: Add support for abort signal

ravicious:
This PR makes it possible to use `useAsync` together with a cleanup function returned from `useEffect`.

```typescript
const [attempt, run] = useAsync((message: string) => doAsyncStuff(message));

useEffect(() => {
  const abortController = new AbortController();
  run(abortController.signal, 'hello');

  return () => {
    abortController.abort();
  };
});
```

# Why

When fetching data in `useEffect`, it's important to [ignore the response from the fetching function in the cleanup function](https://beta.reactjs.org/learn/synchronizing-with-effects#fetching-data).

However, the way `useAsync` used to work meant that we had no control over it – `useAsync` would always call `setState` after the promise from the given callback got resolved.

# How

At first, the solution appear to be simple – let's just somehow not update the state managed by `useAsync` if the result of the async action should be ignored.

There were two problems to tackle. The abort signal should be local to the function passed to `useEffect`. This meant that `useAsync` itself cannot accept an abort signal as `useAsync` cannot be called from within `useEffect`.

Another problem was that `useAsync` has two interfaces through which you can access the result of the async action: the attempt object managed by `useState` & the promise returned from `run`. The attempt object is typically used by the React component and represents the result over time. The promise returned by `run` represents the result of only a single call. Nevertheless, it's still useful if you need to perform some extra actions once the action finishes.

This meant that not only we had to not update the state, the promise should also return 👋 something 👋.

Since the signal needs to be local to `useEffect`, I couldn't come up with any other solution than doing something a bit controversial: letting the `run` function accept a magical argument.

## How it used to work

`useAsync` accepted an arbitrary callback which could accept any arguments. Then the `run` function was returned from `useAsync` which accepted the same args as the callback. It let you control the execution of the async action.

## How it works now

`useAsync` works the same as before, except that the `run` function behaves differently depending on the type of the first argument. If the first argument is an abort signal, then `run` expects that the rest of the args are the full args for the callback. If the first argument is anything else, then `run` expects that it's the first arg of the callback and the rest of the args should also belong to the callback.

This required some slight changes to the types (and thanks to Jan we were able to massively simplify them in the first place). I'm not particularly happy with the type errors returned by the function but I'll leave a comment about that.

---

I didn't have time to add fully fledged tests for `useAsync`, so I added just those that cover the abort signal functionality. I also included some type tests in case we decide to improve the types in the future.


packages/shared/hooks/useAsync.ts
```diff
@@ -58,47 +56,106 @@ import { useCallback, useState } from 'react';
...
+  cb?: Callback
+) {
+  const [state, setState] =
+    useState<Attempt<Awaited<ReturnType<Callback>>>>(makeEmptyAttempt);
 
   const run = useCallback(
-    (...p: Parameters<AsyncCb<R, T>>) =>
-      Promise.resolve()
+    (
+      ...args: Parameters<Callback> | [AbortSignal, ...Parameters<Callback>]
```
ravicious, line 85: So this means that TypeScript simply matches the args provided to `run` against two signatures: either you supply an array where the first element is an abort signal or you supply an array that's equal to the params accepted by the callback.

This leads to error messages which are not that great to be honest.

For example, if I have a callback which accepts no args but pass an extra string to it, this is what I get:

```
packages/shared/hooks/useAsync.test.ts:103:7 - error TS2345: Argument of type '[string]' is not assignable to parameter of type '[] | [AbortSignal]'.
  Type '[string]' is not assignable to type '[AbortSignal]'.
    Type 'string' is not assignable to type 'AbortSignal'.

103   run('foo');
```

The worst error message happens when you specify an extra arg to call where the first arg is not an abort signal. Let's say I have a callback which accepts two args and I supply an extra one:

```
packages/shared/hooks/useAsync.test.ts:179:7 - error TS2345: Argument of type '["foo", 2, "bar"]' is not assignable to parameter of type '[AbortSignal, string, number] | [message: string, count: number]'.
  Type '["foo", 2, "bar"]' is not assignable to type '[AbortSignal, string, number]'.
    Type at position 0 in source is not compatible with type at position 0 in target.
      Type 'string' is not assignable to type 'AbortSignal'.

179   run('foo', 2, 'bar');
```

This all stems from that simple matching that TypeScript performs.

An approach which would result in better error messages would be to add a generic type for the first arg of `run` and destructure the args within the function definition, as in `(firstArg, ...args)`.

Then we could tell TypeScript more or less this: if the first arg is an abort signal, assume that args is the params of the callback. If it's not an abort signal, then it must be the first param of the callback and args must be the rest of them.

I even managed to write such type:

```typescript
  const run = useCallback(
    <FirstArg extends AbortSignal | Parameters<RunFunc>[0]>(
      firstArg?: FirstArg,
      ...restOfArgs: FirstArg extends AbortSignal
        ? Parameters<RunFunc>
        : Parameters<RunFunc> extends [infer _, ...infer Tail]
        ? Tail
        : never
    ) => {
```

However, if you plug it in and look at the type errors in tests, you can see that it doesn't raise an error if the callback accepts one arg and you execute `run` with no args.
This is because the first arg is essentially treated as optional.

I didn't have time to figure out how to make this better, so for now I opted for the simpler solution.

packages/shared/hooks/useAsync.ts
```diff
@@ -58,47 +56,106 @@ import { useCallback, useState } from 'react';
...
+ *   run(abortController.signal, 'hello');
+ *
+ *   return () => {
+ *     abortController.abort();
+ *   };
+ * });
  */
-export function useAsync<R, T extends Function>(cb?: AsyncCb<R, T>) {
-  const [state, setState] = useState<Attempt<R>>(makeEmptyAttempt);
+export function useAsync<Callback extends (...args: unknown[]) => unknown>(
```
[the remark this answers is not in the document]
ravicious: Hmm, I think the solution with `isMounted` doesn't handle situations where there's no unmounting but we simply want to ignore the result of the fetch because, let's say, params for the fetch query have changed.

I forgot to include `run` in `useEffect` deps in the example, but also imagine that the `message` passed to `run` could change when the user changes an input field. `isMounted` wouldn't handle that situation, would it?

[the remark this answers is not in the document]
JanKaczmarkiewicz: No, isMounted only handles unmount cases. 
The solution to the problem you referring to is 
```
if (promise !== asyncTask.current) return // [...] 
```
check.
This will ensure that only the latest promise is handled by then/catch metchods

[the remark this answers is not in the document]
ravicious: Ah, I completely missed that while looking at your solution.

Alright, this seems to be even better! I'm trying to think of a situation where we wouldn't want to return early on unmount or the promise being changed but nothing comes to my mind right now.

Another thing is that this solution would change the default behavior of `useAsync` but I think that's for the better - it'll automatically clean up those calls in flight. Places which depend solely on `attempt` won't see a difference and places which depend on the resolved value of the promise should handle errors anyway.

Thanks a lot!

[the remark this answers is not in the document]
ravicious: And if a certain place doesn't need to ignore stale request, it'd be straightforward to add an option to `useAsync` which simply ignores those checks and does allow stale requests.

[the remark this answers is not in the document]
JanKaczmarkiewicz: > it'd be straightforward to add an option to useAsync which simply ignores those checks and does allow stale requests.

In this case, we can also return current promise instead:
```
if (promise !== asyncTask.current) return asyncTask.current.then(...)
```

packages/shared/hooks/useAsync.ts
```diff
@@ -74,88 +74,64 @@ import { useCallback, useState } from 'react';
...
+        ...prevState,
+        status: 'processing',
+      }));
+
+      const promise = cb(...args);
+      asyncTask.current = promise;
+
+      return promise.then(
+        data => {
+          if (!isMounted) {
```
ravicious, line 96: @JanKaczmarkiewicz I updated `useAsync` but I'm struggling to get the tests to pass.

During the one marked with `test.only`, I see this:

```
 console.error
    Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
        at TestComponent (/Users/rav/Projects/webapps/node_modules/@testing-library/react-hooks/lib/helpers/createTestHarness.js:22:5)
        at Suspense
        at ErrorBoundary (/Users/rav/Projects/webapps/node_modules/react-error-boundary/dist/react-error-boundary.cjs.js:59:35)

      101 |           }
      102 |
    > 103 |           setState(prevState => ({
          |           ^
      104 |             ...prevState,
      105 |             status: 'success',
      106 |             data,
```

So this kinda indicates that on success, isMounted is still set to true and `useAsync` attempt to `setState`. I'll try to come back to this tomorrow, but I think it has to do with `isMounted` being in deps?

JanKaczmarkiewicz: You need to call: 
```
if (!isMounted()) {
```
and change the test to be
```
const returnValue = Symbol();
  const { result, unmount } = renderHook(() =>
    useAsync(() => sleep(100).then(() => returnValue))
  );

  let [, run] = result.current;
  let promise: Promise<[symbol, Error]>;
  act(() => {
    promise = run();
  });
  await sleep(50);

  unmount();

  await expect(promise).resolves.toEqual([null, new AbortedSignalError()]);
```
or 
```
let promise: Promise<[unknown, Error]>;
  const TestComponent = () => {
    const asyncFn = useCallback(() => sleep(1000), []);

    const [state, run] = useAsync(asyncFn);

    useEffect(() => {
      promise = run();
    }, [run]);

    return <>{state.status}</>;
  };

  const { unmount } = render(<TestComponent />);

  await screen.findByText('processing');
  unmount();

  await expect(promise).resolves.toEqual([null, new AbortedSignalError()]);
```

ravicious: Thanks a lot, you're a life saver. I was rushing through and I didn't notice that `useIsMounted` returns a function (which makes total sense).

avatus:
Sure thing! I'll see what I can whip up 👍. Will be traveling tomorrow so if I don't get it by tonight, expect something at beginning of next week.

avatus:
@ravicious I'm getting both tests failing correctly when commenting out the `!== promise` conditional. Fixing some lint errors and then this looks good to me

avatus (approved):

avatus:
@JanKaczmarkiewicz i feel weird approving a PR that i'm the owner of now. When you get a chance, can you give this another look over. Appreciate it

packages/shared/hooks/useAsync.ts
```diff
@@ -58,47 +56,82 @@ import { useCallback, useState } from 'react';
...
+ * const [attempt, run] = useAsync((message: string) => doAsyncStuff(message));
+ *
+ * useEffect(() => {
+ *   const abortController = new AbortController();
+ *   run(abortController.signal, 'hello');
+ *
+ *   return () => {
+ *     abortController.abort();
+ *   };
+ * });
```
JanKaczmarkiewicz, line 75: Lets remove this section
```suggestion

```

packages/shared/hooks/useAsync.ts
```diff
@@ -107,6 +140,28 @@ export function useAsync<R, T extends Function>(cb?: AsyncCb<R, T>) {
...
+
+    return () => {
+      isMounted.current = false;
+    };
+  }, []);
+
+  return useCallback(() => isMounted.current, []);
+}
+
+// TODO: Rename the error.
```
JanKaczmarkiewicz, line 157: We may better rename this to CanceledError since it has nothing to do with Abort signal

JanKaczmarkiewicz:
We may also want to update the PR description to match the implementation. I don't think something long is necessary.

JanKaczmarkiewicz (approved):

ravicious:
> @ravicious I'm getting both tests failing correctly when commenting out the `!== promise` conditional. Fixing some lint errors and then this looks good to me

Those two tests still pass for me if I comment out the promise conditional:


* run does not update state after being re-run when the callback returns a resolved promise
* run does not update state after being re-run when the callback returns a rejected promise

The ones that fail are related to the resolved value of the promise.

avatus:
I misread the failing tests. Will get this updated and fixed

ravicious:
@avatus Before you merge this PR, please check that it indeed fixes the problem with redundant calls in async autocomplete and in the resource tables.

packages/shared/hooks/useAsync.ts
```diff
@@ -107,6 +124,28 @@ export function useAsync<R, T extends Function>(cb?: AsyncCb<R, T>) {
...
```
hatched, line 141: It looks like you've already renamed this so this comment can be removed.

packages/shared/hooks/useAsync.ts
```diff
@@ -107,6 +124,28 @@ export function useAsync<R, T extends Function>(cb?: AsyncCb<R, T>) {
...
+export class CanceledError extends Error {
```
hatched, line 142: A bit of a nit but is this really an error? As this can happen via the normal successful operation of the code it feels more like a status...

avatus: Hmmm... I see your point, but I think it was named this way to think of this is "did the `run` function complete". If we abort early, it means that "no the `run` function DIDNT complete... because we aborted it", rather than asking if the request (or whatever cb) completed. A case could be made either way but, I'm for keeping it as is to have a 'consistent' api. Thoughts?

avatus:
@ravicious this change is awesome. really makes async autocomplete feel a lot nicer.

PR #1383, opened 2022-11-25
Connect: Implement tshd event handlers for db cert renewal

ravicious:
teleport counterpart: gravitational/teleport#17950

# Why

If the user attempts to make a connection through a local db proxy but the db & user certs have expired, we want Connect to bring focus to the Connect app, let the user relogin and then proxy the connection.

This is how the end result looks like:

https://user-images.githubusercontent.com/27113/204036222-e300656f-0040-411c-b937-80232e26d0fd.mov

# How

The backend flow is described in gravitational/teleport#17950. When it comes to the Electron app, tshd needs to be able to ask the Electron app for two things:

* To ask the user to log in again.
* To forward errors from goroutines running gateways to the Electron app in form of a notification. Otherwise those error would be visible only in the logs.

This is done through tshd events sent from tshd.

## Important & regular modals

It's possible that while a db client attempts to make a connection, the user has a modal open in the app, for example to log in to a new cluster or whatever modal we might add in the future.

To avoid discarding the state from such modal, this PR introduces the concept of important modals. The basic idea is that we can have one regular and one important modal being rendered at the same time. The important modal is rendered over the regular modal, so if the user finishes the relogin process they can continue interacting with the old modal.

Opening an important modal should be reserved for situations where the interaction with the app happens outside of its UI and requires us to interrupt the user and show them a modal.

## Structured object vs strings

Both tshd event handlers need to show a specific message to the user. Instead of passing a string from tshd to the Electron app, we pass a structured RPC message. This way the Electron app can decide how exactly the message should look like.


packages/teleterm/src/types.ts
```diff
@@ -42,12 +44,42 @@ export type SubscribeToTshdEvent = <
...
   readonly subscribeToTshdEvent: SubscribeToTshdEvent;
 };
+
+/**
+ * @callback subscribeToTshdEventListener
+ * @param {Object} eventData
+ * @param {Object} eventData.request - The request payload converted to a simple JS object.
+ * @param {subscribeToTshdEventListenerOnCancelled} eventData.onCancelled - A function which lets
+ * you register a callback to be called when the call gets cancelled by the client.
+ */
```
ravicious, line 74: idk, apparently this is the way to use JSDoc in cases like this?

https://jsdoc.app/tags-callback.html

It looks very messy and doesn't actually show the additional docs on symbol lookup.

Maybe I'm doing something wrong, but I think I might move all docs into a single comment… somehow.

packages/teleterm/src/ui/appContext.ts
```diff
@@ -86,10 +96,32 @@ export default class AppContext implements IAppContext {
...
     this.workspacesService.restorePersistedState();
   }
+
+  private setUpTshdEventSubscriptions() {
+    this.subscribeToTshdEvent('relogin', ({ request, onCancelled }) => {
+      // The handler for the relogin event should return only after the relogin procedure finishes.
+      return this.reloginService.relogin(request, onCancelled);
+    });
+
+    this.subscribeToTshdEvent('sendNotification', ({ request }) => {
```
gzdunek, line 123: Are these `notifications` related only to `relogin` events? Or they can occur in other cases? 
I'm wondering if we could combine `ReloginService` and `TshdNotificationsService`.

packages/teleterm/src/ui/types.ts
```diff
@@ -27,6 +27,8 @@ import { NotificationsService } from 'teleterm/ui/services/notifications';
 import { ConnectionTrackerService } from 'teleterm/ui/services/connectionTracker';
 import { FileTransferService } from 'teleterm/ui/services/fileTransferClient';
 import { ResourcesService } from 'teleterm/ui/services/resources';
+import { ReloginService } from 'teleterm/services/relogin';
+import { TshdNotificationsService } from 'teleterm/services/tshdNotificationsService';
```
gzdunek, line 31: Filename shouldn't contain `Service`.

gzdunek (commented):
First pass, looks really good :)

packages/teleterm/src/ui/services/modals/modalsService.ts
```diff
@@ -23,29 +23,38 @@ export class ModalsService extends ImmutableStore<Dialog> {
...
 
-  openDialog(dialog: Dialog) {
+  /*
+   * openDialog opens the given dialog. It returns a function which can be used to close the dialog
+   * and automatically call the dialog's onCancel callback (if present).
+   */
+  openDialog(dialog: Dialog): () => void {
     this.setState(() => dialog);
+
+    return () => {
```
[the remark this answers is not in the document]
ravicious: > IMO it's not obvious that the function that is returned can be used to close the dialog. WDYT about more explicit option?

Yeah I guess an explicit option would be fine too, I can change it. I was inspired here by `useEffect` a little bit.

packages/teleterm/src/ui/services/modals/modalsService.ts
```diff
@@ -16,7 +16,8 @@ limitations under the License.
 
 import { useStore } from 'shared/libs/stores';
 
-import { ImmutableStore } from '../immutableStore';
+import { ImmutableStore } from 'teleterm/ui/services/immutableStore';
```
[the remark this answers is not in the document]
ravicious: > Wasn't `../` better here? I'd treat `services` as a single _module_, so it makes sense to me to use relative imports inside the module.

Hmm, I suppose I don't think of services as a single module, rather than a folder which groups things that fulfill a similar role, so as soon as I want to grab something that's outside of the folder of the specific "unit" (modalsService would be the "unit" in this case), I use a full path.

And I wanted to avoid having another import group (files from parent directories) because I felt that with parent imports it's hard to enforce how deep can you reach. Plus I'm only even interested in what stuff we import from our own package, which stuff we import from other Teleport packages and which stuff we import from built-in/3rd party packages.

Ultimately I don't really care about it as long as the computer can enforce our chosen import style, which I don't think we enforce in that scenario, so since I don't care I can bring back `../immutableStore` here. 😏

packages/teleterm/src/ui/appContext.ts
ravicious: Currently those notifications are tied to the relogin process but that's only because so far it's the only place where we send a notification from tshd to the Electron app. In the future we could add different subjects to `SendNotificationRequest` and handle different notifications through a single RPC.

packages/teleterm/src/types.ts
ravicious: > TBH I would only document params in `SubscribeToTshdEvent` (in `src/types`), however I still don't see any additional docs 🤔.

You mean that even if you move all docs to a single comment, you still don't see docs if you look up an individual argument such as `onCancelled`?

I wonder if this is something that would work correctly if we actually generated HTML docs and the reason it's not shown here is simply a limitation of the language server.

Anyway, if the docs for individual arguments don't work, then having all docs in one comment would be much easier to read.

ravicious: Ah, the docs were visible when all of this was still a proof of concept and almost everything was inline within AppContext. The docs were shown because I added them to the `subscribeToTshdEvent` property of `ElectronGlobals`. Now that it's a property of AppContext, the docs are lost. I'll see what I can do about this.

ravicious: I suppose this is related to this issue: https://github.com/microsoft/TypeScript/issues/51121

There doesn't even seem to be a way to reuse a single typedef across multiple sites, e.g. having the same jsdoc for `appContext.subscribeToTshdEvent` and `ElectronGlobals.subscribeToTshdEvent`. I'm going to move the comment to `appContext.subscribeToTshdEvent` since this is where the function will be actually used so the comment will be most helpful there.

gzdunek: > You mean that even if you move all docs to a single comment, you still don't see docs if you look up an individual argument such as onCancelled?

Hmm, I think so.

I'm unsure about documenting `eventName` and `listener` in `AppContext`. I would do it in the place where they are defined (`src/types`). I bet that after changing them, no one would think about updating the comment in `AppContext`.

packages/teleterm/src/ui/services/modals/modalsService.ts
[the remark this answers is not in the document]
gzdunek: I don't have any strong opinion on it too, it only seemed to me that in this case parent imports was fine :)

gzdunek (approved):
Great work, it is really cool feature!

I'm only wondering if one minute timeout is not too strict? I would consider increasing it to maybe 3-5 minutes.

ravicious:
> I'm only wondering if one minute timeout is not too strict? I would consider increasing it to maybe 3-5 minutes.

I think increasing it wouldn't help much here. pgAdmin's default connection timeout is 10 seconds. [SSMS has a timeout of 15 seconds](https://learn.microsoft.com/en-us/troubleshoot/sql/connect/timeout-expired-error#determine-the-type-of-timeout-expired-errors).

Of course the user is able to change the timeout (at least pgAdmin lets you do that). But in general I'd say the user is supposed to see the modal soon after the db client tries to make the connection and then they should relogin. If they don't manage to do so, then that's too bad but we'll show display a notification.

Also, unfortunately right now the modal will not disappear if the db client drops the connection, see https://github.com/gravitational/teleport/pull/17950#discussion_r1034702009.

packages/teleterm/src/types.ts
ravicious: > I'm unsure about documenting `eventName` and `listener` in `AppContext`. I would do it in the place where they are defined (`src/types`). I bet that after changing them, no one would think about updating the comment in `AppContext`.

Good point, I'm going to move them to `src/types`.

avatus (approved):
I spent a good two hours testing against `teleport-11-ent.asteroid.earth` and I wasn't getting any popup in Connect. Only errors in pgAdmin saying the connection had been terminated. I had a detailed list of reproduction steps, a video and everything only to then realize I hadn't built the `tsh` counterpart to this PR 🤦 

This works great and is REALLY dope!!

PR #1385, opened 2022-11-28
Pass clusterUri rather than documentUri to retryWithRelogin

ravicious:


avatus (approved):

ravicious:
I think that this could've been avoided if we used template literals. https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html

That page doesn't explain it that well, but I think it should be possible to enforce on type level two types: `ResourceUri` and `DocumentUri`. Then we could change the signature of `retryWithRelogin` so that it only accepts `ResourceUri` which would caught this (easy mistake to make after all, both are values are regular strings right now).

ravicious:
https://medium.com/@KevinBGreene/fun-with-typescript-template-literal-types-95580e27e0e1

ravicious:
I swear I looked straight at this function call in code review and I was like "Yeah it looks fine, documentUri should point to the correct resource".

gzdunek:
> I think that this could've been avoided if we used template literals.

I thought about it some time ago, we should definitely try it.

PR #1387, opened 2022-11-28
Fixes `onRemove` console error

ibeckermayer:
Destructures props so that we no longer get a console error about `onRemove`, see https://github.com/gravitational/webapps/pull/1333#discussion_r1031391160


gzdunek (approved):

ravicious (approved):
Thanks for submitting the fix, I didn't have time to do that yesterday.

ibeckermayer:
> Thanks for submitting the fix, I didn't have time to do that yesterday.

No problem, it was my doing. Thanks for pointing it out.

PR #1394, opened 2022-11-29
Connect: Use typed URIs

ravicious:


ravicious:
@avatus @gzdunek This is what I cooked up. There's 35 remaining type errors in 19 files but I'll leave that for later. Submitting this as a WIP because I don't have more time for this today.

In a sound type system, we'd have to validate incoming values to ensure they actually match the type. That is, when making an RPC and getting some objects, we'd have to make a runtime check to verify that the URI indeed matches the type.

In TypeScript, we can cheat a little and just say that we trust that anything that comes from tshd is the type it purports to be.

---

I have to say this is pretty nice. I already made a couple of mistakes while fixing the types in tests and TypeScript immediately caught those.

I'm unsure if there are any unforeseen roadblocks ahead, but I think if we ensure any incoming objects are properly typed (for now it's just stuff from tshd) and that docs & gateways are created with correct types, I think there's nothing which would prevent us from using this approach.

packages/teleterm/src/ui/uri.ts
```diff
@@ -30,6 +30,54 @@ export const paths = {
...
+  `/clusters/${RootClusterId}/leaves/${LeafClusterId}/kubes/${KubeId}`;
+export type LeafClusterDatabaseUri =
+  `/clusters/${RootClusterId}/leaves/${LeafClusterId}/dbs/${DbId}`;
+export type LeafClusterResourceOnlyUri =
+  | LeafClusterServerUri
+  | LeafClusterKubeUri
+  | LeafClusterDatabaseUri;
+export type LeafClusterResourceUri =
+  | LeafClusterUri
+  | LeafClusterResourceOnlyUri;
```
ravicious, line 64: There's lots of repetition here but this stuff is unlikely to change over time any way. It's not like we're going to add a third type of a cluster.

I'm not sure about the `*ResourceOnlyUri` name, but I basically want to be able to say "I accept URIs to resources only" and "I accept URIs to resources and clusters as well".

Maybe `*ResourceOnlyUri` should be renamed to `*ResourceUri` and current `*ResourceUri` to `*ClusterOrResourceUri`.

packages/teleterm/src/ui/uri.ts
```diff
@@ -30,6 +30,54 @@ export const paths = {
...
 };
 
+type RootClusterId = string;
+type LeafClusterId = string;
+type ServerId = string;
+type KubeId = string;
+type DbId = string;
+export type RootClusterUri = `/clusters/${RootClusterId}`;
+export type RootClusterServerUri =
+  `/clusters/${RootClusterId}/servers/${ServerId}`;
```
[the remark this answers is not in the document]
ravicious: I mean, technically the current implementation would at least prohibit us from passing a root cluster URI where a leaf cluster URI is expected. But I don't think this is that useful, we don't have many places where leaf cluster URIs are allowed but root cluster URIs are prohibited. And it doesn't outweigh the false positive of passing leaf cluster URI as a root cluster URI.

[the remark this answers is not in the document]
gzdunek: >It would be great if we could restrict RootClusterId to a stricter type then string.

AFAIK it is not possible, hopefully in the future we will be able to create [regex-validated string types](https://github.com/microsoft/TypeScript/issues/41160).

>If it's not possible, then perhaps we shouldn't define types which cannot be enforced and instead focus just on those that we actually can.

Hmm, I would define them anyway. They would act as [type aliases](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-aliases). I think it is still better for the reader to see `clusterUri: RootClusterUri` than `clusterUri: string`. You could also add a short comment to the type definition with an example.

packages/teleterm/src/ui/uri.ts
gzdunek: > Maybe *ResourceOnlyUri should be renamed to *ResourceUri and current *ResourceUri to *ClusterOrResourceUri.

I agree. Currently, when we refer to a _resource_ we think about servers/dbs/kubes, not clusters.

avatus: repetition is fine with me if it helps us down the road. this is REALLY cool btw. I didn't know that template literals could be used this way. This PR taught me a lot. 

+1 for _ClusterOrResourceUri_

packages/teleterm/src/ui/uri.ts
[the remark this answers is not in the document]
avatus: I don't think our types need to differentiate between root/leaf (i mean... if it was an easy fix sure), but as you said, there is nowhere that expects _leaf_ only, just places that require root only. 

I agree that defining them anyway would help future readers. not only would we have params named stuff like `rootClusterUri` but also typed for double safety.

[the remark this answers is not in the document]
ravicious: > I think it is still better for the reader to see `clusterUri: RootClusterUri` than `clusterUri: string`.

Ah, those would still be typed, we wouldn't use `string`. It's just that because we cannot differentiate between root and leaf URIs, we'd have a single `ClusterUri` type that accepts both. So instead of `clusterUri: RootClusterUri` you'd have `rootClusterUri: ClusterUri`. And as Michael said, I think in those cases the name of the variable should be `rootClusterUri` no matter what the type is. ;)

---

However, we could make the same argument about cluster URIs vs resource URIs. This code works:

```typescript
const serverUri: RootClusterServerUri = '/clusters/foo/servers/bar'
const uri: RootClusterUri = serverUri
const leafServerUri: LeafClusterServerUri = '/clusters/foo/leaves/baz/servers/bar'
const leafUri: LeafClusterUri = leafServerUri
```

But I wouldn't want to remove the distinction between clusters and resources in types simply because we can't prevent resource URIs being passed as cluster URIs.

So yeah, maybe let's keep it as it is for now and see how it works out. Refactoring this will be super easy anyway.

packages/teleterm/src/ui/services/workspacesService/documentsService/documentsService.ts
```diff
@@ -90,17 +89,17 @@ export class DocumentsService {
...
       kubeId: params.kubeId,
       kubeUri: options.kubeUri,
+      // We prepend the name with `rootClusterId/` to create a kube config
+      // inside this directory. When the user logs out of the cluster,
+      // the entire directory is deleted.
       kubeConfigRelativePath:
         options.kubeConfigRelativePath ||
-        // We prepend the name with `rootClusterId/` to create a kube config
-        // inside this directory. When the user logs out of the cluster,
-        // the entire directory is deleted.
```
ravicious, line 97: I moved this comment around because it was messing up syntax highlighting in my editor. ;f

ravicious (commented):
I'm going to take another look at this PR on Monday.

I tried to change just types where possible, I didn't even touch arg names just to reduce the possible bug surface here.

ravicious:
@avatus @gzdunek Anything I can do to make this easier to review? The longer it sits here, the harder it'll be to resolve the conflicts.

packages/teleterm/src/ui/TopBar/Clusters/Clusters.tsx
```diff
@@ -31,7 +33,7 @@ export function Clusters() {
     )
   );
 
-  function selectItem(id: string): void {
+  function selectItem(id: ClusterUri): void {
```
gzdunek, line 36: Can we change `id` to `clusterUri`?

gzdunek (approved):
Looks good to me!

avatus (approved):
This one left my mind. Sorry about missing it. LGTM

ravicious:
Merged. Let me know if you encounter any issues in your open PRs.

PR #1396, opened 2022-11-30
Lazy load Telemetry only when needed

ryanclark:
The introduction of https://github.com/gravitational/webapps/pull/1361 caused our main production entry bundle (the JS the user loads when first opening teleport) to increase by 152 KiB, a 30% increase.

Before the PR:
image

After the PR:
image

This changes the code to only import the `telemetry-boot` file, which in turn includes 13 Telemetry packages, when needed (the user has set `enable-telemetry` to `true` in local storage and refreshed the page).

After this change:
image

This will also make the time to first render faster for those that have enabled telemetry, as the telemetry instantiation is no longer blocking React from rendering the app.

### DIFF
--- a/packages/teleport/src/boot.tsx
+++ b/packages/teleport/src/boot.tsx
@@ -21,7 +21,6 @@ import history from 'teleport/services/history';
 
 import Teleport from './Teleport';
 import TeleportContext from './teleportContext';
-import { instantiateTelemetry } from './telemetry-boot';
 import cfg from './config';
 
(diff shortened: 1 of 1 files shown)


avatus (approved):
THANK YOU

hatched (approved):

zmb3:
Needs a v11 backport too, right?

ryanclark:
Ah I wasn't aware the initial PR was backported. Will backport, cheers @zmb3

PR #1400, opened 2022-11-30
Update xterm to 5.0.0

avatus:
Fixes https://github.com/gravitational/teleport/issues/18784

A big highlight of xterm 5.0.0 is the update to underline rendering. Check it [here](https://github.com/xtermjs/xterm.js/releases/tag/5.0.0). 

Updated the package and the issue was resolved. Here is a before and after (4 on the left, 5 on the right)
image

All tests pass, manual testing shows no issues, and I don't believe any of the breaking changes affect us. Wouldn't mind a double check

### DIFF
--- a/packages/teleport/package.json
+++ b/packages/teleport/package.json
@@ -19,7 +19,7 @@
   "dependencies": {
     "@gravitational/design": "1.0.0",
     "@gravitational/shared": "1.0.0",
-    "xterm": "^4.15.0",
+    "xterm": "^5.0.0",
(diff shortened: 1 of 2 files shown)


zmb3 (approved):
Looks good to me. I also double checked all the v5 breaking changes against our codebase and didn't see anything.

ravicious:
Do we plan to update the xterm version used in Connect as well?

zmb3:
@ravicious - I don't see why not. Let's do it.

PR #1404, opened 2022-12-01
Connect: Add `UsageData` dialog

gzdunek:
This PR adds a dialog that will be used to ask a user about collecting usage data.

I slightly modified the copy, the original was: 
>Are you OK sending anonymized usage data about Teleport Connect? This will help us to improve product 

I think the new one sounds better.

There is a close button that runs `onCancel` - it will close the dialog and show it again on the next launch. I would like to give the user option to postpone the answer.

image


packages/teleterm/src/ui/ModalsHost/modals/UsageData/UsageData.tsx
```diff
@@ -0,0 +1,83 @@
...
+            type="button"
+            onClick={props.onCancel}
+            color="text.secondary"
+          >
+            <Close fontSize={5} />
+          </ButtonIcon>
+        </DialogHeader>
+        <DialogContent mb={4}>
+          <Text typography="body1" color="text.secondary">
+            Do you allow for sending anonymized usage data about Teleport
```
avatus, line 68: This reads a bit weird. I'd suggest something along the lines of
"Do you allow us to collect anonymized usage data...". If you want to keep it as close to what you have now as possible, we should replace "for" with "the", but tbh, both sound a bit odd. 

Also, and this may just be me, but hearing "we are collecting" instead of "we are sending" sounds more like WE are in control of the data instead of just shipping it off elsewhere. idk, maybe I'm over thinking it

avatus (approved):
I'm more of a fan of the original copy but, I won't die on the copy hill and there may be better people to reach out to. The rest of the PR is great!

ravicious (approved):
I also think we should change the copy.

ravicious:
Actually, I'm going to ask during today's design meeting if we want to do it through a modal. I'm working on linking tsh and we want to ask the user about it on startup too.

packages/teleterm/src/ui/ModalsHost/modals/UsageData/UsageData.tsx
gzdunek: Thanks for the suggestions guys!
I decided to go with @ravicious version, it sounds less formal :)

PR #1405, opened 2022-12-02
[Connect] Update autocomplete to use async serverside search

avatus:
After we [enabled serverside search](https://github.com/gravitational/webapps/pull/1307) in the resource tables, this basically "fixed" Connect to work with large clusters. The last chunk to fix was autocomplete because it still used the old paradigm of loading everything at the start and then client side filtering on input.

This PR updates the quickpickers to provide an interface for using serverside search results in the QuickInput. 

https://user-images.githubusercontent.com/5201977/205326673-427c32ad-377c-41d6-a8db-5c4fb86f354c.mov

I've opted for a spinner icon (where the shortcut key normally is) as the loading state, but this is just first pass. Other options would be showing the suggestion menu with skeleton data, or just the suggestion menu saying "loading..." or something. I think spinner looks best but I will not die on that hill.


avatus:
Looks like some commit history got a bit messed around when merging in master and the `useAsync` update branch.

ravicious:
IIRC the branch was based on your branch with advanced search, right?

It'd be nice to clean up the git history of this one anyway since we did a lot of WIP-style commits.

PR #1407, opened 2022-12-05
Connect: Tell fpm to not use symlinks when building the rpm package

ravicious:
Fixes gravitational/teleport#18859.

Grzegorz added the flag but wasn't able to test this. I installed Fedora on my old Intel MBP, verified that Connect 11.1.0 can't be installed if Slack is already installed. Then I built a local version with this change, verified that it fixes the problem, then I built `11.1.2-dev.ravicious.1` and again verified that it works.

### DIFF
--- a/packages/teleterm/README.md
+++ b/packages/teleterm/README.md
@@ -137,6 +137,9 @@ See [Teleport Connect build
 process](https://gravitational.slab.com/posts/teleport-connect-build-process-fu6da5ld) on Slab for
 bulid process documentation that is specific to Gravitational.
 
+### Linux
+To create RPM package for aarch64 target you need to provide `USE_SYSTEM_FPM=1` env var. 
+
 ### macOS
 
 To make a fully-fledged build on macOS with Touch ID support, you need two things:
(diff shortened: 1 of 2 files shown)


packages/teleterm/README.md
```diff
@@ -137,6 +137,16 @@ See [Teleport Connect build
...
 
+### Native dependencies
+
+If node-pty doesn't provide precompiled binaries for your system and the specific Electron version,
+you will need to install [the dependencies required by
+node-pty](https://github.com/microsoft/node-pty#dependencies).
+
+### Linux
+
+To create RPM package for aarch64 target you need to provide `USE_SYSTEM_FPM=1` env var.
```
avatus, line 148:
```suggestion
To create an RPM package for an aarch64 target you need to provide `USE_SYSTEM_FPM=1` env var.
```
I've never seen it called aarch64, TIL. Always thought people just used `arm64`.

avatus (approved):

packages/teleterm/README.md
ravicious: We tend to use arm64 within our codebases too so I'm going to rename it.

PR #1422, opened 2022-12-08
Fix Connect stories

ravicious:


packages/teleterm/src/ui/DocumentCluster/DocumentCluster.tsx
```diff
@@ -38,10 +38,16 @@ export default function Container(props: DocumentProps) {
   );
 
   useEffect(() => {
+    const rootClusterUri = routing.ensureRootClusterUri(clusterUri);
+
+    if (!appCtx.workspacesService.getWorkspace(rootClusterUri)) {
+      return;
+    }
+
```
ravicious, line 46: This shouldn't affect the real app, as far as I can tell. If there's no workspace, `getWorkspaceDocumentService` is going to return a DocumentService that throws the moment you try to update anything.

This won't happen in the real app as before we remove the cluster, we remove all docs that are related to it. But that was throwing an error in tests since we wanted to force a situation where the cluster is not found.

---

Now that I wrote that, perhaps a better move would be to use a leaf cluster here as an example. I don't think we'll ever have a situation in which we render the "not found" page for a root cluster, but it's totally possible for a leaf cluster to stop existing.

avatus (approved):

PR #1430, opened 2022-12-09
[Connect] Use resourcesList in review access request table

avatus:
Protobuf updates to support this [webapps.e PR](https://github.com/gravitational/webapps.e/pull/453).

Small UI update to support https://github.com/gravitational/teleport/pull/19258


gzdunek (approved):
Oops, I totally missed this one 😬

PR #1431, opened 2022-12-12
Display instructions for connecting to a database with passwordless auth

djanda97:
Fixes https://github.com/gravitational/teleport/issues/15353

### DIFF
--- a/packages/teleport/src/Databases/ConnectDialog/ConnectDialog.test.tsx
+++ b/packages/teleport/src/Databases/ConnectDialog/ConnectDialog.test.tsx
@@ -39,14 +39,23 @@ test('correct connect command generated for mysql db', () => {
 
 test('correct tsh login command generated with local authType', () => {
   render(<ConnectDialog {...props} />);
-  const output = 'tsh login --proxy=localhost:443 --auth=local --user=yassine';
+  const output =
+    'tsh login --proxy=localhost:443 --auth=local --user=yassine im-a-cluster';
 
   expect(screen.getByText(output)).toBeInTheDocument();
 });
 
 test('correct tsh login command generated with sso authType', () => {
   render(<ConnectDialog {...props} authType="sso" />);
-  const output = 'tsh login --proxy=localhost:443';
+  const output = 'tsh login --proxy=localhost:443 im-a-cluster';
+
+  expect(screen.getByText(output)).toBeInTheDocument();
+});
+
+test('correct tsh login command generated with passwordless authType', () => {
+  render(<ConnectDialog {...props} authType="passwordless" />);
+  const output =
+    'tsh login --proxy=localhost:443 --auth=passwordless --user=yassine im-a-cluster';
 
   expect(screen.getByText(output)).toBeInTheDocument();
 });
--- a/packages/teleport/src/Databases/ConnectDialog/ConnectDialog.tsx
+++ b/packages/teleport/src/Databases/ConnectDialog/ConnectDialog.tsx
@@ -26,26 +26,16 @@ import { DbProtocol } from 'shared/services/databases';
 
 import { AuthType } from 'teleport/services/user';
 import TextSelectCopy from 'teleport/components/TextSelectCopy';
+import { generateTshLoginCommand } from 'teleport/lib/util';
 
 export default function ConnectDialog({
   username,
+  clusterId,
   dbName,
   onClose,
   authType,
   accessRequestId,
 }: Props) {
-  const { hostname, port } = window.document.location;
-  const host = `${hostname}:${port || '443'}`;
-
-  const connectCmd =
-    authType === 'sso'
-      ? `tsh login --proxy=${host}`
-      : `tsh login --proxy=${host} --auth=local --user=${username}`;
-
-  const requestIdFlag = accessRequestId
-    ? ` --request-id=${accessRequestId}`
-    : '';
-
   return (
     <Dialog
       dialogCss={() => ({
@@ -65,7 +55,15 @@ export default function ConnectDialog({
             Step 1
           </Text>
           {' - Login to Teleport'}
-          <TextSelectCopy mt="2" text={`${connectCmd}${requestIdFlag}`} />
+          <TextSelectCopy
+            mt="2"
+            text={generateTshLoginCommand({
+              authType,
+              clusterId,
+              username,
+              accessRequestId,
+            })}
+          />
         </Box>
         <Box mb={4}>
           <Text bold as="span">
--- a/packages/teleport/src/lib/util.ts
+++ b/packages/teleport/src/lib/util.ts
@@ -48,26 +48,32 @@ export async function Sha256Digest(
 }
 
 export type TshLoginCommand = {
-  accessRequestId?: string;
-  username: string;
   authType: AuthType;
-  clusterId: string;
+  clusterId?: string;
+  username: string;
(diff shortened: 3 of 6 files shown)


ravicious:
/gcbrun

Thanks for the PR!

packages/teleport/src/Databases/ConnectDialog/ConnectDialog.tsx
```diff
@@ -37,10 +37,18 @@ export default function ConnectDialog({
...
+  let connectCmd;
+  switch (authType) {
+    case 'sso':
+      connectCmd = `tsh login --proxy=${host}`;
+      break;
+    case 'local':
+      connectCmd = `tsh login --proxy=${host} --auth=local --user=${username}`;
+      break;
+    case 'passwordless':
+      connectCmd = `tsh login --proxy=${host} --auth=passwordless --user=${username}`;
```
ravicious, line 49: I did a double check to confirm that the `user` flag works together with `--auth=passwordless` and it does.

If I don't pass the user flag and I have multiple passwordless MFAs set up, tsh will ask me to pick one of them. If I pass the user flag, it won't ask and automatically log in as that user.

ravicious (approved):

ravicious:
We should backport this PR to v10 and v11 after it gets merged.

hatched (changes requested):
Thanks for this update @djanda97 

I'd like to see the tests for these components updated to test this new case please.

packages/teleport/src/Databases/AddDatabase/AddDatabase.tsx
```diff
@@ -75,10 +75,19 @@ export function AddDatabase({
...
+    case 'sso':
+      connectCmd = `tsh login --proxy=${host}`;
+      break;
+    case 'local':
+      connectCmd = `tsh login --proxy=${host} --auth=local --user=${username}`;
+      break;
+    case 'passwordless':
+      connectCmd = `tsh login --proxy=${host} --auth=passwordless --user=${username}`;
+      break;
+  }
```
ryanclark, line 89: Thanks for the contribution!

As this is the same as what's used in `ConnectDialog.tsx`, could we make this into a helper function that lives in somewhere like `src/Databases/utils.ts`?

```suggestion
function getConnectCommand(authType: AuthType) {
  switch (authType) {
    case 'sso':
      return `tsh login --proxy=${host}`;
    case 'local':
      return `tsh login --proxy=${host} --auth=local --user=${username}`;
    case 'passwordless':
      return `tsh login --proxy=${host} --auth=passwordless --user=${username}`;
  }
  
  throw new Error(`auth type ${authType} is not supported`);
}
```

Should make it easier to add a test case for the logic 🙂

djanda97: @ryanclark
Good idea!
I couldn't find a `utils.ts` file in the `Databases` directory, but I located an existing function called `generateTshLoginCommand` in `packages/teleport/src/lib/util.ts`.
Do you think it would be appropriate to update the `generateTshLoginCommand` function to include passwordless auth, then we could reference `generateTshLoginCommand` from the database components and tests?

ryanclark: Sorry for the confusion - I meant make `utils.ts` in the `Database` folder.

Good idea on changing `generateTshLoginCommand` to work the same - let's go for that, and it'll make testing easier as there's already tests for that method in https://github.com/gravitational/webapps/blob/master/packages/teleport/src/lib/util.test.ts#L17

packages/teleport/src/Databases/ConnectDialog/ConnectDialog.tsx
```diff
@@ -26,26 +26,16 @@ import { DbProtocol } from 'shared/services/databases';
 
 import { AuthType } from 'teleport/services/user';
 import TextSelectCopy from 'teleport/components/TextSelectCopy';
+import { generateTshLoginCommand } from 'teleport/lib/util';
 
 export default function ConnectDialog({
   username,
+  clusterId,
```
djanda97, line 33: I'm still new to learning TypeScript, so I was wondering if it is alright to add `clusterId` here?
I added `clusterId` here since the `generateTshLoginCommand` requires it and it was already included in the `Props` type for the `ConnectDialog` component.

packages/teleport/src/Databases/AddDatabase/AddDatabase.tsx
djanda97: I've updated the `generateTshLoginCommand` function to check for the passwordless login authType, as well as refactored the `ConnectDialog` component to utilize it.

For the `AddDatabase` component however, I am unsure how to proceed since the `Props` type for the component does not include `clusterId` or `accessRequestId`, which are required for the `generateTshLoginCommand` function.

Could we just add those fields to the `Props` type of the `AddDatabase` component?

ryanclark: You can make `clusterId` and `accessRequestId` optional on the `TshLoginCommand` type

packages/teleport/src/Databases/ConnectDialog/ConnectDialog.test.tsx
```diff
@@ -39,14 +39,23 @@ test('correct connect command generated for mysql db', () => {
...
 
 test('correct tsh login command generated with sso authType', () => {
   render(<ConnectDialog {...props} authType="sso" />);
-  const output = 'tsh login --proxy=localhost:443';
+  const output = 'tsh login --proxy=localhost:443 im-a-cluster';
+
+  expect(screen.getByText(output)).toBeInTheDocument();
+});
+
+test('correct tsh login command generated with passwordless authType', () => {
```
djanda97, line 55: Hi @hatched, I've added a test case to verify that the `tsh login` command is correct when passwordless `authType` is used. Do you have some recommendations for other test cases that could be added/modified?

packages/teleport/src/lib/util.ts
```diff
@@ -48,10 +48,10 @@ export async function Sha256Digest(
 }
 
 export type TshLoginCommand = {
-  accessRequestId?: string;
-  username: string;
   authType: AuthType;
-  clusterId: string;
+  clusterId?: string;
```
hatched, line 52: Now that `clusterId` and `accessRequestId` are optional it's a good practice if they have a default value of the same type. 

Can you define some defaults for them? An empty string should be fine, you can do this in the restructuring in the function signature.

packages/teleport/src/Databases/ConnectDialog/ConnectDialog.test.tsx
hatched: Nope this looks good 👍

hatched (commented):
Thanks for keeping up on this, just one tiny change plz and it'll be ready to go 😁

packages/teleport/src/lib/util.ts
djanda97: Good idea, latest commit provides default values for the optional parameters 😄

hatched (approved):
🎉

ryanclark (approved):

ravicious:
@kimlisa I tried backporting this PR to v11 and v10 but there are some conflicts with `AddDatabase.tsx` that you've recently worked on, do you mind doing the backports?

PR #1434, opened 2022-12-14
Update electron-builder to 24.0.0-alpha.5

ravicious:
At first I thought that the fix for gravitational/teleport#18859 broke symlinking `teleport-connect` in the rpm package but it actually turned out to be a regression in electron-builder. alpha.5 creates the symlink properly and everything else works fine as well.

I remember there was a reason why we had to use an alpha version of electron-builder but I don't remember the details, it was related to the Windows build I think?

By now there's alpha.7 of electron-builder but:

* I made a tag release with alpha.5 (11.1.3-dev.ravicious.1) and went through [the update checklist](https://gravitational.slab.com/posts/electron-builder-update-checklist-wr07avto). Everything seems to work fine.
* The tag build with alpha.6 didn't finish successfully because of an issue with finding a cert for signing.
   * I suspect this is because of a regression in electron-builder. But it also might be due to changes in [the Drone worker pool](https://gravitational.slack.com/archives/C08HF8E9F/p1670612993832519). I asked Fred to verify if the two failed builds that I linked in the thread were done on the same machine as the alpha.5 build.
   * alpha.6 updates quite a few underlying deps so I think once v24 comes out it'll be worthwhile to update electron-builder to stable v24.
* By the time alpha.7 came out, I just gave up on making another round of builds and checks.


gzdunek:
>I remember there was a reason why we had to use an alpha version of electron-builder but I don't remember the details, it was related to the Windows build I think?

electron-builder v23 could not rebuild native deps for Electron 20+ (for all systems), so we updated it to v24 alpha. Under the hood it switched to electron-rebuild for building deps because of this issue.
Now I see that Electron [started to provide the correct headers](https://github.com/electron/electron/pull/36369) on its own, so older tools should work.

I just did a test with electron-builder: 23.6.0 and electron: 21.3.3 (it has the updated headers) and it works, at least on macOS. I wish it worked like this earlier, probably we would save a lot of time 😏

gzdunek (approved):

avatus (approved):

ravicious:
OTOH, I think alpha.5 emits this warning when I installed a deb package made with alpha.5:

> W: Repository is broken: teleport-connect:arm64 (= 1.0.2) has no Size information

Installing 11.1.2 doesn't emit thin message. AFAIK this is just a warning but yeah, another broken thing.

gzdunek:
I guess it can be related to [feat: Introducing deb and rpm auto-updates](https://github.com/electron-userland/electron-builder/releases/tag/v24.0.0-alpha.5). Maybe you want to create an issue in electron-builder for this? :p

PR #1442, opened 2022-12-15
Adds a status code to the closing of the tdp client's websocket

ibeckermayer:
Corresponds to https://github.com/gravitational/teleport/pull/19389


michellescripts (approved):

packages/teleport/src/lib/term/enums.js
```diff
@@ -34,4 +34,5 @@ export const TermEventEnum = {
 
 export const StatusCodeEnum = {
```
zmb3, line 35: Might be worth linking to https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent/code and even renaming this to something like `WebsocketCloseCode` so it's clear that this is a protocol thing and not arbitrary values.

As a side note, I'm not a huge fan of naming every enum `FooEnum`. We don't name every const `FooConst` or every class `FooClass`, so it feels odd to do that only for enumerations. But it appears there's precedent here so I'll leave that part up to you.

packages/teleport/src/lib/tdp/client.ts
```diff
@@ -581,8 +582,8 @@ export default class Client extends EventEmitterWebAuthnSender {
   // so don't call this if your calling object is relying on listeners.
   // It's safe to call this multiple times, calls subsequent to the first call
   // will simply do nothing.
-  nuke() {
+  nuke(statusCode = StatusCodeEnum.NORMAL) {
```
zmb3, line 585: I've always found `nuke` to be a poor/insensitive choice of name here. Would `close` or `shutdown` be more appropriate?

zmb3 (approved):

packages/teleport/src/lib/term/enums.js
ibeckermayer: I believe we're doing that because this is a Javascript file which doesn't support enums. I can convert it to Typescript and remove the `Enum` suffix

ibeckermayer: 416841ec54dfedf06b52c8dc645fb7f4b105f355

PR #1443, opened 2022-12-16
Connect: Store installation ID

gzdunek:
This PR adds a new field to the `RuntimeSettings` called `installationId`. This field will then be sent with each usage event as `distinctId`, so we can distinguish events coming from different app instances.

There is a risk that the user will delete the file, so we will have to generate a new UUID and count the user as new 😞
The alternative can be a library like [systeminformation](https://github.com/sebhildebrandt/systeminformation), that will return OS UUID, but there are [some issues](https://github.com/sebhildebrandt/systeminformation/issues/747) that seem to be important for us too.
EDIT: I don't think this is a big problem though, we will be using metrics such as DAU (Daily Active Users) to count users, so the overall number should not change. We can stay with the approach from the RFD.


packages/teleterm/src/mainProcess/runtimeSettings.ts
```diff
@@ -209,4 +211,32 @@ function getUnixSocketNetworkAddress(socketName: string) {
   return `unix://${path.resolve(app.getPath('userData'), socketName)}`;
 }
 
+const UUID_V4_REGEX =
+  /^[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i;
+
+function loadInstallationId(): string {
+  const filePath = path.resolve(app.getPath('userData'), 'installation_id');
+  const writeNewId = (): string => {
+    try {
```
ravicious, line 220: Wouldn't it be better to wrap the `try` as tightly as possible here, that is only around `fs.writeFileSync`?

packages/teleterm/src/mainProcess/runtimeSettings.ts
```diff
@@ -209,4 +211,32 @@ function getUnixSocketNetworkAddress(socketName: string) {
...
+      fs.writeFileSync(filePath, newId);
+      return newId;
+    } catch (error) {
+      throw new Error(
+        `Could not write installation_id to ${filePath}, ${error.message}`
+      );
+    }
+  };
+
+  try {
```
ravicious, line 231: Same as above, if I understand correctly we want to handle errors only from `fs.readFileSync`, right?

packages/teleterm/src/mainProcess/runtimeSettings.ts
```diff
@@ -209,4 +211,32 @@ function getUnixSocketNetworkAddress(socketName: string) {
   return `unix://${path.resolve(app.getPath('userData'), socketName)}`;
 }
 
+const UUID_V4_REGEX =
+  /^[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i;
+
+function loadInstallationId(): string {
+  const filePath = path.resolve(app.getPath('userData'), 'installation_id');
+  const writeNewId = (): string => {
```
ravicious, line 219: This is more of a personal preference I guess but this function could have been a regular top-level function which accepts path as an argument. 😶‍🌫️

For some reason, I find the pattern of defining ad-hoc functions like this somewhat counterintuitive. It's kinda similar to this common approach where you cannot really follow the execution of the function top to bottom:

```typescript
const onSuccess = (foo, bar) => { … }
const onError = (error) => { … }

something(baz, onSuccess, onError)
```

I know that React kinda popularized this approach, but most of the time those ad-hoc functions you define in React are simple event handlers where you don't really care about the side effects or they just render HTML. Here they directly impact the execution path of the whole function.

ravicious (approved):
I'd say the issue you linked to doesn't necessarily prevent us from using systeminformation. We wouldn't accept input from user which would be then used to call functions from that package, we'd probably just grab the UUID and that's it.

Alas, this is just for telemetry, we're not implementing license key checks or anything like that so your approach here should be fine, I think.

avatus (approved):

packages/teleterm/src/mainProcess/runtimeSettings.ts
gzdunek: Yeah, done.

packages/teleterm/src/mainProcess/runtimeSettings.ts
gzdunek: I thought that `writeNewId` makes sense only in a context of `loadInstallationId` so I wanted to make it available only for that function. I could create a new file for it, but these functions did not seem complex enough to justify it. 
But I'm fine with converting `writeNewId` to a top-level function.

gzdunek:
>I'd say the issue you linked to doesn't necessarily prevent us from using systeminformation. We wouldn't accept input from user which would be then used to call functions from that package, we'd probably just grab the UUID and that's it.

Right, but AFAIK is still requires calling `reg.exe`. If we can avoid it, I would do it, especially if the local file is sufficient.

packages/teleterm/src/mainProcess/runtimeSettings.ts
ravicious: > I could create a new file for it, but these functions did not seem complex enough to justify it.

I'd say in situations like that it's more than fine to do it! `runtimeSettings` is already complex enough so abstracting all this stuff in a separate file is good IMHO.

gzdunek: Done!

PR #1445, opened 2022-12-20
Make bundled tsh available outside of Connect

ravicious:
This PR implements RFD 99 (gravitational/teleport#19284).

TODO:

* [x] Windows
* [x] Linux
* [x] macOS
* [x] Update [electron-builder update checklist](https://gravitational.slab.com/posts/electron-builder-update-checklist-wr07avto) with an item related to the Linux scripts

I'm yet to add macOS support but I figured I'll submit the other changes for review already and then submit the macOS ones as another PR to this PR. macOS support will involve more changes in the application code.

So far there wasn't any substantial changes compared to the RFD but I'll add comments under some of the minor differences.

On Windows I ended up just adding a path to an env var, as I mentioned in the RFD. On Linux, I had to copy the install/uninstall scripts as there's no way to extend the default ones.


packages/teleterm/build_resources/linux/after-remove.tpl
```diff
@@ -0,0 +1,44 @@
...
+# Remove the link to the Electron app binary.
+rm -f "$BIN/${executable}"
+
+# At this point, the app has already been removed from disk. If TSH_SYMLINK_TARGET used to point at
+# tsh bundled with the teleport-connect package, it is a broken symlink now.
+#
+# Is TSH_SYMLINK_TARGET a link that points at a file which doesn't exist?
+if [ -L "$TSH_SYMLINK_TARGET" ] && [ ! -e "$TSH_SYMLINK_TARGET" ]; then
+  rm -f "$TSH_SYMLINK_TARGET"
+fi
```
ravicious, line 42: The RFD says:

> If Connect decides not to create or remove the symlink, it should echo a message explaining the reason.

Here I've decided to not echo a message if Connect doesn't remove the symlink. Idk, it doesn't seem like a necessary thing to me and we already emit a message regarding symlinks during installation.

I'm not even sure if I should emit a message during an upgrade. But I wanted to be able to tell what happened if someone sends logs from an upgrade. OTOH I don't think Linux packages echo stuff from within those scripts?

packages/teleterm/build_resources/linux/after-install.tpl
```diff
@@ -0,0 +1,59 @@
...
+    echo "$message"
+  fi
+fi
+
+# DELETE IN 13.0.0
+# Remove the old symlink from times where we used the default after-install.tpl.
+OLD_SYMLINK=/usr/bin/teleport-connect
+if [ -e "$OLD_SYMLINK" ]; then
+    rm $OLD_SYMLINK
+fi
```
ravicious, line 57: I changed the location from `/usr/bin` to `/usr/local/bin` since this is where our teleport package places the binaries as well.

packages/teleterm/build_resources/linux/after-install.tpl
```diff
@@ -0,0 +1,59 @@
...
+# Custom after-install.tpl script.
+###
+
+APP="/opt/${sanitizedProductName}"
+BIN=/usr/local/bin
+TSH_SYMLINK_SOURCE=$APP/resources/bin/tsh
+TSH_SYMLINK_TARGET=$BIN/tsh
+
+# Create $BIN if it doesn't exist.
+[ ! -d "$BIN" ] && mkdir -p "$BIN"
```
ravicious, line 25: Here I'm just following what [tsh.pkg does already](https://github.com/gravitational/teleport/blob/380b572152b7126b1458e19a1d994c13af0b70cd/build.assets/macos/scripts/postinstall#L6).

packages/teleterm/build_resources/linux/after-install.tpl
```diff
@@ -0,0 +1,59 @@
...
+      message+=" because the symlink already exists and it points to $(realpath $TSH_SYMLINK_TARGET)."
+      echo "$message"
+    fi
+  else
+    message+=" because $TSH_SYMLINK_TARGET already exists and it isn't a symlink."
+    echo "$message"
+  fi
+fi
+
+# DELETE IN 13.0.0
```
gzdunek, line 52: Hmm, wy in v13?

gzdunek:
I have no experience with scripts, but the code looks pro :) 
Tomorrow I'd like to test the app on Linux/Windows.

avatus:
This looks great! I'm excited for this.
Could you make a tag build so we can all individual test it on a bunch of different platforms? Thank you!

packages/teleterm/build_resources/linux/after-install.tpl
ravicious: Ah, I figured that there might be people who would upgrade from, let's say, 11.0.1 straight to 12.x. If we removed this in 12.0, they'd be left with a broken symlink on their system. So hopefully if we do this in v13 there will be less people with that problem.

But! When I was writing this comment I realized that the removal of the old symlink is completely unnecessary. When people upgrade from an old version, their package manager is still going to run after-remove from the old version which is going to remove the symlink.

https://github.com/electron-userland/electron-builder/blob/v24.0.0-alpha.5/packages/app-builder-lib/templates/linux/after-remove.tpl

ravicious:
I just promoted `12.0.0-dev.ravicious.1`.

gzdunek:
Just to make sure: should I see the following message if I remove Connect on Fedora?
```
[parallels@fedora code]$ tsh
bash: /usr/local/bin/tsh: No such file or directory
```

I tested the app also on Ubuntu and Windows, everything works fine!

ravicious:
> Just to make sure: should I see the following message if I remove Connect on Fedora?

I think that's correct because the shell within which you removed tsh still had tsh resolved to /usr/local/bin/tsh but now that file is gone. If you open a new shell and try to run `tsh` it should give you an error along the lines of "command not found".

gzdunek:
> > Just to make sure: should I see the following message if I remove Connect on Fedora?
> 
> I think that's correct because the shell within which you removed tsh still had tsh resolved to /usr/local/bin/tsh but now that file is gone. If you open a new shell and try to run `tsh` it should give you an error along the lines of "command not found".

Ah, right

gzdunek (approved):

avatus:
Working great on win 10/11

avatus (approved):

PR #1449, opened 2022-12-20
[Connect] Update protofiles and webapps.e to support suggestedReviewers and requestableRoles

avatus:
Protofile (and test) updates for https://github.com/gravitational/webapps.e/pull/455


ravicious (approved):

ravicious:
While you're at it, could you add a TODO comment to the call to `this.fetchClusterAssumedRequests` in `ClustersService.syncClusterInfo`? I think that at some point we'd like to return this data from `client.getCluster` rather than having to make a separate call.

gzdunek (approved):

packages/teleterm/src/ui/hooks/useLoggedInUser.ts
```diff
@@ -0,0 +1,23 @@
+import { useAppContext } from 'teleterm/ui/appContextProvider';
+import { LoggedInUser } from 'teleterm/services/tshd/types';
+
+export function useLoggedInUser() {
+  const ctx = useAppContext();
+
+  ctx.clustersService.useState();
+  ctx.workspacesService.useState();
+
+  function getLoggedInUser(): LoggedInUser | undefined {
```
ravicious, line 10: What's the reason behind wrapping this in a function?

avatus: IIRC it was just refactored from the previous iteration of where we were getting `loggedInUser` in a different hook. I'll take it out of a function

gzdunek (approved):

PR #1450, opened 2022-12-21
Connect: Use default config in main process fixtures

ravicious:
This unbreaks some Connect stories which depend on keyboard shortcuts config.

If we ever need to change the config in stories or tests, we should revisit the mock. But it might be enough to call the `set` method on the config service.


gzdunek (approved):

ravicious:
@gzdunek I guess we'll have to remember to backport this one once you backport your PRs related to the config service?

PR #1452, opened 2022-12-21
Add downloads page

mcbattirola:
This PR adds two new fields to the user ACL `download` and `license`. These fields will be used in `webapps.e` to hide or show a new page, Downloads.

Also, this PR adds a new component to display external links in the side menu. Here's what it looks like in the web UI:

Screenshot from 2022-12-23 11-30-46

Both of these features will be used by a PR in `webapps.e`.

Related PRs:

- https://github.com/gravitational/teleport/pull/19565
- https://github.com/gravitational/teleport.e/pull/671
- https://github.com/gravitational/webapps/pull/1452
- https://github.com/gravitational/webapps.e/pull/457


packages/teleport/src/Discover/Discover.test.tsx
```diff
@@ -58,6 +58,8 @@ const fullAcl: Acl = {
   desktopSessionRecordingEnabled: true,
   directorySharingEnabled: true,
   connectionDiagnostic: fullAccess,
+  license: fullAccess,
+  download: fullAccess,
```
michellescripts, line 62: Nit: You didn't introduce this, but I don't love that we need to update this in 3 places. Since we have an exported const that is the same value in `packages/teleport/src/Main/fixtures/index.ts` why aren't we leveraging that for the two tests?

packages/teleport/src/SideNav/SideNavExternalLink.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  return css;
+};
+
+const SideNavExternalLink = ({ children, href, icon }) => {
+  return (
+    <StyledSideNavExternalLink href={href} target="_blank">
+      <SideNavItemIcon as={icon} />
+      {children}
+      <SideNavItemIcon
```
michellescripts, line 73: This is inheriting `margin-right: 16px;` from SideNavItemIcons default props... and it's not exactly lined up with the other icons, If we want them aligned that is.

Screenshot 2022-12-28 at 1 55 03 PM

kimlisa (commented):
i just briefly looked at this and i wanted to understand a couple of things:

- is there a figma design for this? if so, could you link it? (or maybe an issue with proposed design?) 
- so if i understand correctly, we are now having a `Support` and `Downloads` as part of the side nav correct? We already have a `Help & Support` page (image below) where `Create a Support Ticket` links to the same thing (this menu is part of the username click dropdown menu on top right corner of our app). Are we keeping both? 
- Have we decided against putting it in our user dropdown menu for more visibility? (personally it makes more sense to put in user dropdown)
- I wonder how hard it would be to update our `Main` component and feed downloads/license fixtures so we can see it in storybook instead of hooking things up (it seems really complicated to set up). One of the reason for the Main component to exist was originally to help other folks test it out without setting things up (eg running teleport).

image

image

packages/teleport/src/stores/storeUserContext.ts
```diff
@@ -130,6 +130,10 @@ export default class StoreUserContext extends Store<UserContext> {
     return tokens.create;
   }
 
+  getAccessDownloadCenter() {
+    return this.state.acl.license.list || this.state.acl.download.list;
```
kimlisa, line 134: i hunted this down:

```
- Users are able to download binaries and licenses on the web UI
- Downloading binaries is only allowed for users with permission to read and list the resource download in their roles
- Downloading licenses is only allowed for users with permission to read and list the resource license in their roles
```

couple of things:
- the drop down and selector is not centered (image below)
- should user have both `read` and `list` perms?
- first impression of the design (below image), it's pretty confusing, i think it should be like two separate sections `Download License` (and description) and `Download Binaries` especially if we want to handle the below edge case:

i'm probably pointing out a rare edge case where it will rarely be that a user will have `download` perm but not the `license` perm and vice versa, but do we want to handle the case?

image

kimlisa: this also confused me too, the `failure to download license` error (b/c this error seems to belong to the table): 
image

kimlisa: maybe something like this perhaps:
image

packages/teleport/src/Discover/Discover.test.tsx
mcbattirola: Yeah, makes sense.
Added this here https://github.com/gravitational/webapps/pull/1452/commits/f939601a1d04a21e3be3a510a8c6cfc648b8d6cb

packages/teleport/src/stores/storeUserContext.ts
```diff
@@ -130,6 +130,10 @@ export default class StoreUserContext extends Store<UserContext> {
     return tokens.create;
   }
 
+  getAccessDownloadCenter() {
```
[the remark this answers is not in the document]
mcbattirola: Thanks, added in https://github.com/gravitational/webapps/pull/1452/commits/7749c628209698d4112da869511a2920685bfd21 with a comment

packages/teleport/src/stores/storeUserContext.ts
mcbattirola: > the drop down and tab selector is not centered, and the tab icons for OS is also not centered (image below)

Fixed, thanks

> should user have both read and list perms?

I will double-check this with Xin. I think `list` should be all you need to download, while `read` for license (since there is only one license, nothing to list).


> first impression of the design (below image), it's pretty confusing, i think it should be like two separate sections Download License (and description) and Download Binaries especially if we want to handle the below edge case:

I have a sync with @roraback today, will bring this up and show your feedback. I agree that the error state will be confusing.

mcbattirola:
> is there a figma design for this? if so, could you link it? (or maybe an issue with proposed design?)

https://www.figma.com/file/eilCS832k68fSiKNNtg68D/Houston--%3E-SalesCenter?node-id=22%3A4178&t=KvMlT74JTHdhdi6G-0

> so if i understand correctly, we are now having a Support and Downloads as part of the side nav correct? We already have a Help & Support page (image below) where Create a Support Ticket links to the same thing (this menu is part of the username click dropdown menu on top right corner of our app). Are we keeping both?

> Have we decided against putting it in our user dropdown menu for more visibility? (personally it makes more sense to put in user dropdown)

We're keeping both AFAIK, but I'll forward this feedback to @roraback.

mcbattirola:
Update:

I've added a separate check for permissions inside the download page:

User with permission to download license but not binaries will see this (Kenny will provide a better copy, this is a just a placeholder)
Screenshot from 2022-12-29 13-45-42


User with permission to download binaries but not license will see this (same comment about copy as above)

Screenshot from 2022-12-29 13-45-04

packages/teleport/src/stores/storeUserContext.ts
roraback: Thanks for this great feedback @kimlisa ! Some thoughts:
* Support: tldr; I think we should hide the support page from this Houston replacement, because the links are mostly geared toward Teleport _users,_ whereas anyone in this UI is likely focused on deploying their self-hosted cluster. I'm open to other opinions, though, if folks disagree!
  * Original thought behind this was that with fewer options in the nav, it would make sense to include a support link there for greater visibility
  * I think I asked in an early discussion whether a help page would be useful, and the input was along the lines that a help page isn't crucial at this point for getting people off Houston (since they don't currently have a help page in Houston), and that the existing one in Teleport isn't so well suited to helping with the use case at hand.
  * All that said, I *am* somewhat conflicted on this. Part of me would love to reinforce the existing Teleport nav structure around help rather than create a new one. I think longer-term, it could be helpful to mirror the existing help page with something geared specifically toward enterprise deployment (and license update) needs.
* Binaries + License Key Permissions Errors: Agreed that this could confuse users. I'm going to come up with some gentler designs + messaging to inform users if they don't have permissions to access one or both download types.
* Separate tabbed structure for Download types: that feels like overkill to me, since there would just be two tabs and one of them wouldn't need to hold much more than a single button; and users often overlook secondary/tertiary tabs, so this could have the effect of burying the content in whichever tab isn't displayed by default.
  * That said, perhaps a bit of guidance around what to do with the license key—which Houston does NOT do currently—could call it out a bit more clearly as a separate piece from the binaries? My understanding is that our enterprise customers are typically walked through these deployment steps by Teleport during the POV phase, so I'm a bit unsure how much value it adds, but it would be low effort to include and might at least avert some confusion around updating the license.

I tried to respond to everything above, but let me know if I missed anything or if you have additional questions or concerns.

roraback:
@mcbattirola I'm looking at the latest screenshots, and I see that the nav is still loaded up with all of the resource types and other items—the design added downloads/support, but it also removed everything that is irrelevant for the self-hosted use case. Is that changing with a follow-up PR?

mcbattirola:
@roraback This is controlled by the user's permission, so if we create a user with a role that denies access to those screens, they won't show up. When we create the dashboards for users, we'll make sure they can only access downloads and the users screen.

Here's an example of a role that would hide those screens:


```yaml
kind: role
metadata:
  name: regular-guy
spec:
  allow:
    rules:
    - resources:
      - user
      verbs:
      - list
      - create
      - read
      - update
      - delete
    - resources:
      - role
      verbs:
      - list
      - read
  deny:
    rules:
    - resources:
      - node
      - app_server
      - trusted_cluster
      - windows_desktop_service
      - windows_desktop
      - access_request
      - plugin_data
      - kube_server
      - db_server
      - trusted_cluster
      - recordedSessions
      - access_request
      - ssh_session
      - access_request
      verbs:
      - list
      - create
      - read
      - update
      - delete
  options:
    cert_format: standard
    create_host_user: false
    desktop_clipboard: true
    desktop_directory_sharing: true
    enhanced_recording:
    - command
    - network
    forward_agent: false
    max_session_ttl: 30h0m0s
    pin_source_ip: false
    port_forwarding: true
    record_session:
      default: best_effort
      desktop: true
    ssh_file_copy: true
version: v5
```

Now, I noticed that `Access Requests` and `Clusters` are always showing up (hardcoded to `true` in the code). I'll open an issue to fix this.

roraback:
Great, thank you!

packages/teleport/src/stores/storeUserContext.ts
roraback: Here is the design I came up with for clarifying the role of the license key and binary downloads: 
image

mcbattirola:
Downloads page updated to follow new design (actual commit is in the webapps.e PR)

Screenshot from 2023-01-04 10-00-18

mcbattirola:
> Now, I noticed that Access Requests and Clusters are always showing up (hardcoded to true in the code). I'll open an issue to fix this

PRs fixing this:

https://github.com/gravitational/webapps/pull/1484
https://github.com/gravitational/webapps.e/pull/461

kimlisa:
question on this screenshot. from my understanding from this tuesdays meeting, there will never be a case where a user will see both the resource pages and the license/downloads pages? is my understanding correct @roraback 

I think maybe my confusion arise from how sasha gave an example using license, where say you have a houston license, you will only ever see license/downloads page (even if you gave yourself all perms (eg asteriks), which this screenshot implies). With non houston license, you will never be able to see the license/downloads page. Maybe it doesn't matter and I'm taking things too literally :sweat_smile: 

> Downloads page updated to follow new design (actual commit is in the webapps.e PR)
> 
> Screenshot from 2023-01-04 10-00-18

mcbattirola:
@kimlisa we don't want to have Teleport running in different "modes" because it can make things complex, we rather have these screens controlled by RBAC, as any other.

So the idea is just to have an undocumented resource that, when you have permission, will enable these features.

So in the end, having the download page along with other screens is still possible if the user enables everything.

The default roles will not have these resources enabled, so on-prem users will not see this unless they enable it. We will add these permissions when creating a user's dashboard tenant.

Does that make sense? Do you think having this `Downloads` screen showing up for a regular customer that enabled everything is too bad and we should make sure only the dashboards get this?

packages/teleport/src/SideNav/SideNav.tsx
```diff
@@ -94,3 +99,5 @@ export const Content = styled.div`
   flex-direction: column;
   overflow: auto;
 `;
+
+const isExternalLink = (link: string): boolean => link.startsWith('http');
```
kimlisa, line 103: what do you think about putting this here: https://github.com/gravitational/webapps/blob/4da005d774cff0820b50439541a3d6ee1f89d0b3/packages/teleport/src/SideNav/useSideNav.ts#L102, as an optional `externalLink?: boolean` field?

packages/teleport/src/SideNav/SideNavExternalLink.tsx
```diff
@@ -0,0 +1,97 @@
...
+import SideNavItemIcon from './SideNavItemIcon';
+
+const fromTheme = ({ $nested = false, theme = defaultTheme }) => {
+  const css = {
+    position: 'relative',
+    fontSize: '12px',
+    fontWeight: theme.regular,
+    fontFamily: theme.font,
+    paddingLeft: theme.space[9] + 'px',
+    paddingRight: theme.space[2] + 'px',
```
kimlisa, line 33: unless i'm missing something, this file looks almost identical to `SideNavItem.tsx`. If all we wanted was to change the padding (and type of element used which we can use the `as` prop) perhaps we can add `${space}` from design/system so we can pass in our own padding value? (OR, i'm pretty sure we can just use the inline `css` prop as well) 

like this;

```ts
import { color, space } from 'design/system';

const SideNavItem = styled.div`
  box-sizing: border-box;
  display: flex;
  align-items: center;
  justify-content: flex-start;
  border: none;
  border-left: 4px solid transparent;
  cursor: pointer;
  outline: none;
  text-decoration: none;
  width: 100%;
  line-height: 24px;
  ${fromTheme}  // default settings
  ${color}               // customize color
  ${space}            // customize spacing
`;

```
then we can do

```ts
// design system
const SideNavExternalLink = ({ children, href, icon }) => {
  return (
   // note the `as` prop, we can instruct this comp to be something other than a default `div`
    <SideNavItem as="a" href={href} target="_blank" pr={2}> 
      <SideNavItemIcon as={icon} />
      {children}
      <SideNavItemIcon
        css={{ marginLeft: 'auto', transform: 'rotate(-45deg)' }}
        as={ArrowForward}
      />
    </SideNavItem>
  );
};

// inline css (this is what i would do, more simple for one change)
const SideNavExternalLink = ({ children, href, icon }) => {
  return (
    // note the `as` prop, we can instruct this comp to be something other than a default `div`
    <SideNavItem as="a" href={href} target="_blank" css={{paddingRight: `${theme.space[2]}px`}}>  
      <SideNavItemIcon as={icon} />
      {children}
      <SideNavItemIcon
        css={{ marginLeft: 'auto', transform: 'rotate(-45deg)' }}
        as={ArrowForward}
      />
    </SideNavItem>
  );
};

// another way to define inline css, really helpful if you need to use pseudo classes (eg: :hover { stuff })
const SideNavExternalLink = ({ children, href, icon }) => {
  return (
    // note the `as` prop, we can instruct this comp to be something other than a default `div`
    <SideNavItem as="a" href={href} target="_blank" css={`padding-right: ${theme.space[2]}px}>  
      <SideNavItemIcon as={icon} />
      {children}
      <SideNavItemIcon
        css={{ marginLeft: 'auto', transform: 'rotate(-45deg)' }}
        as={ArrowForward}
      />
    </SideNavItem>
  );
};
```

kimlisa (commented):
regarding your reply, i don't see why it would be problem to let admins see the downloads

packages/teleport/src/SideNav/SideNav.tsx
mcbattirola: I think this is better. This way we explicitly define that a nav item is external in `features`, rather than checking for `http`. Added in https://github.com/gravitational/webapps/pull/1452/commits/a2341e7016700db8c9afb309fdf24bb126b1c49f

packages/teleport/src/SideNav/SideNavExternalLink.tsx
mcbattirola: I just copied it at first and forgot to refactor it back. I used your inline CSS approach, looks great now, thank you very much!

michellescripts (approved):

packages/teleport/src/SideNav/SideNavItem.tsx
```diff
@@ -74,6 +74,7 @@ const SideNavItem = styled.div`
   line-height: 24px;
   ${fromTheme}
   ${color}
+  ${space}
```
kimlisa, line 77: don't think we need this since we opted for inline css

kimlisa (approved):
thanks for all the changes!

PR #1464, opened 2022-12-28
Import prehog protos

ravicious:
teleport PR: gravitational/teleport#19700


packages/teleterm/src/services/tshd/v1/usage_events_pb.js
```diff
@@ -17,18 +17,11 @@ var global = (function() { return this || window || global || self || Function('
 
 var google_protobuf_timestamp_pb = require('google-protobuf/google/protobuf/timestamp_pb.js');
 goog.object.extend(proto, google_protobuf_timestamp_pb);
-goog.exportSymbol('proto.teleport.terminal.v1.AccessRequestAssumeRoleEvent', null, global);
-goog.exportSymbol('proto.teleport.terminal.v1.AccessRequestCreateEvent', null, global);
-goog.exportSymbol('proto.teleport.terminal.v1.AccessRequestReviewEvent', null, global);
-goog.exportSymbol('proto.teleport.terminal.v1.ClusterProperties', null, global);
-goog.exportSymbol('proto.teleport.terminal.v1.ConnectUsageEventOneOf', null, global);
-goog.exportSymbol('proto.teleport.terminal.v1.ConnectUsageEventOneOf.EventCase', null, global);
+var prehog_v1alpha_connect_pb = require('../prehog/v1alpha/connect_pb.js');
```
ravicious, line 20: This is how the import paths look like in the generated JS files so that's why I put prehog JS files under `services/tshd/prehog/v1alpha`.

For now I think that's okay but if we ever need to change the paths, there should be a way to achieve this through some config options.

PR #1465, opened 2022-12-28
Improve error message when trying to rename resource

vitorenesduarte:
Fixes https://github.com/gravitational/teleport/issues/18771.

(Previous attempt at https://github.com/gravitational/teleport/pull/19624)

This commit improves the error message when a user tries to rename a resource (role, auth connector and trusted cluster) using the web UI.

Doing this validation on the client-side prevents the situation described in https://github.com/gravitational/teleport/pull/19624#discussion_r1057335629.

#### Testing

Screenshot 2022-12-28 at 20 23 35
Screenshot 2022-12-28 at 20 24 33
Screenshot 2022-12-28 at 20 24 07

### DIFF
--- a/packages/teleport/package.json
+++ b/packages/teleport/package.json
@@ -19,6 +19,7 @@
   "dependencies": {
     "@gravitational/design": "1.0.0",
     "@gravitational/shared": "1.0.0",
+    "yaml": "^2.2.0",
     "xterm": "^5.0.0",
     "xterm-addon-fit": "^0.5.0"
   },
--- a/packages/teleport/src/AuthConnectors/AuthConnectors.tsx
+++ b/packages/teleport/src/AuthConnectors/AuthConnectors.tsx
@@ -46,7 +46,8 @@ export function AuthConnectors(props: State) {
 
   function handleOnSave(content: string) {
     const isNew = resources.status === 'creating';
-    return save(content, isNew);
+    const previousName = resources.item.name;
+    return save(content, isNew, previousName);
   }
 
   return (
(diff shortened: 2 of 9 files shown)


zmb3:
Let's improve the copy on these errors a bit:

- _resource name in "metadata.name" is not found_
  - "Resource metadata is missing the "name" field."
- _resource definition is not a valid YAML_
  - "a valid YAML" sounds awkward to me. How about this? "Invalid YAML. Resource not saved. Correct the resource definition and try again."

packages/teleport/package.json
```diff
@@ -19,6 +19,7 @@
   "dependencies": {
     "@gravitational/design": "1.0.0",
     "@gravitational/shared": "1.0.0",
+    "yaml": "^2.2.0",
```
zmb3, line 22: I'm on the fence about including this, just because it's probably a heavy dependency and the backend has to verify the YAML anyway.

ravicious: @zmb3 Does it mean you'd rather do that on the backend?

vitorenesduarte: We could include the resource name as parameter in the API endpoint so that the validation can be done on the server.
Something like this:
- client: https://github.com/gravitational/webapps/commit/6adee1f637ab62fac8ede1578421c12e84b9d565
- server: https://github.com/gravitational/teleport/commit/f85288ad24572198f1f8fd18fb11ae0bf46ee251

WDYT?

ravicious: That looks better to me, what do you think @zmb3?

vitorenesduarte: I would go checking this server-side so that all validation is done on the server (and no new dependency).

@ravicious @zmb3 If you're ok with this, I'll wait for the webapps repo merge so that we don't have to change 4 repos.

vitorenesduarte:
Closing in favor of https://github.com/gravitational/teleport/pull/20413 and https://github.com/gravitational/teleport.e/pull/713.

PR #1466, opened 2022-12-29
capture additional prehog events

michellescripts:
supports https://github.com/gravitational/cloud/issues/2520
Part of a series of PRs to add the orange-* events belonging to phase 1 reporting
Removes one unused event

Screenshot 2023-01-03 at 9 39 46 AM

Adds mfaType & loginFlow to PreUserOnboardRegisterChallengeSubmitEvent
Moves the credential event to onSuccess
Adds the"PreUserCompleteGoToDashboardClickEvent
Adds "copy" and "paste" PreUserRecoveryCodes<_>ClickEvents
Removes deprecated/unused PreUserOnboardGetStartedClickEvent

### DIFF
--- a/packages/teleport/src/Welcome/NewCredentials/NewCredentials.tsx
+++ b/packages/teleport/src/Welcome/NewCredentials/NewCredentials.tsx
@@ -18,7 +18,7 @@ import React, { useState } from 'react';
 import { Card } from 'design';
 import { PrimaryAuthType } from 'shared/services';
 
-import { StepSlider, NewFlow, StepComponentProps } from 'design/StepSlider';
+import { NewFlow, StepComponentProps, StepSlider } from 'design/StepSlider';
 
 import RecoveryCodes from 'teleport/components/RecoveryCodes';
 import { PrivateKeyLoginDisabledCard } from 'teleport/components/PrivateKeyPolicy';
@@ -76,7 +76,13 @@ export function NewCredentials(props: State & Props) {
   }
 
   if (success) {
-    return <RegisterSuccess redirect={redirect} resetMode={resetMode} />;
+    return (
+      <RegisterSuccess
+        redirect={redirect}
+        resetMode={resetMode}
+        username={resetToken.user}
+      />
+    );
   }
 
   if (recoveryCodes) {
@@ -130,4 +136,5 @@ export function NewCredentials(props: State & Props) {
 
 export type Props = State & {
   resetMode?: boolean;
+  currFlow?: LoginFlow;
 };
--- a/packages/teleport/src/components/RecoveryCodes/RecoveryCodes.tsx
+++ b/packages/teleport/src/components/RecoveryCodes/RecoveryCodes.tsx
@@ -32,6 +32,15 @@ export default function RecoveryCodesDialog({
 }: Props) {
   const codesRef = useRef();
 
+  const captureRecoveryCodeEvent = (event: CaptureEvent) => {
+    if (username) {
+      userEventService.capturePreUserEvent({
+        event: event,
+        username: username,
+      });
+    }
+  };
+
   const onCopyClick = () => {
     copyToClipboard(
       `${recoveryCodes?.codes.join('\n')} \n\nCreated: ${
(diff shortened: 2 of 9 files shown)


packages/teleport/src/Welcome/NewCredentials/NewCredentials.tsx
```diff
@@ -130,4 +136,5 @@ export function NewCredentials(props: State & Props) {
 
 export type Props = State & {
   resetMode?: boolean;
+  currFlow?: LoginFlow;
```
ibeckermayer, line 139: Nit: maybe it would be more clear to name this `loginFlow`?

ibeckermayer (approved):

packages/teleport/src/components/RecoveryCodes/RecoveryCodes.tsx
```diff
@@ -40,16 +49,18 @@ export default function RecoveryCodesDialog({
...
-    if (username) {
-      userEventService.capturePreUserEvent({
-        event: CaptureEvent.PreUserRecoveryCodesContinueClickEvent,
-        username: username,
-      });
-    }
+  const onPrintClick = () => {
+    window.print();
+    captureRecoveryCodeEvent(CaptureEvent.PreUserRecoveryCodesPrintClickEvent);
+  };
```
ryanclark, line 58: What's your thoughts on changing this to an event handler for `beforeprint`? That way we can capture events for people who use a keyboard shortcut to print the page.

packages/teleport/src/components/RecoveryCodes/RecoveryCodes.tsx
```diff
@@ -32,6 +32,15 @@ export default function RecoveryCodesDialog({
 }: Props) {
   const codesRef = useRef();
 
+  const captureRecoveryCodeEvent = (event: CaptureEvent) => {
+    if (username) {
+      userEventService.capturePreUserEvent({
+        event: event,
+        username: username,
+      });
```
kimlisa, line 40: i don't remember if we discussed this before, but components that live in `src/components` are normally shared. I believe this component is used in the following flows:

1) on reset or on invite
2) user is logged in, goes to their account settings, and goes to their recovery code section and clicks on `generate new recovery codes` (so this is not a `preUser` event.)

Not sure if that detail mattered

packages/teleport/src/components/RecoveryCodes/RecoveryCodes.tsx
michellescripts: Resolved via DM; essentially:

> But, considering the question from the perspective of: "how does the user interact with our UI?" Then, capturing an event with a button click is the closest to the truth (including tab/enter methods). The event will open a dialog, and a user may print or click cancel, but they're not "interacting with our UI" at that time. Same if they use a shortcut browser or take a screenshot. That is why the naming is "...recoveryCodesPrint.click", not "...print."
Today, we're trying to see how a user works through the UI to onboard. We may find down the road that a ton of users who were given recovery codes are getting locked out. We can do additional event captures to dig into the issue if it is a problem area. Or add other requirements that the user clicks copy, etc., etc.

ryanclark (approved):

packages/teleport/src/Welcome/NewCredentials/NewCredentials.tsx
michellescripts: I'm not going to rename it here because the[ origin ](https://github.com/gravitational/webapps/blob/b077b510f09b78b8d77fa39a9a43fbc813308488/packages/design/src/StepSlider/StepSlider.tsx#L25)is a few components up. I'm plumbing through one last child component. 

IMO, the pattern: `export type State = ReturnType<typeof useX>;` makes it hard to tell where a rename might impact other components

packages/teleport/src/components/RecoveryCodes/RecoveryCodes.tsx
michellescripts: Yep! Same [story](https://github.com/gravitational/webapps/pull/1344#discussion_r1038570418) as in #1344 

This is fine, we may add more events or logic to this component when we instrument other aspects of UI interaction.

kimlisa (approved):

PR #1474, opened 2023-01-02
Notifications: Add showcase story & fix icon alignment

ravicious:
I noticed that when the notification has a single line with no title, the text and the icon are misaligned.

before | after
--- | ---
before | after

To make sure I can fix this without breaking everything else, I added a story which shows each notification type.

While adding the story, I noticed that the types for the Notification component don't alert me when I forgot to add required props, so I fixed that too.

### DIFF
--- a/packages/shared/components/Notification/Notification.tsx
+++ b/packages/shared/components/Notification/Notification.tsx
@@ -19,8 +19,6 @@ import styled, { css, useTheme } from 'styled-components';
 import { ButtonIcon, Flex, Text } from 'design';
 import { Close } from 'design/Icon';
 
-import type { propTypes } from 'design/system';
-
 import type { NotificationItem, NotificationItemContent } from './types';
 
 interface NotificationProps {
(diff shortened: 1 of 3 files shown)


packages/shared/components/Notification/Notification.tsx
```diff
@@ -19,8 +19,6 @@ import styled, { css, useTheme } from 'styled-components';
 import { ButtonIcon, Flex, Text } from 'design';
 import { Close } from 'design/Icon';
 
-import type { propTypes } from 'design/system';
```
ravicious, line 22: `propTypes` does not include types for props used by our components, but rather [React PropTypes used by styled-system](https://github.com/styled-system/styled-system/blob/v3.1.11/src/index.js#L6-L16). `design/system` itself is not typed as of yet anyway.

Instead, I opted for defining the types for `Notification` in such a way that there's a bunch of props that are required but they also allow passing any extra `styleProps` to `Container`.

packages/shared/components/Notification/Notification.tsx
```diff
@@ -118,7 +124,7 @@ function getRenderedContent(
 
   if (typeof content === 'string') {
     return (
-      <Flex justifyContent="space-between" width="100%">
+      <Flex alignItems="center" justifyContent="space-between" width="100%">
```
ravicious, line 127: Ideally we'd align the icons to the top with `alignItems="baseline"`. However, either I'm using our icons wrong or [the icons in themselves are not aligned to the baseline](https://blog.prototypr.io/align-svg-icons-to-text-and-say-goodbye-to-font-icons-d44b3d7b26b4#6e0c), that is the bottom of the icon gets aligned to the baseline rather than the icon's center falling on the baseline.

This is how replacing both `alignItems="center"` with `"baseline"` looks like:

baseline

I even tried refactoring the component so that it's just three elements within a flex but this didn't help because of the fundamental issue with the icons.

gzdunek: Yeah, it seems like the icons are not properly aligned :/

gzdunek (approved):

zmb3 (approved):

PR #1480, opened 2023-01-03
Add commands for linking tsh on macOS

ravicious:
This will be merged into #1445.

First I added some missing stories in order to be confidently make changes to how suggestions are displayed. Then I added the command bar commands and later added the necessary main process IPC handlers to take care of executing osascript that creates the symlinks.

It's based on how VSCode does this:

* https://github.com/microsoft/vscode/blob/1.74.0/src/vs/workbench/electron-sandbox/actions/installActions.ts
* https://github.com/microsoft/vscode/blob/1.74.0/src/vs/platform/native/electron-main/nativeHostMainService.ts#L261


packages/teleterm/src/mainProcess/mainProcess.ts
```diff
@@ -203,6 +203,52 @@ export default class MainProcess {
...
+        // https://developer.apple.com/library/archive/documentation/AppleScript/Conceptual/AppleScriptLangGuide/reference/ASLR_error_codes.html#//apple_ref/doc/uid/TP40000983-CH220-SW2
+        if (error instanceof Error && error.message.includes('-128')) {
+          return false;
+        }
+        this.logger.error(error);
+        throw error;
+      }
+    });
+
+    ipcMain.handle('main-process-remove-tsh-symlink', async () => {
```
ravicious, line 231: I just copy-pasted the other handler instead of adding some abstractions here, IMHO once we have another use case for osascript we can think about abstracting osascript prompts.

packages/teleterm/src/ui/services/quickInput/parsers.ts
```diff
@@ -136,13 +137,21 @@ export class QuickCommandParser implements QuickInputParser {
   private mapAutocompleteCommandsToSuggestions(
     commands: { name: string; displayName: string; description: string }[]
   ): SuggestionCmd[] {
-    return commands.map(cmd => ({
-      kind: 'suggestion.cmd' as const,
-      token: cmd.displayName,
+    return commands.map(cmd => {
+      const acceptsNoArguments =
+        this.parserRegistry.get(cmd.displayName) instanceof
+        QuickNoArgumentsParser;
```
ravicious, line 143: The way it worked before is that `tsh ssh` and `tsh proxy db` had their own parsers assigned to them:

https://github.com/gravitational/webapps/blob/665f7e7f5cc2bb1cb159b5683231d07c3fad7802/packages/teleterm/src/ui/services/quickInput/quickInputService.ts#L63-L70

However, `tsh install` and `tsh uninstall` don't need custom parsers as they don't accept any extra arguments. So I added `QuickNoArgumentsParser` that I use for those two commands.

If that parser is used for the given command, then QuickInputService is not going to append space to the input.

packages/teleterm/src/mainProcess/fixtures/mocks.ts
```diff
@@ -5,33 +5,17 @@ import { createMockFileStorage } from 'teleterm/services/fileStorage/fixtures/mo
 // Importing electron breaks the fixtures if that's done from within storybook.
 import { createConfigService } from 'teleterm/services/config/configService';
 
-const platform = 'darwin';
```
ravicious, line 8: I had to remove the hardcoded platform in order to be able to test config service for different platforms.

packages/teleterm/src/mainProcess/mainProcess.ts
avatus: I love this thinking. Always reminds me of [this](https://kentcdodds.com/blog/aha-programming).

packages/teleterm/src/mainProcess/mainProcess.ts
```diff
@@ -203,6 +203,52 @@ export default class MainProcess {
...
+      const target = '/usr/local/bin/tsh';
+      const prompt =
+        'Teleport Connect wants to remove a symlink for tsh from /usr/local/bin.';
```
avatus, line 234: I wonder if this language could be changed at all. I know this is exactly what the script is doing, and I know our audience is technical enough to understand thing, but it just seems so... destructive? Idk what the word is. I'm ok with not changing it, but I wonder if there is simpler terms to describe what is happening. Maybe I'm overreacting

avatus (approved):

packages/teleterm/src/mainProcess/mainProcess.ts
```diff
@@ -203,6 +203,52 @@ export default class MainProcess {
       this.windowsManager.forceFocusWindow();
     });
 
+    // Used in the `tsh install` command on macOS to make the bundled tsh available in PATH.
+    // Returns true if tsh got successfully installed, false if the user closed the osascript
+    // prompt. Throws an error when osascript fails.
+    ipcMain.handle('main-process-symlink-tsh', async () => {
```
zmb3, line 209: Since this is specific to macOS, should we indicate that in the name of the event?
```suggestion
    ipcMain.handle('main-process-symlink-tsh-macos', async () => {
```

packages/teleterm/src/ui/commandLauncher.test.ts
```diff
@@ -0,0 +1,36 @@
+import { MockAppContext } from 'teleterm/ui/fixtures/mocks';
```
zmb3, line 1: Missing license here?

zmb3 (approved):

packages/teleterm/src/ui/commandLauncher.ts
```diff
@@ -92,6 +92,50 @@ const commands = {
...
+          if (isSymlinked) {
+            ctx.notificationsService.notifyInfo(
+              'tsh successfully installed in PATH'
+            );
+          }
+        },
+        error => {
+          ctx.notificationsService.notifyError({
+            title: 'Could not install tsh in PATH',
+            description: `Ran into an error: ${error}`,
```
gzdunek, line 110: I didn't check it, but it seems to me that the `error` here is not a plain string, but an object with a `message` field (as we throw `error` from the IPC handler), so we will get a runtime error.

packages/teleterm/src/mainProcess/mainProcess.ts
```diff
@@ -203,6 +203,52 @@ export default class MainProcess {
...
+        this.logger.info(`Removed the symlink under ${target}`);
+        return true;
+      } catch (error) {
+        // Ignore the error if the user canceled the prompt.
+        // https://developer.apple.com/library/archive/documentation/AppleScript/Conceptual/AppleScriptLangGuide/reference/ASLR_error_codes.html#//apple_ref/doc/uid/TP40000983-CH220-SW2
+        if (error instanceof Error && error.message.includes('-128')) {
+          return false;
+        }
+        this.logger.error(error);
+        throw error;
```
gzdunek, line 248: I have mixed feelings about this, it would be nice to return the errors in the same way.
Maybe we could create a new custom error, like `PromptCancelledByUserError` and throw (or even return) it instead of `false`? Then on the renderer side we would have to create our own `instanceOf` (as the the classes are serialized) that would compare errors by `name`.

IMO this would make the code more elegant, but we can live with what we have, it is quite simple and documented.

packages/teleterm/src/mainProcess/fixtures/mocks.ts
```diff
@@ -5,55 +5,66 @@ import { createMockFileStorage } from 'teleterm/services/fileStorage/fixtures/mo
...
+    return true;
+  }
+  async removeTshSymlink() {
+    return true;
+  }
 }
+
+const defaultRuntimeSettings = {
+  platform: 'darwin' as const,
+  dev: true,
```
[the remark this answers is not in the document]
ravicious: > I know this is a mock, but what are the consequences of having our mock be in `dev` mode instead of not? Could this lead to any false positives/negatives?

`dev` is mostly used during the build & startup of the app, it shouldn't have much effect overall on tests.

In general I'm of the opinion that we shouldn't use `dev` for business logic, [I was talking with Grzegorz about this recently](https://github.com/gravitational/webapps/pull/1451#discussion_r1055385302). And for any other scenario, our setup code in tests should probably be closer to the dev build than the prod build (for example in terms of what's getting logged and what's not, etc.).

packages/teleterm/src/mainProcess/mainProcess.ts
ravicious: > I'm ok with not changing it, but I wonder if there is simpler terms to describe what is happening.

We could say "Connect wants to uninstall tsh from PATH" but personally I'd prefer the app to tell me exactly what it wants to do when it asks me for a password.

Another option would be "Connect wants to modify the contents of /usr/local/bin" which doesn't use the notion of removing/uninstalling but again fails at specifying what exactly is going to happen and potentially sounds even more dangerous.

I'd probably keep this, the way I look at it is we ask the user for permission to do something so we better explain that as succinctly as possible.

packages/teleterm/src/mainProcess/mainProcess.ts
ravicious: > Since this is specific to macOS, should we indicate that in the name of the event?

I was thinking about it but in the end I didn't do it for some reason, I'll add that.

packages/teleterm/src/ui/commandLauncher.test.ts
ravicious: I haven't been adding license to any file since I've joined. 🙊

If that's necessary then once the merge is done we should just modify the ignored paths of `make fix-license`.

packages/teleterm/src/ui/QuickInput/QuickInput.story.tsx
```diff
@@ -28,6 +28,35 @@ export default {
 };
 
 export const Story = () => {
+  return (
+    <>
+      {/* Extra bottom margin to accommodate the suggestions displayed under the input. */}
+      <QuickInputDemo description="Pristine state" mb={350} />
```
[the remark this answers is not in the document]
ravicious: > The "scroll into viewport" mechanism in `QuickInput` causes the whole story to jump when you want to use an input at the bottom of the viewport.

Right, it's annoying but the reason I created the story in the first place is that I wanted to have an easy way to inspect all available suggestions.

But I suppose a better way would be to create a separate story which uses `QuickInputList` alone rather than setting up the whole `QuickInput` component. I'll make that change soon.

packages/teleterm/src/ui/QuickInput/useQuickInput.ts
```diff
@@ -122,6 +122,16 @@ export default function useQuickInput() {
...
       }
+      case 'command.tsh-install': {
+        break;
+      }
+      case 'command.tsh-uninstall': {
+        break;
+      }
+      default: {
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        const exhaustiveCheck: never = command;
```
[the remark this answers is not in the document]
ravicious: > Don't you want to use [`assertUnreachable`](https://github.com/gravitational/webapps/blob/master/packages/teleterm/src/ui/utils/assertUnreachable.ts) here?

I was one the fence about it, especially when I remembered the problems we had with throwing errors when the doc type was not included in the switch.

But I think the difference between switching on docs vs commands here is that docs are loaded from a JSON file whereas commands are hardcoded.

We should be able to throw when switching on docs too but the actual issue with docs is that we don't parse the payload coming from the JSON file in any way. This allows objects of invalid type to slip through.

But that's not a problem here so yeah I think we can just always prefer `assertUnreachable`.

packages/teleterm/src/ui/commandLauncher.ts
ravicious: > I didn't check it, but it seems to me that the `error` here is not a plain string, but an object with a `message` field (as we throw `error` from the IPC handler), so we will get a runtime error.

Errors that pass through IPC or the context bridge are still instances of `Error` with `message`, they're just stripped of any custom properties that might be attached to them. See [the _A word on error handling_ section](https://www.electronjs.org/docs/latest/tutorial/ipc#1-listen-for-events-with-ipcmainhandle).

The place where we discussed returning plain objects instead of errors was tshd client but we never got around to that (gravitational/webapps.e#309) 😏.

And I don't think we'd get a runtime error in any case:

```typescript
const err = {message: "test"}; `err: ${err}`
// > "err: [object Object]"
```

packages/teleterm/src/mainProcess/mainProcess.ts
ravicious: > I have mixed feelings about this, it would be nice to return the errors in the same way.

The way I think about this is that closing the prompt is not really an error/exception so I wanted to reserve rejecting the promise for actual errors that might happen during the execution of the script.

As seen in your suggestion, using a custom error would be much more complicated. But even more important is that it wouldn't work because errors are not only stripped of any custom properties, they lose their name as well and become simple `Error` instances. So we couldn't inspect `name` AFAIR.

packages/teleterm/src/ui/commandLauncher.test.ts
```diff
@@ -0,0 +1,36 @@
...
+it('does not return tsh install & uninstall autocomplete command on Windows', () => {
+  const appContext = new MockAppContext({ platform: 'win32' });
+  const commandLauncher = new CommandLauncher(appContext);
+  const autocompleteCommandNames = commandLauncher
+    .getAutocompleteCommands()
+    .map(c => c.name);
+
+  expect(autocompleteCommandNames).not.toContain('autocomplete.tsh-install');
+  expect(autocompleteCommandNames).not.toContain('autocomplete.tsh-uninstall');
+});
```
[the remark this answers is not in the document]
ravicious: > We can avoid repeating the same test code with [`test.each`](https://dev.to/bgord/simplify-repetitive-jest-test-cases-with-test-each-310m).

I think I'm actually going to keep it, it's just three tests and idk how to neatly refactor this with `test.each` without introducing additional logic since one test requires commands to be in the array and others do not.

packages/teleterm/src/ui/commandLauncher.ts
gzdunek: >Errors that pass through IPC or the context bridge are still instances of Error with message, they're just stripped of any custom properties

I see, I was sure that name is passed too, as it is standard `Error` property. 

>And I don't think we'd get a runtime error in any case

Ah right, `Objects are not valid as a react child` happens only when rendering objects directly in JSX.

packages/teleterm/src/mainProcess/mainProcess.ts
gzdunek: So the only way is to return plain objects, but if we don't think of closing the prompt as an error then it's fine to stay with the current solution.

packages/teleterm/src/ui/commandLauncher.test.ts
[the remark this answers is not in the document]
ravicious: Then I'd say it's not worth it for just two short tests. 😏

packages/teleterm/src/mainProcess/mainProcessClient.ts
```diff
@@ -32,10 +32,10 @@ export default function createMainProcessClient(): MainProcessClient {
       return ipcRenderer.invoke('main-process-force-focus-window');
     },
     symlinkTsh() {
```
gzdunek, line 34: I somehow forgot to send a comment earlier, but I'd also add `macOs` part to `symlinkTsh`/`removeTshSymlink`.

gzdunek (approved):

PR #1481, opened 2023-01-04
[Discover] Add database service checker and various db tweaks

kimlisa:

Suggest reviewing by commit, the changes aren't that complicated except maybe for the service checker

#### Description
- Adds a database service checker to check if we can skip the next step of adding a db service or not. If a matching db service exists, then we query for our new db until it shows up on the list (as discussed this should be instant) and skip adding a service 
- Adds tooltip for additional info to help out users
- Adds supporting different informational texts depending on databases (other database types is defined because I was working on it before we decided to just release postgres, figured i just leave it here)

#### Screenshot
tooltip on hover example:
image


packages/shared/components/ToolTip/ToolTip.tsx
```diff
@@ -0,0 +1,82 @@
...
+ * limitations under the License.
+ */
+
+import React, { useState } from 'react';
+import styled from 'styled-components';
+
+import { Text, Popover } from 'design';
+import * as Icons from 'design/Icon';
+
+export function ToolTipInfo({ content }: { content: React.ReactNode }) {
```
ryanclark, line 23: What's your thoughts on using `children` instead so we can have

```typescript jsx
<ToolTipInfo>
content here
</ToolTipInfo>
```

ryanclark (approved):

packages/shared/components/ToolTip/ToolTip.tsx
kimlisa: image

PR #1484, opened 2023-01-04
Make trusted cluster screen hidden based on user roles

mcbattirola:
This PR makes the trusted cluster screen show/hide based on the user ACL.
It also adds methods to do the same thing for enterprise features that were missing it.

Webapps.e PR: https://github.com/gravitational/webapps.e/pull/461


packages/teleport/src/teleportContext.tsx
```diff
@@ -100,6 +100,8 @@ class TeleportContext implements types.Context {
       desktops: userContext.getDesktopAccess().list,
       nodes: userContext.getNodeAccess().list,
       activeSessions: userContext.getActiveSessionsAccess().list,
+      accessRequests: userContext.getAccessRequests().list,
+      newAccessRequest: userContext.getAccessRequests().create,
```
mcbattirola, line 104: It is strange to make this one based on `create`, but it **is** a screen to create access requests. Thoughts?

ibeckermayer: makes sense

ibeckermayer (approved):

michellescripts (approved):

packages/teleport/src/stores/storeUserContext.ts
```diff
@@ -122,6 +122,10 @@ export default class StoreUserContext extends Store<UserContext> {
     return this.state.accessRequestId;
   }
 
+  getAccessRequests() {
```
kimlisa, line 125: to follow naming convention `getAccessRequestAccess` :laughing: (serious i'm okay with it) or `getAccessRequestPerms` (otherwise the current name implies we are getting a list of requests)

kimlisa (approved):

PR #1486, opened 2023-01-04
Update sessionPath value to new endpoint

avatus:
When https://github.com/gravitational/teleport/pull/19593 went in, it removed the deprecated(?) route to `webapi/sessions` in favor of `webapi/sessions/web` but the frontend was never updated since v5 to change this route. It looks like we were always calling to the same old route instead of the updated one. When the old route was removed, you can no longer login on the web ui. 

This change allows the login with the web ui again. I've checked around to find other instances and so far haven't found any but I may have missed something. Some sets of eyes and any historical data around this route would be great!

Teleport buddy: https://github.com/gravitational/teleport/pull/19892

### DIFF
--- a/packages/teleport/src/config.ts
+++ b/packages/teleport/src/config.ts
@@ -121,9 +121,9 @@ const cfg = {
     checkAccessToRegisteredResource: `/v1/webapi/sites/:clusterId/resources/check`,
 
     scp: '/v1/webapi/sites/:clusterId/nodes/:serverId/:login/scp?location=:location&filename=:filename',
-    renewTokenPath: '/v1/webapi/sessions/web/renew',
+    webRenewTokenPath: '/v1/webapi/sessions/web/renew',
     resetPasswordTokenPath: '/v1/webapi/users/password/token',
-    sessionPath: '/v1/webapi/sessions/web',
+    webSessionPath: '/v1/webapi/sessions/web',
     userContextPath: '/v1/webapi/sites/:clusterId/context',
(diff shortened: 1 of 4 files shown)


packages/teleport/src/config.ts
```diff
@@ -121,9 +121,9 @@ const cfg = {
     checkAccessToRegisteredResource: `/v1/webapi/sites/:clusterId/resources/check`,
 
     scp: '/v1/webapi/sites/:clusterId/nodes/:serverId/:login/scp?location=:location&filename=:filename',
-    renewTokenPath: '/v1/webapi/sessions/renew',
+    renewTokenPath: '/v1/webapi/sessions/web/renew',
```
kimlisa, line 124:
```suggestion
    webRenewTokenPath: '/v1/webapi/sessions/web/renew',
```

packages/teleport/src/config.ts
```diff
@@ -121,9 +121,9 @@ const cfg = {
...
     resetPasswordTokenPath: '/v1/webapi/users/password/token',
-    sessionPath: '/v1/webapi/sessions',
+    sessionPath: '/v1/webapi/sessions/web',
```
kimlisa, line 126:
```suggestion
    webSessionPath: '/v1/webapi/sessions/web',
```

kimlisa (approved):
similar name change as backend

hatched (approved):

ibeckermayer (approved):

PR #1488, opened 2023-01-05
[v11] Make bundled tsh available outside of Connect

ravicious:
Backport #1445.


ravicious:
@gzdunek authClusterId and installationId won't be backported to v11, right?

gzdunek:
Hmm the plan is to release reporting with v12, but I think there is a chance that it will go with 11.3 if it is ready.
I can backport these two PRs.

gzdunek (approved):

PR #1491, opened 2023-01-05
Connect: Ask for user job role

gzdunek:
The last frontend part of https://github.com/gravitational/teleport/issues/19000
 
This PR adds a new dialog that asks the user about their job role. Information about showing the dialog is stored in `app_state.json`. When the user clicks "Skip" the dialog is closed and it won't show again (until the variable in `app_state.json` is changed).
image

I also added a new component `RadioGroup` and reused it in `NewMfaDevice.tsx` (it looks almost identical, except the `accent-color`).

| Before      | After |
| ----------- | ----------- |
| image|image|

Best to review commit-by-commit.


packages/teleterm/src/ui/services/statePersistence/statePersistenceService.ts
```diff
@@ -25,6 +25,10 @@ interface ShareFeedbackState {
   hasBeenOpened: boolean;
 }
 
+interface UsageReportingState {
+  askedAboutJobRole: boolean;
```
avatus, line 29: `askedForUserJobRole` maybe?

packages/teleport/src/Welcome/NewCredentials/NewMfaDevice.tsx
```diff
@@ -15,10 +15,10 @@
  */
```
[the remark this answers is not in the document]
gzdunek: Yeah, I didn't see acronyms written in all caps in webapps codebase, so I will leave this as it is.

packages/teleterm/src/ui/services/statePersistence/statePersistenceService.ts
gzdunek: Done

ravicious:
I'm getting an error in dev mode.

```
TypeError: Cannot convert undefined or null to object
    at Function.assign (<anonymous>)
    at eval (documentsService.ts?c4f7:243:14)
    at eval (workspacesService.ts?b2d7:120:15)
    at produce (immer.esm.js?f445:1:16170)
    at WorkspacesService.setState (immutableStore.ts?3631:17:25)
    at WorkspacesService.setState (workspacesService.ts?b2d7:168:11)
    at DocumentsService.eval [as setState] (workspacesService.ts?b2d7:119:18)
    at DocumentsService.update (documentsService.ts?c4f7:241:10)
    at eval (DocumentCluster.tsx?47ee:45:8)
    at commitHookEffectListMount (react-dom.development.js?a814:19731:1)
eval @ boot.tsx?cf68:19
```

Steps to reproduce:

1. `rm -rf ~/Library/Application\ Support/Electron`
2. Switch to `master`, start the dev version, log in to a cluster.
5. Switch to this branch, start the dev version again.
7. Allow telemetry, pick a role (I think I was picking the next to last one).
8. Submit the modal.

ravicious (commented):
I took a cursory look, I'll make a more detailed review tomorrow.

packages/teleterm/src/ui/ModalsHost/modals/UserJobRole/UserJobRole.tsx
```diff
@@ -0,0 +1,147 @@
...
+            placeholder="Other role..."
+            mt={1}
+          />
+        </DialogContent>
+        <DialogFooter>
+          <ButtonPrimary
+            autoFocus
+            mr={3}
+            type="submit"
+            disabled={!jobRoleToSubmit}
```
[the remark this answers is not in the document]
gzdunek: Ah I copied it from another dialog and forgot to remove.
I agree, it's better to set in on radios.

packages/design/src/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,111 @@
...
+}
+
+export function RadioGroup({
+  options,
+  value,
+  onChange,
+  ...styles
+}: RadioGroupProps) {
+  // A radio group is defined by giving each of radio buttons in the group the same name.
+  const [name] = useState(crypto.randomUUID());
```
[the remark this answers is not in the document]
gzdunek: Agree, it was quite weird to generate the name this way :)

Added a prop.

gzdunek:
>I'm getting an error in dev mode.

I have it too.  It took me some time to track down why this is happening and it turned out that the cause is... a bug in  `workspaceService.restorePersistedState()`. 
We create workspaces based on clusters list that we get from `clustersService` without filtering out leaf clusters. Previously it was working because `restorePersistedState()` was called immediately after `await this.clustersService.syncRootClusters()`, so leafs didn't have time to fetch. After the dialog was added, while thinking what option to select in the job role dialog, leaf clusters were fetched in the background, and then we created workspaces for leaf clusters (that caused errors later).
The solution is to just filter out leaf clusters before restoring the workspaces.
Sorry for missing this.

packages/teleterm/src/ui/services/usage/askAboutUserJobRoleIfNeeded.ts
```diff
@@ -0,0 +1,60 @@
...
+function showUserJobRoleDialog(
+  modalsService: ModalsService,
+  usageService: UsageService
+): Promise<void> {
+  return new Promise(resolve => {
+    modalsService.openRegularDialog({
+      kind: 'user-job-role',
+      onSend(jobRole) {
+        usageService.captureUserJobRoleUpdate(jobRole);
+        resolve();
```
ravicious, line 53: Another small nit: It's kind of annoying that we have to write those custom promise wrappers for each modal. In an ideal world, we'd have an uniform way of doing that (though idk how well it'd handle edge cases such as `UsageData` with three different callbacks).

My point being, when we write a custom promise wrapper like this, it might be worthwhile to keep the logic in it to minimum, resolve with a value (`jobRole` in this case) and then implement the rest of the logic outside of the wrapper. Meaning we'd await `jobRole` and then call `usageService.captureUserJobRoleUpdate`.

This way the whole logic could be laid out in `askAboutUserJobRoleIfNeeded` and you could just read that function top to bottom to understand what's going on, without having to jump into `showUserJobRoleDialog` to see that there's a side effect hapenning in `onSend`.

ravicious (commented):
Looks good overall. I'm being extra pedantic about `RadioGroup` in the comments because those shared components are always harder to change once they're merged in.

packages/teleterm/src/ui/ModalsHost/modals/UserJobRole/UserJobRole.tsx
```diff
@@ -0,0 +1,144 @@
...
+            name="jobRole"
+            options={[...JOB_OPTIONS, OTHER_JOB_ROLE]}
+            onChange={handleRadioGroupChange}
+          />
+          <DarkInput
+            ref={inputRef}
+            readOnly={jobRole !== OTHER_JOB_ROLE}
+            value={otherJobRole}
+            onChange={e => setOtherJobRole(e.target.value)}
+            placeholder="Other role..."
```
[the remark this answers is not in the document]
gzdunek: _Today I learned..._

packages/teleterm/src/ui/ModalsHost/modals/UserJobRole/UserJobRole.tsx
```diff
@@ -0,0 +1,144 @@
...
+        <DialogContent mb={3}>
+          <RadioGroup
+            autoFocus={true}
+            name="jobRole"
+            options={[...JOB_OPTIONS, OTHER_JOB_ROLE]}
+            onChange={handleRadioGroupChange}
+          />
+          <DarkInput
+            ref={inputRef}
+            readOnly={jobRole !== OTHER_JOB_ROLE}
```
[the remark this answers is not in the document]
gzdunek: Much better!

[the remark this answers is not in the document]
gzdunek: >Was it something related to readonly having appropriate styles and disabled not having them?

Yes, in `Input.jsx` we have:
```css
  :read-only {
    cursor: not-allowed;
  }
```

packages/design/src/RadioGroup/RadioGroup.story.tsx
```diff
@@ -0,0 +1,65 @@
...
+export default {
+  title: 'Design/RadioGroup',
+};
+
+export const Default = () => {
+  return (
+    <Flex direction="row">
+      <Box mr={6}>
+        <h4>String options</h4>
+        <RadioGroup
```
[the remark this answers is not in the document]
gzdunek: Others provide `value`, so they are controlled (but there are no `onChange` handlers and the value can't change).

packages/teleterm/src/ui/services/usage/askAboutUserJobRoleIfNeeded.ts
gzdunek: Yeah, it is quite inconvenient to have to write that _promisify_ function, but nothing better comes to my mind :(
To have a generic wrapper we should first have a standard set of callbacks passed to modals.

I like the suggestion about moving the logic to the parent function, done.

packages/teleterm/src/ui/ModalsHost/modals/UserJobRole/UserJobRole.tsx
```diff
@@ -0,0 +1,144 @@
...
+            name="jobRole"
+            options={[...JOB_OPTIONS, OTHER_JOB_ROLE]}
+            value={jobRole}
+            onChange={handleRadioGroupChange}
+          />
+          <DarkInput
+            ref={inputRef}
+            value={otherJobRole}
+            onChange={e => setOtherJobRole(e.target.value)}
+            onFocus={() => setJobRole(OTHER_JOB_ROLE)}
```
ravicious, line 111: I just realized that my suggestion was not so great because now if you choose one of the predefined options, you cannot really tab through to the submit button without selecting "Other".

How about this:

```suggestion
            onChange={e => {
              setJobRole(OTHER_JOB_ROLE);
              setOtherJobRole(e.target.value);
            }}
```

ravicious (approved):

packages/teleport/src/Welcome/NewCredentials/NewMfaDevice.tsx
```diff
@@ -132,21 +134,14 @@ export function NewMfaDevice(props: Props) {
...
-                  <input
-                    type="radio"
-                    checked={mfaType.value === opt.value}
-                    onChange={() => onSetMfaOption(index, validator)}
-                  />
-                  <label>{opt.label}</label>
-                </Radio>
-              );
-            })}
+            <RadioGroup
```
ravicious, line 137: I was just comparing the "Mfa Device Optional" story between this branch and master to see if anything has changed. And it looks like this PR improved the form a bit.

All radio buttons are within the same group (because we set the same `name` attribute) which enables navigation through the radio group with arrow keys – that is not present currently on master. Though we focus on another field whenever a radio option is selected so there's not a lot of benefit to it but still, it's a small improvement.

packages/teleterm/src/ui/ModalsHost/modals/UserJobRole/UserJobRole.tsx
ravicious: > WDYT about changing the option also when the user clicks on the input?

Great idea, this seems to cover pretty much all the ways in which you could use a keyboard or a mouse to interact with the form.

gzdunek: Done :)

PR #1495, opened 2023-01-06
[Discover] Temp remove db service checker

kimlisa:

#### Description
- Fixes a bug where when we query for db service (no longer called atm), the labels return can be a `map[string]string OR []string`, so we just check if the label value is an array.
- Temporarily remove the service checker until bug is fixed: https://github.com/gravitational/teleport/issues/19919
- Because of no service checker which guaranteed that there was a active db server that can pick up the newly created database, we now leave it up to the user to add or skip installing the database service
  - If user skipped the installing step, and there was no active db service, the user will have to restart the flow at this time

### DIFF
--- a/packages/teleport/src/Discover/Shared/ActionButtons.tsx
+++ b/packages/teleport/src/Discover/Shared/ActionButtons.tsx
@@ -25,14 +25,17 @@ import cfg from 'teleport/config';
 
 export const ActionButtons = ({
   onProceed = null,
+  onSkip = null,
   proceedHref = '',
   disableProceed = false,
   lastStep = false,
 }: {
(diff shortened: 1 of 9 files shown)


packages/teleport/src/Discover/Shared/ActionButtons.tsx
```diff
@@ -59,6 +62,11 @@ export const ActionButtons = ({
           {lastStep ? 'Finish' : 'Next'}
         </ButtonPrimary>
       )}
+      {onSkip && (
```
ryanclark, line 65:
```suggestion
      {allowSkip && (
```

PR #1497, opened 2023-01-08
Make support links open a new tab

stevenGravy:
Opens support links in new tab instead of current tab.  This is typical for help and support links.


stevenGravy:
Move to teleport repo

PR #1501, opened 2023-01-10
Refactor clusters service

rudream:
## Purpose

This PR refactors our clusters service and moves away from using `lodash`. 

This fixes a bug reported on `master` that would cause a `t is not a function` error whenever trying to list clusters.


hatched (approved):

kimlisa:
if this was caused by lodash version, we have other places that need fixing as well then, which might take a while, and since v12 is about to be released soon, we should just roll back the lodash version? 

think it was introduced here: https://github.com/gravitational/webapps/commit/9f34ffb8da6844b283da01908d68975c272dc622#diff-3752b1e4e3e2032593e21e9268ec0379680a59e4e7b8517d0d5492a71530d3e8R16, i would test going back a commit from this one to make sure it was this commit that introduced the bug

kimlisa (approved):
@rudream can you create an issue to refactor out lodash (in the teleport repo)

from slack discussion: we will find other use of lodash breaking by going through the v12 test plan

zmb3 (approved):

PR #1502, opened 2023-01-10
[v11] Add lazy loading for session playback

ryanclark:
This would fix https://github.com/gravitational/teleport/issues/20044

The immediate rendering of the `Authenticated` component in OSS means that we don't have enough time to receive a token from other tabs on the `BroadcastChannel`, so when opening a new tab (for session playback) the tab becomes unauthenticated during the initial render.

By adding lazy loading, we can delay the rendering of `Authenticated` enough (by falling back to the `Suspense` component) for the token to be received and the tab to be authenticated during the initial render.

### DIFF
--- a/packages/teleport/src/Teleport.tsx
+++ b/packages/teleport/src/Teleport.tsx
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import React from 'react';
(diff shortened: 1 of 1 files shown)


kimlisa (approved):
LGTM!

rudream (approved):

zmb3 (approved):

PR #1505, opened 2023-01-11
Revert "Use sessionStorage for Authentication Bearer Token (#1275)"

rudream:
## Purpose

Some inconsistent bugs have been popping up after starting to use the BroadcastChannel API for managing sessions, this PR reverts those changes to using `localStorage` so we can have stable releases until the issues are addressed.

`e` counterpart: https://github.com/gravitational/webapps.e/pull/467


kimlisa (approved):
out of curiosity, what other problems did we run into?

rudream:
> out of curiosity, what other problems did we run into?

There was someone else reporting an issue where login would sometimes fail with the error `bad bearer token`. It's not consistently reproducible, it likely has to do with some type of delay with the BroadcastChannel API (similar to what was causing the issue with desktop playback).

PR #1509, opened 2023-01-12
Fix clipboard permissions apparent inconsistency

ibeckermayer:
When playing around with desktops, I noticed that even if I denied clipboard permissions at the browser level, the remote clipboard would still get synced to my local clipboard (but not the other way around). This fixes that, see the comment on `shouldTryClipboardRW`.

### DIFF
--- a/packages/teleport/src/DesktopSession/useTdpClientCanvas.tsx
+++ b/packages/teleport/src/DesktopSession/useTdpClientCanvas.tsx
@@ -293,7 +293,8 @@ type Props = {
   setWarnings: Dispatch<SetStateAction<NotificationItem[]>>;
 };
 
-// Returns true if neither 'clipboard-read' nor `clipboard-write' are 'denied'.
+// Returns false if either 'clipboard-read' or `clipboard-write' are 'denied',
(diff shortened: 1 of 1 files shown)


packages/teleport/src/DesktopSession/useTdpClientCanvas.tsx
```diff
@@ -224,9 +229,12 @@ export default function useTdpClientCanvas(props: Props) {
...
   const onContextMenu = () => false;
 
-  const sendLocalClipboardToRemote = (cli: TdpClient) => {
-    // We must check that the DOM is focused or navigator.clipboard.readText throws an error.
-    if (clipboardSharingEnabled && document.hasFocus()) {
+  const sendLocalClipboardToRemote = async (cli: TdpClient) => {
+    if (
+      clipboardSharingEnabled &&
+      document.hasFocus() &&
+      (await shouldTryClipboardRW())
```
jakule, line 236: I think you can move the whole condition to a function and reuse it in both places.

jakule (approved):

hatched (approved):

PR #1524, opened 2023-01-16
Wait for child processes to exit before closing the app

gzdunek:
Requires tshd counterpart to be merged first.

Currently, the Electron app doesn't wait for child processes to exit before the main processes is killed. Because of this, the processes can't run any async cleanup functions (on Windows [even sync functions don't run](https://nodejs.org/api/child_process.html#subprocesskillsignal)). 
This fix is mainly needed to send usage events when tshd exits. Sending the events can take up to a few seconds.

On Windows, where POSIX signals do not exist, the only way to gracefully kill a process is to send Ctrl-Break to its console. This task is done by `tsh daemon stop` program.

For now, the graceful kill is applied only to tshd. In the shared process we don't have any async logic that should be run on exit.

### DIFF
--- a/packages/teleterm/src/main.ts
+++ b/packages/teleterm/src/main.ts
@@ -24,6 +24,7 @@ if (app.requestSingleInstanceLock()) {
 }
 
 function initializeApp(): void {
+  let devRelaunchScheduled = false;
   const settings = getRuntimeSettings();
   const logger = initMainLogger(settings);
   const appStateFileStorage = createFileStorage({
@@ -72,10 +73,25 @@ function initializeApp(): void {
     }
   );
 
-  app.on('will-quit', () => {
+  app.on('will-quit', async event => {
+    event.preventDefault();
+
     appStateFileStorage.putAllSync();
     globalShortcut.unregisterAll();
-    mainProcess.dispose();
+    await mainProcess.dispose();
+    app.exit();
+  });
+
+  app.on('quit', () => {
+    if (devRelaunchScheduled) {
+      const [bin, ...args] = process.argv;
+      const child = spawn(bin, args, {
+        env: process.env,
+        detached: true,
+        stdio: 'inherit',
+      });
+      child.unref();
+    }
   });
 
   app.on('second-instance', () => {
@@ -86,14 +102,7 @@ function initializeApp(): void {
     if (mainProcess.settings.dev) {
       // allow restarts on F6
       globalShortcut.register('F6', () => {
-        mainProcess.dispose();
-        const [bin, ...args] = process.argv;
-        const child = spawn(bin, args, {
-          env: process.env,
-          detached: true,
-          stdio: 'inherit',
-        });
-        child.unref();
+        devRelaunchScheduled = true;
         app.quit();
       });
     }
--- a/packages/teleterm/src/main.ts
+++ b/packages/teleterm/src/main.ts
@@ -81,7 +81,7 @@ function initializeApp(): void {
     try {
       await mainProcess.dispose();
     } catch (e) {
-      logger.error('Failed to dispose main process', e);
+      logger.error('Failed to gracefully dispose of main process', e);
     } finally {
       app.exit();
     }
--- a/packages/teleterm/src/main.ts
+++ b/packages/teleterm/src/main.ts
@@ -78,8 +78,13 @@ function initializeApp(): void {
 
     appStateFileStorage.putAllSync();
     globalShortcut.unregisterAll();
-    await mainProcess.dispose();
-    app.exit();
+    try {
+      await mainProcess.dispose();
+    } catch (e) {
+      logger.error('Failed to dispose main process', e);
+    } finally {
+      app.exit();
+    }
   });
 
   app.on('quit', () => {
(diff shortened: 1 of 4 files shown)


packages/teleterm/src/mainProcess/mainProcess.ts
```diff
@@ -308,10 +318,41 @@ export default class MainProcess {
...
+      {
+        windowsHide: true,
+        timeout: 2_000,
+      }
+    );
+    daemonStop.on('error', error => {
+      logger.error('daemon stop process failed to start', error);
+    });
+    daemonStop.stderr.setEncoding('utf-8');
+    daemonStop.stderr.on('data', logger.error);
```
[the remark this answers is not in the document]
ravicious: Handle what easier? 🙃

packages/teleterm/src/main.ts
```diff
@@ -72,10 +73,25 @@ function initializeApp(): void {
...
+    event.preventDefault();
+
     appStateFileStorage.putAllSync();
     globalShortcut.unregisterAll();
-    mainProcess.dispose();
+    await mainProcess.dispose();
+    app.exit();
+  });
+
+  app.on('quit', () => {
```
ravicious, line 85: Why do we need to handle it within a separate callback? Couldn't we still have that in the handler for F6 where we'd first await `mainProcess.dispose()` and then perform the rest of the logic?

gzdunek: I don't think it can work, because it would result in waiting for `mainProcese.dispose()` two times:
1. F6 is pressed - we wait for the processes to close
2. They are closed, we spawn a new instance and call `app.quit()`
3. `app.on('will-quit')` is called
4. We wait for `mainProcess.dispose()` the second time (it results in a timeout)
5. _Attempted to open a second instance of the app, exiting._

TBH, it seems to me that my changes have cleaned things up a bit:
1. We don't call `mainProcess.dispose` twice
2. Things are called in the correct order - the new process is spawned after the functions from `will-quit` are called, not before.

packages/teleterm/src/mainProcess/mainProcess.ts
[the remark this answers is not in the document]
gzdunek: Ah, handling errors of course :)
I had to add callbacks for `on('error') and `stderr.on('data')`, but now I realized that I would have to read errors from `error` and `stderr` in `spawnSync` too. I wish I could set the encoding when I spawn the process 😏

packages/teleterm/src/main.ts
ravicious: > TBH, it seems to me that my changes have cleaned things up a bit…

Right, I'd say your changes made things more correct but they also increased complexity a little bit as we now have an extra event handler to deal with and an extra variable that's used in different handlers. 😏

I was thinking that in dev mode, we don't really care about gracefully stopping anything so we might as well kill both child processes.

OTOH it's nice that we have a single handler for cleaning up stuff that's called in both cases so this change sounds fine to me.

packages/teleterm/src/main.ts
```diff
@@ -72,10 +73,30 @@ function initializeApp(): void {
...
+  app.on('will-quit', async event => {
+    event.preventDefault();
+
     appStateFileStorage.putAllSync();
     globalShortcut.unregisterAll();
-    mainProcess.dispose();
+    try {
+      await mainProcess.dispose();
+    } catch (e) {
+      logger.error('Failed to dispose main process', e);
```
ravicious, line 84:
```suggestion
      logger.error('Failed to gracefully dispose of main process', e);
```

ravicious (commented):
Looks good but I want to give it a try once the backend part is finished.

avatus (approved):

ravicious (approved):

gzdunek:
I just noticed that this change makes Connect sit in macOS dock until the main process closes (so it can be there for up to 5 seconds after closing the window). On the other platforms the icon is removed just after closing the window.

ravicious:
I suppose we could work around this by having the event handlers on BrowserWindow instead of the Electron app. But it feels safer to have them on the app instead.

gzdunek:
I think I will decrease the batch size (currently it is 25 events). On my machine it takes around 210 ms to send one request, so the entire batch is around 5s (please note it is the worst case scenario, a batch is also sent every hour, so we don't always have to send 25 events). 
Also it requires to click a lot to produce so many events in Connect 😅. I think 15 is enough. 
We can increase it when we have proper batching.

ravicious:
Does this mean we should also increase the time we wait for tshd to shut down?

ravicious:
And perhaps the timeout given to GracefulStop. If it takes 5s for you during perfect conditions and connects to a server in the US, then I can imagine it taking like 10s for some people with spotty connections.

gzdunek:
So with 15 events it takes 3.1 seconds to send in my conditions. 
It seems like 5 seconds timeout is reasonable?
I know that it can take much more, but OTOH I don't think we want people to wait like 10 seconds (in bad conditions) for the app to close 😏 

>Does this mean we should also increase the time we wait for tshd to shut down?

Currently, the timeouts for GracefulStop and waiting for tshd to shut down are equal (5s). Do you think they should be different?

gzdunek:
Reopened in teleport repo.

